PR #86, opened 2020-09-23
[CI-APP][CIAPP-240] Trace Command

juan-fernandez:
### What and why?

[Jira ticket](https://datadoghq.atlassian.net/browse/CIAPP-240)

Create spans associated to CI commands.

An example trace can be found [here](https://app.datadoghq.com/apm/trace/1445452801424440346?spanID=5198930442363958063&env=testing-juan&sort=time&colorBy=service&graphType=flamegraph&shouldShowLegend=true). The corresponding `JenkinsFile` can be found in https://github.com/DataDog/datadog-ci/blob/juan-fernandez/ci-app-trace-command-test-jenkins/Jenkinsfile

For the moment we only support Jenkins for continuing the trace, though this command could be used anywhere (the spans would stand alone). 

### How?

* `spawn` command and create span with exit code and some other metadata. 
* Use `dd-trace` to create the span with the extracted context (only Jenkins supported at the moment). 

### Review checklist

- [x] Feature or bugfix MUST have appropriate tests (unit)
- [ ] Feature or bugfix MUST have appropriate tests (integration)


package.json
```diff
@@ -37,6 +38,7 @@
     "chalk": "2.4.2",
     "clipanion": "2.2.2",
     "datadog-metrics": "https://github.com/DataDog/node-datadog-metrics#12d16a80ea2a8846c91d80f5e127fb7b81b9f347",
+    "dd-trace": "^0.27.0",
```
sdeprez, line 41: Let's prefer fixed version (without `^`)

src/helpers/interfaces.ts
```diff
@@ -0,0 +1,18 @@
...
+    pipeline: {
+      url?: string
+    }
+    provider: {
+      name: string
+    }
+  }
+  git: {
+    branch?: string
+    commit_sha?: string
```
sdeprez, line 12: Inconsistent case, let's rename this one `commitSha`. It's a breaking change but it's ok since it was never used anywhere in Synthetics UI *but* it still requires a change of our schemas so it doesn't break the intake -> https://datadoghq.atlassian.net/browse/SSV-252

src/helpers/tags.ts
```diff
@@ -0,0 +1,19 @@
...
+export const CI_PROVIDER_NAME = 'ci.provider.name'
+
+// Git
+export const GIT_BRANCH = 'git.branch'
+export const GIT_SHA = 'git.commit.sha'
+
+// Other
+export const INSTRUCTION = 'instruction'
+export const EXIT_CODE = 'exit_code'
+export const ERROR = 'error'
```
sdeprez, line 19: Not a big deal but I'm not sure those constants have their place in `helpers`, it looks like they are specific to the `trace` command?

sdeprez (approved):

m-rousse (approved):
LGTM modulo Sdeprez's comments!

yoannmoinet (approved):
LGTM

LICENSE-3rdparty.csv
```diff
@@ -28,3 +28,4 @@ ts-node,import,MIT,Copyright (c) 2014 Blake Embrey (hello@blakeembrey.com)
 tslint,dev,Apache-2.0,Copyright 2013-2019 Palantir Technologies, Inc.
 tslint-config-prettier,dev,MIT,Copyright 2017 Alex Jover Morales <alexjovermorales@gmail.com>
 typescript,dev,Apache-2.0,Copyright (c) Microsoft Corporation.
+dd-trace,import,Apache-2.0,Copyright 2016-Present Datadog, Inc.
```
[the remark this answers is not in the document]
yoannmoinet: But it will fail the 3rdparty test if we change this, right @m-rousse ?
Not sure if we should change this or not.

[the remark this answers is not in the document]
juan-fernandez: it seems to pass without the comma

src/helpers/interfaces.ts
juan-fernandez: ok so to confirm this is blocked until https://datadoghq.atlassian.net/browse/SSV-252 is done, right?

juan-fernandez: ok yeah tests are failing: https://github.com/DataDog/datadog-ci/runs/1320388975?check_suite_focus=true

yoannmoinet: Shouldn't this be done in a separate PR @sdeprez ?

juan-fernandez: that makes sense. If @sdeprez agrees I'll separate this into a different PR

LICENSE-3rdparty.csv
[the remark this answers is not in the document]
m-rousse: The script does not check the number of columns (nor that the license is indeed the right one) but only that there is a line for each dependency listed in the package.json. That's why is does not fail in either case.

src/helpers/interfaces.ts
sdeprez: Whatever works yes, it can be done in another PR to unlock this one (as long as both PRs are done before someones uses this new command so it has proper naming `commitSha` since the beginning!)

PR #88, opened 2020-09-24
[README] add release process

m-rousse:
### What and why?

Add instructions to release `datadog-ci` package.

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,6 +126,19 @@ yarn format
 yarn prepack
 ```
 
+### Release Process
+
+Releasing a new version of `datadog-ci` unfolds as follow:
+
+- create a new branch for the version upgrade
+- update the version using `yarn version [--patch|--minor|--major]` depending on the nature of the changes introduced. You may refer to [Semantic Versioning](https://semver.org/#summary) to determine which to increment.
+- push the branch along with the tag to the upstream (Github), create a Pull Request with the changes introduced detailed in the description and get at least one approval. ([sample Pull Request](https://github.com/DataDog/datadog-ci/pull/78))
+- merge the Pull Request
+- create a Github Release from the [Tags page](https://github.com/DataDog/datadog-ci/tags) with the description of changes introduced
+- finally publish the package using `yarn publish`
+
+
+
 ## License
 
 [Apache License, v2.0](LICENSE)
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -126,6 +126,19 @@ yarn format
...
+
+- create a new branch for the version upgrade
+- update the version using `yarn version [--patch|--minor|--major]` depending on the nature of the changes introduced. You may refer to [Semantic Versioning](https://semver.org/#summary) to determine which to increment.
+- push the branch along with the tag to the upstream (Github), create a Pull Request with the changes introduced detailed in the description and get at least one approval. ([sample Pull Request](https://github.com/DataDog/datadog-ci/pull/78))
+- merge the Pull Request
+- create a Github Release from the [Tags page](https://github.com/DataDog/datadog-ci/tags) with the description of changes introduced
+- finally publish the package using `yarn publish`
+
+
+
```
yoannmoinet, line 141: I don't think that many new lines are necessary 😄 
```suggestion

```

README.md
```diff
@@ -126,6 +126,19 @@ yarn format
...
+### Release Process
+
+Releasing a new version of `datadog-ci` unfolds as follow:
+
+- create a new branch for the version upgrade
+- update the version using `yarn version [--patch|--minor|--major]` depending on the nature of the changes introduced. You may refer to [Semantic Versioning](https://semver.org/#summary) to determine which to increment.
+- push the branch along with the tag to the upstream (Github), create a Pull Request with the changes introduced detailed in the description and get at least one approval. ([sample Pull Request](https://github.com/DataDog/datadog-ci/pull/78))
+- merge the Pull Request
+- create a Github Release from the [Tags page](https://github.com/DataDog/datadog-ci/tags) with the description of changes introduced
+- finally publish the package using `yarn publish`
```
yoannmoinet, line 138: Can we make it a github action?
[It's fairly easy to do](https://github.com/DataDog/build-plugin/blob/master/.github/workflows/publish.yaml), it would be one less step to do manually.

m-rousse: That is nice, I did not know we could trigger on Release creation, I'll add it + update README.md.

yoannmoinet (approved):
LGTM.

Two things to verify for the automatic publish:
- the token used to publish must be from a user that doesn't have 2FA enabled for publish (there are multiple options)
- the `package.json` needs to have this:
```json
    "publishConfig": {
        "access": "public"
    },
```
Because scoped packages (`@datadog/*`) are private by default, so you have to state it's public.

PR #90, opened 2020-09-28
documentation updates for the serverless functions

apigirl:
### What and why?

Updates to the documentation for the serverless configs so that we can pull it into the main documentation site.

### DIFF
--- a/src/commands/lambda/README.md
+++ b/src/commands/lambda/README.md
@@ -6,24 +6,19 @@ You can use the CLI to instrument your AWS Lambda functions with Datadog.
 
 ### Before you begin
 
-Make your AWS credentials `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` available in your environment using the following cmd, or use any of the authentication methods supported by the [AWS js sdk][1].
+Make your AWS credentials `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` available in your environment using the following cmd, or use any of the authentication methods supported by the [AWS JS sdk][1].
 
 ```bash
 # Environment setup
 export AWS_ACCESS_KEY_ID="<ACCESS KEY ID>"
 export AWS_SECRET_ACCESS_KEY="<ACCESS KEY>"
-
-# Or via passing as CLI argument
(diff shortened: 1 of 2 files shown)


src/commands/lambda/README.md
```diff
@@ -17,13 +17,13 @@ export AWS_SECRET_ACCESS_KEY="<ACCESS KEY>"
...
 ```
 
-### API
+Download the [Datadog CI][2].
 
-#### Configuration
+### Configuration
 
-Configuration is done via a json file, by default the tool load `datadog-ci.json` which can be overriden through the `--config` argument.
+Configuration is done using a JSON file. By default AWS Lambda loads `datadog-ci.json` which can be overriden through the `--config` argument.
```
apigirl, line 24: Can we add more specifics here - where is this JSON file - how do you access it - once you update it, do you need to do anything to refresh it?

src/commands/lambda/README.md
```diff
@@ -17,13 +17,13 @@ export AWS_SECRET_ACCESS_KEY="<ACCESS KEY>"
...
 
-The configuration file structure is the following, all fields are optional:
+Use this configuration file structure:
```
apigirl, line 27: should we add a table of all of the supported options like we did for the other repos?

src/commands/lambda/README.md
```diff
@@ -17,13 +17,13 @@ export AWS_SECRET_ACCESS_KEY="<ACCESS KEY>"
 datadog-ci lambda <command>
```
tianchu, line 17: Do you mind helping remove line 15-17? They are redundant.

src/commands/lambda/README.md
tianchu: It's probably unnecessary, since the table from the Commands section probably does a good job of SEO already.

src/commands/lambda/README.md
tianchu: The original sentence was confusing, you need to specify the location of `datadog-ci.json` using the `--config` argument.

tianchu (approved):

yoannmoinet (approved):
LGTM

PR #91, opened 2020-10-05
[serverless] Tag lambdas with ci packege version

Czechh:
### What and why?

This is part of a larger effort to track serverless usage. When a lambda function is provisioned and instrumented with this CLI tool, we want to get data on that and understand how many functions our customers have instrumented with different tools. We tag lambdas already with our macro version, and our serverless plugin version. So, this follows the same pattern to understand which versions are being.

### How?

When a lambda function is being processed, if the lambda has the version tag and it's the same as the current version we don't update at all. But, if no tag is found or a different version exists then we update with the new version tag.

### Review checklist

- [X] Feature or bugfix MUST have appropriate tests (unit, integration)

### DIFF
--- a/src/commands/lambda/tags.ts
+++ b/src/commands/lambda/tags.ts
@@ -1,0 +1,43 @@
+import {Lambda} from 'aws-sdk'
+import {TAG_VERSION_NAME} from './constants'
+import {version} from '../../../package.json'
+
+export interface TagConfiguration {
+  tagResourceRequest?: Lambda.TagResourceRequest
+}
+
+export const applyTagConfig = async (lambda: Lambda, configuration: TagConfiguration) => {
+  const {tagResourceRequest} = configuration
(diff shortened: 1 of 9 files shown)


src/commands/lambda/tags.ts
```diff
@@ -0,0 +1,43 @@
+import {Lambda} from 'aws-sdk'
+import {TAG_VERSION_NAME} from './constants'
+import {version} from '../../../package.json'
```
Czechh, line 3: this might modify the entry point of the script. Will test locally to verify.

yoannmoinet: I'd use something like `const { version } = require(path.join(__dirname, '../../../package.json'));` or something with `require.resolve()`.

yoannmoinet (approved):
TS Config LGTM.

DarcyRaynerDD (commented):
LGTM

PR #95, opened 2020-10-06
[PROF-2111] Add dependencies upload command to integrate with Snyk CLI

piotr-oles:
### What and why?
https://datadoghq.atlassian.net/browse/PROF-2109

Adds a command to upload the dependencies graph generated by the Snyk CLI tool. We need this graph to have vulnerabilities scanning feature in Continous Profiler.

### How?
It's a simple command that uploads file content and some metadata to one of our endpoints. Nothing fancy 😄 

### Review checklist

### TODO:
- [x] Validate input
- [x] Upload file with metadata
- [x] Handle generic error
- [ ] Use API response to check if there is a validation error or upload was successful


src/index.ts
```diff
@@ -18,8 +19,11 @@ const cli = new Cli({
 
 const commandsPath = `${__dirname}/commands`
 for (const commandFolder of fs.readdirSync(commandsPath)) {
-  // tslint:disable-next-line: no-var-requires
-  require(`${commandsPath}/${commandFolder}`).forEach(cli.register.bind(cli))
+  const commandPath = `${commandsPath}/${commandFolder}`
+  if (fs.statSync(commandPath).isDirectory()) {
+    // tslint:disable-next-line: no-var-requires
+    require(commandPath).forEach((command: CommandClass) => cli.register(command))
+  }
```
yoannmoinet, line 26: why is this now needed?

src/commands/dependencies/__tests__/upload.test.ts
```diff
@@ -0,0 +1,405 @@
...
+    )
+    expect(stdout[1]).toEqual(
+      chalk.yellow(`│ Missing optional ${releaseVersion} parameter.                                        │`)
+    )
+    expect(stdout[2]).toEqual(
+      chalk.yellow('│ The analysis may use out of date dependencies and produce false positives/negatives. │')
+    )
+    expect(stdout[3]).toEqual(
+      chalk.yellow('└──────────────────────────────────────────────────────────────────────────────────────┘')
+    )
```
yoannmoinet, line 192: Is this necessary, this seems hard to maintain.
Once you update the layout you'll have to update the test too.
Could we only check for the important part of the output we're looking for?

Something like:
```suggestion
    const output = stdout.join('');
    expect(output.includes('The analysis may use out of date dependencies and produce false positives/negatives.')).toBe(true);
    expect(output.includes(`Missing optional ${releaseVersion} parameter.`)).toBe(true);
```

Or is the layout something we're testing too?

src/commands/dependencies/__tests__/upload.test.ts
```diff
@@ -0,0 +1,405 @@
...
+    const stderr = context.getStderrBuffer().split(os.EOL)
+
+    expect(stderr).toEqual([''])
+    expect(stdout[0]).toEqual(`${chalk.bold('File')}:    ${resolvedFilePath}`)
+    expect(stdout[1]).toEqual(`${chalk.bold('Source')}:  snyk`)
+    expect(stdout[2]).toEqual(`${chalk.bold('Service')}: my-service`)
+    expect(stdout[3]).toEqual(`${chalk.bold('Version')}: 1.234`)
+    expect(stdout[4]).toEqual('')
+    expect(stdout[5]).toEqual('Uploading dependencies...')
+    expect(stdout[6]).toMatch(/Dependencies uploaded in .* seconds\./)
```
yoannmoinet, line 219: Same as before, this feels like a huge overhead to test every line of the output.
Especially including chalk's characters.

src/commands/dependencies/__tests__/upload.test.ts
```diff
@@ -0,0 +1,405 @@
...
+
+      resolve(chunks.join(''))
+    }
+    stream.on('data', handleData)
+    stream.once('error', handleError)
+    stream.once('end', handleEnd)
+
+    stream.resume()
+  })
+}
```
yoannmoinet, line 405: Can we move these helper functions outside the test file?

We could have a `helpers.ignore.ts` somewhere in the `__tests__` directory.
And add the pattern `'.+\\.ignore\\.+'` [here](https://github.com/DataDog/datadog-ci/blob/master/jest.config.js#L16).

src/commands/dependencies/__tests__/upload.test.ts
```diff
@@ -0,0 +1,405 @@
...
+    stdout: new Writable({
+      write(chunk, encoding, callback) {
+        stdoutChunks.push(chunk)
+        callback()
+      },
+    }),
+  }
+}
+
+function streamToString(stream: Readable): Promise<string> {
```
yoannmoinet, line 382: Why using both format of functions? 
(`const a = () => {}` and `function b() {}`)

src/commands/dependencies/api.ts
```diff
@@ -0,0 +1,30 @@
+import {default as axios} from 'axios'
+import FormData from 'form-data'
+import fs from 'fs'
+
+import {APIHelper, Payload} from './interfaces'
+
+export const apiConstructor = (baseIntakeUrl: string, apiKey: string, appKey: string): APIHelper => {
+  function uploadDependencies(payload: Payload) {
```
yoannmoinet, line 8: Why not an arrow function here?

yoannmoinet (commented):
Nice stuff.
I have some comments.

src/index.ts
piotr-oles: I had an issue with Mac OS storing the `.DS_Store` file in the command folder. `require('command/.DS_Store')` causes an error when I wanted to test using `yarn launch` :)

src/commands/dependencies/__tests__/upload.test.ts
piotr-oles: I would like to test if everything is printed correctly to the end-user. I will make it looser by not checking line number and stripping chalk formatting :)

src/commands/dependencies/__tests__/upload.test.ts
piotr-oles: I have no preference - maybe we could have a tslint rule to ban named functions if you want to keep consistency? :)

yoannmoinet: Definitely, I thought we already had. I'll push a PR for it. No need to update your PR.

src/index.ts
yoannmoinet: Oh I see. Makes sense then.

src/commands/dependencies/__tests__/upload.test.ts
piotr-oles: I moved this to the `helpers` directory :)

src/commands/dependencies/__tests__/upload.test.ts
piotr-oles: Oh, I didn't notice your comment - I added a tslint rule in this PR

yoannmoinet (approved):
LGTM.
Very cool addition.

src/commands/dependencies/__tests__/upload.test.ts
yoannmoinet: That's ok, I was afraid of the diff it would produce. 
But it's manageable after all.

src/commands/dependencies/upload.ts
```diff
@@ -0,0 +1,157 @@
...
+      ],
+    ],
+  })
+
+  private static INVALID_INPUT_EXIT_CODE = 1
+  private static MISSING_FILE_EXIT_CODE = 2
+  private static UPLOAD_ERROR_EXIT_CODE = 3
+
+  private config = {
+    apiHost: process.env.DATADOG_API_HOST || 'api.datadoghq.com',
```
m-rousse, line 43: Other commands uses `DATADOG_SITE` environment variable to select the correct datadog instance, could you use this instead?
Also, we should maybe have a `getApiUrlFromSite` function in helpers that all commands could use to obtain the correct api URL.

m-rousse (approved):
Nice PR (the context mock using streams + the linter `only arrow` rule are so nice).
Two nits - otherwise LGTM.

src/commands/dependencies/upload.ts
piotr-oles: I decided to use `DATADOG_API_HOST` for simplicity. There are some cons regarding the `DATADOG_SITE` variable:
 * we can't use `app.datad0g.com` host to test commands on staging (because with `DATADOG_SITE` we have to hardcode `api.` subdomain for prod)
 * we still have to set `DATADOG_API_HOST` to send metrics to the EU. I think it's a bad UX because if someone sends a dependency graph to the EU, he will also want to send metrics to the EU.

Here is a comparison between two README's that illustrates this:
## `sourcemap upload`:
It is possible to configure the tool to use Datadog EU by defining the `DATADOG_SITE` environment variable to datadoghq.eu. By defaut the requests are sent to Datadog US. In order to send the usage metrics to the correct datacenter, the `DATADOG_API_HOST` must also be set if another instance of Datadog is used. For example, for Datadog EU, it should be set to api.datadoghq.eu.

## `dependencies upload`:
By default, the command sends requests to Datadog US. To use Datadog EU, set `DATADOG_API_HOST` environment variable to `api.datadoghq.eu`.

Do you think I should use `DATADOG_SITE` for consistency, even if it's harder to set-up and explain? :)

piotr-oles: Ah, I read your comment again and now I see what you suggest - it makes sense :)

PR #101, opened 2020-10-17
[synthetics] add retry machanism on synthetics command requests

m-rousse:
### What and why?

Sometimes the backend may return 5xx errors on which we want to retry. So this PR adds a retry mechanism on all requests performed by the synthetics command to retry every 500 milliseconds if the status code of the response is in the 5xx class - three times at most

### Review checklist

### DIFF
--- a/src/commands/synthetics/utils.ts
+++ b/src/commands/synthetics/utils.ts
@@ -291,3 +291,28 @@ export const runTests = async (
 }
 
 const definedTypeGuard = <T>(o: T | undefined): o is T => !!o
+
+export async function retry<T, E extends Error>(
+  func: () => Promise<T>,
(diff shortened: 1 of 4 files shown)


src/commands/synthetics/utils.ts
```diff
@@ -291,3 +291,28 @@ export const runTests = async (
...
+    try {
+      return await func()
+    } catch (e) {
+      const waiter = shouldRetryAfterWait(retries, e)
+      if (waiter) {
+        if (typeof waiter === 'number') {
+          await wait(waiter)
+        } else {
+          await waiter
+        }
```
sdeprez, line 309: `waiter` can only be a number

sdeprez (approved):

PR #103, opened 2020-10-21
[doc] add testing process for commands interacting with datadog

m-rousse:
### What and why?

This PR adds testing instructions for 3 commands (interacting with datadog: dependencies, sourcemaps and synthetics) that should be run during the release process to ensure all commands work as expected.


piotr-oles (approved):

Czechh (approved):

yoannmoinet (commented):
I appreciate the initiative and understand why it's there.

But I'm afraid it won't be used at all, this isn't discoverable at all and it unfairly adds a lot of responsibility to the person publishing a new version.

If we have such detailed steps to reproduce and expected results, can't we automate the process instead?

Each team should be responsible for their end-to-end testing, if they don't have one, they shouldn't rely on other people publishing the package testing it manually.

And it doesn't scale at all.

yoannmoinet (approved):

yoannmoinet:
Of course it's not expected in this PR.
Having doc about this is still useful on its own.

piotr-oles:
I had a similar challenge in my open source project. I didn't found any e2e testing library for npm packages, so I wrote some tooling for my use-case: https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/tree/master/test/e2e . The sandbox part should be re-usable. I don't know if it's a direction we would like to take - I'm just pasting if you are interested :)

m-rousse:
We will add end-to-end tests for dependency + sourcemap commands in the CI. I'll do it in another PR.

PR #105, opened 2020-10-21
[serverless] Use GovCloud lambda layer when a GovCloud region is detected

DylanLovesCoffee:
### What and why?

SLS-701 - Adds logic to apply the GovCloud lambda layer when a user is instrumenting a function in GovCloud.

### How?

Returns the GovCloud lambda ARN when a GovCloud region is detected.

### Review checklist


src/commands/lambda/function.ts
```diff
@@ -100,6 +106,10 @@ const getLambdaConfig = async (
 const getLayerArn = (runtime: Runtime, settings: InstrumentationSettings, region: string) => {
   const layerName = RUNTIME_LAYER_LOOKUP[runtime]
   const account = settings.layerAWSAccount ?? DEFAULT_LAYER_AWS_ACCOUNT
+  const isGovCloud = region === 'us-gov-east-1' || region === 'us-gov-west-1'
```
tianchu, line 109: Since CLI can work automatically for new regions (i.e., layer arns not hard coded in a file), it might be better to only check against the prefix `us-gov`?

tianchu (approved):
LGTM, left a small comment.

PR #108, opened 2020-10-30
Add Publish Script

juan-fernandez:
### What and why?

We want to make it as automatic as possible to release the package.

### How?

Add a github actions script to release the package.

### Review checklist

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,18 @@
+name: Publish package on NPM
+on:
+  release:
+    types: [published]
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: '12.x'
+      - run: yarn config set npmAuthToken $YARN_NPM_AUTH_TOKEN
+        env:
+          YARN_NPM_AUTH_TOKEN: ${{ secrets.YARN_NPM_AUTH_TOKEN }}
+      - run: yarn
+      - run: yarn pack
+      - run: yarn publish


.github/workflows/publish.yml
```diff
@@ -0,0 +1,18 @@
+name: Publish package on NPM
+on:
+  release:
+    types: [published]
```
juan-fernandez, line 4: in https://github.com/DataDog/build-plugin/blob/master/.github/workflows/publish.yaml#L4 we have `[created]` but I think we this we avoid creating releases for drafts. This worked quite well for us at scope: https://github.com/DataDog/scope-javascript-agent/blob/master/.github/workflows/npmpublish.yml#L5

.github/workflows/publish.yml
```diff
@@ -0,0 +1,18 @@
...
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: '12.x'
+      - run: yarn config set npmAuthToken $YARN_NPM_AUTH_TOKEN
+        env:
+          YARN_NPM_AUTH_TOKEN: ${{ secrets.YARN_NPM_AUTH_TOKEN }}
+      - run: yarn
+      - run: yarn pack
+      - run: yarn publish
```
juan-fernandez, line 18: if at some point we want to make beta releases we can with the "prerelease" option in github releases: https://github.com/DataDog/scope-javascript-agent/blob/master/.github/workflows/npmpublish.yml#L25

.github/workflows/publish.yml
yoannmoinet: Oh, that's interesting.

yoannmoinet (approved):
LGTM, thanks for doing it!

m-rousse (approved):
🎉

.github/workflows/publish.yml
juan-fernandez: sadly the docs around this are not great 😆 https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#onevent_nametypes

PR #116, opened 2020-11-09
[synthetics] gracefully handle non-parsable start urls

m-rousse:
### What and why?

Gracefully handle non-parsable start urls of synthetics tests.

### How?

The parsing is now in a `try/catch` block and variables replacing is done on a "best-effort" basis.

### Review checklist

### DIFF
--- a/src/commands/synthetics/utils.ts
+++ b/src/commands/synthetics/utils.ts
@@ -33,9 +33,14 @@ const PUBLIC_ID_REGEX = /^[\d\w]{3}-[\d\w]{3}-[\d\w]{3}$/
 const SUBDOMAIN_REGEX = /(.*?)\.(?=[^\/]*\..{2,5})/
 
 const template = (st: string, context: any): string =>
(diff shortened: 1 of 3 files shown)


src/commands/synthetics/utils.ts
```diff
@@ -93,6 +83,32 @@ export const handleConfig = (test: Test, publicId: string, config?: ConfigOverri
...
+    context.HOST = objUrl.host
+    context.HOSTNAME = objUrl.hostname
+    context.ORIGIN = objUrl.origin
+    context.PARAMS = objUrl.search
+    context.PATHNAME = objUrl.pathname
+    context.PORT = objUrl.port
+    context.PROTOCOL = objUrl.protocol
+    context.SUBDOMAIN = subdomainMatch ? subdomainMatch[1] : undefined
+  } catch {
+    write(`Unable to parse start url: ${url}`)
```
sdeprez, line 106: I'm not sure it will be very clear to the user. Maybe `The start url ${url} contains variables so local overrides will be ignored`. Not really about it either because I don't think the user will understand what `local overrides` are so don't hesitate to find something better!

sdeprez (approved):

PR #117, opened 2020-11-09
[synthetics][SW-598] ignore *_PROXY environment variables

m-rousse:
### What and why?

It allows ignoring *_PROXY environment variables, and leverages this in synthetics command.

### How?

It does so by setting the `proxy: false` config in axios when building the axios instance. (cf. https://github.com/axios/axios/pull/691)

### Review checklist

No unit tests were added to test this behavior as it would require setting an HTTP server and was deemed too heavy.

### DIFF
--- a/src/helpers/utils.ts
+++ b/src/helpers/utils.ts
@@ -60,7 +60,13 @@ export interface ProxyConfiguration {
   protocol: ProxyType
 }
 
-export const getRequestBuilder = (baseUrl: string, apiKey: string, appKey?: string, proxyOpts?: ProxyConfiguration) => {
+export const getRequestBuilder = (
+  baseUrl: string,
+  apiKey: string,
+  appKey?: string,
+  proxyOpts?: ProxyConfiguration,
+  disableEnvironmentVariables = false
+) => {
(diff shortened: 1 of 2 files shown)


src/helpers/utils.ts
```diff
@@ -60,7 +60,13 @@ export interface ProxyConfiguration {
...
 }
 
-export const getRequestBuilder = (baseUrl: string, apiKey: string, appKey?: string, proxyOpts?: ProxyConfiguration) => {
+export const getRequestBuilder = (
+  baseUrl: string,
+  apiKey: string,
+  appKey?: string,
+  proxyOpts?: ProxyConfiguration,
+  disableEnvironmentVariables = false
+) => {
```
yoannmoinet, line 69: nit: maybe we should use an object here instead of a growing list of parameters.

yoannmoinet (approved):
LGTM.

Just added a nit if we plan on adding more parameters to this function, maybe we should consider using an object instead.

sdeprez (approved):

src/helpers/utils.ts
m-rousse: Doing it in another PR to ease reviews of other teams.

PR #122, opened 2020-11-25
[Synthetics][SW-635] Don't display unhealthy results URLs

lefebvree:
### What and why?

[SW-635](https://datadoghq.atlassian.net/browse/SW-635)

Unhealthy results and timed-out tests display the regular result URL, but it leads to a 404 since unhealthy results aren't viewable and timed-out results are not yet available if ever.

### How?

Remove displayed line with duration and results URLs for unhealthy test results.
Display `(not yet received)` message after result URL for timed-out executions. 

Also add extra line breaks in the output between all execution results, used to only be present when `resultOutcome` was empty, now consistently split all results.

<details>
  <summary>Example output</summary>

image

</details>

### Review checklist


sdeprez (commented):
Your PR mentions `timed out` results but only covers unhealthy? Which is what we want I think, timeout results still have an URL so the user can see how long did they finally take

m-rousse (approved):
:+1:

PR #125, opened 2020-11-26
[synthetics] fix start url variable substitution

m-rousse:
### What and why?

Fix variable substitution by allowing variables with an empty content.

### Review checklist

### DIFF
--- a/src/commands/synthetics/__tests__/utils.test.ts
+++ b/src/commands/synthetics/__tests__/utils.test.ts
@@ -169,6 +169,22 @@ describe('utils', () => {
       expect(handledConfig.startUrl).toBe(expectedUrl)
       process.env = envVars
     })
+
+    test('startUrl with empty variable is replaced', () => {
+      const publicId = 'abc-def-ghi'
+      const fakeTest = {
+        config: {request: {url: 'http://exmaple.org/path'}},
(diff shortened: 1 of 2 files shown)


sdeprez (approved):

src/commands/synthetics/__tests__/utils.test.ts
```diff
@@ -169,6 +169,22 @@ describe('utils', () => {
...
     })
+
+    test('startUrl with empty variable is replaced', () => {
+      const publicId = 'abc-def-ghi'
+      const fakeTest = {
+        config: {request: {url: 'http://exmaple.org/path'}},
+        public_id: publicId,
+      } as Test
+      const configOverride = {
+        startUrl: 'http://127.0.0.1/newPath{{PARAMS}}',
```
etnbrd, line 180: Where is `PARAMS` defined?

etnbrd (approved):
LGTM

src/commands/synthetics/__tests__/utils.test.ts
m-rousse: It is extracted from the startUrl of the test (so in that case it is empty - it used to not be interpolated, but this PR fixes it and this test makes sure of that)

PR #128, opened 2020-12-10
[synthetics] improve synthetics API results rendering

m-rousse:
### What and why?

This PR improves the rendering of synthetics results for synthetics API SSL/TCP and DNS tests.

### DIFF
--- a/src/commands/synthetics/renderer.ts
+++ b/src/commands/synthetics/renderer.ts
@@ -131,7 +131,7 @@ const renderResultOutcome = (result: Result, test: Test, icon: string, color: ty
 }
 
 const renderApiRequestDescription = (subType: string, request: Test['config']['request']): string => {
-  if (subType === 'dns' && request.host) {
+  if (subType === 'dns') {
     const text = `Query for ${request.host}`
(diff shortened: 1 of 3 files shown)


src/commands/synthetics/renderer.ts
```diff
@@ -130,6 +130,23 @@ const renderResultOutcome = (result: Result, test: Test, icon: string, color: ty
   }
 }
 
+const renderApiRequestDescription = (subType: string, request: Test['config']['request']): string => {
+  if (subType === 'dns' && request.host) {
```
etnbrd, line 134: `request.host` is expected to be always defined?
Otherwise, at first glance, it seems if it's not, the `dns` request will be displayed like the default which looks more like an `http` than a default.

etnbrd (approved):
LGTM

src/commands/synthetics/renderer.ts
m-rousse: 🤦 my apologies, it is always defined for DNS test results, I forgot to remove this check when refactoring my code.

PR #131, opened 2020-12-14
Change the sourcemap URL matching

jybp:
### What and why?

Update the `--minified-path-prefix` flag usage.

### How?

The backend will now be able to match against the path portion of unminified URLs instead of just the full URL.

### Review checklist

### DIFF
--- a/src/commands/sourcemaps/__tests__/upload.test.ts
+++ b/src/commands/sourcemaps/__tests__/upload.test.ts
@@ -34,15 +34,39 @@ describe('upload', () => {
     })
   })
 
-  describe('getMinifiedURL: minifiedPathPrefix has no leading slash', () => {
-    test('should throw an error', () => {
+  describe('isMinifiedPathPrefixValid: full URL', () => {
+    test('should return false', () => {
       const command = new UploadCommand()
-      command['basePath'] = '/js/sourcemaps'
+      command['minifiedPathPrefix'] = 'http://datadog.com/js'
+
+      expect(command['isMinifiedPathPrefixValid']()).toBe(true)
+    })
+  })
+
+  describe('isMinifiedPathPrefixValid: URL without protocol', () => {
+    test('should return false', () => {
+      const command = new UploadCommand()
+      command['minifiedPathPrefix'] = '//datadog.com/js'
+
+      expect(command['isMinifiedPathPrefixValid']()).toBe(true)
+    })
+  })
+
+  describe('isMinifiedPathPrefixValid: leading slash', () => {
+    test('should return false', () => {
+      const command = new UploadCommand()
+      command['minifiedPathPrefix'] = '/js'
+
+      expect(command['isMinifiedPathPrefixValid']()).toBe(true)
+    })
+  })
+
+  describe('isMinifiedPathPrefixValid: no leading slash', () => {
+    test('should return false', () => {
+      const command = new UploadCommand()
       command['minifiedPathPrefix'] = 'js'
 
-      expect(() => command['getMinifiedURL']('/js/sourcemaps/common.min.js.map')).toThrow(
-        'Absolute path must have a leading slash'
-      )
+      expect(command['isMinifiedPathPrefixValid']()).toBe(false)
     })
   })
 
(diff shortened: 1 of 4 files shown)


src/commands/sourcemaps/__tests__/upload.test.ts
```diff
@@ -16,6 +16,15 @@ describe('upload', () => {
...
   })
 
+  describe('getMinifiedURL: minifiedPathPrefix is an absolute path', () => {
+    test('should return correct URL', () => {
+      const command = new UploadCommand()
+      command['basePath'] = '/js/sourcemaps'
+      command['minifiedPathPrefix'] = '/js'
+      expect(command['getMinifiedURL']('/js/sourcemaps/common.min.js.map')).toBe('/js/common.min.js.map')
+    })
+  })
```
lraucy, line 26: @jybp if the prefix is neither a full URL nor an absolute path, it'll fail in the backend. Would that make sense to add prefix validation directly in the command to give feedback directly when running the command, with a meaningful error message?

jybp: Good point. I'd argue in favor of having the leading slash optional and letting the backend handle it properly.

jybp: @lraucy Changed it to check the leading slash.

Should we now make the `--minified-path-prefix` flag optional in the off change no prefix is required?

lraucy: It still make sense to ask for it; `--minified-path-prefix /` says: I want to match everything in this directory as if it was the root directory on my server. I don't think it should be considered as the default behavior, i.e. when the prefix is omitted. Maybe it's worth adding an example with `--minified-path-prefix /` in the documentation though?

jybp: I updated the readme with a sentence on that. Tell my WDYT

src/commands/sourcemaps/upload.ts
```diff
@@ -122,6 +123,17 @@ export class UploadCommand extends Command {
...
+    let protocol
+    try {
+      const objUrl = new URL(this.minifiedPathPrefix!)
+      protocol = objUrl.protocol
+    } catch {
+      // Do nothing.
+    }
+
+    if (!protocol && !this.minifiedPathPrefix!.startsWith('/')) {
+      throw new Error('Absolute path must have a leading slash')
```
lraucy, line 135: With this version if we use the command line incorrectly we only have this error message + a stack trace, which is not very helpful:
```
Internal Error: Absolute path must have a leading slash
    at UploadCommand.getMinifiedURL (/Users/loic.raucy/go/src/github.com/DataDog/datadog-ci/src/commands/sourcemaps/upload.ts:135:13)
    at /Users/loic.raucy/go/src/github.com/DataDog/datadog-ci/src/commands/sourcemaps/upload.ts:116:27
    at Array.map (<anonymous>)
    at UploadCommand.getMatchingSourcemapFiles (/Users/loic.raucy/go/src/github.com/DataDog/datadog-ci/src/commands/sourcemaps/upload.ts:111:27)
    at UploadCommand.<anonymous> (/Users/loic.raucy/go/src/github.com/DataDog/datadog-ci/src/commands/sourcemaps/upload.ts:89:27)
    at Generator.next (<anonymous>)
    at /Users/loic.raucy/go/src/github.com/DataDog/datadog-ci/src/commands/sourcemaps/upload.ts:8:71
    at new Promise (<anonymous>)
    at __awaiter (/Users/loic.raucy/go/src/github.com/DataDog/datadog-ci/src/commands/sourcemaps/upload.ts:4:12)
    at UploadCommand.execute (/Users/loic.raucy/go/src/github.com/DataDog/datadog-ci/src/commands/sourcemaps/upload.ts:40:16)
error Command failed with exit code 1.
```

We could avoid rendering the stack trace, render a more complete error message (something that gives more context such as `--minified-path-prefix should either be an URL (such as "http://example.com/static") or an absolute path starting with a `/` such as "/static")` displayed in red and exit with a non-zero status code.

We already have some rendering code in `src/commands/sourcemaps/renderer.ts` and other commands might have more examples.

jybp: Thanks for the `renderer.ts` tip. Changed the code a bit to accommodate the proper error message.

lraucy (approved):
LGTM

PR #136, opened 2020-12-22
Git support for sourcemap upload

jybp:
### What and why?

This PR changes the sourcemap upload command to upload additional repository data alongside the sourcemaps.

### How?

By invoking git commands using the new `simple-git` dependency.

### Review checklist


src/commands/sourcemaps/upload.ts
```diff
@@ -113,21 +116,46 @@ export class UploadCommand extends Command {
...
+      const simpleGit = newSimpleGit()
+      repositoryData = await gitInfos(simpleGit, this.context.stdout, this.repositoryURL)
+    }
 
-    return sourcemapFiles.map((sourcemapPath) => {
-      const minifiedFilePath = getMinifiedFilePath(sourcemapPath)
+    return Promise.all(
+      sourcemapFiles.map(async (sourcemapPath) => {
+        const minifiedFilePath = getMinifiedFilePath(sourcemapPath)
+        let infos: string | undefined
```
lraucy, line 131: Maybe we could change `infos` into a slightly less generic name such as `gitMetadata` or `gitInfo`?

src/commands/sourcemaps/upload.ts
```diff
@@ -113,21 +116,46 @@ export class UploadCommand extends Command {
...
+        return {
+          minifiedFilePath,
+          minifiedUrl: this.getMinifiedURL(minifiedFilePath),
+          projectPath: this.projectPath,
+          repository: infos,
+          service: this.service!,
+          sourcemapPath,
+          version: this.releaseVersion!,
+        }
+      })
```
lraucy, line 157: In terms of readability, the function's purpose is less clear than before. Ideally the scope of this function should be very high-level:
- look for source maps
- generate something that can be used as a payload for the backend

Now the details about git handling make this function a bit harder to comprehend.

Could we have a function `getGitInfos` that would wraps all those details?
It would be also nice to document this function to explain exactly the behavior implemented (since it has a few edge cases for now, better be explicit about how we are handling this)

src/commands/sourcemaps/git.ts
```diff
@@ -0,0 +1,208 @@
...
+
+// GitTrackedFiles returns the tracked files of the current repository.
+export const gitTrackedFiles = async (git: simpleGit.SimpleGit): Promise<string[]> => {
+  const files = await git.raw('ls-files')
+
+  return files.split(/\r\n|\r|\n/)
+}
+
+// TrimStart trims from a set of characters from the start of a string.
+export const trimStart = (str: string, chars: string[]) => {
```
lraucy, line 60: Do we need to export all the functions defined in the file? A lot of them looks like internals?

src/commands/sourcemaps/git.ts
```diff
@@ -0,0 +1,208 @@
...
+  // Query parmeter
+  const pos = source.lastIndexOf('?')
+  if (pos > 0) {
+    source = source.slice(0, pos)
+  }
+
+  return [trimStart(source, ['/', '.']), projectPathFound]
+}
+
+// TrackedFilesMap transforms a list of tracked files into a map to look up sources.
```
lraucy, line 114: Maybe we could add a bit more details to the docstring? More generally speaking I think the algorithm used to do the matching could be more documented.

lraucy (commented):
Left a couple of comments, but I'd like to review it again with more documentation about the different functions written here!

src/commands/sourcemaps/git.ts
jybp: Just cleaned up the `export` in the file. (Some are still exported because of tests)

src/commands/sourcemaps/git.ts
jybp: I added several comments explaining what the functions do and how to use them. Tell me if it makes more sense.

src/commands/sourcemaps/upload.ts
jybp: I reworked the functions in the `upload.ts` file and added comments to make everything more clear. Tell me what you think.

src/commands/sourcemaps/upload.ts
```diff
@@ -94,7 +97,11 @@ export class UploadCommand extends Command {
       )
     )
     const metricsLogger = getMetricsLogger(this.releaseVersion, this.service)
-    const payloads = this.getMatchingSourcemapFiles()
+    let payloads = await this.getPayloadsToUpload()
+    const useGit = this.disableGit === undefined || !this.disableGit
+    if (useGit) {
+      payloads = await this.addRepositoryDataToPayloads(payloads)
+    }
```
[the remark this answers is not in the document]
yoannmoinet: Seeing @lraucy's comment about the `getPayloadsToUpload()`, maybe we should add a new function that would merge the two.

Like, keep `getPayloadsToUpload(useGit)` for this high-level function where we have both `getMatchingSourcemapFiles` and `addRepositoryDataToPayloads` when needed:

```js
private getPayloadsToUpload = async (useGit: boolean = false): Promise<Payload[]> => {
  const payloads = await this.getMatchingSourcemapFiles();
  if (!useGit) {
    return payloads;
  }
  return this.addRepositoryDateToPayloads(payloads);
}
```

It's just that this logic shouldn't be in the main function IMHO.
WDYT?

src/commands/sourcemaps/git.ts
```diff
@@ -0,0 +1,239 @@
...
+// GitRemote returns the remote of the current repository.
+const gitRemote = async (git: simpleGit.SimpleGit): Promise<string> => {
+  const remotes = await git.getRemotes(true)
+  if (remotes.length === 0) {
+    throw new Error('No git remotes available')
+  }
+  remotes.forEach((remote) => {
+    if (remote.name === 'origin') {
+      return remote.refs.push
+    }
```
yoannmoinet, line 30: Why do we need to do this?
Maybe a comment about this particularity would be useful.

src/commands/sourcemaps/git.ts
```diff
@@ -0,0 +1,239 @@
...
+const gitHash = async (git: simpleGit.SimpleGit): Promise<string> => git.revparse('HEAD')
+
+// GitTrackedFiles returns the tracked files of the current repository.
+export const gitTrackedFiles = async (git: simpleGit.SimpleGit): Promise<string[]> => {
+  const files = await git.raw('ls-files')
+
+  return files.split(/\r\n|\r|\n/)
+}
+
+// TrimStart trims from a set of characters from the start of a string.
```
yoannmoinet, line 59: NIT: This was confusing for the following function. No need for the name of the function here.
The same comment for other functions as well, or at least follow the same casing as the function name.

```suggestion
// trims from a set of characters from the start of a string.
```

src/commands/sourcemaps/git.ts
```diff
@@ -0,0 +1,239 @@
...
+  let start = 0
+  const end = str.length
+  while (start < end && chars.indexOf(str[start]) >= 0) {
+    ++start
+  }
+
+  return start > 0 ? str.substring(start, end) : str
+}
+
+// Trim trims from a set of characters from a string.
```
yoannmoinet, line 70: NIT: Confusing.

```suggestion
// trims from a set of characters from a string.
```

src/commands/sourcemaps/git.ts
```diff
@@ -0,0 +1,239 @@
...
+
+  return start > 0 || end < str.length ? str.substring(start, end) : str
+}
+
+// CleanupSource generates a proper source file path from a sourcemap:
+// - Strip a set of hard-coded prefixes ('webpack:///./')
+// - Strip the eventual projectPath
+// - Removes query parameters
+//
+// It returns the new source as well as wether are not the specified projectPath was stripped.
```
yoannmoinet, line 89:
```suggestion
// It returns the new source as well as whether or not the specified projectPath was stripped.
```

yoannmoinet (commented):
Mostly nits.
Well done.

src/commands/sourcemaps/README.md
```diff
@@ -46,9 +46,11 @@ Absolute paths must always begin with a leading slash. Simply specify `/` if the
 
 In addition, some optional parameters are available:
 
-* `--project-path` (default: empty): the path of the project on the filesystem where the sourcemaps were built. This will be stripped off from the file name in the displayed stack traces.
 * `--concurrency` (default: `20`): number of concurrent upload to the API.
 * `--dry-run` (default: `false`): it will run the command without the final step of upload. All other checks are performed.
+* `--disable-git` (default: false): Presence of flag prevents the command from sending any repository related data to Datadog (hash, remote URL and the correct paths within the repository of the source paths referenced in the sourcemap).
+* `--repository-url` (default: empty): overrides the repository remote with a custom URL. For example: https://github.com/my-company/my-project
+* `--project-path` (default: empty): the path of the project where the sourcemaps were built. This will be stripped off from sources paths referenced in the sourcemap so they can be properly matched against tracked files paths.
```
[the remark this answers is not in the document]
jybp: Right, it's more logical 👍

src/commands/sourcemaps/git.ts
jybp: That was an arbitrary choice made when I wrote the RFC. The idea is that remotes don't really have any priority attached to them in git. So trying to pick the one called 'origin' since it's the default used by git before falling back to picking the first one seemed to make more sense to me.

I'll add a comment.

src/commands/sourcemaps/git.ts
jybp: I could not keep the casing because of linting errors `comment must start with uppercase letter`. I'll adjust the comments 👍

src/commands/sourcemaps/git.ts
jybp: I'll update the comments.

src/commands/sourcemaps/upload.ts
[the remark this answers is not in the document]
jybp: Yes looks better indeed. Just made the changes.

src/commands/sourcemaps/git.ts
yoannmoinet: Oh I see, but are you sure that when you return within a forEach, it's returned to the parent function?

jybp: You're right, fixed the function and covered it with tests 🤦

m-rousse (approved):
LGTM - minor nit on the 3rd party csv

LICENSE-3rdparty.csv
```diff
@@ -28,4 +28,5 @@ ts-node,import,MIT,Copyright (c) 2014 Blake Embrey (hello@blakeembrey.com)
 tslint,dev,Apache-2.0,Copyright 2013-2019 Palantir Technologies, Inc.
 tslint-config-prettier,dev,MIT,Copyright 2017 Alex Jover Morales <alexjovermorales@gmail.com>
 typescript,dev,Apache-2.0,Copyright (c) Microsoft Corporation.
-dd-trace,import,Apache-2.0,Copyright 2016-Present Datadog Inc.
\ No newline at end of file
+dd-trace,import,Apache-2.0,Copyright 2016-Present Datadog Inc.
+simple-git,import,MIT,Copyright (c) 2015 Steve King
```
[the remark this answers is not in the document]
jybp: Done 👍

src/commands/sourcemaps/git.ts
```diff
@@ -0,0 +1,242 @@
...
+
+  for (const remote of remotes) {
+    // We're trying to pick the remote called with the default git name 'origin'.
+    if (remote.name === 'origin') {
+      return remote.refs.push
+    }
+  }
+
+  // Falling back to picking the first remote in the list if 'origin' is not found.
+  return remotes[0].refs.push
```
lraucy, line 36: Is there a particular reason to use `remotes[0].refs.push` and not `remotes[0].refs.fetch`?

yoannmoinet (approved):
LGTM!

src/commands/sourcemaps/git.ts
```diff
@@ -0,0 +1,242 @@
...
+    return remote
+  }
+}
+
+// Returns the hash of the current repository.
+const gitHash = async (git: simpleGit.SimpleGit): Promise<string> => git.revparse('HEAD')
+
+// Returns the tracked files of the current repository.
+export const gitTrackedFiles = async (git: simpleGit.SimpleGit): Promise<string[]> => {
+  const files = await git.raw('ls-files')
```
lraucy, line 57: Is this command run from current location or from the base of the repository?

Does it work correctly if the command `datadog-ci` is run from a subfolder?

src/commands/sourcemaps/git.ts
```diff
@@ -0,0 +1,242 @@
...
+  // Prefixes
+  const prefixesToRemove = ['webpack:']
+  for (const p of prefixesToRemove) {
+    if (source.startsWith(p)) {
+      source = source.slice(p.length)
+    }
+  }
+  source = trimStart(source, ['/', '.'])
+  // ProjectPath
+  projectPath = trim(projectPath, ['/', '.'])
```
lraucy, line 118: This will remove `/../` and this will transform `/.hidden` into `hidden`. Is that what we want?

src/commands/sourcemaps/git.ts
```diff
@@ -0,0 +1,242 @@
...
+}
+
+// Looks up the source paths declared in the sourcemap and try to match each of them
+// with a tracked file. The list of matching tracked files is returned.
+export const filterTrackedFiles = async (
+  stdout: Writable,
+  srcmapPath: string,
+  projectPath: string,
+  trackedFiles: Map<string, string>
+): Promise<string[] | undefined> => {
```
lraucy, line 216: How do we match tracked files with original source files in the source maps? Both are list, but they can be of different sizes, depending on the number of files that are matched, right?

src/commands/sourcemaps/git.ts
```diff
@@ -0,0 +1,242 @@
...
+    return undefined
+  }
+
+  // Only keep tracked files that match sources inside the sourcemap.
+  const filteredTrackedFiles: string[] = new Array()
+  let projectPathFoundInAllSources = false
+  for (const source of sources) {
+    const [cleanedupSource, projectPathFound] = cleanupSource(source, projectPath)
+    if (projectPathFound) {
+      projectPathFoundInAllSources = true
```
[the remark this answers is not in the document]
jybp: Yes the name of the variable is just wrong. Changed it 👍

src/commands/sourcemaps/git.ts
jybp: None :)

src/commands/sourcemaps/git.ts
jybp: Good point. The output if ran is a subfolder would be different. Changed it to make sure the baseDir is set to the root of the repo.

src/commands/sourcemaps/git.ts
jybp: Yes they can be of different size. The idea I had in mind was to have a similar logic to match on the backend-side.

lraucy: What do you think about using a different data structure that allows us to skip matching? Gaving two places where we do matching seems error-prone. One way could be to represents files with no matches by empty strings, and consider the original source files array and this array always have the same size?

src/commands/sourcemaps/git.ts
jybp: Probably not. I updated the code and added some tests to cover those edge cases.

src/commands/sourcemaps/git.ts
```diff
@@ -0,0 +1,242 @@
...
+// file path.
+//
+// For example the following tracked file path:
+// project/folder1/folder2/src.js
+//
+// Will be matched by any of:
+// project/folder1/folder2/src.js
+// folder1/folder2/src.js
+// folder2/src.js
+// src.js
```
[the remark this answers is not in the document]
jybp: Yes it could allow some false positives because of the effort to create a match whenever possible. In the example you described `src.js` would indeed be resolved to the `lib/src.js` tracked file.

What you are saying is that we should only have an entry `src.js` => `lib/src.js` if the sourcemap we're currently uploading has `lib` as a prefix for all of the 'sources'?

To boil the issue down to two cases:

- `webpack:///./project/folder/src.js` with `folder/src.js` as tracked file => 
    To have a proper match we ask the user to specify in the `--project-path` flag the `project` value.

- `webpack:///./folder/src.js` with `project/folder/src.js` as tracked file (this is the monorepo case for example) =>
    To have a proper match we don't ask anything but try the `trackedFilesMap` method of removing folders from the tracked files paths.

Assuming the second case only ever happens in a monorepo-like case, then indeed all sources of a sourcemap should have the same prefix to remove. It's tricky tbh.

We could also require the user to provide yet another flag to cover the second case and therefore have zero missing information and just do perfect matches. wdyt?

src/commands/sourcemaps/git.ts
jybp: I agree. The starting point of all this was the idea that we'd have the full list of tracked files available on the backend and do matches there. Allowing to apply any update to the logic to everyone instantaneously. 
Then came the idea of only sending only the strict minimal amount of tracked files since most of the tracked files (assets, ...) would be useless and sourcemaps only reference a small subset.
I'll have to think about it more thoroughly, I'll come up with something.

src/commands/sourcemaps/git.ts
jybp: It is now covered in e832d79 cc @lraucy

src/commands/sourcemaps/git.ts
jybp: It is now covered in e832d79 cc @lraucy

src/commands/sourcemaps/git.ts
jybp: It is now covered in e832d79 cc @lraucy 
As we talked about, we're now just trying to filter out unrelated tracked files.

src/commands/sourcemaps/git.ts
jybp: This should not be relevant anymore with the changes in e832d79 cc @lraucy

jybp:
I added two commits to the PR
e832d79 to address @lraucy 's reviews and change a bit the behavior of the tracked files filtering.

bcd124b to send the datadog-ci version inside the HTTP request and as a tag in the metrics send by the upload command. 
I only found that way of doing it. It's the same as in https://github.com/DataDog/datadog-ci/blob/jb.pinalie/git-support/src/index.ts#L17:L17

lraucy (approved):
Only one tiny comment about what we display when git is not available, otherwise LGTM!

src/commands/sourcemaps/renderer.ts
```diff
@@ -8,6 +8,12 @@ const ICONS = {
   WARNING: chalk.bold.green('⚠️'),
 }
 
+export const renderGitError = (errorMessage: string) =>
+  chalk.red(`${ICONS.FAILED} An error occured while invoking git: ${errorMessage}\n`)
```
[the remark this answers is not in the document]
jybp: 100% agreed! Here is the improved wording I came up with:

```
⚠️ An error occured while invoking git: Error: fatal: not a git repository (or any of the parent directories): .git

Make sure the command is running within your git repository to fully leverage Datadog's git integration.
To ignore this warning use the --disable-git flag.
```

Tell me if you have other ideas.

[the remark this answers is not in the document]
lraucy: No that's great!

lraucy (approved):

jybp:
219b0c4 updates this PR so that it can be merged to master and released in the stable channel.
The feature is now disabled by default. It can be enabled through the undocumented  `--enable-git-experimental` flag.

yoannmoinet (approved):
Didn't review the code.
Just the infra and the integration.
LGTM.
Some documentation is maybe missing.
(EDIT: Just saw your comment about "undocumented flag" 😄 )

src/commands/sourcemaps/__tests__/upload.test.ts
```diff
@@ -81,6 +82,46 @@ describe('upload', () => {
...
+        minifiedUrl: 'http://example/empty.min.js',
+        projectPath: '',
+        service: 'svc',
+        sourcemapPath: 'src/commands/sourcemaps/__tests__/fixtures-empty/empty.min.js.map',
+        version: '1.2.3',
+      })
+      await command['addRepositoryDataToPayloads'](payloads)
+      expect(payloads[0].gitRepositoryURL).toBeDefined()
+      expect(payloads[0].gitCommitSha).toHaveLength(40)
+      expect(payloads[0].gitRepositoryPayload).toBeUndefined()
```
lraucy, line 103: Reading this test I don't understand exactly what's going on (not super familiar with this test codebase, that might also be a reason).

Could you add some doc to describe why we expect different results in both cases?
What do we expect to be in `getRepositoryURL`, `gitCommitSha`? Is it data from the current repository?

lraucy (commented):
One comment about documentation we should address before merging imo, otherwise lgtm!

src/commands/sourcemaps/README.md
```diff
@@ -44,9 +44,9 @@ Example: if you're uploading `dist/file.js` to `https://example.com/static/file.
 
 In addition, some optional parameters are available:
 
-* `--project-path` (default: empty): the path of the project on the filesystem where the sourcemaps were built. This will be stripped off from the file name in the displayed stack traces.
 * `--concurrency` (default: `20`): number of concurrent upload to the API.
 * `--dry-run` (default: `false`): it will run the command without the final step of upload. All other checks are performed.
+* `--project-path` (default: empty): the path of the project where the sourcemaps were built. This will be stripped off from sources paths referenced in the sourcemap so they can be properly matched against tracked files paths.
```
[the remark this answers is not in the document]
jybp: My bad, I forgot that flag documentation was updated. Reverting to the original wording for now.

src/commands/sourcemaps/__tests__/git.test.ts
```diff
@@ -0,0 +1,172 @@
...
+
+      expect(stripCredentials(input)).toBe('https://gitlab.com/user/project.git')
+    })
+  })
+
+  describe('TrackedFilesMatcher', () => {
+    describe('related cases', () => {
+      test('related path', () => {
+        const sources = ['webpack:///./src/file.ts']
+        const trackedFiles = ['src/file.ts']
```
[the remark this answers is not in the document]
jybp: I don't think it will. I will plan a PR to support it, thanks for mentioning it.

src/commands/sourcemaps/__tests__/upload.test.ts
jybp: Yes exactly. They are tests that run against the current repository with two kind of sourcemaps: an empty one and one which matches a tracked filename. The repositoryURL / commit is whatever the values are for the current repository when tests run.
I've added some comments to make it clearer.

src/commands/sourcemaps/README.md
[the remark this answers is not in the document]
jybp: Yes for now let's leave it undocumented.

lraucy (approved):
LGTM!

PR #147, opened 2021-01-18
[Synthetics][SW-797] Fix config override for tests run by public ID

lefebvree:
### What and why?

When triggering tests with `--public-id`, the global config is ignored and tests are run with their original configuration.

### How?

Build `publicIdsTriggers` with global override configuration as config for each test instead of `{}`.

### Review checklist


m-rousse (approved):
LGTM

PR #150, opened 2021-01-25
[Synthetics][SW-843] Update test trigger payload for CI results batches

lefebvree:
### What and why?

[SW-843](https://datadoghq.atlassian.net/browse/SW-843)

Update test trigger payload with necessary info for CI results batches.

### How?

- Add `pollingTimeout` to forwarded config overrides.
- Remove `metadata` from all tests' payload and only send it in payload root.
- Always send strictest `executionRule` in test config, stop overriding it in test `options.ci`.
- Send `skipped` tests (will now be skipped in backend only) evenfor batches of skipped-only tests.

Also appends `(non-blocking)` to non blocking tests trigger message.

### Review checklist

### DIFF
--- a/src/commands/synthetics/__tests__/utils.test.ts
+++ b/src/commands/synthetics/__tests__/utils.test.ts
@@ -46,6 +46,11 @@ describe('utils', () => {
   describe('runTest', () => {
     const processWrite = process.stdout.write.bind(process.stdout)
     const fakeTest = {
+      config: {
+        request: {
+          url: 'http://example.org/',
+        },
+      },
       name: 'Fake Test',
       public_id: '123-456-789',
     }
--- a/src/commands/synthetics/__tests__/utils.test.ts
+++ b/src/commands/synthetics/__tests__/utils.test.ts
@@ -105,15 +105,11 @@ describe('utils', () => {
       expect(hasThrown).toBeTruthy()
     })
 
-    test('skipped tests should not be run', async () => {
-      let hasThrown = false
-      try {
-        const config = {executionRule: ExecutionRule.SKIPPED}
-        await utils.runTests(api, [{id: fakeTest.public_id, config}], processWrite)
-      } catch (e) {
-        hasThrown = true
-      }
-      expect(hasThrown).toBeTruthy()
+    test("batch of skipped tests only doesn't trigger run", async () => {
+      await expect(utils.runTests(api, [], processWrite)).rejects.toEqual(new Error('No tests to trigger'))
+
+      const test = [{config: {executionRule: ExecutionRule.SKIPPED}, id: fakeTest.public_id}]
+      await expect(utils.runTests(api, test, processWrite)).rejects.toEqual(new Error('No tests to trigger'))
     })
   })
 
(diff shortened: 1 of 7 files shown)


src/commands/synthetics/__tests__/utils.test.ts
```diff
@@ -46,6 +46,11 @@ describe('utils', () => {
   describe('runTest', () => {
     const processWrite = process.stdout.write.bind(process.stdout)
     const fakeTest = {
+      config: {
+        request: {
+          url: 'http://example.org/',
+        },
+      },
```
lefebvree, line 53: This test was missing this making it [throw in `runTests`](https://github.com/DataDog/datadog-ci/blob/19836a630f094797880f6df367b119fc0125ec03/src/commands/synthetics/utils.ts#L73), this was the reason [this test was successful](https://github.com/DataDog/datadog-ci/pull/150/files#diff-f1ccd70098fb2de33a36d62bcfdefdedef3440975916ed0812c55b55fe9fe7ecL103) and not because of the expected `No test to trigger error`.

src/commands/synthetics/renderer.ts
```diff
@@ -230,18 +230,22 @@ export const renderResults = (test: Test, results: PollResult[], baseUrl: string
 }
 
 // Other rendering
-export const renderTrigger = (test: Test | undefined, testId: string, config: ConfigOverride) => {
+export const renderTrigger = (test: Test, testId: string, executionRule: ExecutionRule, config: ConfigOverride) => {
   const idDisplay = `[${chalk.bold.dim(testId)}]`
 
   const getMessage = () => {
-    if (!test) {
-      return chalk.red.bold(`Could not find test "${testId}"`)
```
lefebvree, line 238: Should already be handled here: https://github.com/DataDog/datadog-ci/blob/19836a630f094797880f6df367b119fc0125ec03/src/commands/synthetics/utils.ts#L272-L279

src/commands/synthetics/utils.ts
```diff
@@ -270,37 +274,37 @@ export const runTests = async (
       } catch (e) {
         const errorMessage = formatBackendErrors(e)
         write(`[${chalk.bold.dim(id)}] Test not found: ${errorMessage}\n`)
-      }
-
-      if (!test || config.executionRule === ExecutionRule.SKIPPED) {
-        write(`[${chalk.bold.dim(id)}] Test skipped as per test or global configuration.\n`)
```
lefebvree, line 276: Outputted in `renderTrigger`.

m-rousse (commented):
This looks good overall, I have a doubt regarding the handling of `executionRule`, for which we want to compute the `strictestRule` to decide whether we'll run the test or not, but I expected we would send the `executionRule` from the `configOverride` to the backend, wdyt?

src/commands/synthetics/interfaces.ts
```diff
@@ -178,8 +178,13 @@ export interface ConfigOverride {
...
 }
 
-export interface Payload extends ConfigOverride {
+export interface Payload {
   metadata?: Metadata
+  tests: TestPayload[]
+}
+
+export interface TestPayload extends ConfigOverride {
+  executionRule: ExecutionRule
```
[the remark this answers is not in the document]
lefebvree: Since it is needed for all `skipped` tests (due to config override or CI options) it is easier to always send it, it doesn't change anything on the backend so we can keep it like this.

m-rousse (approved):
LGTM

PR #152, opened 2021-01-28
Pre release process

yoannmoinet:
### What and why?

We have the automated release process.
Let's have a pre-release automated process as well.

### How?

Adding a new workflow for pre-released that will publish the package on a different channel on NPM.

### Review checklist


.github/workflows/release.yml
```diff
@@ -1,7 +1,7 @@
 name: Publish package on NPM
 on:
   release:
-    types: [published]
+    types: [released]
```
yoannmoinet, line 4: This is updated because the `published` event is trigger for both release and pre-release.
Using `released` instead will ensure we only target releases and not pre-releases.

.github/workflows/prerelease.yml
```diff
@@ -0,0 +1,24 @@
...
+          node-version: '12.x'
+      - run: |
+          # Isolate the tag and exit if none found
+          IFS='-' read -ra ARR_TAG <<< $TAG
+          if ((${#ARR_TAG[@]} <= 1)); then echo "Missing channel in given tag '$TAG'. Should be in the format '0.0.0-channel'" && exit 1; fi
+          TAG_TO_PUBLISH=${ARR_TAG[${#ARR_TAG[@]} - 1]}
+
+          # Publish on npm in the correct channel
+          echo "Publishing on $TAG_TO_PUBLISH channel."
+          yarn publish --tag=$TAG_TO_PUBLISH
```
m-rousse, line 21: In the release workflow we have `yarn` and `yarn pack` as well, are they not necessary with yarn publish?
If yes, could you remove them in the release workflow, else add them in this workflow?

yoannmoinet: I'll check what's necessary.
And update accordingly.
Good catch!

yoannmoinet: Ok, `yarn pack` is included in `yarn publish` I'll update both workflows.

m-rousse (approved):
:+1:

.github/workflows/prerelease.yml
arcanis: Note that, even if fairly widespread in the ecosystem, I wouldn't recommend using dist-tags to mark whether a package is stable or not. The reason for that is that semver ranges don't care about tags, only about versions - so for instance, if you have `latest` set to `2.0.0` and `beta` set to `2.1.0`, then `^2.0.0` may resolve to `2.1.0`, even if it's still beta.

The cleaner approach for now would be to publish the beta on a different package name, then use package renaming (`"foo": "npm:foo-beta@^1.0.0"`) to explicitly require them.

Still, good enough for now, just to keep in mind in case you see weird behaviours.

arcanis: (Oh and as long as you add the proper `-alpha` / `-beta` suffixes to the versions, then things should be mostly fine as well 👍)

yoannmoinet: We're naming channel versions after their channel name.
Like `0.10.2` on the alpha channel is actually version `0.10.2-alpha`.
Even with this, you may end up with the alpha version if you had `^0.10.0`?

PR #153, opened 2021-02-01
Git support: improve performance

jybp:
### What and why?

Improve the performance of the sourcemap upload command with git support.

- Replace the nested loop with one loop + hash map lookups
- Encapsulate the whole tracked files matching logic into a single class for more clarity
- Use `fs.readFileSync` directly instead of "promisifying" `fs.readFile`
- Change the output of successful uploads to also include the time spent building payloads

### How?

Leverage hash maps to reduce the time complexity of the algorithm.

### Review checklist

### DIFF
--- a/src/commands/sourcemaps/git.ts
+++ b/src/commands/sourcemaps/git.ts
@@ -156,13 +156,13 @@ export class TrackedFilesMatcher {
 
   public matchSources(sources: string[]): string[] {
     let filtered: string[] = new Array()
-    const filenameAlreadyMatched = new Map<string, boolean>()
+    const filenameAlreadyMatched = new Set<string>()
     for (const source of sources) {
       const filename = this.getFilename(source)
       if (filenameAlreadyMatched.has(filename)) {
         continue
       }
-      filenameAlreadyMatched.set(filename, true)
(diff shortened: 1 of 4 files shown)


yoannmoinet (commented):
I'm not sure how to review this 😄 

I don't think I have enough background to fully review it thoroughly, so I'll let your team do a more valuable review and simply give you a light LGTM.

Were you able to compare this with the previous implementation?
What kind of number did you get from this refactor?

src/commands/sourcemaps/git.ts
```diff
@@ -133,47 +105,96 @@ export const getRepositoryData = async (
...
+
+    return filtered
   }
 
-  // Only keep tracked files that may be related to the sources declared in the sourcemap
-  const filtered: string[] = new Array()
-  for (const trackedFile of trackedFiles) {
+  public matchSources(sources: string[]): string[] {
+    let filtered: string[] = new Array()
+    const filenameAlreadyMatched = new Map<string, boolean>()
```
lraucy, line 159: nit: is there a reason to use a map instead of a set here?

lraucy (approved):
LGTM!

PR #156, opened 2021-02-04
[synthetics] do not parse url if type or subtype is not correct

KevinYeramian:
### What and why?

Parse `startUrl` for `browser` and `http` tests

### DIFF
--- a/src/commands/synthetics/utils.ts
+++ b/src/commands/synthetics/utils.ts
@@ -70,9 +70,8 @@ export const handleConfig = (
     ]),
   }
 
-  const context = parseUrlVariables(test.config.request.url, write)
-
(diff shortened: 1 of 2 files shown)


src/commands/synthetics/utils.ts
```diff
@@ -72,7 +72,7 @@ export const handleConfig = (
 
   const context = parseUrlVariables(test.config.request.url, write)
 
-  if (config.startUrl) {
+  if ((test.type === 'browser' || test.subtype === 'http') && config.startUrl) {
     handledConfig.startUrl = template(config.startUrl, context)
   }
```
m-rousse, line 77: The `context` variable can be computed in the condition body as well

m-rousse (approved):

PR #157, opened 2021-02-08
Allow additional extensions to be uploaded

jybp:
### What and why?

Upload any file ending with `js.map` to accommodate for more extensions.

### How?

### Review checklist

### DIFF
--- a/src/commands/sourcemaps/upload.ts
+++ b/src/commands/sourcemaps/upload.ts
@@ -114,7 +114,7 @@ export class UploadCommand extends Command {
   }
 
   private getMatchingSourcemapFiles(): Payload[] {
-    const sourcemapFiles = glob.sync(buildPath(this.basePath!, '**/*.js.map'))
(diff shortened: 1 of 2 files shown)


src/commands/sourcemaps/upload.ts
```diff
@@ -114,7 +114,7 @@ export class UploadCommand extends Command {
   }
 
   private getMatchingSourcemapFiles(): Payload[] {
-    const sourcemapFiles = glob.sync(buildPath(this.basePath!, '**/*.js.map'))
+    const sourcemapFiles = glob.sync(buildPath(this.basePath!, '**/*js.map'))
```
lraucy, line 117: Looks like we don't have matching tests... should we add some?

jybp: I added a test case in `upload.test.ts` that covers it. 👍

lraucy (approved):

PR #158, opened 2021-02-11
Use latest datatog-metrics instead of Git dependency

rpelliard:
### What and why?

Uses [datadog-metrics 0.9.0](https://www.npmjs.com/package/datadog-metrics/v/0.9.0) instead of Datadog's fork. https://github.com/dbader/node-datadog-metrics/pull/50 has been merged, including the changes that we wanted. There is no need to have a dependency though git anymore, which is causing security concerns for some customers.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
     "axios": "0.21.1",
     "chalk": "2.4.2",
     "clipanion": "2.2.2",
-    "datadog-metrics": "https://github.com/DataDog/node-datadog-metrics#12d16a80ea2a8846c91d80f5e127fb7b81b9f347",
+    "datadog-metrics": "0.9.0",
     "dd-trace": "0.30.1",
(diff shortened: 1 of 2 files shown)


yoannmoinet (commented):
You'll have to run `yarn install` in the repo and commit the changes.

yoannmoinet:
@m-rousse maybe we should use `yarn install --immutable` here:
https://github.com/DataDog/datadog-ci/blob/2ed8ffbb495c4354f9179a3ba29c62f713180deb/.github/workflows/ci.yml#L31

So we don't merge dependency updates without the lock updates.
What do you think?

m-rousse:
Agreed -> https://github.com/DataDog/datadog-ci/pull/159

yoannmoinet (approved):
LGTM

PR #169, opened 2021-02-19
[synthetics] add support for connection tunneling

m-rousse:
### What and why?

In order to ease testing in local environments a tunneling feature has been added to the synthetics command available through the `--tunnel` flag with the `synthetics run-tests` command.

### How?

When triggering tests, we issue a request to datadog to open a tunnel between datadog-ci and Datadog's synthetics workers so that local resources are accessible to the workers performing the tests.

### Review checklist


m-rousse:
Note: This PR was already reviewed in internal repository when first created (cf. https://github.com/DataDog/datadog-ci-tunnel)

package.json
```diff
@@ -40,28 +40,36 @@
     "async-retry": "1.3.1",
     "aws-sdk": "2.682.0",
     "axios": "0.21.1",
-    "chalk": "2.4.2",
+    "chalk": "^3.0.0",
```
yoannmoinet, line 43: Can you use fixed version please?

src/commands/synthetics/run-test.ts
```diff
@@ -19,27 +20,52 @@ export class RunTestCommand extends Command {
...
 
     if (!testsToTrigger.length) {
       this.context.stdout.write('No test suites to run.\n')
 
       return 0
     }
 
-    const {tests, triggers} = await runTests(api, testsToTrigger, this.context.stdout.write.bind(this.context.stdout))
+    let tunnel: Tunnel | undefined
+    if (this.config.tunnel || this.shouldOpenTunnel) {
```
yoannmoinet, line 48: Don't we want one of these setting superseding the other?

Like if I have the tunnel setting in my config, but I decide to overwrite it with the flag for instance, it would always work as long as one of the two is truthy.

yoannmoinet (approved):
LGTM, just some nits.

src/commands/synthetics/run-test.ts
m-rousse: Agreed - I updated it to have the CLI flag superseeding the configuration.

lefebvree (approved):

PR #171, opened 2021-03-02
Mrousse/sw 910/add proxy support tunnel

m-rousse:
### What and why?

The tunnel uses a websocket connection to encapsulate streams to datadog's infrastructure. This PR updates the tunnel to use the proxy defined by the proxy configuration of datadog-ci if there is one.

### How?

It passes the proxy options to the WebSocketWithReconnection object which then creates an agent and feeds it to the `ws` library. As `proxy-agent` typings are not up-to-date compared to latest nodes one, there is a `as unknown as Agent` cast before passing it to `ws`.

### Review checklist

### DIFF
--- a/src/commands/synthetics/websocket.ts
+++ b/src/commands/synthetics/websocket.ts
@@ -1,7 +1,11 @@
 import {EventEmitter, once} from 'events'
+import type {Agent} from 'http'
 import {Writable} from 'stream'
 
+import ProxyAgent from 'proxy-agent'
 import WebSocket, {createWebSocketStream} from 'ws'
+
+import {ProxyConfiguration} from '../../helpers/utils'
 
 /**
  * TODO: test websocket class:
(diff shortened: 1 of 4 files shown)


src/commands/synthetics/websocket.ts
```diff
@@ -104,7 +109,11 @@ export class WebSocketWithReconnect extends EventEmitter {
   private establishWebsocketConnection(resolve: (value: void) => void, reject: (error: Error) => void) {
     if (!this.websocket) {
       this.reconnectRetries++
-      this.websocket = new WebSocket(this.url)
+      const options: WebSocket.ClientOptions = {}
+      if (this.proxyOpts.host && this.proxyOpts.port) {
```
KevinYeramian, line 113: Should you also check `protocol` ?

KevinYeramian (approved):

PR #188, opened 2021-03-24
[synthetics] improve tunnel connection and bump yamux-js

m-rousse:
### What and why?

This PR does:
- bump `yamux-js` to version `0.0.4` to benefit from latest bug fixes preventing performing tunneled tests with files >100KB
- modify the websocket handler to be able to obtain the first message sent over the connection. Before this PR, we had a race-condition where the first message was already sent but the client did not register the event listener yet and the `once` (cf. [code](https://github.com/DataDog/datadog-ci/blob/9eae78e9d07ffca6c4a51145486a0e156455137d/src/commands/synthetics/tunnel.ts#L226)) callback was never called.

### DIFF
--- a/src/commands/synthetics/tunnel.ts
+++ b/src/commands/synthetics/tunnel.ts
@@ -6,6 +6,7 @@ import {Duplex, Writable} from 'stream'
 import chalk from 'chalk'
 import {AuthContext, Connection as SSHConnection, Server as SSHServer, ServerChannel as SSHServerChannel} from 'ssh2'
 import {ParsedKey, SSH2Stream, SSH2StreamConfig} from 'ssh2-streams'
+import WebSocket from 'ws'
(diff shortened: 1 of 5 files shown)


src/commands/synthetics/tunnel.ts
```diff
@@ -223,7 +230,7 @@ export class Tunnel {
   }
 
   private async getConnectionInfo() {
-    const [rawConnectionInfo] = (await once(this.ws!, 'message')) as (string | Buffer)[]
+    const rawConnectionInfo = await this.firstMessagePromise
```
sdeprez, line 233: Rather than storing the promise here, can't you store in `this.ws` and expose a `await this.ws.waitForFirstMessage`?

yoannmoinet (approved):
infra LGTM.

lefebvree (approved):

sdeprez (approved):

PR #195, opened 2021-04-01
[synthetics] remove websocket reconnection + add debug logs

m-rousse:
### What and why?

Currently the websocket reconnection is not supported in the worker, this PR removes it until properly supported in the product.
Also add error logging for the websocket connection, the multiplexer (yamux) and finally display the tunnel identifier when connecting to it.


lefebvree (approved):
LGTM

PR #205, opened 2021-04-15
[synthetics] Add --files parameter

yoannmoinet:
### What and why?

I need to run multiple suites in parallel, so we can reduce the number of tests per run.
But I still want to use a global configuration file.

### How?

Add a `--files`, `-f` parameter so you can override the glob pattern used by the global configuration file.
This way, I can run multiple suites using the same configuration on each run.

I also renamed the function `getTestsToTrigger` in the class, it was quite confusing to have two different functions using the same name in the same file.

### Review checklist

### DIFF
--- a/src/commands/synthetics/run-test.ts
+++ b/src/commands/synthetics/run-test.ts
@@ -35,7 +35,7 @@ export class RunTestCommand extends Command {
 
     const api = this.getApiHelper()
     const publicIdsFromCli = this.publicIds.map((id) => ({config: this.config.global, id}))
-    const testsToTrigger = publicIdsFromCli.length ? publicIdsFromCli : await this.getTestsToTrigger(api)
+    const testsToTrigger = publicIdsFromCli.length ? publicIdsFromCli : await this.getTestsList(api)
 
     if (!testsToTrigger.length) {
       this.context.stdout.write('No test suites to run.\n')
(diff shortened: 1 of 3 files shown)


src/commands/synthetics/run-test.ts
```diff
@@ -174,13 +175,19 @@ export class RunTestCommand extends Command {
...
-  private async getTestsToTrigger(api: APIHelper) {
+  private async getTestsList(api: APIHelper) {
     if (this.testSearchQuery) {
       const testSearchResults = await api.searchTests(this.testSearchQuery)
 
       return testSearchResults.tests.map((test) => ({config: this.config.global, id: test.public_id}))
     }
-    const suites = (await getSuites(this.config.files, this.context.stdout.write.bind(this.context.stdout)))
+
+    const files = this.files || [this.config.files]
```
m-rousse, line 185: The `files` variable actually is a list of globs (the naming in the config may be confusing in the first place), could we rename it to `fileGlobs` so it is more explicit it is a list of globs?

m-rousse (approved):
LGTM + naming nit

src/commands/synthetics/run-test.ts
yoannmoinet: I'll rename the variable internally.
But I'll keep the one exposed as `files`, to keep the consistency with the configuration file.

PR #213, opened 2021-04-21
Enable git support by default in the alpha channel

jybp:
### What and why?

- Merge master in the alpha channel
- Keep git support enabled by default in the alpha channel


lraucy:
Isn't the diff a bit weird?

jybp:
It's because of the 4bef0f4 commit I added after the merge. 

This PR ensures the alpha branch will remain with the same behavior as today (`--disable-git`).
Otherwise someone might merge master to alpha and override the behavior by mistake.

lraucy (approved):

PR #214, opened 2021-04-23
Display more information in error messages when api key is faulty.

lraucy:
### What and why?

Right now error messages caused by wrong API keys are not super helpful: a raw axios exception is logged to stdout and nothing really tells you what's going on.

This PR aims to fix this by:
- displaying a meaningful message when it's clear the API key is faulty
- trying to validate the API key using [this API](https://docs.datadoghq.com/api/latest/authentication/) when the API key could be faulty, and display a meaningful message otherwise

Additionally, errors 400 are possibly related to a single sourcemap upload, not all of them, so this PR also changes the behavior when a 400 is returned: the failed upload is not retried but other uploads continue (previous behavior was to stop the CLI entirely).

Here are a few screenshots of CLI output in different cases:

Successful upload:
Screen Shot 2021-04-27 at 11 23 35

Multiple causes of failure/skipped payloads:
Screen Shot 2021-04-27 at 11 22 22

Other example:
Screen Shot 2021-04-27 at 11 20 56

Everything failed because of an invalid API key:
Screen Shot 2021-04-26 at 21 04 32

### How?

### Review checklist


lraucy:
I'm not so satisfied by the CLI output: the information we need is displayed, but the way it is outputted is not super clear.

src/commands/sourcemaps/renderer.ts
```diff
@@ -32,8 +34,45 @@ export const renderRetriedUpload = (payload: Payload, errorMessage: string, atte
...
+  }
+  if (results.get(UploadStatus.Failure)) {
+    output.push(chalk.red(`${ICONS.FAILED} ${results.get(UploadStatus.Failure)} files failed to upload.`))
+  }
+  if (results.get(UploadStatus.Skipped)) {
+    output.push(chalk.yellow(`${ICONS.WARNING}  ${results.get(UploadStatus.Skipped)} files were ignored.`))
+  }
+
+  return output.join('\n') + '\n'
+}
```
yoannmoinet, line 75: **Super NIT**: you could pluralize outputs with an helper function to simplify the process, I think it would make the output more readable.

```js
const pluralize = (nb, singular, plural) => {
  if (nb >= 2) {
    return `${nb} ${plural};
  }
  return `${nb} ${singular}`;
} 
```

And then use it like:
```js
output.push(chalk.yellow(`${ICONS.WARNING}  ${pluralize(results.get(UploadStatus.Skipped), 'file was', 'files were')} ignored.`))
```

yoannmoinet (approved):
LGTM.

> I'm not so satisfied by the CLI output: the information we need is displayed, but the way it is outputted is not super clear.

Is it still the case after the PR?
If so, what I'd do would be to treat that output as a report, with more spaces and new lines.
Titleing it a `Upload report` even.
It would make the information more readable and understandable I think.
Having everything on single lines makes it difficult to read it quickly.

src/commands/sourcemaps/renderer.ts
lraucy: ah nice! Will do, it will look much better, thanks!

m-rousse (approved):
This looks really nice, we may include it in the synthetics command as well.

src/commands/sourcemaps/renderer.ts
```diff
@@ -32,8 +35,67 @@ export const renderRetriedUpload = (payload: Payload, errorMessage: string, atte
...
+  statuses.forEach((status) => {
+    if (!results.has(status)) {
+      results.set(status, 0)
+    }
+    results.set(status, results.get(status)! + 1)
+  })
+
+  const output = ['', chalk.bold('Command summary:')]
+  if (results.get(UploadStatus.Failure)) {
+    output.push(chalk.red(`${ICONS.FAILED} All sourcemaps have not been uploaded correctly.`))
```
rpelliard, line 49: Small nit: I find `Some sourcemaps have not been uploaded correctly` a bit clearer and less cumbersome that `All sourcemaps have not..` - WDYT ?

rpelliard (approved):
LGTM !

jybp (approved):
Looks much better!

src/commands/sourcemaps/renderer.ts
lraucy: Indeed, I updated this using your sentence, thank you!

PR #215, opened 2021-04-28
[s8s] refactor reporter mechanism (prepare for jUnit reporter)

yoannmoinet:
### What and why?

We'd like to add a jUnit reporter to the run-tests command so it would export a jUnit file at the end of the run.

I'm opening the PR so we can discuss it if needed.

### How?

I've refactored the way we had to log so it's more extendable with custom reporters.

We now have a `Reporter` interface that is being used across the command lifecycle to report what it's doing and what is happening.

```javascript
export interface Reporter {
// Only used in parseUrlVariables() to log errors
  renderError?(error: string): void
// Used right before the trigger, will rename.
  renderGlobalErrors?(errors: string[]): void
// Renders the header at the beginning of the run.
  renderHeader?(timings: {startTime: number}): void
// Used across the app to log stuff.
  renderLog?(log: string): void
// When we get results of a test.
  renderResults?(test: Test, results: PollResult[], baseUrl: string, locationNames: LocationsMapping): void
// At the end of the run.
  renderSummary?(summary: Summary): void
// Whenever a test is triggered.
  renderTrigger?(test: Test, testId: string, executionRule: ExecutionRule, config: ConfigOverride): void
// After the trigger.
  renderWait?(test: Test): void
}
```

### Review checklist

- [x] Add testing
- [x] Add documentation
- [x] Make it work with our integration


m-rousse (commented):
I'm OK with this update, I'd argue to:
- have Writer being defined from Reporter (or the other way around, using `Partial` or `Required`)
- use the same method names for Writer and Reporter
- rename Writer (we used `Writable` as it is a property of the logger we are using, but the `Writer` class has more duties here as it handles various data from the synthetics tests run). I suggest Writer would become Reporter and Reporter would become Transport. (the Reporter gets the data of the different steps and passes it to the various transports that prints/export the data in the proper format)

src/commands/synthetics/reporters/logs.ts
```diff
@@ -1,4 +1,5 @@
 import chalk from 'chalk'
+import {Writable} from 'node:stream'
```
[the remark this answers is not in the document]
yoannmoinet: oh, good catch, that's a bad auto-import, will update.

src/commands/synthetics/interfaces.ts
```diff
@@ -11,6 +11,28 @@ interface Timings {
   total: number
 }
 
+export interface Reporter {
```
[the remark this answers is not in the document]
yoannmoinet: agreed, will update this part.

src/commands/synthetics/README.md
```diff
@@ -213,3 +214,22 @@ Took 11546ms
...
+| :------------ | :-------------------------------------------------------------------------------------- | :-------------------------------------------------------------- |
+| `error`       | `(error: string)`                                                                       | called whenever an error occurs.                                |
+| `initErrors`  | `(errors: string[])`                                                                    | called whenever an error occurs during the tests parsing phase. |
+| `log`         | `(log: string)`                                                                         | called for logging.                                             |
+| `runEnd`      | `(summary: Summary)`                                                                    | called at the end of the run.                                   |
+| `start`       | `(timings: {startTime: number})`                                                        | called at the begining of the run.                              |
```
m-rousse, line 232: It seems to be called at the beginning of the report, could you update the description so it is clearer? (or rename it to `reportStart`?)

src/commands/synthetics/reporters/default.ts
```diff
@@ -233,73 +236,83 @@ const getTestResultColor = (success: boolean, isNonBlocking: boolean) => {
...
-        return `>> Skipped test "${chalk.yellow.dim(test.name)}"`
-      } else {
-        return `>> Skipped test "${chalk.yellow.dim(test.name)}" because of execution rule configuration in Datadog`
-      }
+export class DefaultReporter implements Reporter {
+  private write: Writable['write']
+  constructor(command: RunTestCommand) {
+    this.write = command.context.stdout.write.bind(command.context.stdout)
+  }
+  public error(error: string) {
```
m-rousse, line 244: Nit: could you add blank lines before each function definition?

m-rousse (approved):
LGTM :+1:

src/commands/synthetics/reporters/default.ts
yoannmoinet: this should be a lint rule.

yoannmoinet: added the new rule and fixed it.

yoannmoinet:
Confirmed that the integration in our CI is working exactly the same, with the exact same output as before.
Merging this now.

PR #217, opened 2021-04-29
0.12.0-alpha.0

juan-fernandez:
### What and why?

Merge master into alpha


yoannmoinet (approved):
LGTM.
Just so we're sure, the release has to be called `0.12.0-alpha` and NOT `0.12.0-alpha.0`

PR #223, opened 2021-05-03
[sourcemaps] Display meaningful errors when files are empty

lraucy:
### What and why?

Current code tries to send them to the API, and receives an error 400 without much details. This PR provides prior validation and display better output.

The code has been tested using mac OS X and Windows.

### Review checklist


src/commands/sourcemaps/upload.ts
```diff
@@ -238,11 +238,20 @@ export class UploadCommand extends Command {
...
-      )
-      metricsLogger.increment('skipped_missing_js', 1)
+    try {
+      validatePayload(sourcemap)
+    } catch (error) {
+      if (error instanceof InvalidPayload) {
+        this.context.stdout.write(renderFailedUpload(sourcemap, error.message))
+        metricsLogger.increment(error.reason, 1)
+      } else {
+        this.context.stdout.write(
```
jybp, line 248: Do we want to add something like `metricsLogger.increment("unknown", 1)` in the else case?

jybp (approved):
LGTM

rpelliard (approved):

src/commands/sourcemaps/upload.ts
lraucy: Agreed, I pushed another commit with better metrics.

PR #225, opened 2021-05-05
[sourcemaps] Fix metrics reporting

lraucy:
### What and why?

This PR has several goals:
- use the `DATADOG_SITE` config to decide which Datadog API to report metrics to
- display a warning message when those metrics cannot be reported to Datadog

Additionally I moved the axios and datadog-metrics mocks to `src/`, since they were considered for all tests anyway (as far as my understanding goes).

Screen Shot 2021-05-05 at 18 53 38


lraucy:
@piotr-oles some of these changes might interest you

src/__mocks__/datadog-metrics.js
```diff
@@ -3,7 +3,11 @@ const metrics = require('datadog-metrics')
 const BufferedMetricsLogger = metrics.BufferedMetricsLogger
 
 function NullReporter() {}
-NullReporter.prototype.report = jest.fn()
```
m-rousse, line 6: Why is it needed to move the `__mocks__` folder to the `src/` folder?
(looking at [jest doc](https://jestjs.io/docs/manual-mocks) it seems it should be next to the file using the module, but I may have missed some details)

lraucy: My understanding is that for dependencies it should be at the project root:

> If the module you are mocking is a Node module (e.g.: lodash), the mock should be placed in the __mocks__ directory adjacent to node_modules (unless you configured roots to point to a folder other than the project root) and will be automatically mocked. There's no need to explicitly call jest.mock('module_name').

During my tests, I noticed that the `src/commands/dependencies/__mocks__/datadog-metrics.js` was used both in `dependencies` test but also in `sourcemaps` test, which was unexpected to me. In both cases, the mock seems to be applied to all tests, but at least if it is in `src/__mocks__` it's transparent that it will be used everywhere.

I only have a very basic understanding of this though.

m-rousse: Ok, after re-reading the doc I understand the same behavior so it LGTM, thanks for this change.

m-rousse (approved):
Good for infra

jybp (approved):

piotr-oles (approved):

PR #227, opened 2021-05-06
Handle sourcemaps containing backslashes in paths

jybp:
### What and why?

Properly handles sourcemaps that contain paths with backslashes (windows).

### How?

### Review checklist

### DIFF
--- a/src/commands/sourcemaps/git.ts
+++ b/src/commands/sourcemaps/git.ts
@@ -182,6 +182,10 @@ export class TrackedFilesMatcher {
   // Example: webpack:///./src/folder/ui/select.vue?821e
   private getFilename(s: string): string {
     let start = s.lastIndexOf('/')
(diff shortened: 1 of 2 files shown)


src/commands/sourcemaps/git.ts
```diff
@@ -182,6 +182,10 @@ export class TrackedFilesMatcher {
   // Example: webpack:///./src/folder/ui/select.vue?821e
   private getFilename(s: string): string {
     let start = s.lastIndexOf('/')
+    const backslashStart = s.lastIndexOf('\\')
```
lraucy, line 185: What if we have a unix file with a backslash in its name?

jybp: Those particular files won't be reported.  Not sure there are better ways to handle that case.
We could use the current env separator but the paths in the sourcemaps 'sources' could still use any separator.

PR #228, opened 2021-05-06
[dependencies] metrics flushing fix.

lraucy:
### What and why?

Port changes from https://github.com/DataDog/datadog-ci/pull/225 to dependencies command.


piotr-oles (approved):
Thanks!

PR #230, opened 2021-05-07
[dependencies] upgrade proxy agent to version 4.0.1

m-rousse:
### What and why?

Goal of this PR is to update the `netmask` nested dependency to fix a critical vulnerability.
It does so by updating the `proxy-agent` dependency from major 3.x to major 4.x. It should not introduce breaking changes from our perspective, it upgrades the underlying dependency `agent-base` from v4 to v6 (cf. https://github.com/TooTallNate/node-proxy-agent/pull/55) which itself changes the way options are passed to the proxy agent.

### Review checklist


src/commands/synthetics/websocket.ts
```diff
@@ -103,7 +102,8 @@ export class WebSocket extends EventEmitter {
     if (!this.websocket) {
       const options: WebSocketModule.ClientOptions = {}
       if (this.proxyOpts.host && this.proxyOpts.port) {
-        options.agent = (new ProxyAgent(this.proxyOpts) as unknown) as Agent // Proxy-agent typings are incomplete
+        const proxyUrl = getProxyUrl(this.proxyOpts)
+        options.agent = new ProxyAgent(proxyUrl)
```
yoannmoinet, line 106: So much better 😆

yoannmoinet (approved):
LGTM

PR #245, opened 2021-05-11
[ci-app] JUnit XML Upload – Use Safe File Names

juan-fernandez:
### What and why?

Use a function to make sure the file name we use to be stored in S3 is safe.

Also, improve rendered message when using a single file.

### How?

* Create `getSafeFileName`. 
* Render a slightly different name if the input is a single file.

### Review checklist

### DIFF
--- a/src/commands/junit/api.ts
+++ b/src/commands/junit/api.ts
@@ -10,7 +10,10 @@ import {renderUpload} from './renderer'
 
 import {CI_JOB_URL, CI_PIPELINE_URL, GIT_SHA} from '../../helpers/tags'
 
-const getSafeFileName = (unsafeFileName: string) => unsafeFileName.replace(/[^a-z0-9]/gi, '_')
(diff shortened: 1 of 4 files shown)


src/commands/junit/api.ts
```diff
@@ -10,6 +10,8 @@ import {renderUpload} from './renderer'
 
 import {CI_JOB_URL, CI_PIPELINE_URL, GIT_SHA} from '../../helpers/tags'
 
+const getSafeFileName = (unsafeFileName: string) => unsafeFileName.replace(/[^a-z0-9]/gi, '_')
```
drodriguezhdez, line 13: Can we have a test of this function? Also, it'd worth it to have a comment explaining why this change in the name is required. (Are there some requirements in the name of the file for S3)?

juan-fernandez: done

drodriguezhdez (approved):

PR #251, opened 2021-05-18
DD Lambda Extension support on `lambda instrument` command

zARODz11z:
### What and why?
This PR introduces support for the DD lambda library extension to the `datadog-ci lambda instrument` command. The DD Lambda Extension is a special version of the Datadog Agent that the Serverless & Agent-Core team are working on releasing to GA, therefore the Serverless team would like all our instrumentation integrations to have support for the DD Lambda Extension.
Summary of changes:
* Addition of DD Lambda Extension via a layer
* Addition of DD_API_KEY/DD_API_KMS_KEY environment variables on lambda configs
* Block extension & forwarder to be included at the same time
* Make sure the the lambda function config already has a DD_API_KEY/DD_KMS_API_KEY or is given an api key through the users exported environment variables. Making it so the api keys never have to be placed in a config file, they can just be exported in the users environment.
### How?
The DD Lambda Extension is added to a lambda function as a layer therefore I followed the convention of how the DD Lambda Library layers were added with a few code changes to allow the extension to be added.

### Testing
I manually tested my changes by running the datadog-ci lambda instrument command on a sample lambda function in sandbox and confirmed the layers (lambda library and extension) were added. 
These are the cases I manually tested: 
- [x] Forwarder with lambda library layer
- [x] Extension and Forwarder at same time (correctly returns a 1 exit code)
- [x] ApiKey & ApiKMS key (always added if given and not already present on lambda config)
- [x] Extension with lambda library (eg. no problem adding a node lambda library layer with extension)
- [x] Extension with DD_API_KEY & DD_KMS_API_KEY exported in environment
- [x] Extension with DD_KMS_API_KEY exported in environment
- [x] Extension with DD_API_KEY exported in environment
- [x] Extension with no apiKey given or already on lambda config(correctly throws an error)
### Review checklist


zARODz11z:
p.s. I've read the release [process](https://github.com/DataDog/datadog-ci#release-process) for the datadog-ci, i'm waiting to update the version using `yarn version` until I get an approval on my PR

zARODz11z:
Github actions are degraded ATM https://www.githubstatus.com/#

src/commands/lambda/function.ts
```diff
@@ -141,12 +162,50 @@ const calculateUpdateRequest = (
...
+  let needsLayerUpdate = false
+  if (fullLambdaLibraryLayerARN !== undefined) {
+    if (!layerARNs.includes(fullLambdaLibraryLayerARN)) {
+      needsUpdate = true
+      needsLayerUpdate = true
+      // Remove any other versions of the layer
+      layerARNs = [...layerARNs.filter((l) => !l.startsWith(lambdaLibraryLayerArn)), fullLambdaLibraryLayerARN]
+    }
+  }
+  if (fullExtensionLayerARN !== undefined) {
```
DarcyRaynerDD, line 191: nit: seems like most of this logic is duplicated from line 183. Could it be moved into a function?

DarcyRaynerDD (approved):
Nice work, LGTM overall

src/commands/lambda/function.ts
zARODz11z: yeah I agree that its crowding the function, I moved it to `determineIfLayerShouldBeAdded`

src/commands/lambda/function.ts
```diff
@@ -244,6 +233,28 @@ export const calculateUpdateRequest = (
   return needsUpdate ? updateRequest : undefined
 }
 
+const determineIfLayerShouldBeAdded = (
```
DarcyRaynerDD, line 236: The name is kind of confusing given the return type. I would expect it to return a boolean based on that name. Maybe just addLayerARN? I think you could also keep the extension/env var check in calculateUpdateRequest, since it's not related to what the function indicates it's doing.

DarcyRaynerDD (approved):
LGTM overall, had just 1 nit

src/commands/lambda/function.ts
zARODz11z: yeah I was a bit iffy on the name, thanks for pointing it out. I also moved the extension stuff back into calculateUpdateRequest

PR #261, opened 2021-06-02
[Synthetics] Tunnel: Reject SSH forwarding on socket error

lefebvree:
### What and why?

This PR denies traffic forwarding to the socket when an error occurs and it hasn't been accepted yet, this prevents hanging on DNS errors due to unresolvable hosts.

### How?

Reject forwarding on `error` or `close` when it hasn't been initiated yet and log a custom error message on unresolvable host error.

### Review checklist

### DIFF
--- a/src/commands/synthetics/tunnel.ts
+++ b/src/commands/synthetics/tunnel.ts
@@ -169,11 +169,20 @@ export class Tunnel {
           })
         })
         dest.on('error', (error) => {
-          console.error('Forwarding error', error.message)
+          if (!src) {
+            if ((error as any).code === 'ENOTFOUND') {
+              this.logError(`Unable to resolve host ${(error as any).hostname}`)
+            } else {
+              this.logError(`Forwarding channel error: "${error.message}"`)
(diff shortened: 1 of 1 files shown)


src/commands/synthetics/tunnel.ts
```diff
@@ -169,11 +169,20 @@ export class Tunnel {
           })
         })
         dest.on('error', (error) => {
-          console.error('Forwarding error', error.message)
+          if (!src) {
+            if ((error as any).code === 'ENOTFOUND') {
```
m-rousse, line 173: Nit: I'd argue to use `'code' in error && error.code === 'ENOTFOUND'` though I do not feel strongly about this

m-rousse (approved):

src/commands/synthetics/tunnel.ts
lefebvree: The typings prevented that but it seems we just needed to specify the error parameter as `NodeJS.ErrnoException`.

PR #264, opened 2021-06-02
Report commits command

jybp:
### What and why?

Introduces a new command to upload git commits to datadog.
The command will only be added to the alpha channel for now.

### How?

It uses the same overall structure as the [sourcemap upload](https://github.com/DataDog/datadog-ci/tree/master/src/commands/sourcemaps) command. It gathers git data the same way.
Instead of factoring out the `git` part between the two command I kept them both separated.

### Review checklist


src/commands/commit/README.md
```diff
@@ -0,0 +1,49 @@
...
+### Setup
+
+You need to have `DATADOG_API_KEY` in your environment.
+
+```bash
+# Environment setup
+export DATADOG_API_KEY="<API KEY>"
+```
+
+It is possible to configure the tool to use Datadog EU by defining the `DATADOG_SITE` environment variable to `datadoghq.eu`. By defaut the requests are sent to Datadog US. In order to send the usage metrics to the correct datacenter, the `DATADOG_API_HOST` must also be set if another instance of Datadog is used. For example, for Datadog EU, it should be set to `api.datadoghq.eu`.
```
m-rousse, line 16: Will the setup of sending git commits details to one site and usage metrics to another one be common?
If not, could we infer the API host from the `DATADOG_SITE` so changing site is only one variable away?

(Also it seems the `DATADOG_SOURCEMAP_INTAKE_URL` variable is mostly internal and will not be used by customers so maybe should not be documented here?)

src/commands/commit/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+It is also possible to override the full URL for the intake endpoint by defining the `DATADOG_SOURCEMAP_INTAKE_URL` environment variable.
+
+### Commands
+
+#### `upload`
+
+This command will upload the current commit details to Datadog in order to create links to your repositories inside DataDog's UI
```
m-rousse, line 24:
```suggestion
This command will upload the current commit details to Datadog in order to create links to your repositories inside DataDog's UI.
```

src/commands/commit/apikey.ts
```diff
@@ -0,0 +1,42 @@
...
+export class ApiKeyValidator {
+  public apiKey: string | undefined
+  public datadogSite: string
+
+  private isValid?: boolean
+
+  constructor(apiKey: string | undefined, datadogSite: string) {
+    this.apiKey = apiKey
+    this.datadogSite = datadogSite
+  }
```
m-rousse, line 12: Nit: Typescript has some sugar to pass arguments as attributes to a class

```suggestion
  private isValid?: boolean

  constructor(public apiKey: string | undefined, public datadogSite: string) {}
```

src/commands/commit/api.ts
```diff
@@ -0,0 +1,41 @@
...
+import {getRequestBuilder} from '../../helpers/utils'
+
+import {Payload} from './interfaces'
+import {renderUpload} from './renderer'
+
+// Dependency follows-redirects sets a default maxBodyLength of 10 MB https://github.com/follow-redirects/follow-redirects/blob/b774a77e582b97174813b3eaeb86931becba69db/index.js#L391
+// We don't want any hard limit enforced by the CLI, the backend will enforce a max size by returning 413 errors.
+const maxBodyLength = Infinity
+
+export const uploadRepository = (request: (args: AxiosRequestConfig) => AxiosPromise<AxiosResponse>) => async (
```
m-rousse, line 14: It does not seem to be imported in other files:
```suggestion
const uploadRepository = (request: (args: AxiosRequestConfig) => AxiosPromise<AxiosResponse>) => async (
```

src/commands/commit/apikey.ts
```diff
@@ -0,0 +1,42 @@
...
+    return `https://api.${this.datadogSite}/api/v1/validate`
+  }
+
+  private async validateApiKey(): Promise<boolean> {
+    try {
+      const response = await axios.get(this.getApiKeyValidationURL(), {
+        headers: {
+          'DD-API-KEY': this.apiKey,
+        },
+      })
```
m-rousse, line 32: I think this should be in `api.ts` with other calls to the API (also it avoids duplicating the API base URL crafting)

src/commands/commit/git.ts
```diff
@@ -0,0 +1,105 @@
...
+  hash: string
+  remote: string
+  trackedFiles: string[]
+}
+
+// Returns the current hash, remote URL and tracked files paths.
+export const getRepositoryData = async (
+  git: simpleGit.SimpleGit,
+  stdout: Writable,
+  repositoryURL: string | undefined
```
m-rousse, line 77:
```suggestion
  repositoryURL?: string
```

src/commands/commit/git.ts
```diff
@@ -0,0 +1,105 @@
...
+    return undefined
+  }
+
+  const data = {
+    hash,
+    remote,
+    trackedFiles,
+  }
+
+  return data
```
m-rousse, line 104:
```suggestion
  return {
    hash,
    remote,
    trackedFiles,
  }
```

src/commands/commit/git.ts
```diff
@@ -0,0 +1,105 @@
...
+    if (repositoryURL) {
+      ;[hash, trackedFiles] = await Promise.all([gitHash(git), gitTrackedFiles(git)])
+      remote = repositoryURL
+    } else {
+      ;[remote, hash, trackedFiles] = await Promise.all([gitRemote(git), gitHash(git), gitTrackedFiles(git)])
+    }
+  } catch (e) {
+    stdout.write(renderGitError(e))
+
+    return undefined
```
m-rousse, line 95:
```suggestion
    return
```

src/commands/commit/upload.ts
```diff
@@ -0,0 +1,182 @@
...
+
+  public async execute() {
+    const initialTime = new Date().getTime()
+    const api = this.getApiHelper()
+    this.context.stdout.write(renderCommandInfo(this.dryRun))
+
+    const cliVersion = require('../../../package.json').version
+    const metricsLogger = getMetricsLogger(cliVersion)
+    const payload = await this.getPayloadToUpload(cliVersion)
+    if (payload === undefined) {
```
m-rousse, line 55: You can replace all `abc === undefined` with `!abc` and `return undefined` with `return`
```suggestion
    if (!payload) {
```

m-rousse (approved):
This LGTM overall, I made few coding style comments.

src/commands/commit/apikey.ts
jybp: Good point, URL crafting was a bit scattered around. There's now only one source of truth inside `api.ts`.

src/commands/commit/README.md
jybp: I don't think it's common but i'll ask to double check just in case.

lraucy: It's not common indeed, and doing so should probably not be documented. I didn't look at the code for this new command yet, but this is already what is done for sourcemaps upload (inferring both srcmap intake and metrics intake url from `DATADOG_SITE`).

jybp: Indeed it's inferred in both cases.
The README from the sourcemaps upload command that was copied over was outdated.
This PR updates it: https://github.com/DataDog/datadog-ci/pull/284

PR #266, opened 2021-06-08
[ci-app] Add Rich Commit Metadata

juan-fernandez:
### What and why?

We want the full CI App experience even for CIs that are not instrumented. This allows for that. 

### How?

* Add extra git metadata to the git extraction. 

### Review checklist

### DIFF
--- a/src/helpers/tags.ts
+++ b/src/helpers/tags.ts
@@ -15,15 +15,15 @@ export const CI_BUILD_LEVEL = '_dd.ci.build_level'
 
 // Git
 export const GIT_BRANCH = 'git.branch'
-export const GIT_SHA = 'git.commit.sha'
-export const GIT_TAG = 'git.tag'
-export const GIT_COMMIT_MESSAGE = 'git.commit.message'
-export const GIT_COMMIT_COMMITTER_DATE = 'git.commit.committer.date'
-export const GIT_COMMIT_COMMITTER_EMAIL = 'git.commit.committer.email'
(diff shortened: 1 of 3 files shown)


src/helpers/tags.ts
```diff
@@ -17,6 +17,13 @@ export const CI_BUILD_LEVEL = '_dd.ci.build_level'
 export const GIT_BRANCH = 'git.branch'
 export const GIT_SHA = 'git.commit.sha'
 export const GIT_TAG = 'git.tag'
+export const GIT_COMMIT_MESSAGE = 'git.commit.message'
+export const GIT_COMMIT_COMMITTER_DATE = 'git.commit.committer.date'
+export const GIT_COMMIT_COMMITTER_EMAIL = 'git.commit.committer.email'
+export const GIT_COMMIT_COMMITTER_NAME = 'git.commit.committer.name'
+export const GIT_COMMIT_AUTHOR_DATE = 'git.commit.author.date'
+export const GIT_COMMIT_AUTHOR_EMAIL = 'git.commit.author.email'
+export const GIT_COMMIT_AUTHOR_NAME = 'git.commit.author.name'
```
m-rousse, line 26: Nit: could we sort alphabetically these constants?

m-rousse (approved):
LGTM :+1:

src/helpers/tags.ts
juan-fernandez: sure thing

PR #270, opened 2021-06-10
[Synthetics] Tunnel: Reduce `Proxy opened` log verbosity

lefebvree:
### What and why?

Only display `[Tunnel] Proxy opened for test ...` log once per test instead on every proxy connection to reduce output spam for browser tests.

### Review checklist

### DIFF
--- a/src/commands/synthetics/tunnel.ts
+++ b/src/commands/synthetics/tunnel.ts
@@ -23,11 +23,11 @@ export interface TunnelInfo {
 }
 
 export class Tunnel {
+  private connected = false
   private forwardSockets: Socket[] = []
   private log: (message: string) => void
   private logError: (message: string) => void
   private multiplexer?: Multiplexer
-  private openedTunnels: Set<string> = new Set()
   private privateKey: string
   private publicKey: ParsedKey
   private sshStreamConfig: SSH2StreamConfig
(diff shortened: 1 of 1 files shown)


src/commands/synthetics/tunnel.ts
```diff
@@ -142,7 +143,11 @@ export class Tunnel {
     }
 
     // Username is allowed and key authentication was successful
-    this.log(`Proxy opened for test ${user}`)
+    if (!this.openedTunnels.has(ctx.username)) {
+      // Limit to one log per test
+      this.openedTunnels.add(ctx.username)
+      this.log(`Proxy opened for test ${ctx.username}`)
+    }
```
etnbrd, line 150: Nit, maybe removing the `ctx.username` at the end of the life-cycle of the tunnel?
Another solution would be to throttle based on the username?

etnbrd (approved):
LGTM

src/commands/synthetics/tunnel.ts
lefebvree: The life-cycle of the tunnel is basically the process itself, there won't be more than one opened in the command life-cycle so I think we're good to leave them there

m-rousse: These logs will not bring much value to an user at a test level (it is unlikely that one test amongst a suite will not have its tunnel opened), could we make it be displayed only once or at most once per tunnel?

m-rousse (approved):

PR #271, opened 2021-06-11
[command-synthetics] add defaultConfig and resolveConfig

KevinYeramian:
### What and why?

- Add `defaultConfig` to resolve the config one time and only use one source of truth (`this.config`)  

### How?

-  add `defaultConfig` function

### Review checklist

### DIFF
--- a/src/commands/synthetics/__tests__/run-test.test.ts
+++ b/src/commands/synthetics/__tests__/run-test.test.ts
@@ -1,10 +1,9 @@
 // tslint:disable: no-string-literal
-
 import * as ciUtils from '../../../helpers/utils'
 
 import {ExecutionRule} from '../interfaces'
 import {DefaultReporter} from '../reporters/default'
-import {DEFAULT_COMMAND_CONFIG, RunTestCommand} from '../run-test'
+import {DEFAULT_COMMAND_CONFIG, removeUndefinedValues, RunTestCommand} from '../run-test'
 import * as utils from '../utils'
 import {mockReporter} from './fixtures'
 
@@ -374,5 +373,10 @@ describe('run-test', () => {
         datadogSite: 'datadog.config.file',
       })
     })
+  })
+
+  test('removeUndefinedValues', () => {
+    // tslint:disable-next-line: no-null-keyword
+    expect(removeUndefinedValues({a: 'b', c: 'd', e: undefined, g: null})).toEqual({a: 'b', c: 'd', g: null})
   })
 })
--- a/src/commands/synthetics/__tests__/run-test.test.ts
+++ b/src/commands/synthetics/__tests__/run-test.test.ts
@@ -1,11 +1,10 @@
 // tslint:disable: no-string-literal
-jest.mock('fs')
 
 import * as ciUtils from '../../../helpers/utils'
 
 import {ExecutionRule} from '../interfaces'
 import {DefaultReporter} from '../reporters/default'
-import {RunTestCommand} from '../run-test'
+import {DEFAULT_COMMAND_CONFIG, RunTestCommand} from '../run-test'
 import * as utils from '../utils'
 import {mockReporter} from './fixtures'
 
@@ -27,9 +26,17 @@ export const assertAsyncThrow = async (func: any, errorRegex?: RegExp) => {
 }
 
 describe('run-test', () => {
+  beforeEach(() => {
+    jest.resetAllMocks()
+    jest.spyOn(ciUtils, 'getConfig').mockImplementation(async () => ({}))
+    process.env = {}
+  })
+
   describe('execute', () => {
+    beforeEach(() => {
+      jest.resetAllMocks()
+    })
     test('should apply config override for tests triggered by public id', async () => {
-      jest.spyOn(ciUtils, 'parseConfigFile').mockImplementation(async (config, _) => config)
       const getTestsToTriggersMock = jest.spyOn(utils, 'getTestsToTrigger').mockReturnValue(
         Promise.resolve({
           overriddenTestsToTrigger: [],
@@ -40,15 +47,17 @@ describe('run-test', () => {
       jest.spyOn(utils, 'runTests').mockImplementation()
 
       const startUrl = '{{PROTOCOL}}//myhost{{PATHNAME}}{{PARAMS}}'
-      const locations = ['location1', 'location2']
+      const locations = ['location1', 'location22']
       const configOverride = {locations, startUrl}
 
       const apiHelper = {}
       const command = new RunTestCommand()
       command.context = {stdout: {write: jest.fn()}} as any
       command['getApiHelper'] = (() => apiHelper) as any
-      command['config'].global = configOverride
-      command['publicIds'] = ['public-id-1', 'public-id-2']
+      // Override with config file
+      jest
+        .spyOn(ciUtils, 'getConfig')
+        .mockImplementation(async () => ({global: configOverride, publicIds: ['public-id-1', 'public-id-2']}))
       await command.execute()
 
       expect(getTestsToTriggersMock).toHaveBeenCalledWith(
@@ -62,7 +71,6 @@ describe('run-test', () => {
     })
 
     test('should not wait for `skipped` only tests batch results', async () => {
-      jest.spyOn(ciUtils, 'parseConfigFile').mockImplementation(async (config, _) => config)
       const getTestsToTriggersMock = jest.spyOn(utils, 'getTestsToTrigger').mockReturnValue(
         Promise.resolve({
           overriddenTestsToTrigger: [],
@@ -81,9 +89,9 @@ describe('run-test', () => {
       const configOverride = {executionRule: ExecutionRule.SKIPPED}
       command.context = {stdout: {write}} as any
       command['getApiHelper'] = (() => apiHelper) as any
-      command['config'].global = configOverride
-      command['publicIds'] = ['public-id-1', 'public-id-2']
-
+      jest
+        .spyOn(ciUtils, 'getConfig')
+        .mockImplementation(async () => ({global: configOverride, publicIds: ['public-id-1', 'public-id-2']}))
       await command.execute()
 
       expect(getTestsToTriggersMock).toHaveBeenCalledWith(
@@ -111,6 +119,7 @@ describe('run-test', () => {
     test('subdomain should be overridable', async () => {
       process.env = {DATADOG_SUBDOMAIN: 'custom'}
       const command = new RunTestCommand()
+      await command['resolveConfig']()
 
       expect(command['getAppBaseURL']()).toBe('https://custom.datadoghq.com/')
     })
@@ -121,6 +130,7 @@ describe('run-test', () => {
         DATADOG_SUBDOMAIN: 'custom',
       }
       const command = new RunTestCommand()
+      await command['resolveConfig']()
 
       expect(command['getAppBaseURL']()).toBe('https://custom.datadoghq.eu/')
     })
@@ -130,6 +140,7 @@ describe('run-test', () => {
     test('should default to datadog us api', async () => {
       process.env = {}
       const command = new RunTestCommand()
+      await command['resolveConfig']()
 
       expect(command['getDatadogHost']()).toBe('https://api.datadoghq.com/api/v1')
       expect(command['getDatadogHost'](true)).toBe('https://intake.synthetics.datadoghq.com/api/v1')
@@ -138,6 +149,7 @@ describe('run-test', () => {
     test('should be tunable through DATADOG_SITE variable', async () => {
       process.env = {DATADOG_SITE: 'datadoghq.eu'}
       const command = new RunTestCommand()
+      await command['resolveConfig']()
 
       expect(command['getDatadogHost']()).toBe('https://api.datadoghq.eu/api/v1')
       expect(command['getDatadogHost'](true)).toBe('https://api.datadoghq.eu/api/v1')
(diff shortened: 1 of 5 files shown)


src/commands/synthetics/run-test.ts
```diff
@@ -177,29 +179,60 @@ export class RunTestCommand extends Command {
...
-        config: {...this.config!.global, ...test.config},
+        config: {...this.config.global, ...test.config},
         id: test.id,
       }))
 
     return testsToTrigger
   }
 
+  private async resolveConfig() {
+    // Default < ENV < file < CLI
```
m-rousse, line 204: Let's have environment defined config supersede file-defined config
eg. `Default < file < ENV < CLI`

src/commands/synthetics/run-test.ts
```diff
@@ -1,42 +1,47 @@
...
-    appKey: process.env.DATADOG_APP_KEY,
-    datadogSite: process.env.DATADOG_SITE || 'datadoghq.com',
-    files: '{,!(node_modules)/**/}*.synthetics.json',
-    global: {} as ConfigOverride,
-    pollingTimeout: 2 * 60 * 1000,
-    proxy: {protocol: 'http'} as ProxyConfiguration,
-    subdomain: process.env.DATADOG_SUBDOMAIN || 'app',
-    tunnel: false,
-  }
+  private config: CommandConfig = JSON.parse(JSON.stringify(DEFAULT_COMMAND_CONFIG))
```
m-rousse, line 29: The deep copy does not seem to be needed

src/commands/synthetics/run-test.ts
```diff
@@ -1,42 +1,47 @@
...
+export const DEFAULT_COMMAND_CONFIG: CommandConfig = {
+  apiKey: '',
+  appKey: '',
+  configPath: 'datadog-ci.json',
+  datadogSite: 'datadoghq.com',
+  fileGlobs: ['{,!(node_modules)/**/}*.synthetics.json'],
+  global: {},
+  pollingTimeout: 2 * 60 * 1000,
+  proxy: {protocol: 'http'},
+  publicIds: [],
```
m-rousse, line 21: Is `CommandConfig` expected to match all options available through the config file?
In that case, both `publicIds` and `configPath` should not be here.

src/commands/synthetics/interfaces.ts
```diff
@@ -267,3 +267,18 @@ export interface APIConfiguration {
   baseUrl: string
   proxyOpts: ProxyConfiguration
 }
+
+export interface CommandConfig {
+  apiKey: string
+  appKey: string
+  configPath: string
+  datadogSite: string
+  fileGlobs: string[]
```
yoannmoinet, line 276: We should keep the same name as before, aka `files`, as it is documented [here](https://github.com/DataDog/datadog-ci/tree/master/src/commands/synthetics#configuration) and [here](https://docs.datadoghq.com/synthetics/ci/?tab=apitest#setup-the-client).
`fileGlobs` was the CLI parameter.

m-rousse (commented):
This looks good overall, I'm not sure about adding `publicIds` to the config file though

src/commands/synthetics/run-test.ts
yoannmoinet, line 29: hacky hacky.
At the very least, I'd add a comment explaining why we do this.

src/commands/synthetics/run-test.ts
```diff
@@ -220,11 +253,18 @@ export class RunTestCommand extends Command {
...
+}
+
 RunTestCommand.addPath('synthetics', 'run-tests')
 RunTestCommand.addOption('apiKey', Command.String('--apiKey'))
 RunTestCommand.addOption('appKey', Command.String('--appKey'))
 RunTestCommand.addOption('configPath', Command.String('--config'))
+RunTestCommand.addOption('fileGlobs', Command.Array('-f,--files'))
 RunTestCommand.addOption('publicIds', Command.Array('-p,--public-id'))
-RunTestCommand.addOption('testSearchQuery', Command.String('-s,--search'))
 RunTestCommand.addOption('shouldOpenTunnel', Command.Boolean('-t,--tunnel'))
```
yoannmoinet, line 269: This name doesn't need to be as explicit anymore IMO.
Should be unified as `tunnel`.

src/commands/synthetics/run-test.ts
```diff
@@ -220,11 +253,18 @@ export class RunTestCommand extends Command {
...
+  Object.keys(newObject).forEach((k) => newObject[k] === undefined && delete newObject[k])
+
+  return newObject
+}
+
 RunTestCommand.addPath('synthetics', 'run-tests')
 RunTestCommand.addOption('apiKey', Command.String('--apiKey'))
 RunTestCommand.addOption('appKey', Command.String('--appKey'))
 RunTestCommand.addOption('configPath', Command.String('--config'))
+RunTestCommand.addOption('fileGlobs', Command.Array('-f,--files'))
```
yoannmoinet, line 267: Should be renamed as `files` to unify the naming with the internal configuration.

src/commands/synthetics/run-test.ts
lefebvree: I think it's still good to copy it to avoid mutating the default, even if we don't refer to `DEFAULT_COMMAND_CONFIG` anywhere else for now, we might in the future. But we can use `deepCopy()` in `helpers.ts`

src/helpers/__tests__/utils.test.ts
```diff
@@ -1,67 +1,52 @@
...
 
-      await expect(parseConfigFile({})).rejects.toEqual(Error('Config file is not correct JSON'))
+      await expect(ciUtils.parseConfigFile({})).rejects.toEqual(Error('Config file is not correct JSON'))
     })
 
     test('config file should overwrite default configuration', async () => {
-      ;(fs.readFile as any).mockImplementation((_path: string, _opts: any, callback: any) =>
-        callback(undefined, '{"configKey":"newconfigvalue"}')
-      )
+      jest.spyOn(ciUtils, 'getConfig').mockImplementation(async () => ({configKey: 'newconfigvalue'}))
```
yoannmoinet, line 46: NIT: Does it need to be `async`?

src/helpers/utils.ts
```diff
@@ -16,13 +16,18 @@ export const pick = <T extends object, K extends keyof T>(base: T, keys: K[]) =>
   return pickedObject
 }
 
-export const parseConfigFile = async <T>(baseConfig: T, configPath?: string) => {
+export const getConfig = async (configPath: string) => {
+  const configFile = await promisify(fs.readFile)(configPath, 'utf-8')
```
yoannmoinet, line 20: NIT: Isn't it equivalent to this?
```suggestion
export const getConfig = (configPath: string) => {
  const configFile = fs.readFileSync(configPath, 'utf-8')
```

Agreed that it's not parallelizable anymore, but it was always used with `await` aka blocking anyways.
I may be missing something here.

I just feel like that it's used as blocking, but is promisified just for the kicks of it.
Could be simplified if synchronous IMO, WDYT @m-rousse ?

If we're looking for optimisation, then the file content should be cached.

src/commands/synthetics/run-test.ts
KevinYeramian: `CommandConfig` should match all options ( `ENV` U `file` U `CLI`)

yoannmoinet (commented):
Looks super good.
Great improvement for the config handling.

Shouldn't we globalize this for other commands too @m-rousse ?
So they all have the same behavior, could be misleading if other commands decide to overwrite configurations differently.

src/commands/synthetics/run-test.ts
KevinYeramian: `DEFAULT_COMMAND_CONFIG` is used in tests and some tests modify it. I needed to deep copy to avoid confit between tests.

src/commands/synthetics/run-test.ts
KevinYeramian: I am not sure about this one. The file is parsed after so it may change the override for customers.

src/commands/synthetics/run-test.ts
KevinYeramian: N.B: I merged `this.files` and `this.fileGlobs` to `this.fileGlobs` because `fileGlobs` was already used as a CLI option and `file` was only use internally for a  default value.

yoannmoinet: But doesn't it change the config file API?
So we also need to change all the docs we have around it, and also be sure we bump a minor on this.

IMO, not worth the pain.

src/commands/synthetics/run-test.ts
yoannmoinet: +1  that ENV should overwrite the file.
If we did the other way around before, we should bump a minor and notify the breaking change in the release.

src/helpers/__tests__/utils.test.ts
KevinYeramian: Yes because `getConfig` returns a promise

src/helpers/utils.ts
m-rousse: `readFileSync` blocks the thread while it waits for the result of the system call, whereas during an `await` the event loop is not blocked and asynchronous calls can be performed in other parts of the code (callbacks, events from other parts of the code and so on).
It is generally a good practice to use `await` rather than `*Sync` for this reason, though I agree it does not change much here.
Also it is not needed to use `promisify` as standard lib already has promisified versions of each functions:
```suggestion
import {promises as fs} from 'fs'

[...]

export const getConfig = async (configPath: string) => {
  const configFile = await fs.readFile(configPath, 'utf-8')
```

src/commands/synthetics/__tests__/run-test.test.ts
```diff
@@ -1,11 +1,10 @@
 // tslint:disable: no-string-literal
-jest.mock('fs')
```
m-rousse, line 2:
```suggestion
```

src/commands/synthetics/run-test.ts
```diff
@@ -177,29 +179,65 @@ export class RunTestCommand extends Command {
...
+        configPath: this.configPath,
+        files: this.files,
+        publicIds: this.publicIds,
+        testSearchQuery: this.testSearchQuery,
+        tunnel: this.tunnel,
+      })
+    )
+
+    if (typeof this.config.files === 'string') {
+      this.reporter!.log('[DEPRECATED] "files" should not be a string, array of string expected. The conversion will be automatic')
```
m-rousse, line 236:
```suggestion
      this.reporter!.log('[DEPRECATED] "files" should be an array of string instead of a string. The conversion will be automatic')
```

src/commands/synthetics/run-test.ts
```diff
@@ -220,11 +258,18 @@ export class RunTestCommand extends Command {
   }
 }
 
+const removeUndefinedValues = <T extends {[key: string]: any}>(object: T): T => {
```
m-rousse, line 261: Could you add unit tests for this function?

m-rousse (commented):
Agreed that it should be globalized to other commands, we'll do that in another PR.

Kevin, could you merge this branch to alpha and release a new version so we can battle-test it?

src/helpers/utils.ts
yoannmoinet: Of course, it was a way to simplify the code by making it synchronous.
While being a good practice, it still depends on the context and I just wanted to trigger the discussion as some food for thought.

Not needed for this PR of course.

yoannmoinet (approved):
LGTM

m-rousse (commented):
This almost LGTM, additionally I made it run in frontends CI and it worked correctly as well.

src/commands/synthetics/run-test.ts
```diff
@@ -205,7 +205,11 @@ export class RunTestCommand extends Command {
 
     // Override with file config variables
     const configPath = this.configPath ?? this.config.configPath
-    this.config = await parseConfigFile(this.config, configPath)
+    try {
+      this.config = await parseConfigFile(this.config, configPath)
+    } catch (error) {
+      this.reporter!.log(error.message)
+    }
```
m-rousse, line 212: It is missing a new line at end of the error message.
Could you also make it work like current behavior does: if no `configPath` has been passed, we try to load the default config file path but do not throw an error if we do not find it, if a `configPath` has been passed, we throw an error if it is not found.
In both cases, we throw an error if it fails to be parsed.

src/commands/synthetics/run-test.ts
```diff
@@ -177,29 +179,71 @@ export class RunTestCommand extends Command {
...
+  private async resolveConfig() {
+    // Default < file < ENV < CLI
+
+    // Override with file config variables
+    const configPath = this.configPath ?? this.config.configPath
+    try {
+      this.config = await parseConfigFile(this.config, configPath)
+    } catch (error) {
+      if (configPath) {
+        this.reporter!.log(`${error.message}\n)
```
m-rousse, line 212: Missing backtick
```suggestion
        this.reporter!.log(`${error.message}\n`)
```

m-rousse (commented):
One last thing: the command does not stop in case a config file is specified but not found.
(cf. test cases in this screenshot: https://a.cl.ly/Wnux7xE7)

src/commands/synthetics/run-test.ts
```diff
@@ -12,7 +12,7 @@ import {getReporter, getSuites, getTestsToTrigger, hasTestSucceeded, runTests, w
 export const DEFAULT_COMMAND_CONFIG: CommandConfig = {
   apiKey: '',
   appKey: '',
-  configPath: 'datadog-ci.json',
+  configPath: '',
```
m-rousse: This breaks the behavior of trying to load `datadog-ci.json` by default (cf. [documentation](https://github.com/DataDog/datadog-ci/tree/master/src/commands/synthetics#configuration))

It should:
- try to load `datadog-ci.json`, if it does not exist, continue without error (neither log a message nor throw an exception)
- if a `--configPath` argument was passed, load this configured file, if it does not exist, throw an exception and stop the command

m-rousse (approved):

PR #274, opened 2021-06-14
[s8s] Add a `--junit` flag to create a jUnit report

yoannmoinet:
### What and why?

It's very much needed to be able to instrumentalize the results on the integration side.
Right now, the only thing we can do is parse the log and hope for the best.

### How?

- Add a `--junit` flag to export a jUnit report.
- Add a `--runName` flag to indentify the current run.
- Add more metadata on suites so we can identify them in the report.

### Local testing

```
cd $DATADOG_ROOT/datadog-ci
git fetch && git checkout origin/yoann/junit-report
yarn build

cd $DATADOG_ROOT/web-ui
yarn link $DATADOG_ROOT/datadog-ci 
yarn datadog-ci synthetics run-tests -s "tag:test:junit" --jUnitReport "junit.xml" --runName "My Local Run" 

# For uploading the jUnit report back to CIApp
export DATADOG_API_KEY=***
yarn datadog-ci junit upload --service test-synthetics-junit-report junit.xml
```
And navigate to [the list](https://app.datadoghq.com/ci/test-runs?query=%40test.service%3Atest-synthetics-junit-report&index=citest&start=1631893756722&end=1632498556722&paused=false)

It's running [these tests](https://app.datadoghq.com/synthetics/tests?query=tag%3A%28%22test%3Ajunit%22%29) which is a list of a bit of everything (browser, api, multi steps,...).
Once the run is done, it will have created a `junit.xml` file at the root of `web-ui`.

### Review checklist


src/commands/synthetics/reporters/junit.ts
```diff
@@ -0,0 +1,280 @@
...
+})
+
+// Return the stats from a given object
+// based on getDefaultStats
+const getStats = (obj: any): Stats => {
+  const baseStats = getDefaultStats()
+  for (const entry of Object.entries(baseStats)) {
+    const [key] = entry as [keyof Stats, number]
+    baseStats[key] = obj[key] || baseStats[key]
+  }
```
juan-fernandez, line 84: nitpick, but would this be a bit simpler? 
```suggestion
  for (const [key, value] of Object.entries(baseStats)) {
    baseStats[key] = value || baseStats[key]
  }
```

src/commands/synthetics/reporters/junit.ts
```diff
@@ -0,0 +1,280 @@
...
+
+  private getStep(stepDetail: Step): XMLStep[] {
+    const mainStep: XMLStep = {
+      $: {
+        allow_failure: stepDetail.allowFailure,
+        is_skipped: stepDetail.skipped,
+        name: stepDetail.description,
+        substep_public_id: stepDetail.subTestPublicId,
+        time: stepDetail.duration / 1000,
```
juan-fernandez, line 180: `stepDetail.duration` is in ms?

src/commands/synthetics/reporters/junit.ts
```diff
@@ -0,0 +1,280 @@
...
+        url: stepDetail.url,
+        ...this.getStepStats(stepDetail),
+      },
+      browser_error: [],
+      error: [],
+      warning: [],
+    }
+    const steps = [mainStep]
+
+    if (stepDetail.subTestStepDetails && stepDetail.subTestStepDetails.length) {
```
juan-fernandez, line 191: do you need the `.length` check?

juan-fernandez (commented):
looking good! I only left nitpicks :)

src/commands/synthetics/reporters/junit.ts
yoannmoinet: Yep

src/commands/synthetics/reporters/junit.ts
yoannmoinet: I think I had to do this for the typing.
Let me double check.

juan-fernandez: yeah don't sweat it, it is also ok this way

yoannmoinet: Yep, typing issue.
image

src/commands/synthetics/reporters/junit.ts
yoannmoinet: Good catch, probably not since we're directly looping.

src/commands/synthetics/__tests__/reporters/junit.test.ts
```diff
@@ -0,0 +1,202 @@
...
+    context: {
+      stdout: {
+        write: writeMock,
+      },
+    },
+    jUnitReport: 'junit',
+  }
+
+  // Using "any" to access private properties
+  let reporter: any
```
etnbrd, line 28: Isn't there any other way possible?
In the worker code base, to keep the type that is being tested, we usually access private properties through the square-bracket notation and we let ESLint accept this notation: https://github.com/DataDog/synthetics-worker/blob/prod/worker/test/.eslintrc.js).

yoannmoinet: Ah makes sense. Will update.

yoannmoinet: Though, the typing is really getting in the way of testing if I don't use `any`.

src/commands/synthetics/reporters/junit.ts
```diff
@@ -0,0 +1,280 @@
...
+          {$: {name: 'result_id', value: result.resultID}},
+          {$: {name: 'type', value: test.type}},
+          {$: {name: 'message', value: test.message}},
+          {$: {name: 'monitor_id', value: test.monitor_id}},
+          {$: {name: 'tags', value: test.tags.join(',')}},
+          {$: {name: 'locations', value: test.locations.join(',')}},
+          {$: {name: 'start_url', value: result.result.startUrl}},
+          {$: {name: 'device', value: result.result.device.id}},
+          {$: {name: 'width', value: result.result.device.width}},
+          {$: {name: 'height', value: result.result.device.height}},
```
m-rousse, line 273: These values do not always exist depending on the test type (in that case, it is only for browser, `startUrl` only for browser and api-http, else it is `host` for the others, ...)
I do not know how jUnit handles that and also it may try to access undefined properties as `result.result.device` may be undefined, accessing `device.id` should throw an error.

src/commands/synthetics/reporters/junit.ts
```diff
@@ -0,0 +1,280 @@
...
+
+    return steps
+  }
+
+  private getStepStats(step: Step): Stats {
+    const errors = step.browserErrors ? step.browserErrors.length : 0
+
+    return {
+      allowfailures: step.allowFailure ? 1 : 0,
+      assertions: step.subTestStepDetails ? step.subTestStepDetails.length : 1,
```
m-rousse, line 234: I am not sure to understand which value `assertions` represents, if it is the number of assertions per step, it should be 0/1 according to the step type and the sum of steps containing an assertion when the step is a subtest.
WDYT?

src/commands/synthetics/reporters/junit.ts
```diff
@@ -0,0 +1,280 @@
...
+      this.write(`✅ Created a jUnit report at ${c.bold.green(this.destination)}\n`)
+    } catch (e) {
+      this.write(`❌ Couldn't write the report to ${c.bold.green(this.destination)}:\n${e.toString()}\n`)
+    }
+  }
+
+  public testEnd(test: Test, results: PollResult[]) {
+    const suiteRunName = test.suite || 'Undefined suite'
+
+    let suiteRun = this.json.testsuites.testsuite.find((suite: any) => suite.$.name === suiteRunName)
```
m-rousse, line 121:
```suggestion
    let suiteRun = this.json.testsuites.testsuite.find((suite: XMLRun) => suite.$.name === suiteRunName)
```

src/commands/synthetics/reporters/junit.ts
```diff
@@ -0,0 +1,280 @@
...
+  errors: 0,
+  failures: 0,
+  skipped: 0,
+  tests: 0,
+  warnings: 0,
+})
+
+// Return the stats from a given object
+// based on getDefaultStats
+const getStats = (obj: any): Stats => {
```
m-rousse, line 79:
```suggestion
const getStats = (obj: Stats): Stats => {
```

src/commands/synthetics/reporters/junit.ts
```diff
@@ -0,0 +1,280 @@
...
+  const baseStats = getDefaultStats()
+  for (const entry of Object.entries(baseStats)) {
+    const [key] = entry as [keyof Stats, number]
+    baseStats[key] = obj[key] || baseStats[key]
```
m-rousse, line 83:
```suggestion
    const [key, value] = entry as [keyof Stats, number]
    baseStats[key] = value || baseStats[key]
```

src/commands/synthetics/__tests__/fixtures.ts
```diff
@@ -52,3 +52,39 @@ export const getApiTest = (publicId: string): Test => ({
...
+  ],
+  skipped: false,
+  description: 'description',
+  warnings: [],
+  url: 'about:blank',
+  value: 'value',
+  duration: 1000,
+  allowFailure: false,
+  type: 'type',
+  stepId: -1,
```
m-rousse, line 73: Nit: could you sort the object in alphabetical order?

m-rousse (commented):
This looks good, my main/only concern is how will the reporter behave with different types of results (it seems to support mostly browser tests)

src/commands/synthetics/interfaces.ts
```diff
@@ -108,6 +130,7 @@ export interface Test {
   status: string
   stepCount: number
   subtype: string
+  suite?: string
```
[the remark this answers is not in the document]
yoannmoinet: Good idea.
I was wondering if we could use shared types from the worker maybe? (for another PR)
It's getting difficult to keep it updated here.

src/commands/synthetics/utils.ts
```diff
@@ -361,11 +362,14 @@ export const getTestsToTrigger = async (api: APIHelper, triggerConfigs: TriggerC
...
   const tests = await Promise.all(
-    triggerConfigs.map(async ({config, id}) => {
+    triggerConfigs.map(async ({config, id, suite}) => {
       let test: Test | undefined
       id = PUBLIC_ID_REGEX.test(id) ? id : id.substr(id.lastIndexOf('/') + 1)
       try {
-        test = await api.getTest(id)
+        test = {
+          suite,
+          ...(await api.getTest(id)),
```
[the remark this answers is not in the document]
yoannmoinet: Good catch.

src/commands/synthetics/run-test.ts
```diff
@@ -228,3 +252,5 @@ RunTestCommand.addOption('publicIds', Command.Array('-p,--public-id'))
 RunTestCommand.addOption('testSearchQuery', Command.String('-s,--search'))
 RunTestCommand.addOption('shouldOpenTunnel', Command.Boolean('-t,--tunnel'))
 RunTestCommand.addOption('fileGlobs', Command.Array('-f,--files'))
+RunTestCommand.addOption('jUnitReport', Command.String('-j,--junitReport'))
```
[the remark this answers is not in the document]
yoannmoinet: Definitely, will update.

src/commands/synthetics/reporters/junit.ts
yoannmoinet: Totally forgot about the API test 😭 
Will adapt to cover both type of tests.

src/commands/synthetics/reporters/junit.ts
yoannmoinet: I think we can remove it.
I saw a jUnit that was using it, but I don't think it's that useful for us.

src/commands/synthetics/interfaces.ts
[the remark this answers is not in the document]
m-rousse: We discussed it with @sdeprez a few weeks back and it means open-sourcing the package for "few types" for which we do not need them to be complete (we can only add needed fields as we use them), so we'd prefer not to.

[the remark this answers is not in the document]
yoannmoinet: Ah right, the open source thing is definitely a blocker.
What about a different package that type public things, open sourced, and used in this and other private/public packages?

gh0stonio (approved):

src/commands/synthetics/reporters/junit.ts
```diff
@@ -0,0 +1,335 @@
...
+      browser_error: browserError,
+      error,
+      warning,
+    }
+  }
+
+  private getResultStats(result: PollResult, stats: Stats | undefined = getDefaultStats()): Stats {
+    let stepsStats: Stats[] = []
+    if ('stepDetails' in result.result) {
+      // It's a browser test.
```
m-rousse, line 261: Adding `checkType` would make this code more straightforward to read as well

m-rousse (commented):
This LGTM, I only have typings suggestions

src/commands/synthetics/interfaces.ts
```diff
@@ -45,40 +32,98 @@ export interface Result {
     message: string
   }
   passed: boolean
+}
```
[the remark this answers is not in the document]
yoannmoinet: Ah, didn't know that, I'll update.

src/commands/synthetics/reporters/default.ts
```diff
@@ -204,15 +204,16 @@ const renderExecutionResult = (
   const isSuccess = hasResultPassed(result, failOnCriticalErrors, failOnTimeout)
   const color = getTestResultColor(isSuccess, test.options.ci?.executionRule === ExecutionRule.NON_BLOCKING)
   const icon = isSuccess ? ICONS.SUCCESS : ICONS.FAILED
-  const locationName = !!result.tunnel ? 'Tunneled' : locationNames[dc_id] || dc_id.toString()
-  const device = test.type === 'browser' && result.device ? ` - device: ${chalk.bold(result.device.id)}` : ''
+
+  const locationName = !!('tunnel' in result && result.tunnel) ? 'Tunneled' : locationNames[dc_id] || dc_id.toString()
```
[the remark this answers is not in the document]
yoannmoinet: Will do.

src/commands/synthetics/reporters/junit.ts
```diff
@@ -0,0 +1,335 @@
...
+  return baseStats
+}
+
+export class JUnitReporter implements Reporter {
+  private builder: Builder
+  private destination: string
+  private json: XMLJSON
+  private write: Writable['write']
+
+  constructor(command: RunTestCommand) {
```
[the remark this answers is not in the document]
yoannmoinet: I'll see what we can do here.
I really wanted to give full access to the command to the reporters, to give greater freedom for their implementation.

src/commands/synthetics/__tests__/reporters/junit.test.ts
```diff
@@ -0,0 +1,126 @@
...
+    jUnitReport: 'junit',
+  }
+
+  let reporter: JUnitReporter
+
+  describe('constructor', () => {
+    beforeEach(() => {
+      reporter = new JUnitReporter(commandMock as RunTestCommand)
+    })
+    it('should update destination with .xml if needed', () => {
```
juan-fernandez, line 30: I don't understand the name of this test: when is the destination updated?

juan-fernandez (commented):
code looks good, but I just did an end to end test including `datadog-ci junit upload` and it does not seem to upload the results correctly. My suspicion is that it's because`<testcase>` tag is missing, but I'll ask

juan-fernandez:
> code looks good, but I just did an end to end test including `datadog-ci junit upload` and it does not seem to upload the results correctly. My suspicion is that it's because `<testcase>` tag is missing, but I'll ask. If it's an easy addition it'd be great to have it, so that we can upload the results to CI Visibility

just confirmed it, it does need a `<testcase>` tag

yoannmoinet:
@juan-fernandez I think we can call the lowest child a `testcase` instead of a `testsuite`.
It was making sense this way before, when I was including substeps, but not anymore.
I'll update.

src/commands/synthetics/__tests__/reporters/junit.test.ts
yoannmoinet: It's just when we use `--jUnitReport=jUnit-report` for instance (without a `.xml`), it will append the `.xml` part.

juan-fernandez: oh that makes sense and it's good that it does it. Maybe a clearer description would be 
`should append '.xml' to destination if isn't there` ?

src/commands/synthetics/reporters/default.ts
```diff
@@ -136,7 +136,7 @@ const renderResultOutcome = (
   }
 
   if (test.type === 'browser') {
-    if (!hasResultPassed(result, failOnCriticalErrors, failOnTimeout) && result.stepDetails) {
+    if (!hasResultPassed(result, failOnCriticalErrors, failOnTimeout) && 'stepDetails' in result) {
```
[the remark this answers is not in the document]
yoannmoinet: 🤔 I tried, but couldn't get something working here.
Maybe I'm not doing it right though, I'm a bit rusty in TS.

m-rousse (approved):
This LGTM, great addition 🙇

src/commands/synthetics/reporters/junit.ts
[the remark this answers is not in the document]
yoannmoinet: Updated to only type what we use. 
I think I didn't understand the comment at first 😄  and though you were saying to only pass what was needed.
But I suppose you meant from the other side (reporter's side) only.

juan-fernandez (approved):
as mentioned in slack, the only thing missing to be able to use ci visibility with the generated junit xml reports is either change the timestamp format to RFC3339 / ISO8601 or just remove it altogether (ci visibility's backend will assign a timestamp automatically, which would be very close to what's generated now anyway). Feel free to ignore this for now though, we can do it later.

src/commands/synthetics/reporters/junit.ts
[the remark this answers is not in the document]
m-rousse: Yes, that's exactly that, thanks :+1:

PR #277, opened 2021-06-14
[ci-app] Bump `dd-trace`

juan-fernandez:
### What and why?

* Keep `dd-trace` up to date. 

### How?

* Mainly fix bugs, like instrumentation of snapshot errors and retries in timeouts. 

### Review checklist

### DIFF
--- a/testEnvironment.js
+++ b/testEnvironment.js
@@ -1,8 +1,9 @@
-require('dd-trace').init({
-  startupLogs: false,
-  enabled: !!process.env.CI,
-  debug: true,
-  logLevel: 'error',
-})
+if (process.env.CI) {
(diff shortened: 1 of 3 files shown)


testEnvironment.js
```diff
@@ -1,8 +1,9 @@
-require('dd-trace').init({
-  startupLogs: false,
-  enabled: !!process.env.CI,
-  debug: true,
-  logLevel: 'error',
-})
+if (process.env.CI) {
```
juan-fernandez, line 1: this has proved better for local dev, so you don't even require `'dd-trace'`

yoannmoinet (approved):

PR #279, opened 2021-06-15
[ci-app] Compress JUnit XML Report Files

juan-fernandez:
⚠️ **Do not merge**: Backend is not yet ready for this. 

### What and why?

* Compress `.xml` files before uploading them.

### How?

* Pipe file reading stream to `'zlib'`. 

### Review checklist

### DIFF
--- a/src/commands/junit/api.ts
+++ b/src/commands/junit/api.ts
@@ -3,6 +3,7 @@ import FormData from 'form-data'
 import fs from 'fs'
 import path from 'path'
 import {Writable} from 'stream'
+import {createGzip} from 'zlib'
 
 import {getRequestBuilder} from '../../helpers/utils'
 import {Payload} from './interfaces'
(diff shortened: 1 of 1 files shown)


src/commands/junit/api.ts
```diff
@@ -36,6 +37,7 @@ export const uploadJUnitXML = (request: (args: AxiosRequestConfig) => AxiosPromi
   const spanTags = {
     service: jUnitXML.service,
     ...jUnitXML.spanTags,
+    '_dd.cireport_version': '2',
```
yoannmoinet, line 40: NIT: I suppose it is to differentiate zipped input form unzipped ones.
Don't we have the version of the CLI for this maybe?

yoannmoinet (approved):
Very nice.
LGTM.

src/commands/junit/api.ts
juan-fernandez: indeed, it is for that. 

We could send the cli version indeed. We hadn't thought about it 😅 . It'd be a nice improvement though. For now we'll leave it as is, as the backend deployments take quite a bit

AdrianLC (approved):

PR #295, opened 2021-07-16
[ci-app] Add `--env` Input to `junit` Command.

juan-fernandez:
### What and why?

It seems weird that you can only configure the api key and env through environment variables. This PR changes this.

### How?

* Add `--apiKey` input to `junit`.
* Add `--env` input to `junit`.

### Review checklist

### DIFF
--- a/src/commands/junit/README.md
+++ b/src/commands/junit/README.md
@@ -23,7 +23,7 @@ datadog-ci junit upload --service my-service --tags key1:value1 --tags key2:valu
 ```
 
 - The positional arguments are the directories or file paths in which the jUnit XML reports are located. If you pass a folder, the CLI will look for all `.xml` files in it.
-- `--apiKey` (default: `DATADOG_API_KEY` env var) is the API Key used to authenticate the requests. **Important**: the API key needs to be available either through this parameter or `DATADOG_API_KEY` env var.
+- `--api-key` (default: `DATADOG_API_KEY` env var) is the API Key used to authenticate the requests. **Important**: the API key needs to be available either through this parameter or `DATADOG_API_KEY` env var.
 - `--service` (default: `DD_SERVICE` env var) should be set as the name of the service you're uploading jUnit XML reports for.
 - `--tags` is a array of key value pairs of the shape `key:value`. This will set global tags applied to all spans.
   - The resulting dictionary will be merged with whatever is in the `DD_TAGS` environment variable. If a `key` appears both in `--tags` and `DD_TAGS`, whatever value is in `DD_TAGS` will take precedence.
--- a/src/commands/junit/README.md
+++ b/src/commands/junit/README.md
@@ -35,7 +35,7 @@ datadog-ci junit upload --service my-service --tags key1:value1 --tags key2:valu
 
 Additionally you might configure the `junit` command with environment variables:
 
-- `DATADOG_API_KEY`: API key used to authenticate the requests. **Important**: the API key needs to be available either through this env var or through `--apiKey` parameter.
+- `DATADOG_API_KEY`: API key used to authenticate the requests. **Important**: the API key needs to be available either through this env var or through `--api-key` parameter.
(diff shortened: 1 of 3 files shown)


src/commands/junit/README.md
```diff
@@ -23,17 +23,19 @@ datadog-ci junit upload --service my-service --tags key1:value1 --tags key2:valu
 ```
 
 - The positional arguments are the directories or file paths in which the jUnit XML reports are located. If you pass a folder, the CLI will look for all `.xml` files in it.
+- `--apiKey` (default: `DATADOG_API_KEY` env var) is the API Key used to authenticate the requests. **Important**: the API key needs to be available either through this parameter or `DATADOG_API_KEY` env var.
```
AdrianLC, line 26:
```suggestion
- `--api-key` (default: `DATADOG_API_KEY` env var) is the API Key used to authenticate the requests. **Important**: the API key needs to be available either through this parameter or `DATADOG_API_KEY` env var.
```

src/commands/junit/upload.ts
```diff
@@ -90,7 +99,9 @@ export class UploadJUnitXMLCommand extends Command {
 
   private getApiHelper(): APIHelper {
     if (!this.config.apiKey) {
-      this.context.stdout.write(`Missing ${chalk.red.bold('DATADOG_API_KEY')} in your environment.\n`)
+      this.context.stdout.write(
+        `Missing ${chalk.red.bold('DATADOG_API_KEY')} in your environment and --apiKey was not passed.\n`
```
AdrianLC, line 103:
```suggestion
        `Missing ${chalk.red.bold('DATADOG_API_KEY')} in your environment and --api-key was not passed.\n`
```

src/commands/junit/upload.ts
```diff
@@ -175,6 +186,8 @@ export class UploadJUnitXMLCommand extends Command {
 }
 UploadJUnitXMLCommand.addPath('junit', 'upload')
 UploadJUnitXMLCommand.addOption('service', Command.String('--service'))
+UploadJUnitXMLCommand.addOption('env', Command.String('--env'))
+UploadJUnitXMLCommand.addOption('apiKey', Command.String('--apiKey'))
```
AdrianLC, line 190:
```suggestion
UploadJUnitXMLCommand.addOption('apiKey', Command.String('--api-key'))
```

AdrianLC (changes requested):
Let's make it consistent with the other parameters that do not have camel case

AdrianLC (approved):

alxch- (changes requested):
I'd say the fact that the key can only be configured by setting an environment variable is a safeguard against leaking a Datadog API key in our customers' CI by logging it, and sending it to anywhere customers send their logs (including Datadog).
Also, setting an environment variable is very easy, well documented, and other commands in `datadog-ci` are not supporting this.
Otherwise setting the environment via the CLI seems great.

juan-fernandez:
> I'd say the fact that the key can only be configured by setting an environment variable is a safeguard against leaking a Datadog API key in our customers' CI by logging it, and sending it to anywhere customers send their logs (including Datadog).
> Also, setting an environment variable is very easy, well documented, and other commands in `datadog-ci` are not supporting this.
> Otherwise setting the environment via the CLI seems great.

agree 👍

src/commands/junit/README.md
juan-fernandez: the problem is that other commands use `apiKey`: https://github.com/DataDog/datadog-ci/blob/master/src/commands/synthetics/run-test.ts#L338. It's probably better to keep consistency among commands @AdrianLC

AdrianLC: I'm not so sure if the other commands are not even consistent themselves. Anyway it's not important whatever you think it's best

juan-fernandez: we're going to remove `--api-key` so that there's less risk when dealing with secrets, so this is irrelevant

PR #300, opened 2021-07-21
RUMM-1470: `dsyms upload` command is implemented

buranmert:
### What and why?

iOS developers need to upload their app's dSYM files in order to symbolicate their crash reports.

### How?

`datadog-ci` tool now has `dsyms upload` command.

It takes either a file path or a folder path
1. File: `dsyms upload ~/Downloads/appDsyms.zip`
2. Folder: `dsyms upload ~/Library/Developer/Xcode/DerivedData/`

then looks for `*.dSYM` files in their content and uploads each of them by zipping.

### Review checklist


yarn.lock
```diff
@@ -3,40 +3,47 @@
```
yoannmoinet, line 3: This file shouldn't have changed.

yoannmoinet (changes requested):

yoannmoinet:
> this command is mostly copy&pasted from sourcemaps upload command.

Please don't do that.
Instead, factorise what both commands need.

yarn.lock
buranmert: reverted ✅

buranmert:
@yoannmoinet i started off with existing `sourcemaps`, then removed unneeded parts (such as `git.ts` and some other stuff)

yoannmoinet:
@buranmert I understand that, but we should not have duplicated code in the project.

If you need stuff that the sourcemaps command implemented, you move the functions you need in a separate file that is accessible from both commands.
And you import it from there.
We have an `./src/helpers/` folder specifically for this purpose.

src/commands/dsyms/README.md
```diff
@@ -0,0 +1,71 @@
...
+To verify this command works as expected, you can trigger a test run and verify it returns 0:
+
+```bash
+export DATADOG_API_KEY='<API key>'
+
+TEMP_DIR=$(mktemp -d)
+echo '{}' > $TEMP_DIR/fake.js
+echo '{"version":3,"file":"out.js","sourceRoot":"","sources":["fake.js"],"names":["src"],"mappings":"AAgBC"}' > $TEMP_DIR/fake.js.map
+yarn launch sourcemaps upload $TEMP_DIR/ --service test_datadog-ci --release-version 0.0.1 --minified-path-prefix https//fake.website
+rm -rf $TEMP_DIR
```
yoannmoinet, line 59: I don't think this is correct.

src/commands/dsyms/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+Successful output should look like this:
+
+```bash
+Starting upload with concurrency 20.
+Will look for sourcemaps in /var/folders/s_/ds1hc9g54k7ct8x7p3kwsq1h0000gn/T/tmp.fqWhNgGdn6/
+Will match JS files for errors on files starting with https//fake.website
+version: 0.0.1 service: test_datadog-ci project path:
+Uploading sourcemap /var/folders/s_/ds1hc9g54k7ct8x7p3kwsq1h0000gn/T/tmp.fqWhNgGdn6/fake.js.map for JS file available at https//fake.website/fake.js
+✅ Uploaded 1 files in 0.68 seconds.
```
yoannmoinet, line 70: This is not correct.

src/commands/dsyms/api.ts
```diff
@@ -0,0 +1,43 @@
...
+  })
+}
+
+export const apiConstructor = (baseIntakeUrl: string, apiKey: string) => {
+  const requestIntake = getRequestBuilder({baseUrl: baseIntakeUrl, apiKey})
+
+  return {
+    uploadDSYM: uploadDSYM(requestIntake),
+  }
+}
```
yoannmoinet, line 43: This should be factored IMO.

src/commands/dsyms/renderer.ts
```diff
@@ -0,0 +1,125 @@
...
+
+export const renderGitWarning = (errorMessage: string) =>
+  chalk.yellow(`${ICONS.WARNING} An error occured while invoking git: ${errorMessage}
+Make sure the command is running within your git repository to fully leverage Datadog's git integration.
+To ignore this warning use the --disable-git flag.\n`)
+
+export const renderConfigurationError = (error: Error) => chalk.red(`${ICONS.FAILED} Configuration error: ${error}.\n`)
+
+export const renderInvalidPrefix = chalk.red(
+  `${ICONS.FAILED} --minified-path-prefix should either be an URL (such as "http://example.com/static") or an absolute path starting with a / such as "/static"\n`
```
yoannmoinet, line 21: Do you really need this?

src/commands/dsyms/renderer.ts
```diff
@@ -0,0 +1,125 @@
...
+import {Payload} from './interfaces'
+import {pluralize} from './utils'
+
+const ICONS = {
+  FAILED: chalk.bold.red('❌'),
+  SUCCESS: chalk.bold.green('✅'),
+  WARNING: chalk.bold.green('⚠️'),
+}
+
+export const renderGitWarning = (errorMessage: string) =>
```
yoannmoinet, line 13: Do you really need this?

src/commands/dsyms/renderer.ts
```diff
@@ -0,0 +1,125 @@
...
+    }
+    if (results.get(UploadStatus.Failure)) {
+      output.push(chalk.red(`  * ${pluralize(results.get(UploadStatus.Failure)!, 'dSYM', 'dSYMs')} failed to upload`))
+    }
+  }
+
+  return output.join('\n') + '\n'
+}
+
+export const renderCommandInfo = (
```
yoannmoinet, line 94: Do you need this?

src/commands/dsyms/upload.ts
```diff
@@ -0,0 +1,167 @@
...
+      [
+        'Upload all sourcemaps in current directory',
+        'datadog-ci sourcemaps upload . --service my-service --minified-path-prefix https://static.datadog.com --release-version 1.234',
+      ],
+      [
+        'Upload all sourcemaps in /home/users/ci with 50 concurrent uploads',
+        'datadog-ci sourcemaps upload . --service my-service --minified-path-prefix https://static.datadog.com --release-version 1.234 --concurency 50',
+      ],
+    ],
+  })
```
yoannmoinet, line 42: This has been copy/pasted with no update.

yoannmoinet (changes requested):
cc @m-rousse, @juan-fernandez 
We probably want to factorise our way of uploading files.

Even going further, maybe having a single upload command to which you can give any payload, typed based on the `type` property of that payload.

Otherwise, we have to copy/paste everything to do it (api, renderer, metrics, ...) for a very slight difference.

IMO, it would be way cleaner to have something like `datadog-ci upload --file ./whatever.zip --type dsyms`, implemented in a single `src/commands/upload/*` with a variation on the `type` input.

buranmert:
fccd4e86a1bd062f2ba2ebe556a2cb2a4817567a uploaded dSYM files and tested in Datadog UI: stack trace is symbolicated as shown below ✅ 
Screenshot 2021-07-28 at 16 21 40

src/commands/dsyms/README.md
```diff
@@ -0,0 +1,75 @@
...
+You need to have `DATADOG_API_KEY` in your environment.
+
+```bash
+# Environment setup
+export DATADOG_API_KEY="<API KEY>"
+```
+
+It is possible to configure the tool to use Datadog EU by defining the `DATADOG_SITE` environment variable to `datadoghq.eu`. By defaut the requests are sent to Datadog US.
+
+It is also possible to override the full URL for the intake endpoint by defining the `DATADOG_SOURCEMAP_INTAKE_URL` environment variable.
```
yoannmoinet, line 18: `SOURCEMAP`?
```suggestion
It is also possible to override the full URL for the intake endpoint by defining the `DATADOG_DSYM_INTAKE_URL` environment variable.
```

src/commands/dsyms/README.md
```diff
@@ -0,0 +1,75 @@
...
+* `--dry-run` (default: `false`): it will run the command without the final step of upload. All other checks are performed.
+
+#### Bitcode
+
+With bitcode enabled, you should download your app's dSYM files from [App Store Connect](https://appstoreconnect.apple.com/).
+They come in the form of a zip file, named `appDsyms.zip`. In that case, you can run `datadog-ci` by pointing to the zip file.
+
+```bash
+// if appDsyms.zip path: ~/Downloads/appDsyms.zip
+datadog-ci dsyms upload ~/Downloads/
```
yoannmoinet, line 47: NIT: why not target the file directly?

src/commands/dsyms/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+To upload the dSYM files in your Derived Path, this command should be run:
+
+```bash
+datadog-ci dsyms upload ~/Library/Developer/Xcode/DerivedData/
+```
+
+In addition, some optional parameters are available:
+
+* `--concurrency` (default: `20`): number of concurrent upload to the API.
```
yoannmoinet, line 37: This is not the correct flag.
```suggestion
* `--max-concurrency` (default: `20`): number of concurrent upload to the API.
```

src/commands/dsyms/utils.ts
```diff
@@ -0,0 +1,127 @@
...
+import path from 'path'
+import {promisify} from 'util'
+import {buildPath} from '../../helpers/utils'
+import {Payload} from './interfaces'
+import {checkNonEmptyFile} from './validation'
+
+const UUID_REGEX = '[0-9A-F]{8}-([0-9A-F]{4}-){3}[0-9A-F]{12}'
+
+const getMatchingDSYMFiles = async (absoluteFolderPath: string): Promise<Payload[]> => {
+  const dSYMFiles = glob.sync(buildPath(absoluteFolderPath, '**/*.dSYM'))
```
yoannmoinet, line 13: NIT: Why using `.sync` here?

src/commands/dsyms/utils.ts
```diff
@@ -0,0 +1,127 @@
...
+        try {
+          await unzip(zipFile, unzipPath)
+
+          return unzipPath
+        } catch (err) {
+          console.log(err)
+        }
+      })
+      .filter((item) => item !== undefined)
+  )) as string[]
```
yoannmoinet, line 50: Why do you need to do this?

src/commands/dsyms/utils.ts
```diff
@@ -0,0 +1,127 @@
...
+  const searchPaths = (await Promise.all(
+    zipFiles
+      .map(async (zipFile: string) => {
+        const unzipPath = buildPath(unzipFolder, path.basename(zipFile, '.zip'), Date.now().toString())
+        try {
+          await unzip(zipFile, unzipPath)
+
+          return unzipPath
+        } catch (err) {
+          console.log(err)
```
yoannmoinet, line 46: Is this how we'll handle the errors?

src/commands/dsyms/utils.ts
```diff
@@ -0,0 +1,127 @@
+import {exec} from 'child_process'
+import {mkdirSync} from 'fs'
+import glob from 'glob'
+import path from 'path'
+import {promisify} from 'util'
+import {buildPath} from '../../helpers/utils'
+import {Payload} from './interfaces'
+import {checkNonEmptyFile} from './validation'
```
yoannmoinet, line 8: Why a different file for this one?

src/commands/dsyms/utils.ts
```diff
@@ -0,0 +1,127 @@
...
+const unzip = (sourcePath: string, targetPath: string) => {
+  const dirPath = path.dirname(targetPath)
+  mkdirSync(dirPath, {recursive: true})
+
+  return execute('unzip', ['-o', sourcePath, '-d', targetPath])
+}
+
+const execProc = promisify(exec)
+const execute = (cmd: string, args: string[], {env = process.env} = {}) =>
+  execProc(`${cmd} ${args.join(' ')}`, {
```
yoannmoinet, line 98: You should simplify this and only use `cmd`.
```suggestion
const execute = (cmd: string, {env = process.env} = {}) =>
  execProc(cmd, {
```

src/commands/sourcemaps/upload.ts
```diff
@@ -7,11 +7,12 @@ import path from 'path'
 import asyncPool from 'tiny-async-pool'
 import {URL} from 'url'
 
+import {ApiKeyValidator} from '../../helpers/apikey'
+import {UploadStatus} from '../../helpers/interfaces'
```
yoannmoinet, line 11: :+1:

src/commands/dsyms/utils.ts
```diff
@@ -0,0 +1,127 @@
...
+    env,
+    maxBuffer: 5 * 1024 * 5000,
+  }).then((output: {stderr: string; stdout: string}) => output)
+
+export const getBaseIntakeUrl = () => {
+  if (process.env.DATADOG_SOURCEMAP_INTAKE_URL) {
+    return process.env.DATADOG_SOURCEMAP_INTAKE_URL
```
yoannmoinet, line 105: It shouldn't be called this way.
```suggestion
  if (process.env.DATADOG_DSYM_INTAKE_URL) {
    return process.env.DATADOG_DSYM_INTAKE_URL
```

src/commands/dsyms/utils.ts
```diff
@@ -0,0 +1,127 @@
...
+  const dirPath = path.dirname(targetPath)
+  mkdirSync(dirPath, {recursive: true})
+
+  const sourceDir = path.dirname(sourcePath)
+  const sourceFile = path.basename(sourcePath)
+
+  // `zip -r foo.zip f1/f2/f3/foo.dSYM`
+  // this keeps f1/f2/f3 folders in foo.zip, we don't want this
+  // `cd ${sourceDir}` is called to avoid that
+  return execute(`cd ${sourceDir} && zip`, ['-r', targetPath, sourceFile])
```
yoannmoinet, line 86: Would the [`cwd` option](https://nodejs.org/api/child_process.html#child_process_child_process_exec_command_options_callback) of `exec` prevent the `cd` ?
```suggestion
  // this keeps f1/f2/f3 folders in foo.zip, we don't want this
  // `cd ${sourceDir}` is called to avoid that
  return execute(`zip`, ['-r', targetPath, sourceFile], {cwd: sourceDir})
```

src/commands/dsyms/README.md
buranmert: my first implementation supported targeting both folder and file path, but after some internal discussion it was decided to be confusing for the users. 
that's why i ended up with targeting folder only

src/commands/dsyms/utils.ts
buranmert: i took this from `getMatchingSourcemapFiles` in `sourcemaps/upload.ts`
i can make it `async`, should i?

src/commands/dsyms/utils.ts
buranmert: i couldn't get TS compiler to compile without this, otherwise TS sees the resulting array as `[string | undefined]` type whereas it should be `[string]` despite `filter((item) => item !== undefined)`

i'll be more than happy to solve this in a better way, open to suggestions

src/commands/dsyms/utils.ts
buranmert: i took `validation.ts` from `sourcemaps/validation.ts`
`checkNonEmptyFile` can be moved to `utils.ts` 👍

src/commands/dsyms/utils.ts
buranmert: i went with the simplest way that i knew. 

open to suggestions: should i pass `this.context.stderr` from `upload.ts` and use it here?

src/commands/dsyms/utils.ts
yoannmoinet: Since the function is `async` already, I think it's safe to use async/await code in it.
It just felt out of place here, I won't block for it.

src/commands/dsyms/utils.ts
yoannmoinet: I think it might come from the `console.log(err)`, try to `return []` right after it.

src/commands/dsyms/utils.ts
yoannmoinet: I think it would be more user friendly to let the error throw here and deal with it in `upload.ts` and craft an error message that the user will understand.

src/commands/dsyms/utils.ts
```diff
@@ -0,0 +1,127 @@
+import {exec} from 'child_process'
+import {mkdirSync} from 'fs'
+import glob from 'glob'
+import path from 'path'
+import {promisify} from 'util'
+import {buildPath} from '../../helpers/utils'
```
m-rousse, line 6:
```suggestion

import {buildPath} from '../../helpers/utils'

```

src/commands/dsyms/utils.ts
```diff
@@ -0,0 +1,127 @@
...
+
+export const getPayloads = async (searchPaths: string[]): Promise<Payload[]> => {
+  const nestedPayloads = await Promise.all(searchPaths.map((searchPath: string) => getMatchingDSYMFiles(searchPath)))
+
+  return nestedPayloads.flat()
+}
+
+export const getSearchPaths = async (baseSearchPath: string): Promise<string[]> => {
+  const zipFiles = getZipFiles(baseSearchPath)
+  const unzipFolder = buildPath('/tmp', 'datadog-ci', 'dsyms')
```
m-rousse, line 36: Same comment than at https://github.com/DataDog/datadog-ci/pull/300/files#diff-68b1d45d8e3483efa65f9faf279b01a6e96851cdf787a2875132f5575e36930bR24

src/commands/dsyms/utils.ts
```diff
@@ -0,0 +1,127 @@
...
+
+  return execute('unzip', ['-o', sourcePath, '-d', targetPath])
+}
+
+const execProc = promisify(exec)
+const execute = (cmd: string, args: string[], {env = process.env} = {}) =>
+  execProc(`${cmd} ${args.join(' ')}`, {
+    env,
+    maxBuffer: 5 * 1024 * 5000,
+  }).then((output: {stderr: string; stdout: string}) => output)
```
m-rousse, line 101: This `then` callback does not do anything
```suggestion
  })
```

src/commands/dsyms/utils.ts
m-rousse, line 98: As the `execProc` variable is used only once, it can be inlined when used,
Also this function is async as it returns a promise so it should be marked `async` :
```suggestion
const execute = async (cmd: string, args: string[], {env = process.env} = {}) =>
  promisify(exec)(`${cmd} ${args.join(' ')}`, {
```

src/commands/dsyms/utils.ts
```diff
@@ -0,0 +1,127 @@
...
+  // this keeps f1/f2/f3 folders in foo.zip, we don't want this
+  // `cd ${sourceDir}` is called to avoid that
+  return execute(`cd ${sourceDir} && zip`, ['-r', targetPath, sourceFile])
+}
+
+const unzip = (sourcePath: string, targetPath: string) => {
+  const dirPath = path.dirname(targetPath)
+  mkdirSync(dirPath, {recursive: true})
+
+  return execute('unzip', ['-o', sourcePath, '-d', targetPath])
```
m-rousse, line 93: This function returns the output of a function (`execute`) that returns a Promise so it should be marked `async`
Also you can then used the async version of `mkdir` as well

```suggestion
import {promises} from 'fs'
[...]

const unzip = async (sourcePath: string, targetPath: string) => {
  const dirPath = path.dirname(targetPath)
  await promises.mkdir(dirPath, {recursive: true})

  return execute('unzip', ['-o', sourcePath, '-d', targetPath])
```

src/commands/dsyms/utils.ts
```diff
@@ -0,0 +1,127 @@
...
+    const regexMatches = line.match(UUID_REGEX)
+    if (regexMatches && regexMatches.length > 0) {
+      uuids.push(regexMatches[0])
+    }
+  })
+
+  return uuids
+}
+
+export const zip = (sourcePath: string, targetPath: string) => {
```
m-rousse, line 76: Same comment than for `unzip`: it should be an async function and the asynchronous counterpart of `mkdir` should be used as well

src/commands/dsyms/utils.ts
```diff
@@ -0,0 +1,127 @@
...
+        const unzipPath = buildPath(unzipFolder, path.basename(zipFile, '.zip'), Date.now().toString())
+        try {
+          await unzip(zipFile, unzipPath)
+
+          return unzipPath
+        } catch (err) {
+          console.log(err)
+        }
+      })
+      .filter((item) => item !== undefined)
```
m-rousse, line 49:
```suggestion
      .filter((item) => !!item)
```

src/commands/dsyms/utils.ts
```diff
@@ -0,0 +1,127 @@
...
+  )) as string[]
+  searchPaths.push(baseSearchPath)
+
+  return searchPaths
+}
+
+export const getZipFiles = (absoluteFolderPath: string): string[] => {
+  const zipFiles = glob.sync(buildPath(absoluteFolderPath, '*.zip'))
```
m-rousse, line 57: This function is only called in this file so should not be exported.
Also it is called in an already asynchronous function so you could use the async version of `glob` instead.

(Using asynchronous functions avoids blocking the thread of node.js, it is important to do it as much as possible, especially when dealing with slow resources like I/O)

src/commands/dsyms/validation.ts
```diff
@@ -0,0 +1,12 @@
+import {statSync} from 'fs'
+
+export const checkNonEmptyFile = (path: string) => {
+  try {
+    const stats = statSync(path)
```
m-rousse, line 5: Could you make this function async as well? (using `fs.promises.stat` instead)

src/commands/dsyms/utils.ts
buranmert: we may have one single error among multiple zip files, the error case shouldn't stop this function.
then i need to return both `searchPaths` and `errors` from this function, is this okay?

```
...
return { searchPaths: searchPaths, errors: errors }
```

yoannmoinet: sure.

src/commands/dsyms/utils.ts
buranmert: why is that? `/helpers/**` should be separate from the others?

src/commands/dsyms/README.md
ncreated: IMO it is more convenient and cleaner if we expect a list of paths (e.g. a single `.zip` file or a list of `.dsym` files) instead of a folder path. 

Given the example used above:
```bash
datadog-ci dsyms upload ~/Downloads/
```
will it unzip and inspect all `*.zip` archives in the `~/Downloads/` folder? If so, this might be needless if the user has other random archives in there.

buranmert: i agree. we can re-open this discussion

src/commands/dsyms/utils.ts
buranmert: https://github.com/DataDog/datadog-ci/pull/300#discussion_r680956626
now we are reverting back to "either folder or a zip path": we won't be unzipping every zip file.
making this change will simplify `getSearchPaths` function and we won't need to return both paths and errors.

src/commands/dsyms/utils.ts
m-rousse: We sort dependencies with stdlibs, 3p libs, then all imports from projects files grouped by their amount of `../`. (as in https://github.com/DataDog/datadog-ci/blob/master/src/commands/junit/upload.ts) though it is not enforced by the linter.

src/commands/dsyms/api.ts
```diff
@@ -0,0 +1,43 @@
...
+  write: Writable['write']
+) => {
+  const form = new FormData()
+  write(renderUpload(dSYM))
+  form.append('type', dSYM.type)
+  const concatUUIDs = dSYM.uuids.join()
+  form.append('uuids', concatUUIDs)
+
+  const tmpFilePath = buildPath('/tmp', 'datadog-ci', 'dsyms', `${concatUUIDs}.zip`)
+  await zip(dSYM.path, tmpFilePath)
```
lraucy, line 25: Should we delete the temp directories after use, to avoid the tmp dir to run out of space over time?

src/commands/dsyms/README.md
buranmert: ℹ️ we decided to support both `~/Downloads/appDsyms.zip` (file) and `~/Library/Xcode/Developer/DerivedData/` (folder) as inputs

src/commands/dsyms/api.ts
```diff
@@ -0,0 +1,43 @@
...
+  dSYM: Payload,
+  write: Writable['write']
+) => {
+  const form = new FormData()
+  write(renderUpload(dSYM))
+  form.append('type', dSYM.type)
+  const concatUUIDs = dSYM.uuids.join()
+  form.append('uuids', concatUUIDs)
+
+  const tmpFilePath = buildPath('/tmp', 'datadog-ci', 'dsyms', `${concatUUIDs}.zip`)
```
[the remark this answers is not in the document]
buranmert: i switched to `tmpdir()` (results in `/var/{random string}` folders in mac OS) 👍 
but i kept appending timestamp to the folder name instead of using `mkdtemp`'s 6 random digits.

imo timestamp might make debugging slighlty easier. do you think i still should use `mkdtemp`?

src/commands/dsyms/api.ts
buranmert: i switched to [`os.tmpdir()`](https://nodejs.org/api/os.html#os_os_tmpdir), OS should take care of this now

lraucy: It looks safer if the program is handling the clean-up itself rather than relying on the OS to clean up (each OS might clean up at a different rate). It looks like [macOS cleans it up every 3 days](https://superuser.com/questions/187071/in-macos-how-often-is-tmp-deleted).

Since we use  the `tmpFolder` const for the whole execution, maybe we can simply delete that temp file when the command exits?

src/commands/dsyms/api.ts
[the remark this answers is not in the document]
m-rousse: Nope, timestamps are fine in my opinion.

src/commands/dsyms/utils.ts
```diff
@@ -0,0 +1,112 @@
...
+
+export const zipToTmpDir = async (sourcePath: string, targetFilename: string): Promise<string> => {
+  await promises.mkdir(tmpFolder, {recursive: true})
+  const targetPath = buildPath(tmpFolder, targetFilename)
+  const sourceDir = path.dirname(sourcePath)
+  const sourceFile = path.basename(sourcePath)
+  // `zip -r foo.zip f1/f2/f3/foo.dSYM`
+  // this keeps f1/f2/f3 folders in foo.zip, we don't want this
+  // `cwd: sourceDir` is passed to avoid that
+  await execute(`zip -r ${targetPath} ${sourceFile}`, sourceDir)
```
m-rousse, line 67: This does not seem to be platform independent and I suspect it would not work under Windows.
You could use `archiver` to do so in a platform agnostic way (cf. https://www.npmjs.com/package/archiver)
[As a side, it would also remove two occurrences of `execute` which is always a red flag security-wise)

m-rousse (approved):
This LGTM overall, I would push in favor of using a dedicated library to compress/decompress instead of using the shell (which would also have the advantage of being platform-independant), though `dwarfdump` may not be available on windows as well (maybe you could add an environment variable or CLI flag to specify the dwarfdump executable path?).

That being said, I won't block the PR for that.

src/commands/dsyms/utils.ts
buranmert: i agree, we can use a library in the upcoming releases to un/zip operations instead of spinning a new process.

but platform dependency shouldn't be a major concern for us as we target iOS devs.
i couldn't find `dwarfdump` for Windows and it is essential for `dsyms upload` command, we can't process dsym files without their UUID. 

as `dsyms upload` won't work in Windows, i added `if (os.platform() === 'darwin')` condition to the command

src/commands/dsyms/api.ts
buranmert: i think having the files for a while can be useful. for example, we can ask the user to look at that folder in case of any problem.
and if the disk space gets too low, this folder should be deleted immediately (that's how it works in iOS).

so i can't think of any problems with using `tmpdir()`, can you?

src/commands/dsyms/utils.ts
buranmert: i had to revert `if (os.platform() === 'darwin')` condition because our CI runs in a non-darwin platform.

now we have `dsyms upload` in all platforms, but it will crash in non-darwin platforms.
and i mock `dwarfdumpUUID` function in our unit tests.

src/commands/dsyms/upload.ts
```diff
@@ -0,0 +1,172 @@
...
+      ['Upload all dSYM files in Derived Data path', 'datadog-ci dsyms upload ~/Library/Developer/Xcode/DerivedData'],
+      [
+        'Upload all dSYM files in a zip file (this is usually the case if your app has Bitcode enabled)',
+        'datadog-ci dsyms upload /path/to/folder/my_file.zip',
+      ],
+    ],
+  })
+
+  private apiKeyValidator: ApiKeyValidator
+  private basePath?: string
```
yoannmoinet, line 44: Is it really possible to not have it?
I see it marked as required.
```suggestion
  private basePath: string
```

src/commands/dsyms/upload.ts
```diff
@@ -0,0 +1,172 @@
...
+  private dryRun = false
+  private maxConcurrency = 20
+
+  constructor() {
+    super()
+    this.apiKeyValidator = new ApiKeyValidator(this.config.apiKey, this.config.datadogSite)
+  }
+
+  public async execute() {
+    this.basePath = path.posix.normalize(this.basePath!)
```
yoannmoinet, line 58: Since it's required, I think you can ditch the `!`.
```suggestion
    this.basePath = path.posix.normalize(this.basePath)
```

src/commands/dsyms/upload.ts
```diff
@@ -0,0 +1,172 @@
...
+    const cliVersion = require('../../../package.json').version
+    const metricsLogger = getMetricsLogger(cliVersion)
+    const api = this.getApiHelper()
+
+    this.context.stdout.write(renderCommandInfo(this.basePath!, this.maxConcurrency, this.dryRun))
```
yoannmoinet, line 63:
```suggestion
    this.context.stdout.write(renderCommandInfo(this.basePath, this.maxConcurrency, this.dryRun))
```

yoannmoinet (approved):
LGTM.
Just some nits left.
Thanks for sticking around with the lengthy reviews :+1:

src/commands/dsyms/api.ts
buranmert: i still think we should let OS deal with this but i've just tried to clear `tmpFolder` at the end of `execute()` as explained [here in Node.js docs](https://nodejs.org/api/fs.html#fs_fspromises_rmdir_path_options)
```typescript
import { promises } from 'fs'
public async execute() {
...
  await promises.rm(tmpFolder, { recursive: true, force: true })
}
```
but compiler gives me the error below
> Property 'rm' does not exist on type 'typeof promises'.ts(2339)

i can only import `rmdir` from `fs` and `fsPromises` packages.
it looks like `rm` was added in v14.14.0, are we using an older version of Node?

buranmert: @lraucy we discussed with @m-rousse over slack and our Node version doesn't have an easy `rm -rf` like functionality.
given that we create temporary files and not in huge volumes, we decided to leave that for later. 
Datadog iOS app's dSYMs are ~25Mb, i imagine the worst case scenario could be ~250Mb.

we will probably use another 3rd party for this (eg: [rimraf](https://www.npmjs.com/package/rimraf))

lraucy (approved):
lgtm!

PR #306, opened 2021-07-28
[ci-app] Improve help for non-US1 use cases

edrevo:
### What and why?

Using the junit command with a non-US1 endpoint isn't straightforward: the command will fail with a 403 error without any more context, and the CLI help doesn't mention anything about the DATADOG_SITE env var.

### How?

Added a new example that shows how to send data to EU1

### Review checklist

### DIFF
--- a/src/commands/junit/upload.ts
+++ b/src/commands/junit/upload.ts
@@ -41,6 +41,10 @@ export class UploadJUnitXMLCommand extends Command {
         'Upload all jUnit XML test report files in current directory and add extra tags',
         'datadog-ci junit upload --service my-service --tags key1:value1 --tags key2:value2 .',
       ],
+      [
+        'Upload all jUnit XML test report files in current directory to the datadoghq.eu site',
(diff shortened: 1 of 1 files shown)


src/commands/junit/upload.ts
```diff
@@ -41,6 +41,10 @@ export class UploadJUnitXMLCommand extends Command {
         'Upload all jUnit XML test report files in current directory and add extra tags',
         'datadog-ci junit upload --service my-service --tags key1:value1 --tags key2:value2 .',
       ],
+      [
+        'Upload all jUnit XML test report files in current directory to the datadoghq.eu site',
+        'DATADOG_SITE=datadoghq.eu datadog-ci junit upload --service my-service --tags key1:value1 --tags key2:value2 .',
```
juan-fernandez, line 46:
```suggestion
        'DATADOG_SITE=datadoghq.eu datadog-ci junit upload --service my-service .',
```
if we want to highlight the datadog site we probably can drop the `tags`

juan-fernandez (approved):

PR #309, opened 2021-07-30
[synthetics] Allow spaces between variable brackets

lefebvree:
### What and why?

Match tests variables definition rule to allow whitespace between variable name and brackets.

### How?

Update variable template regex.

### Review checklist

### DIFF
--- a/src/commands/synthetics/utils.ts
+++ b/src/commands/synthetics/utils.ts
@@ -34,9 +34,10 @@ import {Tunnel} from './tunnel'
 const POLLING_INTERVAL = 5000 // In ms
 const PUBLIC_ID_REGEX = /^[\d\w]{3}-[\d\w]{3}-[\d\w]{3}$/
 const SUBDOMAIN_REGEX = /(.*?)\.(?=[^\/]*\..{2,5})/
+const TEMPLATE_REGEX = /{{\s*([^{}]*?)\s*}}/g
 
 const template = (st: string, context: any): string =>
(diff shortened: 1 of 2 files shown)


src/commands/synthetics/utils.ts
```diff
@@ -36,7 +36,7 @@ const PUBLIC_ID_REGEX = /^[\d\w]{3}-[\d\w]{3}-[\d\w]{3}$/
 const SUBDOMAIN_REGEX = /(.*?)\.(?=[^\/]*\..{2,5})/
 
 const template = (st: string, context: any): string =>
-  st.replace(/{{([A-Z_]+)}}/g, (match: string, p1: string) => (p1 in context ? context[p1] : match))
+  st.replace(/{{\s*([A-Z_]+)\s*}}/g, (match: string, p1: string) => (p1 in context ? context[p1] : match))
```
etnbrd, line 39: This RegEx seems already available in synthetics common: https://github.com/DataDog/synthetics-worker/blob/prod/common-public/src/variable.ts#L1
Though, it's way more permissive, so I am not sure we should use it here, but we might want to synchronize the two, no?

etnbrd (approved):
LGTM

src/commands/synthetics/utils.ts
lefebvree: We don't use synthetics-common in datadog-ci but we should certainly synchronize them yes

m-rousse (approved):
Awesome :+1:

PR #311, opened 2021-08-03
[ci-app] Simple JUnit XML Validation

juan-fernandez:
### What and why?

Validate jUnit XML before uploading:
* We skip uploading useless files.
* We signal to the user that they're doing something wrong.

### How?
* Add new dependency `fast-xml-parser`.
* Check that the file is valid XML by using `fast-xml-parser`.
* Check that the root of the XML includes a `testsuite` tag.

### Review checklist

### DIFF
--- a/src/commands/junit/upload.ts
+++ b/src/commands/junit/upload.ts
@@ -33,8 +33,8 @@ const validateXml = (xmlFilePath: string) => {
     return validationOutput.err.msg
   }
   const xmlFileJSON = xmlParser.parse(String(xmlFileContentString))
-  if (!('testsuites' in xmlFileJSON)) {
-    return '<testsuites> is not the root tag.'
+  if (!xmlFileJSON.testsuites && !xmlFileJSON.testsuite) {
+    return 'Neither <testsuites> nor <testsuite> is not the root tag.'
   }
 
   return undefined
--- a/src/commands/junit/upload.ts
+++ b/src/commands/junit/upload.ts
@@ -1,6 +1,7 @@
 import retry from 'async-retry'
 import chalk from 'chalk'
 import {Command} from 'clipanion'
+import xmlParser from 'fast-xml-parser'
 import fs from 'fs'
 import glob from 'glob'
 import path from 'path'
@@ -12,6 +13,7 @@ import {
   renderCommandInfo,
   renderDryRunUpload,
   renderFailedUpload,
+  renderInvalidFile,
   renderRetriedUpload,
   renderSuccessfulCommand,
 } from './renderer'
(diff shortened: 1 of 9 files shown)


yoannmoinet (approved):
LGTM infra wise.

src/commands/junit/upload.ts
```diff
@@ -24,6 +26,20 @@ import {buildPath} from '../../helpers/utils'
 const errorCodesNoRetry = [400, 403, 413]
 const errorCodesStopUpload = [400, 403]
 
+const validateXml = (xmlFilePath: string) => {
+  const xmlFileContentString = String(fs.readFileSync(xmlFilePath))
+  const validationOutput = xmlParser.validate(xmlFileContentString)
+  if (validationOutput !== true) {
+    return validationOutput.err.msg
+  }
+  const xmlFileJSON = xmlParser.parse(String(xmlFileContentString))
```
AdrianLC, line 35: This looks like it parses the full XML. It could get a bit slow for huge reports. Does this library have something to check only the root element?

juan-fernandez: I don't think it does. I've checked alternative libraries and they don't either. We need to go through the whole report for validation anyway, don't we?

juan-fernandez: tested with a rather big file: upload is still a much bigger factor, so I don't think we should worry about it

AdrianLC (approved):

src/commands/junit/upload.ts
```diff
@@ -24,6 +26,20 @@ import {buildPath} from '../../helpers/utils'
...
+  if (!('testsuites' in xmlFileJSON)) {
```
AdrianLC, line 36: Can this be `testsuites` || `testsuite`? some exporters inline a single test suite directly. This would match the parsing we do in backend

AdrianLC (changes requested):

src/commands/junit/upload.ts
```diff
@@ -24,6 +26,20 @@ import {buildPath} from '../../helpers/utils'
...
 
+const validateXml = (xmlFilePath: string) => {
+  const xmlFileContentString = String(fs.readFileSync(xmlFilePath))
+  const validationOutput = xmlParser.validate(xmlFileContentString)
+  if (validationOutput !== true) {
+    return validationOutput.err.msg
+  }
+  const xmlFileJSON = xmlParser.parse(String(xmlFileContentString))
+  if (!xmlFileJSON.testsuites && !xmlFileJSON.testsuite) {
+    return 'Neither <testsuites> nor <testsuite> is not the root tag.'
```
AdrianLC, line 37:
```suggestion
    return 'Neither <testsuites> nor <testsuite> are the root tag.'
```

AdrianLC (approved):

PR #316, opened 2021-08-09
RUMM-1470 dsyms upload handles whitespaces in path

buranmert:
### What and why?

`dsyms upload` command fails with any path containing whitespaces.
This is due to the fact that `dsyms upload` command executes a few operations in the shell and commands passed are not sanitized.

### How?

Now parameters passed to shell are wrapped with `' '` so that whitespaces don't break the execution anymore.

In order to test that, `dsyms/__tests__/files/` folder renamed to `/test files/`

### Review checklist


m-rousse (approved):
This LGTM.
As a sidenote, you may want to update the code owners of the `src/commands/dsyms` folder to notify the RUM mobile team!

yoannmoinet (approved):

PR #318, opened 2021-08-11
[ci-app] Allow Users to Provide `git` Metadata

juan-fernandez:
### What and why?

We want to allow users to provide git metadata manually, so that there's a fallback when other approaches fail.

⚠️ **Important**: do not merge until datadog-ci-spec is updated.

### How?

* Read certain environment variables and pass them as span tags. These tags take precedence over metadata extracted by other methods. 

### Review checklist


yoannmoinet (approved):

src/helpers/__tests__/user-provided-git.test.ts
```diff
@@ -0,0 +1,76 @@
...
+      [GIT_COMMIT_MESSAGE]: 'DD_GIT_COMMIT_MESSAGE',
+      [GIT_COMMIT_COMMITTER_DATE]: 'DD_GIT_COMMIT_COMMITTER_NAME',
+      [GIT_COMMIT_COMMITTER_EMAIL]: 'DD_GIT_COMMIT_COMMITTER_EMAIL',
+      [GIT_COMMIT_COMMITTER_NAME]: 'DD_GIT_COMMIT_COMMITTER_DATE',
+      [GIT_COMMIT_AUTHOR_DATE]: 'DD_GIT_COMMIT_AUTHOR_DATE',
+      [GIT_COMMIT_AUTHOR_EMAIL]: 'DD_GIT_COMMIT_AUTHOR_EMAIL',
+      [GIT_COMMIT_AUTHOR_NAME]: 'DD_GIT_COMMIT_AUTHOR_NAME',
+    })
+  })
+  it('returns an empty object is no user git is defined', () => {
```
edrevo, line 61:
```suggestion
  it('returns an empty object if no user git is defined', () => {
```

src/helpers/interfaces.ts
```diff
@@ -9,8 +9,12 @@ import {
   CI_STAGE_NAME,
   CI_WORKSPACE_PATH,
   GIT_BRANCH,
+  GIT_COMMIT_AUTHOR_DATE,
   GIT_COMMIT_AUTHOR_EMAIL,
   GIT_COMMIT_AUTHOR_NAME,
+  GIT_COMMIT_COMMITTER_DATE,
+  GIT_COMMIT_COMMITTER_EMAIL,
+  GIT_COMMIT_COMMITTER_NAME,
```
edrevo, line 17: shouldn't we update getGitMetadata() to read these new tags from the .git folder?

m-rousse (approved):
LGTM infra-wise

src/helpers/interfaces.ts
juan-fernandez: it does already, but you made me realise that we gain from adding an output type to `getGitMetadata`

src/helpers/__tests__/user-provided-git.test.ts
juan-fernandez: good catch

edrevo (approved):

PR #323, opened 2021-08-13
Extract upload code in a shared helpers module.

lraucy:
### What and why?

Refactor `multipart/form-data` payload uploads in a single shared module to reduce code duplication.

### How?

- Introduce a new `helpers/upload` module with helpers to upload `multipart/form-data` payloads.
- Use this new module in `sourcemap upload`, `dsyms upload` and `commit upload`
- Logging & metrics reporting are still specific to each commands
- `junit upload` and `dependencies upload` are not using this helper yet but could be migrated afterwards.

### Review checklist


src/commands/sourcemaps/upload.ts
```diff
@@ -141,47 +137,34 @@ export class UploadCommand extends Command {
...
       payloads.map(async (payload) => {
-        payload.gitRepositoryPayload = this.getRepositoryPayload(repositoryData, payload.sourcemapPath)
-        payload.gitRepositoryURL = repositoryData.remote
-        payload.gitCommitSha = repositoryData.hash
+        const repositoryPayload = this.getRepositoryPayload(repositoryData!, payload.sourcemapPath)
+        payload.addRepositoryData({
+          gitCommitSha: repositoryData.hash,
+          gitRepositoryPayload: repositoryPayload,
+          gitRepositoryURL: repositoryData.remote,
+        })
```
lraucy, line 153: I introduced a new interface GitData to group all git-related information into a single object (the object is either present entirely or not, no real use case to have these fields partially defined).

src/commands/sourcemaps/upload.ts
```diff
@@ -141,47 +137,34 @@ export class UploadCommand extends Command {
...
-        return {
-          cliVersion,
-          minifiedFilePath,
-          minifiedUrl: minifiedURL,
-          projectPath: this.projectPath,
-          service: this.service!,
-          sourcemapPath,
-          version: this.releaseVersion!,
-        }
+        return new Sourcemap(minifiedFilePath, minifiedURL, sourcemapPath)
```
lraucy, line 168: `cliVersion`, `projectPath`, `service`, `version` are all common to all sourcemaps a single run, so no need to duplicate the info in all `Sourcemap` object.

buranmert (approved):
regarding `dsyms` folder, LVGTM 👍

src/helpers/upload.ts
```diff
@@ -0,0 +1,129 @@
...
+  content: Map<string, MultipartValue>
+}
+
+export interface MultipartValue {
+  options?: FormData.AppendOptions | string
+  value: string | ReadStream
+}
+
+export interface UploadOptions {
+  /** ApiKeyValidator (optional) throws an InvalidConfiguratonException when upload fails because
```
m-rousse, line 23:
```suggestion
  /** ApiKeyValidator (optional) throws an InvalidConfigurationException when upload fails because
```

src/helpers/upload.ts
```diff
@@ -0,0 +1,129 @@
...
+   */
+  retries: number
+
+  /** Callback when upload fails (retries are not considered as failure)
+   */
+  onError(error: Error): void
+
+  /** Callback to execute before retries
+   */
+  onRetry(e: Error, attempts: number): void
```
m-rousse, line 40: Nit: to be consistent with other callback definition
```suggestion
  onRetry(error: Error, attempts: number): void
```

src/helpers/apikey.ts
```diff
@@ -1,28 +1,57 @@
-import axios from 'axios'
+import {AxiosError, default as axios} from 'axios'
+import chalk from 'chalk'
+import {BufferedMetricsLogger} from 'datadog-metrics'
 
+import {InvalidConfigurationError} from './errors'
+import {} from './utils'
```
m-rousse, line 6: I'm not sure if our linter allows this, but you should be able to write this as `import './utils'`.
Also, why is it needed to import the file but no function from it?

jybp (approved):
Thanks for this refactoring, LGTM 👍

src/helpers/apikey.ts
lraucy: Strange, that must be a leftover. I removed it!

src/commands/commit/upload.ts
```diff
@@ -30,32 +32,50 @@ export class UploadCommand extends Command {
...
+    if (payload === undefined) {
       return 0
     }
     try {
-      await this.uploadRepository(api, metricsLogger.logger, payload)
+      const requestBuilder = this.getRequestBuilder()
+      const status = await this.uploadRepository(requestBuilder)(payload, {
+        apiKeyValidator,
+        onError: (e) => {
+          this.context.stdout.write(renderFailedUpload(e.message)), metricsLogger.logger.increment('failed', 1)
```
yoannmoinet, line 61: is this correct?
Shouldn't it be:
```suggestion
          this.context.stdout.write(renderFailedUpload(e.message))
          metricsLogger.logger.increment('failed', 1)
```

yoannmoinet (commented):
LGTM.

Just a weird formatting thing.
Also, shouldn't `src/helpers/upload.ts` be tested as well?

lraucy:
> Also, shouldn't `src/helpers/upload.ts` be tested as well?

Very good point. I added some tests [in this commit](https://github.com/DataDog/datadog-ci/pull/323/commits/bb286e649a8784e5381b0cdba1ff987ad0391115). I also changed the exported ApiKeyValidator into an interface to ease testing without doing weird type casting.

src/helpers/__tests__/upload.test.ts
```diff
@@ -0,0 +1,193 @@
...
+          },
+          onError: errorCallback,
+          onRetry: retryCallback,
+          onUpload: uploadCallback,
+          retries: 1,
+        }
+      )
+      expect(result).rejects.toMatch('errorApiKey')
+      expect(mockCreate).toBeCalledTimes(1)
+      expect(uploadCallback).toBeCalledTimes(1)
```
jybp, line 187: I was about to suggest to add `expect(verifyKey).toBeCalledTimes(1)` but it doesn't seem to work, no idea why.

jybp (approved):

src/helpers/__tests__/upload.test.ts
lraucy: Yes I initially added this check but couldn't make it work, I wonder if this is related to how exception/rejected promise handling work with jest. I didn't dig further because I realized if the rejected promise matched `errorApiKey` that proves that the mocked `verifyKey` function has been called.

PR #324, opened 2021-08-17
Honor lambda states (optional now, rolled out everywhere in September)

astuyve:
- feat: Honor lambda state in instrument method
- feat: double escape quotes to pass snapshots

### What and why?

This PR verifies that the function can be updated, which according to this [post](https://aws.amazon.com/blogs/compute/coming-soon-expansion-of-aws-lambda-states-to-all-functions/) is when the `State` is `Active` and `LastUpdateStatus` is `Successful.

### How?

When we fetch the function configuration, this is attribute is returned. Because this command is likely ran in CI, immediately after updating a function, we will try 3 extra times if the first state is not valid, with exponential backoff between attempts.

I believe _all_ function configurations return `State` and `LastUpdateStatus`, but I've included a guard clause in cause. To be removed 1 Oct 2021, as per comment and schedule.

### Review checklist

### DIFF
--- a/src/commands/lambda/__tests__/instrument.test.ts
+++ b/src/commands/lambda/__tests__/instrument.test.ts
@@ -235,6 +235,30 @@ describe('lambda', () => {
                                         `)
       })
 
+      test("aborts if a function is not in an Active state with LastUpdateStatus Successful", async () => {
+        ;(fs.readFile as any).mockImplementation((a: any, b: any, callback: any) => callback({code: 'ENOENT'}))
+        ;(Lambda as any).mockImplementation(() => makeMockLambda({
+          'arn:aws:lambda:us-east-1:123456789012:function:lambda-hello-world': {
+            FunctionArn: 'arn:aws:lambda:us-east-1:123456789012:function:lambda-hello-world',
+            Handler: 'index.handler',
+            Runtime: 'nodejs12.x',
+            State: 'Failed',
+            LastUpdateStatus: 'Unsuccessful'
+          }
+        }))
+
+        const cli = makeCli()
+        const context = createMockContext() as any
+        const code = await cli.run(['lambda', 'instrument', '--function', 'arn:aws:lambda:us-east-1:123456789012:function:lambda-hello-world', '--layerVersion', '10'], context)
+
+        const output = context.stdout.toString()
+        expect(code).toBe(1)
+        expect(output).toMatchInlineSnapshot(`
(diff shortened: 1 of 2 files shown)


src/commands/lambda/function.ts
```diff
@@ -40,6 +40,46 @@ export interface InstrumentationSettings {
...
+
+/**
+ * Waits for n ms
+ * @param ms
+ * @returns
+ */
+const wait = (ms: number): Promise<void> => new Promise((res) => setTimeout(res, ms))
+
+/**
+ *
```
DarcyRaynerDD, line 53: Is this doc comment just here for linting reasons? Seems redundant with the types?

src/commands/lambda/__tests__/instrument.test.ts
```diff
@@ -235,6 +235,42 @@ describe('lambda', () => {
                                         `)
       })
 
+      test('aborts if a function is not in an Active state with LastUpdateStatus Successful', async () => {
```
DarcyRaynerDD, line 238: Nice test

src/commands/lambda/function.ts
```diff
@@ -40,6 +40,46 @@ export interface InstrumentationSettings {
...
+ * @param config
+ * @param functionArn
+ * @param attempts
+ * @returns bool
+ */
+const checkLambdaState = async (
```
agocs, line 59: Maybe rename to `isLambdaActive` or `isUpdatable`. `checkLambdaState` isn't a clear name, especially given that it returns a boolean.

src/commands/lambda/function.ts
```diff
@@ -40,6 +40,46 @@ export interface InstrumentationSettings {
...
+): Promise<boolean> => {
+  // TODO remove 1 Oct 2021 https://aws.amazon.com/blogs/compute/tracking-the-state-of-lambda-functions/
+  if (!config.State || !config.LastUpdateStatus) {
+    return true
+  }
+  if (config.LastUpdateStatus === 'Successful' && config.State === 'Active') {
+    return true
+  }
+  if (config.State === 'Pending' && attempts <= MAX_LAMBDA_STATE_CHECKS) {
+    await wait(2 ** attempts * 1000)
```
DarcyRaynerDD, line 73: At a glance I wasn't sure what the order of operations between ** and * is. Maybe add parenthesis? Or use  Math.pow over **.

DarcyRaynerDD (approved):
LGTM, left some nits

agocs (approved):
Looks good! Just a nitpick about the name `checkLambdaState`

src/commands/lambda/function.ts
astuyve: Bad editor config, need to remove jsdoc for `.ts`

PR #339, opened 2021-08-24
Support proxy config from env var in sourcemaps upload.

lraucy:
### What and why?

This PR:
- allow the possibility for commands to configure proxy through an URL instead of host/port/protocol
- allow proxy configuration in sourcemaps upload through `https_proxy` and `HTTPS_PROXY`.

Open questions:
- should we add support for proxy configuration through command-line flags?
- should we support `https_proxy` and `HTTPS_PROXY` by default for all commands?

Still left to do in this PR:
- support proxy in commits upload and dsyms upload commands

What's not addressed in this PR:
- proxy support for reported datadog metrics

### How?

### Review checklist

### DIFF
--- a/src/commands/sourcemaps/upload.ts
+++ b/src/commands/sourcemaps/upload.ts
@@ -52,6 +52,7 @@ export class UploadCommand extends Command {
   private config = {
     apiKey: process.env.DATADOG_API_KEY,
     datadogSite: process.env.DATADOG_SITE || 'datadoghq.com',
+    httpsProxy: process.env.https_proxy || process.env.HTTPS_PROXY,
   }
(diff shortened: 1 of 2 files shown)


src/commands/sourcemaps/upload.ts
```diff
@@ -228,11 +229,15 @@ export class UploadCommand extends Command {
     return getRequestBuilder({
       apiKey: this.config.apiKey!,
       baseUrl: getBaseIntakeUrl(),
+      disableEnvironmentVariables: true,
```
lraucy, line 232: Is there a reason not to enable this setting by default for all commands @m-rousse ?

PR #341, opened 2021-08-25
Add support for custom spans in CI Visibility

edrevo:
### What and why?

Add a new "trace" command to support custom spans in the CI Visibility product

### How?

The trace command wraps another command, measures its duration and sends a custom span to the webhooks pipeline.

### Review checklist


src/commands/trace/api.ts
```diff
@@ -0,0 +1,41 @@
...
+  const gitSpanTags = await getGitMetadata()
+  const userGitSpanTags = getUserGitMetadata()
+  customSpan.tags = {
+    ...gitSpanTags,
+    ...userGitSpanTags,
+    ...customSpan.tags,
+  }
+
+  return request({
+    data: customSpan,
```
juan-fernandez, line 25: minor, but I'd avoid modifying `customSpan` if possible 
```suggestion
  return request({
    data: {
      ...customSpan,
      tags: {
         ...gitSpanTags,
         ...userGitSpanTags,
         ...customSpan.tags,
      }
    },
```

src/commands/trace/trace.ts
```diff
@@ -0,0 +1,121 @@
...
+
+export class TraceCommand extends Command {
+  public static usage = Command.Usage({
+    description: 'Trace a command with a custom span and report it to Datadog.',
+    details: `
+            This command wraps another command, which it will launch, and report a custom span to Datadog.
+            See README for details.
+        `,
+    examples: [
+      ['Trace a command and report to Datadog', 'datadog-ci trace echo "Hello World"'],
```
juan-fernandez, line 19: I'm not sure if this is helpful, but .NET does something like this:
```
datadog-ci trace -- echo "Hello World"
```
https://docs.datadoghq.com/continuous_integration/setup_tests/dotnet/#instrumenting-tests 

I believe the idea being that we can more easily tell arguments apart, e.g.
```
datadog-ci trace --service=my-service -- echo --help
```

src/commands/trace/trace.ts
```diff
@@ -0,0 +1,121 @@
...
+
+      return 1
+    }
+
+    const [command, ...args] = this.command
+    const id = crypto.randomBytes(5).toString('hex')
+    const t0 = performance.now()
+    const spawnResult = spawnSync(command, args, {env: {...process.env, DD_CUSTOM_PARENT_ID: id}, stdio: 'inherit'})
+    const t1 = performance.now()
+    const duration = t1 - t0
```
juan-fernandez, line 46: this is in seconds, isn't it? 
I think this is fine and feel free to leave it like this, but the approach I've seen the most for this is: 
```javascript
const start = process.hrtime()
const spawnResult = spawnSync(command, args, {env: {...process.env, DD_CUSTOM_PARENT_ID: id}, stdio: 'inherit'})
const [secondsPassed, nanosecondsPassed] = process.hrtime(start);
const durationNanoseconds = secondsPassed*1e9 + nanosecondsPassed
```
this way you don't need to import anything and duration is already in nanoseconds, which is probably what you want anyway (I believe it's the standard for span duration, isn't it?)

src/commands/trace/trace.ts
```diff
@@ -0,0 +1,121 @@
...
+        tags: this.config.envVarTags ? parseTags(this.config.envVarTags.split(',')) : {},
+      },
+      provider
+    )
+
+    return exitCode
+  }
+
+  public getData(): [string, Record<string, string>] {
+    if (process.env.CIRCLECI === 'true') {
```
juan-fernandez, line 66: I think you can get away with 
```suggestion
    if (process.env.CIRCLECI) {
```

src/commands/trace/trace.ts
```diff
@@ -0,0 +1,121 @@
...
+  private getBaseIntakeUrl() {
+    if (process.env.DATADOG_SITE || process.env.DD_SITE) {
+      return `https://webhooks-http-intake.logs.${process.env.DATADOG_SITE || process.env.DD_SITE}`
+    }
+
+    return 'https://webhooks-http-intake.logs.datadoghq.com'
+  }
+
+  private getEnvironmentVars(keys: string[]): Record<string, string> {
+    return Object.fromEntries(keys.filter((key) => key in process.env).map((key) => [key, process.env[key]!]))
```
juan-fernandez: `Object.fromEntries` is not available in `10.24.1`, which is the oldest version we support, so we can't use it. 

```suggestion
    return keys.reduce((envVarsDict, key) => {
       if (process.env[key]) {
         envVarsDict[key] = process.env[key]
       }
       return envVarsDict
    })
```

src/commands/trace/trace.ts
```diff
@@ -0,0 +1,121 @@
...
+      return 1
+    }
+
+    const [command, ...args] = this.command
+    const id = crypto.randomBytes(5).toString('hex')
+    const t0 = performance.now()
+    const spawnResult = spawnSync(command, args, {env: {...process.env, DD_CUSTOM_PARENT_ID: id}, stdio: 'inherit'})
+    const t1 = performance.now()
+    const duration = t1 - t0
+    const exitCode = spawnResult.status ?? this.signalToNumber(spawnResult.signal!) ?? 127
```
juan-fernandez, line 47: if `.status` is defined, `.signal` is always truthy, right? That's what I read from this. But `signalToNumber` receives a potentially `undefined` value

src/commands/trace/trace.ts
```diff
@@ -0,0 +1,121 @@
...
+  private getEnvironmentVars(keys: string[]): Record<string, string> {
+    return Object.fromEntries(keys.filter((key) => key in process.env).map((key) => [key, process.env[key]!]))
+  }
+
+  private signalToNumber(signal?: string): number | undefined {
+    if (!signal) {
+      return undefined
+    }
+
+    return (os.constants.signals as any)[signal!] + 128
```
juan-fernandez, line 116: do you need `signal!` ? Doesn't typescript pick up that signal is truthy already?

juan-fernandez (commented):
Looking real good! Good job! 😄 . I left some mostly minor comments

src/commands/trace/trace.ts
edrevo: If you try to run a command that doesn't exist (so no child proc is actually launched), then both status and signal are undefined. The `!` in signal should be removed. I'll fix that.

src/commands/trace/trace.ts
edrevo: So I did start with that syntax, but I got a message from yarn saying that `--` was deprecated or something like that. I'll add it back.

src/commands/trace/trace.ts
edrevo: It does!

src/commands/trace/trace.ts
edrevo: I ended up changing this and there's no duration anymore, just a start time and an end time

src/commands/trace/trace.ts
```diff
@@ -0,0 +1,127 @@
...
+
+    return 'https://webhooks-http-intake.logs.datadoghq.com'
+  }
+
+  private getEnvironmentVars(keys: string[]): Record<string, string> {
+    return keys.filter((key) => key in process.env).reduce((accum, key) => ({...accum, [key]: process.env[key]!}), {})
+  }
+
+  private signalToNumber(signal: string | null): number | undefined {
+    if (!signal) {
```
juan-fernandez, line 117: FYI: an empty string `''` will also map to `undefined`. I don't know if that's an issue

src/commands/trace/__tests__/trace.test.ts
```diff
@@ -0,0 +1,54 @@
...
+describe('trace', () => {
+  describe('signalToNumber', () => {
+    test('should map undefined to undefined', () => {
+      process.env = {}
+      const write = jest.fn()
+      const command = new TraceCommand()
+      command.context = {stdout: {write}} as any
+
+      expect(command['signalToNumber'].call(undefined, null)).toBeUndefined()
+    })
```
juan-fernandez, line 13: by using `.call` you're setting the `this` argument, but I don't think you need it in this case. The name of the test should also be `map null to undefined`. And you probably don't need the `.context`. We could do: 
```suggestion
    test('should map null to undefined', () => {
      const command = new TraceCommand()
      expect(command['signalToNumber'](null)).toBeUndefined()
    })
```

src/commands/trace/__tests__/trace.test.ts
```diff
@@ -0,0 +1,54 @@
...
+    test('should map SIGKILL to 137', () => {
+      process.env = {}
+      const write = jest.fn()
+      const command = new TraceCommand()
+      command.context = {stdout: {write}} as any
+
+      expect(command['signalToNumber'].call(undefined, 'SIGKILL')).toEqual(137)
```
juan-fernandez, line 20: same as above with the context, and `.call`

src/commands/trace/trace.ts
```diff
@@ -0,0 +1,127 @@
...
+        start_time: startTime,
+        tags: this.config.envVarTags ? parseTags(this.config.envVarTags.split(',')) : {},
+      },
+      provider
+    )
+
+    return exitCode
+  }
+
+  public getData(): [string, Record<string, string>] {
```
juan-fernandez, line 70: `getData` does not tell much of what the function is doing. Maybe `getCIMetadata` ?

src/commands/trace/trace.ts
```diff
@@ -0,0 +1,127 @@
...
+          'CIRCLE_NODE_TOTAL',
+          'CIRCLE_PROJECT_REPONAME',
+          'CIRCLE_PULL_REQUEST',
+          'CIRCLE_REPOSITORY_URL',
+          'CIRCLE_SHA1',
+          'CIRCLE_TAG',
+          'CIRCLE_WORKFLOW_ID',
+        ]),
+      ]
+    }
```
juan-fernandez, line 89: we only want to support circleCI at the moment?

src/commands/trace/__tests__/trace.test.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  describe('getData', () => {
+    test('should throw if no CI is detected', () => {
+      process.env = {}
+      const write = jest.fn()
+      const command = new TraceCommand()
+      command.context = {stdout: {write}} as any
+
+      expect(command['getData'].bind(command)).toThrow(
+        'Cannot detect any CI Provider. This command only works if run as part of your CI.'
```
juan-fernandez, line 32: is this what we want? This would break under an unsupported CI provider, which might not be a great user experience. It might make more sense to just run the command as is and log a warning?

src/commands/trace/__tests__/trace.test.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    test('should correctly detect the circleci environment', () => {
+      process.env = {
+        CIRCLECI: 'true',
+        CIRCLE_WORKFLOW_ID: 'test',
+        NON_CIRCLE_ENV: 'bar',
+      }
+      const write = jest.fn()
+      const command = new TraceCommand()
+      command.context = {stdout: {write}} as any
```
juan-fernandez, line 44: you might not need this at all

src/commands/trace/api.ts
```diff
@@ -0,0 +1,43 @@
...
+import {getRequestBuilder} from '../../helpers/utils'
+import {Payload} from './interfaces'
+
+// Dependency follows-redirects sets a default maxBodyLength of 10 MB https://github.com/follow-redirects/follow-redirects/blob/b774a77e582b97174813b3eaeb86931becba69db/index.js#L391
+// We don't want any hard limit enforced by the CLI, the backend will enforce a max size by returning 413 errors.
+const maxBodyLength = Infinity
+
+export const reportCustomSpan = (request: (args: AxiosRequestConfig) => AxiosPromise<AxiosResponse>) => async (
+  customSpan: Payload,
+  provider: string
```
juan-fernandez, line 14: we could gain from having a typescript's type of supported CI providers, so we could type this `provider` here

src/commands/trace/interfaces.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export interface CustomIDs {
+  id: string
+  parent_id?: string
+}
+
+export interface Payload {
+  command: string
+  custom: CustomIDs
+  data: Record<string, string>
```
juan-fernandez, line 11: this data are the env vars dictionary, isn't it? Could we update the name so that it reflects that? If the key is already defined in the backend, either: 
* have interfaces with different (more meaningful) keys here and map them at the end, when doing `request` or
* add comments

src/commands/trace/interfaces.ts
```diff
@@ -0,0 +1,21 @@
+import {AxiosPromise, AxiosResponse} from 'axios'
+
+export interface CustomIDs {
+  id: string
+  parent_id?: string
+}
+
+export interface Payload {
+  command: string
+  custom: CustomIDs
```
juan-fernandez, line 10: are we using `CustomIDs` somewhere else? If not maybe just put it in `Payload` ?

src/commands/trace/interfaces.ts
```diff
@@ -0,0 +1,21 @@
+import {AxiosPromise, AxiosResponse} from 'axios'
+
+export interface CustomIDs {
```
juan-fernandez, line 3: the more usual way to do this is to define this as singular `CustomId`

src/commands/trace/interfaces.ts
```diff
@@ -0,0 +1,21 @@
...
+  parent_id?: string
+}
+
+export interface Payload {
+  command: string
+  custom: CustomIDs
+  data: Record<string, string>
+  end_time: string
+  is_error: boolean
+  name: string
```
juan-fernandez, line 14: name here is mapped to the span's operation name? I'd update the `name` to make it more explicit

src/commands/trace/trace.ts
```diff
@@ -0,0 +1,127 @@
...
+      this.context.stderr.write('Missing command to run\n')
+
+      return 1
+    }
+
+    const [command, ...args] = this.command
+    const id = crypto.randomBytes(5).toString('hex')
+    const startTime = new Date().toISOString()
+    const spawnResult = spawnSync(command, args, {env: {...process.env, DD_CUSTOM_PARENT_ID: id}, stdio: 'inherit'})
+    const endTime = new Date().toISOString()
```
juan-fernandez, line 45: this will only have down to milliseconds resolution. Are we OK with that?

src/commands/trace/trace.ts
```diff
@@ -0,0 +1,127 @@
...
+    const exitCode = spawnResult.status ?? this.signalToNumber(spawnResult.signal) ?? 127
```
juan-fernandez, line 46: why is `127` the default? Maybe extract it into a constant?

src/commands/trace/trace.ts
```diff
@@ -0,0 +1,127 @@
...
+    const api = this.getApiHelper()
+    const [provider, data] = this.getData()
```
juan-fernandez, line 48: again, `data` does not tell much of what's inside

juan-fernandez (commented):
great job! I left some comments. I'm not going to request changes since I'm going on PTO and I don't want to block the PR, but I think the code would gain from applying the feedback 😄

src/commands/trace/trace.ts
edrevo: `signal` should never be an empty string, according to the documentation. If it turns out to be an empty string, I think it is OK to map to undefined

src/commands/trace/trace.ts
edrevo: Yes, in this PR there's only support for CircleCI. Other CI Providers will come in different PRs

src/commands/trace/trace.ts
edrevo: It's the bash and zsh default for when a command does not exist:

```
❯ fdas
zsh: command not found: fdas

datadog-ci on  edrevo/custom-spans [!] is 📦 v0.16.0 via  v16.5.0 on ☁️  (us-east-1) on ☁️  ximo.guanter@datadoghq.com
❯ echo $?
127
```

I'll create a constant for it.

src/commands/trace/interfaces.ts
edrevo: The decision on how to map this is a backend decision. At the datadog-ci level, the user can provide a human-readable name to represent the span, which is what is represented by this field. Does that make sense?

(At the moment, this value is mapped to the `resource` value of the span)

src/commands/trace/__tests__/trace.test.ts
edrevo: Not sure about this one. I understand your point, but at the same time I think an explicit error is easier to debug during onboarding. So, if someone is trying out custom spans for the first time and we just log a warning, they might overlook it and open up a support ticket saying "why aren't custom spans working?".

However, I think there's 2 scenarios where having a warning is better than failing:
1. If we remove support for a specific CI provider in datadog-ci: the user experience here would be that they upgrade the CLI and it suddenly fails. Not a great UX. However, I don't think this is a likely scenario.
2. If the user is migrating from one CI provider to another one, and they have a period of time where they are running the jobs in both CI providers, and one is supported and the other one isn't. Again, not a super-likely scenario, but more likely that the first one.

How about if I keep the failing behavior by default and add a `--no-fail` param that people can opt into if they are migration CI Providers?

juan-fernandez: that sounds good to me. You can add the `--no-fail` at a later point 👍

edrevo:
Feedback addressed

src/commands/trace/__tests__/trace.test.ts
```diff
@@ -0,0 +1,40 @@
+// tslint:disable: no-string-literal no-null-keyword
+import {TraceCommand} from '../trace'
+
+describe('trace', () => {
+  describe('signalToNumber', () => {
+    test('should map undefined to undefined', () => {
```
juan-fernandez, line 6:
```suggestion
    test('should map null to undefined', () => {
```

src/commands/trace/__tests__/trace.test.ts
```diff
@@ -0,0 +1,40 @@
...
+      const command = new TraceCommand()
+      expect(command['signalToNumber'](null)).toBeUndefined()
+    })
+    test('should map SIGKILL to 137', () => {
```
juan-fernandez, line 10: maybe use the constant you've defined for this? `badCommandExitCode`. Maybe `should map SIGKILL to $badCommandExitCode`

src/commands/trace/__tests__/trace.test.ts
```diff
@@ -0,0 +1,40 @@
...
+      const command = new TraceCommand()
+      expect(command['signalToNumber']('SIGKILL')).toEqual(137)
```
juan-fernandez, line 12: probably better to import the constant here if possible

src/commands/trace/interfaces.ts
```diff
@@ -0,0 +1,24 @@
+import {AxiosPromise, AxiosResponse} from 'axios'
+
+export const CIRCLECI = 'circleci'
+
+export const SUPPORTED_PROVIDERS = [CIRCLECI] as const
+export type Provider = typeof SUPPORTED_PROVIDERS[number]
```
juan-fernandez, line 6: nice

src/commands/trace/interfaces.ts
juan-fernandez: was this addressed?

src/commands/trace/interfaces.ts
juan-fernandez: yeah it makes sense, I just worry that `name` is not meaningful enough. It's minor though, we can keep it as is

src/commands/trace/trace.ts
```diff
@@ -0,0 +1,144 @@
+import chalk from 'chalk'
+import {spawnSync} from 'child_process'
+import {Command} from 'clipanion'
+import crypto from 'crypto'
+import os from 'os'
+import {parseTags} from '../../helpers/tags'
+import {apiConstructor} from './api'
+import {APIHelper, CIRCLECI, Provider, SUPPORTED_PROVIDERS} from './interfaces'
+
+// We use 127 as exit code for invalid commands since that is what *sh terminals
```
juan-fernandez, line 10: is this phrase finished?

src/commands/trace/trace.ts
```diff
@@ -0,0 +1,144 @@
...
+const badCommandExitCode = 127
```
juan-fernandez, line 11: I'd probably rename this to `BAD_COMMAND_EXIT_CODE`. It is more conventional

juan-fernandez: and probably export it to be used for tests

src/commands/trace/trace.ts
```diff
@@ -0,0 +1,144 @@
...
+
+export class TraceCommand extends Command {
+  public static usage = Command.Usage({
+    description: 'Trace a command with a custom span and report it to Datadog.',
+    details: `
+            This command wraps another command, which it will launch, and report a custom span to Datadog.
+            See README for details.
+        `,
+    examples: [
+      ['Trace a command and report to Datadog', 'datadog-ci trace --name "Say Hello" -- echo "Hello World"'],
```
juan-fernandez, line 21: maybe add here something like 
```suggestion
      ['Trace a command with name "Say Hello" and report to Datadog', 'datadog-ci trace --name "Say Hello" -- echo "Hello World"'],
```

src/commands/trace/trace.ts
```diff
@@ -0,0 +1,144 @@
...
+
+    return apiConstructor(this.getBaseIntakeUrl(), this.config.apiKey)
+  }
+
+  private getBaseIntakeUrl() {
+    if (process.env.DATADOG_SITE || process.env.DD_SITE) {
+      return `https://webhooks-http-intake.logs.${process.env.DATADOG_SITE || process.env.DD_SITE}`
+    }
+
+    return 'https://webhooks-http-intake.logs.datadoghq.com'
```
juan-fernandez, line 125: I believe this was going to check. Can you double check with @AdrianLC ?

juan-fernandez (approved):
🚀 great job!

src/commands/trace/trace.ts
```diff
@@ -0,0 +1,144 @@
...
+      )
+      throw new Error('API key is missing')
+    }
+
+    return apiConstructor(this.getBaseIntakeUrl(), this.config.apiKey)
+  }
+
+  private getBaseIntakeUrl() {
+    if (process.env.DATADOG_SITE || process.env.DD_SITE) {
+      return `https://webhooks-http-intake.logs.${process.env.DATADOG_SITE || process.env.DD_SITE}`
```
AdrianLC, line 122:
```suggestion
      return `https://webhook-http-intake.${process.env.DATADOG_SITE || process.env.DD_SITE}`
```

src/commands/trace/trace.ts
AdrianLC:
```suggestion
    return 'https://webhook-http-intake.datadoghq.com'
```

src/commands/trace/api.ts
```diff
@@ -0,0 +1,43 @@
...
+        ...userGitSpanTags,
+        ...customSpan.tags,
+      },
+    },
+    headers: {
+      'X-Datadog-CI-Custom-Event': provider,
+    },
+    maxBodyLength,
+    method: 'POST',
+    url: 'v1/input',
```
AdrianLC, line 33:
```suggestion
    url: 'api/v2/webhook',
```

AdrianLC (changes requested):
Just and update of webhook URLs to the new ones

src/commands/trace/__tests__/trace.test.ts
edrevo: This is a different value (137), not badCommandExitCode (127). 137 is 128 + 9 (which is the SIGKILL number)

juan-fernandez: ops my bad

src/commands/trace/interfaces.ts
edrevo: it has, now! 🙈

src/commands/trace/__tests__/trace.test.ts
edrevo: I changed the expected value to `128 + 9` to make it more explicit

src/commands/trace/trace.ts
edrevo: Fixed.

src/commands/trace/trace.ts
edrevo: Done

edrevo:
All feedback was addressed.

AdrianLC (approved):

src/commands/trace/trace.ts
```diff
@@ -0,0 +1,147 @@
...
+    if (!this.command || !this.command.length) {
+      this.context.stderr.write('Missing command to run\n')
+
+      return 1
+    }
+
+    const [command, ...args] = this.command
+    const id = crypto.randomBytes(5).toString('hex')
+    const startTime = new Date().toISOString()
+    const spawnResult = spawnSync(command, args, {env: {...process.env, DD_CUSTOM_PARENT_ID: id}, stdio: 'inherit'})
```
m-rousse, line 52: As your function is already asynchronous, could you `spawn` from within a promise?
As much as possible we use async functions to avoid blocking node's main thread. (Though I reckon code is a bit heavier to read/write)
```suggestion
    const childProcess = spawn(command, args, {env: {...process.env, DD_CUSTOM_PARENT_ID: id}, stdio: 'inherit'})
    const spawnResult = await new Promise((resolve, reject) => {
      childProcess.on('error', (error: Error) => {
        reject(error)
      })

      childProcess.on('close', (exitCode: number) => {
        resolve(exitCode)
      }
    })
```

src/commands/trace/trace.ts
```diff
@@ -0,0 +1,147 @@
...
+
+    return apiConstructor(this.getBaseIntakeUrl(), this.config.apiKey)
+  }
+
+  private getBaseIntakeUrl() {
+    if (process.env.DATADOG_SITE || process.env.DD_SITE) {
+      return `https://webhook-intake.${process.env.DATADOG_SITE || process.env.DD_SITE}`
+    }
+
+    return 'https://webhook-intake.datadoghq.com'
```
m-rousse, line 128:
```suggestion
    const site = process.env.DATADOG_SITE || process.env.DD_SITE || 'datadoghq.com'

    return `https://webhook-intake.${site}`
```

src/commands/trace/trace.ts
```diff
@@ -0,0 +1,147 @@
...
+  private getEnvironmentVars(keys: string[]): Record<string, string> {
+    return keys.filter((key) => key in process.env).reduce((accum, key) => ({...accum, [key]: process.env[key]!}), {})
+  }
+
+  private signalToNumber(signal: string | null): number | undefined {
+    if (!signal) {
+      return undefined
+    }
+
+    return (os.constants.signals as any)[signal] + 128
```
m-rousse, line 140: As spawn typings are not precise here, I'd rather cast the variable as close to its declaration as possible, you could cast `spawnResult.signal` to `NodeJS.signal` on [line 54](https://github.com/DataDog/datadog-ci/pull/341/files#diff-af7806b72ff44c02eecebce7695ed34326d9f73f964f873ac8f639d9eaf165bbR54) to avoid using `any`/have variable types as exact as possible

```suggestion
  private signalToNumber(signal: NodeJS.Signals | null): number | undefined {
    if (!signal) {
      return undefined
    }

    return os.constants.signal[signal] + 128
```

src/commands/trace/api.ts
```diff
@@ -0,0 +1,43 @@
...
+      'X-Datadog-CI-Custom-Event': provider,
+    },
+    maxBodyLength,
+    method: 'POST',
+    url: 'api/v2/webhook',
+  })
+}
+
+export const apiConstructor = (baseIntakeUrl: string, apiKey: string) => {
+  const requestIntake = getRequestBuilder({baseUrl: baseIntakeUrl, apiKey})
```
m-rousse, line 38: Nit:
```suggestion
export const apiConstructor = (baseUrl: string, apiKey: string) => {
  const requestIntake = getRequestBuilder({baseUrl, apiKey})
```

m-rousse (commented):
Almost all good for me, my two questions are:
- why is this command not enabled? (you may document it as beta as well, or if it _should not_ be used, you may add it to an alpha release of the command)
- would it be possible to use the async version of `spawn` instead

edrevo:
Thanks for the feedback and the code suggestions @m-rousse! I've added the README and specified it is an alpha command (like the junit one). I have enabled the command too.

m-rousse (approved):
Good to go from infra PoV :+1:

PR #350, opened 2021-09-01
Add tests for getRequestBuilder proxy behavior.

lraucy:
### What and why?

Add some tests around the proxy configuration of getRequestBuilder.

### How?

Creates local http server and proxy for testing end-to-end behavior.

### Review checklist


tsconfig.json
```diff
@@ -17,5 +17,8 @@
     "lib": ["es2019"],
     "resolveJsonModule": true
   },
-  "include": ["src/**/*"]
+  "include": [
+    "src/**/*",
+    "proxy.d.ts"
```
lraucy, line 22: I don't know where's the best place to put this type file, probably not at the root, waiting for some suggestion?

src/helpers/__tests__/utils.test.ts
```diff
@@ -211,4 +215,224 @@ describe('utils', () => {
       expect(ciUtils.getProxyUrl({auth, protocol: 'http'})).toBe('')
     })
   })
+
+  // Test the different possibilities of proxy configuration of getRequestHelper.
+  // All the calls to getRequestHelpers should be https calls, but to keep the test suite
+  // simple tests are using http calls (testing with https would require us to add http certs
+  // and configure axios to trust these https certs, which requires an agent config, which
```
m-rousse, line 222:
```suggestion
  // simple tests are using http calls (testing with https would require us to add ssl certs
  // and configure axios to trust these certs, which requires an agent config, which
```

src/helpers/__tests__/utils.test.ts
```diff
@@ -211,4 +215,224 @@ describe('utils', () => {
...
+  // interferes a bit with how the proxies are configured since they are configured through an
+  // agent themselves.
+  // Proxy of https requests is still tested in the proxy-agent library itself.
+  describe('Proxy configuration', () => {
+    const initialHttpProxyEnv = process.env.HTTP_PROXY
```
m-rousse, line 227: Really nit: It does not really change the behavior but for symmetry you could put this in a `beforeAll` block

tsconfig.json
m-rousse: You can put it in `src/` or if you prefer to have all declarations in a folder you may create a `types` folder

m-rousse (approved):
LGTM 👍

package.json
```diff
@@ -78,6 +78,7 @@
     "jest": "27.0.5",
     "jest-environment-node": "27.0.5",
     "prettier": "2.0.5",
+    "proxy": "^1.0.2",
```
yoannmoinet, line 81: @lraucy can we get a fixed version please?

PR #353, opened 2021-09-06
[CIAPP-2679] Update DNS and url for junit upload with v2 API

AdrianLC:
### What and why?
Current DNS and URL are deprecated

### DIFF
--- a/src/commands/junit/api.ts
+++ b/src/commands/junit/api.ts
@@ -39,7 +39,7 @@ export const uploadJUnitXML = (request: (args: AxiosRequestConfig) => AxiosPromi
     ...jUnitXML.spanTags,
     '_dd.cireport_version': '2',
   }
-  form.append('message', JSON.stringify(spanTags))
+  form.append('event', JSON.stringify(spanTags))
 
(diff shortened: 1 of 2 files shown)


src/commands/junit/api.ts
```diff
@@ -39,7 +39,7 @@ export const uploadJUnitXML = (request: (args: AxiosRequestConfig) => AxiosPromi
     ...jUnitXML.spanTags,
     '_dd.cireport_version': '2',
   }
-  form.append('message', JSON.stringify(spanTags))
+  form.append('event', JSON.stringify(spanTags))
```
juan-fernandez, line 42: this should make it work:
```suggestion
  form.append('event', JSON.stringify(spanTags), {filename: 'event.json'})
```
I don't know which is the file name we'd want there though

juan-fernandez (approved):

src/commands/junit/api.ts
AdrianLC: Doesn't matter they won't upload it they are just validating

PR #354, opened 2021-09-07
Fix datadog-ci not working with configuration file

duncanista:
### What and why?

CLI not working with the following configuration file.
```
{
    "lambda": {
        "layerVersion": 60,
        "extensionVersion": 10,
        "functions": ["arn:aws:lambda:ap-southeast-1:601427279990:function:tian-test-ci-extension"],
        "region": "ap-southeast-1"
    }
}
```
There was a condition that checked if the functions array was empty, but if you specified a config file, it won't bother to check the config functions array.

### How?

Adding a ternary conditional to specify the functions array.

### Review checklist

- [x] Created unit test that mocks a case where config file is used but no functions are provided

### DIFF
--- a/src/commands/lambda/__tests__/mock-configs/no-functions.json
+++ b/src/commands/lambda/__tests__/mock-configs/no-functions.json
@@ -1,8 +1,8 @@
 {
-    "lambda": {
-        "layerVersion": 60,
-        "extensionVersion": 10,
-        "functions": [],
-        "region": "ap-southeast-1"
-    }
+  "lambda": {
+    "layerVersion": 60,
+    "extensionVersion": 10,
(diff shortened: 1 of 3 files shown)


src/commands/lambda/__tests__/mock-configs/no-functions.json
```diff
@@ -0,0 +1,8 @@
+{
```
maxday, line 1: Do you still need this file since `fs.readFile` is mocked?

duncanista: sorry, thought I deleted them, fixed

maxday (approved):
LGTM!  🎉

PR #356, opened 2021-09-09
Fix cannot disable options that default to true 

duncanista:
### What and why?

See original customer report https://a.cl.ly/P8uGkpmE. 
The real issue seems to be that we are using the presence of --xxx to determine whether xxx’s value (present meaning true, otherwise false). The problem with that is for --tracing and --flushMetricsToLogs, they default to true, so whether you add them to your command or not, their values will always be true. To avoid confusion, probably the best fix is to let them take a value, e.g., --tracing false --flushMetricsToLogs false, so customers can disable them if needed.The other option is to change their defaults to false and document more clearly how their values are determined, apparently the customer is expecting them to take a value. Therefore I’d prefer the other fix. 

### How?

The first option was taken, therefore, `tracing` and `flushMetricsToLogs` are now strings, so they are taking values.
Edit: Since --mergeXrayTraces follows the same path, I decided to change its type as well, the default value will remain the same, but if the users wants to change it, they need to provide a valid value.

### Review checklist

- [x] Updated unit tests


src/commands/lambda/instrument.ts
```diff
@@ -149,9 +150,29 @@ export class InstrumentCommand extends Command {
...
+      return
+    }
+
+    if (!['true', 'false', undefined].includes(this.tracing?.toLowerCase())) {
+      this.context.stdout.write('No valid value specified for tracing.\n')
+
+      return
+    }
+
+    const flushMetricsToLogs = this.flushMetricsToLogs
```
maxday, line 165: A nice way to test it could be to extract this logic a pure function taking two parameters (two strings representing flushMetricsToLogs et config.flushMetricsToLogs) then apply all for possible inputs and assert the result.
This method could be reused for tracing since it seems that it's the exact same logic

src/commands/lambda/instrument.ts
```diff
@@ -149,9 +150,29 @@ export class InstrumentCommand extends Command {
...
+      ? this.flushMetricsToLogs.toLowerCase() === 'true'
+      : this.config.flushMetricsToLogs
+      ? this.config.flushMetricsToLogs.toLowerCase() === 'true'
+      : true
     const mergeXrayTraces = this.mergeXrayTraces ?? this.config.mergeXrayTraces ?? false
-    const tracingEnabled = this.tracing ?? this.config.tracing ?? true
+    const tracingEnabled = this.tracing
+      ? this.tracing === 'true'
```
maxday, line 172: maybe we should lowercase it as well

PR #361, opened 2021-09-13
Add Jenkins support for custom spans

edrevo:
### What and why?

Add support for custom spans in Jenkins

### How?

Gather the necessary env vars so the backend can create the span

### Review checklist

### DIFF
--- a/src/commands/trace/__tests__/trace.test.ts
+++ b/src/commands/trace/__tests__/trace.test.ts
@@ -56,7 +56,7 @@ describe('trace', () => {
 
     test('should not detect the jenkins environment if it is not instrumented', () => {
       process.env = {
-        // DD_CUSTOM_TRACE_ID: 'abc',
+        // DD_CUSTOM_TRACE_ID not defined to simulate a non-instrumented instance
         JENKINS_HOME: '/root',
         NON_JENKINS_ENV: 'bar',
         WORKSPACE: 'def',
--- a/src/commands/trace/__tests__/trace.test.ts
+++ b/src/commands/trace/__tests__/trace.test.ts
@@ -36,5 +36,22 @@ describe('trace', () => {
         'circleci',
       ])
     })
+
+    test('should correctly detect the jenkins environment', () => {
+      process.env = {
+        DD_CUSTOM_TRACE_ID: 'abc',
+        JENKINS_HOME: '/root',
+        NON_JENKINS_ENV: 'bar',
+        WORKSPACE: 'def',
+      }
+      const command = new TraceCommand()
+      expect(command['getCIEnvVars']()).toEqual([
+        {
+          DD_CUSTOM_TRACE_ID: 'abc',
+          WORKSPACE: 'def',
+        },
+        'jenkins',
+      ])
+    })
   })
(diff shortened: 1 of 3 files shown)


juan-fernandez (approved):

juan-fernandez:
is this something we want to merge now before I do the alpha release?

src/commands/trace/__tests__/trace.test.ts
```diff
@@ -36,5 +36,22 @@ describe('trace', () => {
         'circleci',
       ])
     })
+
+    test('should correctly detect the jenkins environment', () => {
+      process.env = {
+        DD_CUSTOM_TRACE_ID: 'abc',
```
juan-fernandez, line 42: two things: 
* this will be injected by the jenkins plugin? How about running this in an "uninstrumented" jenkins instance? 
* this works differently from gitlab, right? In gitlab we do generate the id here, don't we?

edrevo: - Yes, it is generated by the Jenkins plugin. Uninstrumented Jenkins instances would generate a custom span which we would not be able to use in the backend. I'll add code to check for the presence of this variable and print a warning saying that you need to instrument the Jenkins instance if the variable does not exist.
- It works differently from circleci, yes. For circleci, the trace id is generated in the backend, so we don't need to report it in the custom span (we just need to report enough env var information to be able to re-calculate it in the backend too)

juan-fernandez: I see, thanks for the explanation. We do generate the parent span id, don't we? And we pass it to the process we spawn:

https://github.com/DataDog/datadog-ci/blob/master/src/commands/trace/trace.ts#L50-L52 

Also, this [`process.env. DD_CUSTOM_PARENT_ID`](https://github.com/DataDog/datadog-ci/blob/master/src/commands/trace/trace.ts#L73) would be empty, wouldn't it? We only set it for the child process.

edrevo: In Jenkins we always generate a parent id for the same reason we generate a trace id: everything happens plugin-side

In CircleCI we do not generate a parent id if the parent is a job. If the parent is another custom id, since the custom ids are generated in datadog-ci, we populate the DD_CUSTOM_PARENT_ID variable so we can support customs spans having children.

src/commands/trace/__tests__/trace.test.ts
```diff
@@ -36,5 +36,33 @@ describe('trace', () => {
...
+          DD_CUSTOM_TRACE_ID: 'abc',
+          WORKSPACE: 'def',
+        },
+        'jenkins',
+      ])
+    })
+
+    test('should not detect the jenkins environment if it is not instrumented', () => {
+      process.env = {
+        // DD_CUSTOM_TRACE_ID: 'abc',
```
m-rousse, line 59: This comment does not seem to add much here, could it be removed or if it is to show the environment variable is not defined, could it be made explicit that this env var is not supposed to be defined?

m-rousse (approved):
LGTM :+1:

PR #365, opened 2021-09-13
[synthetics-command] Add DATADOG_LOCATIONS env variable to override locations

KevinYeramian:
### What and why?

This feature was requested by front-end teams to spread tests over DCs 

### How?

- Add `DATADOG_LOCATIONS` env variable (not SYNTHETICS_LOCATIONS because it is not consistent with the rest)

### Review checklist

### DIFF
--- a/src/commands/synthetics/README.md
+++ b/src/commands/synthetics/README.md
@@ -25,7 +25,7 @@ If the org uses a custom sub-domain to access Datadog app, it needs to be set in
 export DATADOG_SUBDOMAIN="myorg"
 ```
 
-You can use the `DATADOG_SYNTHETICS_LOCATIONS` to override the locations where your tests run. Locations should be separated with `;`. Note that the configuration of the test will always the precedent over others overrides.  
+You can use the `DATADOG_SYNTHETICS_LOCATIONS` to override the locations where your tests run. Locations should be separated with `;`. Note that the configuration in test files takes precedence over others overrides.  
 
 ```bash
 export DATADOG_SYNTHETICS_LOCATIONS="aws:us-east-1;aws:us-east-2"
--- a/src/commands/synthetics/README.md
+++ b/src/commands/synthetics/README.md
@@ -25,6 +25,12 @@ If the org uses a custom sub-domain to access Datadog app, it needs to be set in
 export DATADOG_SUBDOMAIN="myorg"
 ```
 
+You can use the `DATADOG_SYNTHETICS_LOCATIONS` to override the locations where your tests run. Locations should be separated with `;`.
+
+```bash
+export DATADOG_SYNTHETICS_LOCATIONS="aws:us-east-1;aws:us-east-2"
+```
+
 ### API
 
 By default it runs at the root of the working directory and finds `{,!(node_modules)/**/}*.synthetics.json` files (every files ending with `.synthetics.json` except those in the `node_modules` folder).
--- a/src/commands/synthetics/README.md
+++ b/src/commands/synthetics/README.md
@@ -25,7 +25,7 @@ If the org uses a custom sub-domain to access Datadog app, it needs to be set in
 export DATADOG_SUBDOMAIN="myorg"
 ```
 
-You can use the `DATADOG_SYNTHETICS_LOCATIONS` to override the locations where your tests run. Locations should be separated with `;`.
+You can use the `DATADOG_SYNTHETICS_LOCATIONS` to override the locations where your tests run. Locations should be separated with `;`. Note that the configuration of the test will always the precedent over others overrides.  
 
 ```bash
 export DATADOG_SYNTHETICS_LOCATIONS="aws:us-east-1;aws:us-east-2"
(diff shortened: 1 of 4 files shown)


src/commands/synthetics/__tests__/run-test.test.ts
```diff
@@ -281,6 +281,68 @@ describe('run-test', () => {
...
+      command['getApiHelper'] = (() => apiHelper) as any
+      command['publicIds'] = ['publicId']
+
+      expect(await command.execute()).toBe(0)
+      expect(triggerTests).toHaveBeenCalledWith(
+        expect.objectContaining({tests: [{executionRule: 'blocking', locations: ['edge.us1.prod.dog'], public_id: 'publicId'}]})
+      )
+
+      process.env = {
+        DATADOG_LOCATIONS: 'edge.us2.prod.dog',
```
m-rousse, line 321: Locations should be referred to using their short format in datadog-ci, could you use `aws:us-east-1` and the likes in the tests so its closer to real data?

src/commands/synthetics/run-test.ts
```diff
@@ -349,6 +355,7 @@ export class RunTestCommand extends Command {
         apiKey: process.env.DATADOG_API_KEY,
         appKey: process.env.DATADOG_APP_KEY,
         datadogSite: process.env.DATADOG_SITE,
+        locations: process.env.DATADOG_LOCATIONS?.split(';'),
```
m-rousse, line 358: Could you rename it to `DATADOG_SYNTHETICS_LOCATIONS` as this environment variable is specific to the synthetics command?

m-rousse (approved):
LGTM - could you document it in the synthetics command README?

src/commands/synthetics/run-test.ts
```diff
@@ -124,6 +125,11 @@ export class RunTestCommand extends Command {
       return safeExit(1)
     }
     const {tests, overriddenTestsToTrigger, summary} = testsToTriggerResult
+    if (this.config.locations.length) {
+      overriddenTestsToTrigger.forEach((overriddenTestToTrigger) => {
+        overriddenTestToTrigger.locations = this.config.locations
+      })
+    }
```
etnbrd, line 132: Sorry for the late review.
It seems most of the config override is done in (or around) `handleConfig` in `synthetics/utils`: https://github.com/DataDog/datadog-ci/blob/master/src/commands/synthetics/utils.ts#L43-L85
Out of curiosity, what is the reason for doing the `locations` override so far from the original override?

src/commands/synthetics/run-test.ts
```diff
@@ -323,7 +325,11 @@ export class RunTestCommand extends Command {
     const testsToTrigger = suites
       .reduce((acc, suiteTests) => acc.concat(suiteTests), [])
       .map((test) => ({
-        config: {...this.config.global, ...test.config},
+        config: {
+          ...this.config.global,
+          ...test.config,
+          ...(this.config.locations?.length ? {locations: this.config.locations} : {}),
```
m-rousse, line 331: Could you extract this object to a variable `configFromEnvironment`?
(If it could be created/populated outside of `this.config`/`resolveConfig` it'd be nice as well so that we know it is only used for overrides from env vars)

m-rousse (commented):
Could you also add an entry in the README of the `synthetics` command?

m-rousse (approved):

src/commands/synthetics/run-test.ts
```diff
@@ -320,10 +322,15 @@ export class RunTestCommand extends Command {
...
 
+    const configFromEnvironment = this.config.locations?.length ? {locations: this.config.locations} : {}
     const testsToTrigger = suites
       .reduce((acc, suiteTests) => acc.concat(suiteTests), [])
       .map((test) => ({
-        config: {...this.config.global, ...test.config},
+        config: {
+          ...this.config.global,
+          ...test.config,
+          ...configFromEnvironment,
```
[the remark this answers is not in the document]
m-rousse: It is not, and I agree, doing otherwise would suit web-ui/datadog's usage, but is not intuitive. :+1:

src/commands/synthetics/README.md
```diff
@@ -25,7 +25,7 @@ If the org uses a custom sub-domain to access Datadog app, it needs to be set in
 export DATADOG_SUBDOMAIN="myorg"
 ```
 
-You can use the `DATADOG_SYNTHETICS_LOCATIONS` to override the locations where your tests run. Locations should be separated with `;`.
+You can use the `DATADOG_SYNTHETICS_LOCATIONS` to override the locations where your tests run. Locations should be separated with `;`. Note that the configuration of the test will always the precedent over others overrides.
```
m-rousse, line 28:
```suggestion
You can use the `DATADOG_SYNTHETICS_LOCATIONS` to override the locations where your tests run. Locations should be separated with `;`. Note that the configuration in test files takes precedence over others overrides.  
```

m-rousse (approved):

PR #366, opened 2021-09-14
[lambda][instrument] Datadog CLI can set up many Lambda functions

duncanista:
### What and why?

Currently the datadog-ci lambda instrument command asks customers to list out all the Lambda functions to the instrument one-by-one. In some cases, customers prefer to let AWS generate a unique function name for them each time, and it’s hard to predict the actual function names.

### How?

1. Allow the --function argument to support wildcard like *, or regular expressions

2. Instead of supplying --function, the customer can specify functions matching certain tags using an option, e.g.,  --function-with-tags service:web,env:prod instruments functions tagged with BOTH service:web AND env:prod. If the customer provides more than one --function-with-tags option, they are OR’d.

### Review checklist


m-rousse:
This branch seems to contain unrelated changes, could you merge/rebase of latest `master`?

duncanista:
Thinking on implementing the `--function-with-tags` command in this PR, let me know what you think @maxday

src/commands/lambda/function.ts
```diff
@@ -96,38 +97,55 @@ export const getLambdaConfigs = async (
...
+  settings: InstrumentationSettings
+): Promise<FunctionConfiguration[]> => {
+  const regEx = new RegExp(pattern)
+  const matchedFunctions: Lambda.FunctionConfiguration[] = []
+  let retryCount = 0
+  let listFunctionsResponse: Lambda.ListFunctionsResponse
+  let nextMarker: string | undefined
+  let firstCall = true
+
+  while (nextMarker || firstCall) {
```
maxday, line 123: I think nextMarker + firstCall boolean are a bit redundant. I understand that you need to check whether or not this is the first call to handle the undefined marker possibility properly but it makes the code a bit harder to read.
Maybe you can wrap the listFunction to add a boolean which indicates whether or not an undefined value is OK (should be OK for the first call, but should stop the loop in other cases)

src/commands/lambda/instrument.ts
```diff
@@ -307,6 +348,7 @@ export const sentenceMatchesRegEx = (sentence: string, regex: RegExp) => sentenc
 
 InstrumentCommand.addPath('lambda', 'instrument')
 InstrumentCommand.addOption('functions', Command.Array('-f,--function'))
+InstrumentCommand.addOption('regExPattern', Command.String('-fR,--functions-regex'))
```
maxday, line 351: Don't forget to update the lambda command documentation as well here https://github.com/DataDog/datadog-ci/blob/master/src/commands/lambda/README.md

src/commands/lambda/function.ts
duncanista: Mmmm weird, I thought I had already solved this, maybe the merge stashed that code, I'll fix it again

duncanista: I do understand why it may seem redundant. The thing is that `nextMarker` can only be `undefined` at the beginning and in the end (when we've finished fetching all the functions). So if it's at the beginning, we do have to know if it's the first call to keep on retrying until it succeeds or it just reaches the retry limit.

duncanista: refactored using your recommendation from todays coding session

maxday (approved):
LGTM 👍

PR #367, opened 2021-09-14
Add --tags to trace command

edrevo:
### What and why?

Add support for `--tags` to the `trace` command

### How?

Same as in the `junit` command

### Review checklist


juan-fernandez (approved):
maybe add an example with `--tags` in `Command.Usage` ?

m-rousse (approved):

PR #368, opened 2021-09-14
Datadog CLI sets default for unified service tagging

duncanista:
### What and why?

To demonstrate the full capabilities of Datadog serverless monitoring, we should provide options for customers to easily tag their applications, like --service, --env, --version and --extra-tags.

If the customer misses one of --service, --env, --version tags, we should print out a message to encourage users to set them up.

### How?

Add four commands as described above, modified some tests to comply with new feature, and created unit tests for the feature.

### Review checklist

- [x] Modified existing unit tests
- [x] Created unit test

### DIFF
--- a/src/commands/lambda/function.ts
+++ b/src/commands/lambda/function.ts
@@ -33,19 +33,22 @@ export interface FunctionConfiguration {
   updateRequest?: Lambda.UpdateFunctionConfigurationRequest
 }
 
-export interface InstrumentationSettings {
+interface InstrumentationTags {
   environment?: string
-  extensionVersion?: number
   extraTags?: string
+  service?: string
+  version?: string
+}
+
+export interface InstrumentationSettings extends InstrumentationTags {
+  extensionVersion?: number
   flushMetricsToLogs: boolean
   forwarderARN?: string
   layerAWSAccount?: string
   layerVersion?: number
   logLevel?: string
   mergeXrayTraces: boolean
-  service?: string
   tracingEnabled: boolean
-  version?: string
 }
 
 const MAX_LAMBDA_STATE_CHECKS = 3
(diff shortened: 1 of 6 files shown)


src/commands/lambda/instrument.ts
```diff
@@ -177,15 +182,46 @@ export class InstrumentCommand extends Command {
...
+    }
+    let tagMissing = false
+    for (const [tag, value] of Object.entries(tagsMap)) {
+      if (!value) {
+        tagMissing = true
+        this.context.stdout.write(`No value found for the ${tag} tag.\n`)
+      }
+    }
+    if (tagMissing) {
+      this.context.stdout.write('It is recommended to set tags, try to do it later.\n')
```
maxday, line 202: I think we need to warn the customer only if service/env/version is missing, tags should remain optional

src/commands/lambda/function.ts
```diff
@@ -251,6 +259,23 @@ export const calculateUpdateRequest = (
     changedEnvVars[FLUSH_TO_LOG_ENV_VAR] = settings.flushMetricsToLogs.toString()
   }
 
+  if (settings.environment && oldEnvVars[ENVIRONMENT_ENV_VAR] !== settings.environment.toString()) {
```
maxday, line 262: maybe this chunk of code could be refactored? it seems that the logic for each env variable is the same?

src/commands/lambda/instrument.ts
```diff
@@ -177,15 +182,46 @@ export class InstrumentCommand extends Command {
...
+    }
+
+    const extraTags = this.extraTags ?? this.config.extraTags
+    if (extraTags && !extraTags.match(EXTRA_TAGS_REG_EXP)) {
```
maxday, line 206: Could be great to extract this part to a pure function with a string + regex as input and a boolean as a result.
That way you could test some valid/invalid inputs against the resulting boolean

maxday: Also we are lowercasing the tags downstream in layers/extensions but we should do it here as well

maxday (changes requested):
looks nice! I left some minor comments

src/commands/lambda/function.ts
duncanista: Yup, was thinking the same, maybe just a map for every env variable?

src/commands/lambda/instrument.ts
duncanista: Oh yeah, that code is from a map that contains those, maybe should just rename the map and the boolean

maxday (approved):

PR #370, opened 2021-09-16
RUMM-1608 Dsym upload: if dwarfdump throws, error is caught

buranmert:
### What and why?

In the case of invalid dSYM files, dwarfdump can throw.
This resulted in breaking the execution.

### How?

Now we continue running by skipping that particular dsym files.

### Review checklist

### DIFF
--- a/src/commands/dsyms/upload.ts
+++ b/src/commands/dsyms/upload.ts
@@ -68,10 +68,11 @@ export class UploadCommand extends Command {
       metricsLogger: metricsLogger.logger,
     })
     const payloads = await getMatchingDSYMFiles(searchPath)
+    const validPayloads = payloads.filter((payload) => payload !== undefined) as Dsym[]
     const requestBuilder = this.getRequestBuilder()
     const uploadDSYM = this.uploadDSYM(requestBuilder, metricsLogger, apiKeyValidator)
     try {
-      const results = await asyncPool(this.maxConcurrency, payloads, uploadDSYM)
+      const results = await asyncPool(this.maxConcurrency, validPayloads, uploadDSYM)
       const totalTime = (Date.now() - initialTime) / 1000
       this.context.stdout.write(renderSuccessfulCommand(results, totalTime, this.dryRun))
       metricsLogger.logger.gauge('duration', totalTime)
--- a/src/commands/dsyms/upload.ts
+++ b/src/commands/dsyms/upload.ts
@@ -67,7 +67,7 @@ export class UploadCommand extends Command {
       datadogSite: this.config.datadogSite,
       metricsLogger: metricsLogger.logger,
     })
-    const payloads = await getMatchingDSYMFiles(searchPath)
+    const payloads = await getMatchingDSYMFiles(searchPath, this.context)
     const validPayloads = payloads.filter((payload) => payload !== undefined) as Dsym[]
     const requestBuilder = this.getRequestBuilder()
     const uploadDSYM = this.uploadDSYM(requestBuilder, metricsLogger, apiKeyValidator)
--- a/src/commands/dsyms/utils.ts
+++ b/src/commands/dsyms/utils.ts
@@ -28,7 +28,7 @@ export const isZipFile = async (filepath: string) => {
 
 export const getMatchingDSYMFiles = async (
   absoluteFolderPath: string,
-  context: BaseContext | undefined
+  context: BaseContext
 ): Promise<(Dsym | undefined)[]> => {
   const dSYMFiles = await globAsync(buildPath(absoluteFolderPath, '**/*.dSYM'), {})
 
@@ -38,7 +38,7 @@ export const getMatchingDSYMFiles = async (
 
       return new Dsym(dSYMPath, uuids)
     } catch {
-      context?.stdout.write(renderInvalidDsymWarning(dSYMPath))
+      context.stdout.write(renderInvalidDsymWarning(dSYMPath))
 
       return undefined
     }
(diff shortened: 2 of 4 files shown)


src/commands/dsyms/utils.ts
```diff
@@ -24,16 +24,20 @@ export const isZipFile = async (filepath: string) => {
...
-    dSYMFiles.map(async (dSYMPath) => {
+  const allDsyms = dSYMFiles.map(async (dSYMPath) => {
+    try {
       const uuids = await dwarfdumpUUID(dSYMPath)
 
       return new Dsym(dSYMPath, uuids)
-    })
-  )
+    } catch {
+      return undefined
```
ncreated, line 36: This leaves us and the user blind on the issue as we're silently ignoring it - can we emit a CLI warning if we find a file that we cannot process?

Also, is there any in-build way for collecting telemetry from `datadog-ci`? cc @lraucy

lraucy: Yes we can use the metricsLogger: https://github.com/DataDog/datadog-ci/blob/d46f5675a280362797deb66fbe4276f0a6e43544/src/commands/dsyms/upload.ts#L50

buranmert: i added a user-facing warning message for invalid dsym files but i haven't used `metricsLogger`, because i don't see how we can improve our tool with this information. do you have any ideas?

src/commands/dsyms/upload.ts
```diff
@@ -67,11 +67,12 @@ export class UploadCommand extends Command {
       datadogSite: this.config.datadogSite,
       metricsLogger: metricsLogger.logger,
     })
-    const payloads = await getMatchingDSYMFiles(searchPath)
+    const payloads = await getMatchingDSYMFiles(searchPath, this.context)
+    const validPayloads = payloads.filter((payload) => payload !== undefined) as Dsym[]
```
m-rousse, line 71: It is not necessary to cast as `Dsym[]` as typescript infers it correctly
```suggestion
    const validPayloads = payloads.filter((payload) => payload !== undefined)
```

src/commands/dsyms/utils.ts
```diff
@@ -24,16 +26,25 @@ export const isZipFile = async (filepath: string) => {
   }
 }
 
-export const getMatchingDSYMFiles = async (absoluteFolderPath: string): Promise<Dsym[]> => {
-  const dSYMFiles = await globAsync(buildPath(absoluteFolderPath, '**/*.dSYM'))
+export const getMatchingDSYMFiles = async (
+  absoluteFolderPath: string,
+  context: BaseContext | undefined
```
m-rousse, line 31: It'd be best to always define `context` and use a mock in the tests instead, cf [upload.test.ts from sourcemaps](https://github.com/DataDog/datadog-ci/blob/2ba2e690a11f5974c4c22c2fa36b0b9c8446b6cb/src/commands/sourcemaps/__tests__/upload.test.ts#L89-L90).

m-rousse (commented):
Ok from infra PoV

ncreated (approved):

src/commands/dsyms/utils.ts
ncreated: > because i don't see how we can improve our tool with this information. do you have any ideas?

I'm not sure if we're handling all possible cases. For example, when working on the Python symbolicator POC I found that [some symbols need a different (nested) path being passed to `dwarfdump`](https://github.com/DataDog/ios-crash-symbolication-poc/blob/542fc5f54339d796429b0f20237c1661f32928a5/backend-symbolication/src/symbols_registry.py#L119-L120). Not sure if this corresponds only to system symbols (like `UIKitCore.framework` in my POC), or maybe customer apps might also produce it differently in some edge-case setup. Having sort of visibility might help us spot issues earlier, otherwise we'd need to receive feedback and work with the customer to catch cases like that. As we don't use `datadog-ci` for system symbols, and I couldn't observe it for user symbols, we can consider printed warning "good enough" for now.

buranmert: @ncreated that's a good point. in fact, by default `Debug` configuration doesn't produce `DWARF with dSYM` and puts debug information (`DWARF`) to the executable binary.

```
➜ dwarfdump --uuid {BUILD_DIR}/Debug-iphonesimulator/Datadog.framework/Datadog
UUID: 02C05A85-1536-382E-B1CE-A3BC83F791C9 (x86_64) Datadog
```

**can we process such executables with DWARF in our backend? should we upload such binaries as well?**

[looks like Crashlytics doesn't work with non-dSYM files](https://stackoverflow.com/questions/29209582/make-sure-your-project-build-settings-are-generating-a-dsym-file-debug-informat#31511058) so i guess we can ignore this case. wdyt?

src/commands/dsyms/upload.ts
buranmert: maybe this is a problem with my local but it can't infer properly and gives me an error without casting.
is this a blocker issue? if so, i can try to investigate further

src/commands/dsyms/utils.ts
ncreated: > can we process such executables with DWARF in our backend? should we upload such binaries as well?

IMO no need to look into this, as (any) crash reporting doesn't make much sense in `Debug` configuration (it usually runs with debugger attached; `Debug` builds cannot be submitted to App Store). But `DWARF with dSYM` option is a necessary thing to mention in our Crash Reporting doc 👍 even though it's a default in `Release` configuration.

buranmert: i thought for some reason someone may use `DWARF` in their non-Debug builds. this should be an edge case anyway 👌 .

i left a comment about `Debug Information Format` settings in the related JIRA ✅ 
Resolving this convo.

src/commands/dsyms/upload.ts
m-rousse: This not a blocking issue, feel free to discard!

PR #374, opened 2021-09-22
[lambda][uninstrument] Create uninstrument command

duncanista:
### What and why?

Customers can use the instrument command to enable instrumentation, but when they are done with the trial or need to disable the instrumentation for some reasons, there is no easy way to revert the changes made by the instrument command.

### How?

Implement the `uninstrument` command which should revert what the `instrument` command does, mostly remove environment variables, layers and probably tags.

### Review checklist

### DIFF
--- a/src/commands/lambda/index.ts
+++ b/src/commands/lambda/index.ts
@@ -1,3 +1,4 @@
 import {InstrumentCommand} from './instrument'
+import { UninstrumentCommand } from './uninstrument'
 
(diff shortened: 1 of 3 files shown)


maxday:
It would be interesting to have a before/after code coverage report to make sure that the refactoring is not decreasing the coverage of the existing command instrument

maxday:
This looks great!
Can you paste a before/after coverage result?

PR #375, opened 2021-09-23
SLS-1665: Support for git sha upload

ajsaavedra:
### What and why?

This enables setting `git.commit.sha` on lambda functions and uploads git data.

### How?

Users must use the `--sourceCodeIntegration` flag (or simply `-sci`) when calling `lambda instrument`. This will automatically fetch their git status and latch onto the latest commit sha. Users must have a clean git status for the upload to function.

### Testing Guidelines
In my personal repo, I deployed a broken lambda function using CDK: https://github.com/ajsaavedra/sample-python-cdk/tree/082acbafab49472eb66b0e1f0d6d84ec581b7055

Using the latest commit sha, I called the updated `datadog-ci` library using
```
aws-vault exec sandbox-account-admin -- node ../datadog-ci/dist/index.js lambda instrument -r us-east-1 -v 46 -e 10 -f Tony-Sample-Python-CdkSta-samplepythonlambda3C9E28-b0WuiNPvLDgl --extra-tags custom-tag:hello,other:tag:to:apply,test_tag:applied,this-tag:is:valid --sci
```

This tags the lambda telemetry with the git commit sha:

https://a.cl.ly/z8urgk1d

and enriches the Datadog UI:

https://a.cl.ly/kpunx4W4

Support link to example traces [shown here](https://support-admin.us1.prod.dog/admin/switch_handle_get/org_id/433174?next_url=%2Fapm%2Ftraces%3Fquery%3Denv%253Anone%2520service%253Aaws.lambda%2520operation_name%253Aaws.lambda%2520resource_name%253ATony-Sample-Python-CdkSta-samplepythonlambda3C9E28-b0WuiNPvLDgl%26cols%3Dcore_service%252Ccore_resource_name%252Clog_duration%252Clog_http.method%252Clog_http.status_code%26historicalData%3Dtrue%26messageDisplay%3Dinline%26sort%3Ddesc%26spanViewType%3Dmetadata%26streamTraces%3Dtrue%26start%3D1632356861333%26end%3D1632360461333%26paused%3Dtrue).

### Review checklist


src/commands/lambda/instrument.ts
```diff
@@ -323,3 +401,4 @@ InstrumentCommand.addOption('service', Command.String('--service'))
 InstrumentCommand.addOption('environment', Command.String('--env'))
 InstrumentCommand.addOption('version', Command.String('--version'))
 InstrumentCommand.addOption('extraTags', Command.String('--extra-tags'))
+InstrumentCommand.addOption('sourceCodeIntegration', Command.Boolean('-sci,--sourceCodeIntegration'))
```
nhinsch, line 404: Nit: Let's keep the flag consistent with the others, i.e. `--source-code-integration` instead of `--sourceCodeIntegration`

src/commands/lambda/instrument.ts
```diff
@@ -55,6 +61,16 @@ export class InstrumentCommand extends Command {
       return 1
     }
 
+    if (this.sourceCodeIntegration) {
+      if (!process.env.DATADOG_API_KEY) {
```
nhinsch, line 65: I think we want to support `DATADOG_KMS_API_KEY` as well here

src/commands/lambda/instrument.ts
```diff
@@ -127,6 +143,46 @@ export class InstrumentCommand extends Command {
...
+    const status = await simpleGit.status()
+
+    return {isClean: status.isClean(), ahead: status.ahead, files: status.files, hash: gitCommitInfo?.hash}
+  }
+
+  private async getGitDataAndUpload(settings: InstrumentationSettings) {
+    const currentStatus = await this.getCurrentGitStatus()
+
+    if (currentStatus === 1) {
+      return 1
```
nhinsch, line 161: Instead of returning 0 or 1 from a function to indicate whether there was a problem, it would be more idiomatic in TypeScript to throw an error if something goes wrong, and return nothing if all is well.

src/commands/lambda/instrument.ts
IvanTopolcic: It looks like the cases are actually more mixed here than we think:

```
InstrumentCommand.addPath('lambda', 'instrument')
InstrumentCommand.addOption('functions', Command.Array('-f,--function'))
InstrumentCommand.addOption('regExPattern', Command.String('--functions-regex'))
InstrumentCommand.addOption('region', Command.String('-r,--region'))
InstrumentCommand.addOption('extensionVersion', Command.String('-e,--extensionVersion'))
InstrumentCommand.addOption('layerVersion', Command.String('-v,--layerVersion'))
InstrumentCommand.addOption('layerAWSAccount', Command.String('-a,--layerAccount', {hidden: true}))
InstrumentCommand.addOption('tracing', Command.String('--tracing'))
InstrumentCommand.addOption('mergeXrayTraces', Command.String('--mergeXrayTraces'))
InstrumentCommand.addOption('flushMetricsToLogs', Command.String('--flushMetricsToLogs'))
InstrumentCommand.addOption('dryRun', Command.Boolean('-d,--dry'))
InstrumentCommand.addOption('configPath', Command.String('--config'))
InstrumentCommand.addOption('forwarder', Command.String('--forwarder'))
InstrumentCommand.addOption('logLevel', Command.String('--logLevel'))

InstrumentCommand.addOption('service', Command.String('--service'))
InstrumentCommand.addOption('environment', Command.String('--env'))
InstrumentCommand.addOption('version', Command.String('--version'))
InstrumentCommand.addOption('extraTags', Command.String('--extra-tags'))
InstrumentCommand.addOption('sourceCodeIntegration', Command.Boolean('-sci,--sourceCodeIntegration'))
```

Should we go with kebab-case or camel-case? I'm leaning towards the former but I'm biased because more of the CLI tools i've used use hyphens.

src/commands/lambda/instrument.ts
IvanTopolcic: Agreed, but it looks like this file actually has 15 cases of `return 1` and 3 cases of `return 0`. I think it might make more sense to keep this convention, at least for now. Do you agree?

src/commands/lambda/instrument.ts
IvanTopolcic: Fixed, let me know if it looks good!

IvanTopolcic:
Just re-verified that these changes work with a test lambda:

image

Everything from my end looks good here but I'm not sure what else we should be testing.

IvanTopolcic:
Verified that these changes are working after merging the latest master and fixing some parts of the code. From my perspective, this looks good to merge (barring addressing the PR comments) but let me know what y'all think!

src/commands/lambda/instrument.ts
```diff
@@ -53,6 +58,18 @@ export class InstrumentCommand extends Command {
       return 1
     }
 
+    if (this.sourceCodeIntegration) {
+      if (!process.env.DATADOG_API_KEY && !process.env.DATADOG_KMS_API_KEY) {
```
nhinsch, line 62: I think we are adding support for a secret ARN as well -- I would coordinate with @duncanista on this

IvanTopolcic: Sounds good -- maybe we should pull this out into a helper function then so we don't need to re-implement everywhere.

src/commands/lambda/instrument.ts
nhinsch: I don't really think so -- I don't think returning 0 and 1 from functions is idiomatic in TypeScript, we should fix it or at the very least not make the problem worse.

src/commands/lambda/instrument.ts
nhinsch: Per real-life convo, never mind (this is a bit of a different use case)

src/commands/lambda/instrument.ts
nhinsch: Per convo, let's use kebab case

IvanTopolcic: Going with kebab-case after discussion

src/commands/lambda/instrument.ts
nhinsch: Thinking more about this, we should check to make sure that DATADOG_KMS_API_KEY is supported here as well (it might not be)

nhinsch: In fact, I'm almost sure it's not

nhinsch:
I would suggest adding tests for this feature.

src/commands/lambda/instrument.ts
IvanTopolcic: After a discussion in person, it probably doesn't make sense to keep DATADOG_KMS_API_KEY here. Thanks for clarifying!

src/commands/lambda/instrument.ts
```diff
@@ -343,3 +415,4 @@ InstrumentCommand.addOption('service', Command.String('--service'))
 InstrumentCommand.addOption('environment', Command.String('--env'))
 InstrumentCommand.addOption('version', Command.String('--version'))
 InstrumentCommand.addOption('extraTags', Command.String('--extra-tags'))
+InstrumentCommand.addOption('sourceCodeIntegration', Command.Boolean('-sci,--source-code-integration'))
```
nhinsch, line 418: I think the short version of the flag (`-sci`) should be one letter long (`-s`). Usually these short flags are only one letter long, and can be combined, so `-sci` could also be `-s -c -i`. See https://nullprogram.com/blog/2020/08/01/

nhinsch (approved):
This looks good! Let's just switch the short version of the flag to `-s`.

src/commands/lambda/instrument.ts
IvanTopolcic: Agreed, the previous PR had `-sci` so I didn't know whether or not this discussion was already had. 👍

PR #384, opened 2021-09-28
[s8s] Create junit report even in directory

yoannmoinet:
### What and why?

When trying to create a jUnit report in a directory, it's not working if the directory in question doesn't exist.

image

Also, when reporting errors, new steps where overwriting over previous steps.

### How?

- First create the directory.
- Push errors instead of overwriting the full array.

### Review checklist

Published in `0.17.2-alpha`

### DIFF
--- a/src/commands/synthetics/reporters/junit.ts
+++ b/src/commands/synthetics/reporters/junit.ts
@@ -41,16 +41,16 @@ interface XMLTestCaseProperties extends Stats {
   timestamp: string
 }
 
-interface XMLTestCase {
+export interface XMLTestCase {
   $: XMLTestCaseProperties
   // These are singular for a better display in the XML format of the report.
-  browser_error?: XMLError[]
+  browser_error: XMLError[]
   error: XMLError[]
   properties: {
     property: {$: {name: string; value: any}}[]
   }
   testcase: XMLStep[]
-  warning?: XMLError[]
+  warning: XMLError[]
 }
 
 interface XMLStepProperties extends Stats {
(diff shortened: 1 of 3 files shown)


jakub-g (approved):

src/commands/synthetics/reporters/junit.ts
```diff
@@ -163,15 +165,15 @@ export class JUnitReporter implements Reporter {
         // It's a browser test.
         for (const stepDetail of result.result.stepDetails) {
           const {browser_error, error, warning} = this.getBrowserTestStep(stepDetail)
-          testCase.browser_error = browser_error
-          testCase.error = error
-          testCase.warning = warning
+          testCase.browser_error.push(...browser_error)
+          testCase.error.push(...error)
+          testCase.warning.push(...warning)
```
m-rousse, line 170: Does it mean errors from all steps will be merged in the test case? (I do not know junit really well but it seems difficult to interpret afterwards if it is the case)

yoannmoinet: Yes, that's what it does.
Since we're only reporting at the test level (and not the step level) we're reporting all the error at the top level.

src/commands/synthetics/reporters/junit.ts
```diff
@@ -131,6 +132,7 @@ export class JUnitReporter implements Reporter {
     // Write the file
     try {
       const xml = this.builder.buildObject(this.json)
+      await fs.mkdir(path.dirname(this.destination), {recursive: true})
```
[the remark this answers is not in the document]
yoannmoinet: I'll add a test case to confirm that.

src/commands/synthetics/reporters/junit.ts
m-rousse: Ok, as long as the step tied to the error is reported, then I think it is OK.

PR #388, opened 2021-09-30
[s8s] Fix jUnit report with Timeout errors

yoannmoinet:
### What and why?

Currently, the jUnit report doesn't consider timeouts as errors in its reporting.

### How?

- Type errors
- Catch timeouts as errors in the reporting

### Review checklist

### DIFF
--- a/src/commands/synthetics/interfaces.ts
+++ b/src/commands/synthetics/interfaces.ts
@@ -20,6 +20,12 @@ export interface MainReporter {
   testWait(test: Test): void
 }
 
+export enum ERRORS {
+  TIMEOUT = 'Timeout',
+  ENDPOINT = 'Endpoint Failure',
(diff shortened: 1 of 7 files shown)


src/commands/synthetics/interfaces.ts
```diff
@@ -43,7 +49,7 @@ export interface BrowserTestResult extends TestResult {
     width: number
   }
   duration: number
-  error?: string | 'Endpoint Failure' | 'Timeout' | 'Tunnel Failure'
+  error?: string | ERRORS
```
m-rousse, line 52: No change required: we moved away from enum to union, however in that case, having `string | 'value' | ...` defeats the purpose of the union, we should change that (and this is another point in favor of sharing synthetics typings)

m-rousse (approved):
LGTM :+1:

PR #395, opened 2021-10-04
[ci-app] Capture error message

edrevo:
### What and why?

- Adds error messages and exit code to the payload for custom spans
- Adds retries to HTTP intake for the trace command

### How?

The upload code in the utils package has been refactored into a retry package to have a common retry policy among commands. The trace command as well as the source code integration commands and the junit commands are now using this common retry logic.

### Review checklist

### DIFF
--- a/src/commands/trace/trace.ts
+++ b/src/commands/trace/trace.ts
@@ -1,14 +1,12 @@
-import retry from 'async-retry'
 import chalk from 'chalk'
 import {spawn} from 'child_process'
 import {Command} from 'clipanion'
 import crypto from 'crypto'
 import os from 'os'
+import {retryRequest} from '../../helpers/retry'
 import {parseTags} from '../../helpers/tags'
 import {apiConstructor} from './api'
 import {APIHelper, CIRCLECI, JENKINS, Payload, Provider, SUPPORTED_PROVIDERS} from './interfaces'
-
-const errorCodesNoRetry = [400, 403, 413]
 
 // We use 127 as exit code for invalid commands since that is what *sh terminals return
 const BAD_COMMAND_EXIT_CODE = 127
@@ -111,39 +109,6 @@ export class TraceCommand extends Command {
     return exitCode
   }
 
-  private async reportCustomSpan(payload: Payload, provider: Provider) {
-    const api = this.getApiHelper()
-    try {
-      await retry(
-        async (bail) => {
-          try {
-            await api.reportCustomSpan(payload, provider)
-          } catch (error) {
-            const util = require('util')
-            if (error.response) {
-              // If it's an axios error
-              if (!errorCodesNoRetry.includes(error.response.status)) {
-                // And a status code that is not excluded from retries, throw the error to retry
-                throw error
-              }
-            }
-            // If it's another error or an axios error let us retry just in case
-            // This will catch DNS resolution errors and connection timeouts
-            throw error
-          }
-        },
-        {
-          onRetry: (e, attempt) => {
-            this.context.stderr.write(chalk.yellow(`[attempt ${attempt}] Could not report custom span. Retrying...: ${e.message}\n`))
-          },
-          retries: 5,
-        }
-      )
-    } catch (error) {
-      this.context.stderr.write(chalk.red(`Failed to report custom span: ${error.message}\n`))
-    }
-  }
-
   public getCIEnvVars(): [Record<string, string>, Provider?] {
     if (process.env.CIRCLECI) {
       return [
@@ -236,6 +201,22 @@ export class TraceCommand extends Command {
 
   private getEnvironmentVars(keys: string[]): Record<string, string> {
     return keys.filter((key) => key in process.env).reduce((accum, key) => ({...accum, [key]: process.env[key]!}), {})
+  }
+
+  private async reportCustomSpan(payload: Payload, provider: Provider) {
+    const api = this.getApiHelper()
+    try {
+      await retryRequest(() => api.reportCustomSpan(payload, provider), {
+        onRetry: (e, attempt) => {
+          this.context.stderr.write(
+            chalk.yellow(`[attempt ${attempt}] Could not report custom span. Retrying...: ${e.message}\n`)
+          )
+        },
+        retries: 5,
+      })
+    } catch (error) {
+      this.context.stderr.write(chalk.red(`Failed to report custom span: ${error.message}\n`))
+    }
   }
 
   private signalToNumber(signal: NodeJS.Signals | null): number | undefined {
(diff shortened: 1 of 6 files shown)


src/helpers/__tests__/retry.test.ts
```diff
@@ -0,0 +1,86 @@
...
+    try {
+      await retryRequest(createResultWithErrors([buildHttpError(413), buildHttpError(413)]), {
+        onRetry: retryCallback,
+        retries: 1,
+      })
+    } catch (error) {
+      threwError = true
+    }
+    expect(threwError).toBeTruthy()
+    expect(retryCallback).toBeCalledTimes(0)
```
lraucy, line 76: Shouldn't `retryCallback` be `1`? (probably need `buildHttpError(500)` for it to work)

Also I wonder if we should add a test where `retryCallback > 1`? (all the tests here only show 0 or 1 retry)

src/helpers/retry.ts
```diff
@@ -0,0 +1,28 @@
...
+      await requestBuilder()
+    } catch (error) {
+      if (error.response) {
+        // If it's an axios error
+        if (!errorCodesNoRetry.includes(error.response.status)) {
+          // And a status code that is not excluded from retries, throw the error so that upload is retried
+          throw error
+        }
+      } else {
+        // If it's another error or an axios error let us retry just in case
```
lraucy, line 18: For the sake of clarity what do you think about the following:

```typescript
if (error.response) {
    if (errorCodesNoRetry.includes(error.response.status)) {
        bail(error)
    }
}
// Other cases are retried: other axios errors as well as
// non-axios errors such as DNS resolution errors and connection timeouts
throw error

```

lraucy (commented):
Thank you for doing this! A few comments, especially around the tests, otherwise lgtm for the shared upload code.

src/helpers/__tests__/retry.test.ts
edrevo: This test was buggy because 413 is a non-retryable error. Good catch. I have fixed it now.

src/helpers/retry.ts
edrevo: Fixed

src/helpers/__tests__/retry.test.ts
```diff
@@ -0,0 +1,99 @@
...
+        return Promise.resolve({
+          config: {},
+          data: {},
+          headers: undefined,
+          status: 200,
+          statusText: '',
+        })
+      } else {
+        return Promise.reject(errors[i])
+      }
```
juan-fernandez, line 21:
```suggestion
      }
      return Promise.reject(errors[i])
```
no need for else here

src/helpers/__tests__/retry.test.ts
```diff
@@ -0,0 +1,99 @@
...
+    return () => {
+      i = i + 1
+      if (errors[i] === undefined) {
+        return Promise.resolve({
+          config: {},
+          data: {},
+          headers: undefined,
+          status: 200,
+          statusText: '',
+        })
```
juan-fernandez, line 18: are you ever reading the resolved value? We could just 
```suggestion
        return Promise.resolve()
```

src/helpers/__tests__/retry.test.ts
```diff
@@ -0,0 +1,99 @@
...
+  test('should not retry if the call was successful', async () => {
+    await retryRequest(createResultWithErrors([undefined]), {
+      maxTimeout: 50,
+      minTimeout: 10,
+      onRetry: retryCallback,
+      retries: 5,
+    })
+    expect(retryCallback).toBeCalledTimes(0)
+  })
+})
```
juan-fernandez, line 99: great tests!

juan-fernandez (changes requested):
looking good! Only real concern is a potential exception. The rest are minor comments

lraucy (approved):

src/commands/trace/trace.ts
```diff
@@ -53,7 +54,16 @@ export class TraceCommand extends Command {
     const [command, ...args] = this.command
     const id = crypto.randomBytes(5).toString('hex')
     const startTime = new Date().toISOString()
-    const childProcess = spawn(command, args, {env: {...process.env, DD_CUSTOM_PARENT_ID: id}, stdio: 'inherit'})
+    const childProcess = spawn(command, args, {
+      env: {...process.env, DD_CUSTOM_PARENT_ID: id},
+      stdio: ['inherit', 'inherit', 'pipe'],
```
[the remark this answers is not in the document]
edrevo: Correct. If you're pipeing stderr is null

src/helpers/__tests__/retry.test.ts
edrevo: It will complain that it is not the correct type.

src/commands/trace/trace.ts
```diff
@@ -63,15 +73,17 @@ export class TraceCommand extends Command {
         resolve([exitStatus, exitSignal])
       })
     })
+
+    const stderr: string = await strerrCatcher
+    this.context.stderr.write(stderr)
```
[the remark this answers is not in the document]
edrevo: Very cool. Didn't know you could do that!

src/commands/trace/trace.ts
```diff
@@ -53,7 +54,16 @@ export class TraceCommand extends Command {
...
+    })
+    const chunks: Buffer[] = []
+    const strerrCatcher: Promise<string> = new Promise((resolve, reject) => {
+      childProcess.stderr.on('data', (chunk) => chunks.push(Buffer.from(chunk)))
+      childProcess.stderr.on('error', (err) => reject(err))
+      childProcess.stderr.on('end', () => resolve(Buffer.concat(chunks).toString('utf8')))
+    })
     const [status, signal] = await new Promise((resolve, reject) => {
       childProcess.on('error', (error: Error) => {
         reject(error)
```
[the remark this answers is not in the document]
juan-fernandez: that's right, sry

juan-fernandez (approved):

src/helpers/retry.ts
```diff
@@ -0,0 +1,25 @@
+import retry from 'async-retry'
+
+const errorCodesNoRetry = [400, 403, 413]
+
+export const retryRequest = async (requestBuilder: () => Promise<any>, retryOpts: retry.Options): Promise<void> => {
+  // Request function, passed to async-retry
+  const doRequest = async (bail: (e: Error) => void) => {
+    try {
+      await requestBuilder()
```
m-rousse, line 9: I think it was not introduced by your PR, however the naming `request builder` is used to designate functions to build "functions that will perform queries", which does not seem to be the case here as it is directly a function performing a request.
I'd suggest to rename it to `requestPerformer`

m-rousse (approved):
Thanks for this PR and factorization :+1:

PR #412, opened 2021-10-25
[lambda] support ARM layers

duncanista:
### What and why?

Refer to [SLS-1390](https://datadoghq.atlassian.net/browse/SLS-1390)

### How?

Refactor `getLayerArn` and `getExtensionArn` to support this feature, and move them into `commons.ts`.
Moved tests to their corresponding file, and added some more according to the feature.

### Review checklist

### DIFF
--- a/src/commands/lambda/functions/commons.ts
+++ b/src/commands/lambda/functions/commons.ts
@@ -1,7 +1,17 @@
 import {CloudWatchLogs, Lambda} from 'aws-sdk'
 import {GetFunctionRequest} from 'aws-sdk/clients/lambda'
-import {MAX_LAMBDA_STATE_CHECK_ATTEMPTS, Runtime, RUNTIME_LAYER_LOOKUP} from '../constants'
-import {FunctionConfiguration} from '../interfaces'
+import {
+  ARM64_ARCHITECTURE,
+  ARM_LAYER_SUFFIX,
+  ARM_RUNTIMES,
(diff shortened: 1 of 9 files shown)


src/commands/lambda/functions/commons.ts
```diff
@@ -76,6 +112,36 @@ export const getLambdaFunctionConfigs = (
...
+/**
+ * Returns the correct ARN of a **Specific Runtime Layer** given its configuration, region,
+ * and settings (optional).
+ *
+ * @param config a Lambda FunctionConfiguration.
+ * @param region a region where the layer is hosted.
+ * @param settings instrumentation settings, mainly used to change the AWS account that contains the Layer.
+ * @returns the ARN of a **Specific Runtime Layer** with the correct region, account, architecture, and name.
+ */
+export const getLayerArn = (
```
maxday, line 124: it seems that this function looks a lot like `getExtensionArn`? Maybe we should refactor?

maxday (changes requested):
it looks like two functions are doing the same thing? can you have a 👀 ?

src/commands/lambda/functions/commons.ts
duncanista: I refactored the function to be able to get the Extension, let me know what you think about the followed approach.

maxday (approved):

m-rousse (approved):

PR #421, opened 2021-11-01
[lambda][uninstrument] Add multiple uninstrumentation with regex pattern

duncanista:
### What and why?

Refer to [SLS-1646](https://datadoghq.atlassian.net/jira/people/5d4b49dba23f060c108396ed/boards/1491?modal=detail&selectedIssue=SLS-1646).

### How?

Add condition to check if regex pattern has been set in order to uninstrument searching by function name if it matches the pattern. Update `uninstrument.ts` to function properly. Add tests for uninstrumenting with regex pattern and update current ones.

### Review checklist


src/commands/lambda/functions/uninstrument.ts
```diff
@@ -70,6 +71,45 @@ export const getFunctionConfig = async (
   }
 }
 
+export const getFunctionConfigsFromRegEx = async (
```
maxday, line 74: it looks like this function is very similar to : https://github.com/DataDog/datadog-ci/blob/e56249423b2b6bf637b66aea5e36471b3d50e047/src/commands/lambda/functions/instrument.ts#L91
maybe this could be refactored?

maxday (changes requested):
I left a question :)

src/commands/lambda/functions/uninstrument.ts
```diff
@@ -8,6 +8,7 @@ import {
   HANDLER_LOCATION,
   KMS_API_KEY_ENV_VAR,
   LAMBDA_HANDLER_ENV_VAR,
+  LIST_FUNCTIONS_MAX_RETRY_COUNT,
```
IvanTopolcic, line 11: Do we remove the sourceCodeIntegration flag? Now that the flag is merged it, we should make sure that we're also uninstrumenting it.

src/commands/lambda/instrument.ts
```diff
@@ -4,7 +4,7 @@ import {Command} from 'clipanion'
 import {parseConfigFile} from '../../helpers/utils'
```
IvanTopolcic, line 4: What's supposed to happen if uninstrumentation fails on one lambda but succeeds on the others?

IvanTopolcic (commented):
Looks good! Just some small questions.

src/commands/lambda/instrument.ts
duncanista: I think it will tell the user that everything failed, we might have to consider catching the error and stating that it failed only in certain ones, will take a look into it, but won't do it in this PR.

IvanTopolcic: That sounds great!

src/commands/lambda/functions/uninstrument.ts
duncanista: I just checked, and we are already uninstrumenting it by removing the DD_TAGS where the flags might appear.

src/commands/lambda/functions/uninstrument.ts
duncanista: Just refactored the function, there was some code that could be used by both instrument/uninstrument, even though I reduced the use of repetitive code, there might be still some of it, please let me know what you think of the modification

maxday (approved):
LGTM!

PR #423, opened 2021-11-04
[SW-1695] explicitely export module required externally

etnbrd:
### What and why?

https://github.com/DataDog/synthetics-ci-github-action depends on some modules from `datadog-ci`. Until now, these dependencies were imported directly from the files, instead of going through a proper export.
Partly because of that, a change in `datadog-ci` led to a bug in the source of https://github.com/DataDog/synthetics-ci-github-action.

### Review checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,6 +19,7 @@
   "publishConfig": {
     "access": "public"
   },
+  "main": "dist/main.js",
   "resolutions": {
     "ini": "1.3.7",
     "kind-of@^6.0.0": "6.0.3"
--- a/src/main.ts
+++ b/src/main.ts
@@ -1,0 +1,4 @@
+import * as Synthetics from './commands/synthetics/main'
+import {parseConfigFile} from './helpers/utils'
+
+export {parseConfigFile, Synthetics}
--- a/src/commands/synthetics/main.ts
+++ b/src/commands/synthetics/main.ts
@@ -1,0 +1,5 @@
+export {CiError} from './errors'
+export * from './interfaces'
+export {DefaultReporter} from './reporters/default'
+export {executeTests} from './run-test'
+export * as utils from './utils'


m-rousse (approved):
This LGTM from Synthetics' PoV, though I'd like to have some 👍 of other commands before merging.
Could you send a message to ensure that's OK for them as well?

package.json
```diff
@@ -19,6 +19,7 @@
   "publishConfig": {
     "access": "public"
   },
+  "main": "dist/main.js",
```
yoannmoinet, line 22: There's already a `main` in there.
Is it correct?

etnbrd: Ah, I completely missed this 🤦 
Though, I am not sure we want to build all the CLI when importing `datadog-ci`. Maybe we could rename `index` to `cli` and keep `index` be what is know `main`? What do you think?

yoannmoinet: I think `dist/index.js` was never meant to be imported.
So I think you can remove it altogether, and keep `"main": "dist/main.js"` only.
The `cli` part is covered by `bin` already.

yoannmoinet: Ah misread your `cli` comment.
I'm fine either way.

For the sake of this PR, I'd keep `dist/index.js` and investigate the renaming in a different PR if ever needed.

src/main.ts
```diff
@@ -0,0 +1,4 @@
+import * as Synthetics from './commands/synthetics/main'
+import {parseConfigFile} from './helpers/utils'
+
+export {parseConfigFile, Synthetics}
```
yoannmoinet, line 4: Not sure about that design though.

I don't see it scaling very well with other commands.

Also, why the uppercase on `Synthetics`? It's not a class.

I believe this needs to be discussed a bit more and prepared for the future as well.
This will become the main entrypoint of datadog-ci as a lib, it needs to be thought out. cc @m-rousse 

IMO, we should simplify this, having a `index.ts` in each command directory that will handle what's getting exported to the lib (and rename the current `index.ts` as `cli.ts` for the automatic cli registration).

Then, in the root `package.json` we keep the `"main"` to be `"dist/index.ts"`.
Renaming the current `index.ts` to `cli.ts` and update the `"bin"`, as suggested by @etnbrd in another comment.

etnbrd: > Also, why the uppercase on `Synthetics`? It's not a class.
I thought it would be some kind of namespace. I really don't have any opinion on the casing of this.

> IMO, we should simplify this, having an `index.ts` in each command directory that will handle what's getting exported to the lib (and rename the current index.ts as cli.ts for the automatic cli registration).

How would you change the current design from the existing? `index.ts` is currently being used by the commands: https://github.com/DataDog/datadog-ci/blob/master/src/index.ts#L22. I'm happy to turn all the `index.ts` into `cli.ts` but indeed, we might want to give this some thoughts.

An alternative would be to extract all the synthetics part into another package, which would be a lib and not a CLI, and import this package in `datadog-ci` which is only a CLI and not a lib.

m-rousse: Agreed it need further planning and I think the best way to do so, both laying all solutions + discussing them would be a RFC. (not necessarily a long one, but I think it's the best way to communicate through this)
Let's create one and discuss it with all owners/maintainers of commands.

etnbrd:
Closed in favor of https://github.com/DataDog/datadog-ci/pull/433

PR #427, opened 2021-11-12
[CIAPP-3236] Use correct env vars to build GitHub URLs

cg14823:
### What and why?

- Use `${GITHUB_SERVER_URL}` instead of hardcoding `https://github.com` this will allow support for GitHub Enterpise.
- Use correct url for GitHub action workflows.

### Review checklist


src/helpers/__tests__/ci-env/github.json
```diff
@@ -8,14 +8,16 @@
...
       "GITHUB_SHA": "ghactions-commit",
       "GITHUB_WORKFLOW": "ghactions-pipeline-name",
-      "GITHUB_WORKSPACE": "/foo/bar"
+      "GITHUB_WORKSPACE": "/foo/bar",
+      "GITHUB_SERVER_URL": "https://github.com",
+      "GITHUB_RUN_ATTEMPT": "10"
     },
     {
-      "ci.job.url": "https://github.com/ghactions-repo/commit/ghactions-commit/checks",
+      "ci.job.url": "https://github.com/ghactions-repo/actions/runs/ghactions-pipeline-id/attempts/10",
```
drodriguezhdez, line 16: is this aligned with the `spec_tests`?

cg14823: In this case the code just makes the pipeline URL and the Job URL match wasn't sure if I should change it or not. Should I put the job url back to wha tit was so it matches the spec?

drodriguezhdez: I'd say whatever we decide, we should be consistent between pipelines and tests. wdyt?

cg14823: Makes sense I updated this as well as the PR to the spec so that they match.  We still use the check run url but it will always use `GITHUB_SERVER_URL` so that it works on on-prem GitHub enterprise instances

drodriguezhdez (approved):
LGTM

juan-fernandez (changes requested):
it looks good, but I'm blocking to avoid a potential issue. Is the `github.json` here generated automatically from `datadog-ci-spec`? I'm wondering because we added a test case in https://github.com/DataDog/datadog-ci-spec/pull/157, but it doesn't show up here. What I'd do: 

* Merge https://github.com/DataDog/datadog-ci-spec/pull/157 
* Generate updated json files (including `github.json`)
* Use them here 

Let me know if you don't know how to generate the json files 😄

cg14823:
@juan-fernandez I did not know this was autogenerated, could you point me to instructions on how to generate it?

Regarding DataDog/datadog-ci-spec#157  bunch of test for the different tracers are failing not sure what I have to do about it or if it can be merged with the test failing.

juan-fernandez:
> @juan-fernandez I did not know this was autogenerated, could you point me to instructions on how to generate it?
> 
> Regarding [DataDog/datadog-ci-spec#157](https://github.com/DataDog/datadog-ci-spec/pull/157) bunch of test for the different tracers are failing not sure what I have to do about it or if it can be merged with the test failing.

yes, spec PRs can be merged ahead of the tracers. We normally work like this: update spec, break the tests, fix the tests. 
As for how to generate the json files: 
You can either generate the json files locally, with `make json` (https://github.com/DataDog/datadog-ci-spec#generate-json-files), or download the artifacts that are generated within the GHA, for example: 

https://github.com/DataDog/datadog-ci-spec/actions/runs/1452538332#artifacts 

The artifact should always be there when the GHA has run

juan-fernandez (approved):

PR #429, opened 2021-11-12
[sourcemaps] Rework wording and add example for --project-path

rpelliard:
This is an attempt at improving the documentation, so that it's easier to understand the goal of the `--project-path` argument.
- Rework the wording
- Add a sub-section to the _Link errors with your source code_ section to mention the `--project-path` argument, with an usage example

### DIFF
--- a/src/commands/sourcemaps/README.md
+++ b/src/commands/sourcemaps/README.md
@@ -52,11 +52,11 @@ Errors in Datadog UI can be enriched with links to GitHub/GitLab/Bitbucket if th
 - `git` executable is installed
 - `datadog-ci` is run within the git repository
 
-When these requirements are met, the upload command reports Git information such as the current commit hash, the repository URL, and the list of tracked file paths in the code repository.
-Each sourcemap uploaded gets such information associated with it.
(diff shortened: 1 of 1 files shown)


src/commands/sourcemaps/README.md
```diff
@@ -69,6 +69,14 @@ The value can be overriden with `--repository-url`.
 Example: With a remote `git@github.com:Datadog/example.git`, links pointing to `https://github.com/Datadog/example` are generated.
 This behavior can be overriden with links to `https://gitlab.com/Datadog/example` with the flag `--repository-url=https://gitlab.com/Datadog/example`.
 
+#### Setting the project path
```
lraucy, line 72: Maybe we can add a link to this section in the description of the flag above?

lraucy (approved):
Great change, just a minor suggestion.

jybp (approved):

src/commands/sourcemaps/README.md
rpelliard: Just added one 👍

PR #432, opened 2021-11-16
Export datadog-ci as a library - 1/2

etnbrd:
### What and why?

This PR replace the root and all command `index.ts` with `cli.ts` to later leave the `index.ts` be able to export `datadog-ci` as a library.

---

part 1/2 https://github.com/DataDog/datadog-ci/pull/432
part 2/2 https://github.com/DataDog/datadog-ci/pull/433

PR #433, opened 2021-11-16
Export datadog-ci as a library - 2/2

etnbrd:
### What and why?

This PR adds some index files to exports specific modules, and import them with `datadog-ci` as a library.

---

part 1/2 https://github.com/DataDog/datadog-ci/pull/432
part 2/2 https://github.com/DataDog/datadog-ci/pull/433


src/index.ts
```diff
@@ -0,0 +1,4 @@
+import * as synthetics from './commands/synthetics'
+import {parseConfigFile} from './helpers/utils'
+
+export {parseConfigFile, synthetics}
```
yoannmoinet, line 4: I'd namespace `utils`.
```suggestion
import * as utils from './helpers/utils'

export {utils, synthetics}
```

WDYT?

src/commands/synthetics/index.ts
```diff
@@ -0,0 +1,5 @@
+export {CiError} from './errors'
+export * from './interfaces'
```
yoannmoinet, line 2: Should we also namespace as `types` maybe?
```suggestion
export * as types from './interfaces'
```

yoannmoinet (commented):
I think we should be mindful about the way we expose the API of the lib. 
Have namespaces that makes sense and not have everything top-level.

src/commands/synthetics/index.ts
etnbrd: What would that change? Wouldn't that just be more tedious to reference?

src/index.ts
etnbrd: I thought that the config file is important enough to have some function in the root namespace or a `config` namespace, rather than pushed in a `util` namespace which might get crowded and doesn't convey a strong meaning.
No strong opinion, though.

src/commands/synthetics/index.ts
yoannmoinet: Discoverability and general organisation of the library's API.

In the end, I'm not even sure we need these `index.js` files and we should simply import from the actual arborescence, which is less confusing (there is not a secondary, somewhat duplicated API to access some part of the code), less maintenance (you don't have to maintain a public API, it's relying on the current arbo), less documentation (you don't have to document what's exported to the public API and what's not, you only document what's actually usable as a library)...

For instance, I think this is clearer:
```javascript
const {executeTests} = require('@datadog/datadog-ci/commands/synthetics/run-test');
const Reporter = require('@datadog/datadog-ci/commands/synthetics/interfaces');

await executeTests();
const report: synthetics.Reporter = new synthetics.DefaultReporter();
```

The less we try to explicitly control what's exported, the easier it will be for us to maintain and the more freedom the user will have.

yoannmoinet: This is being discussed with @m-rousse and @sdeprez on Slack.
I'll follow along their take on this. 
Just wanted to be sure that we present something clean and explicit to the customer.

m-rousse: FTR as discussed on slack, we prefer exposing only functions that should be used by users of the package.

m-rousse (approved):
LGTM, we'll do a documentation pass in a future PR to make it clear `@datadog/datadog-ci` can be used as a package + which functions are available

PR #437, opened 2021-11-18
[Hackathon] Add GHA for Creating and Distributing Binary

juan-fernandez:
### What and why?
This introduces standalone binary releases. The goal is to remove barrier of entry for customers for which having to install a recent version of node in their CI is problematic. We have seen this, even internally for usage from datadog-agent.

### How?
- Built with https://github.com/vercel/pkg, one binary per OS.
- Had to make some small changes to dynamic requires to make it work.
- A github action will build and upload the assets to GitHub when a release is created.
- We also added a command `datadog-ci version` for quick testing of the binary.
- We also added a quick install script, so customers can just paste one line in their CI or docker image.


juan-fernandez:
closing in favor of https://github.com/DataDog/datadog-ci/pull/485

Issue #438, opened 2021-11-21
Allow configuration of custom junit intake URL


emilgelman:
We have a proxy service that we use to upload Junit tests to Datadog, so we need a way to set a custom value for the junit intake URL - for example `DATADOG_JUNIT_INTAKE_URL`, similarly to `DATADOG_SOURCEMAP_INTAKE_URL`

emilgelman:
https://github.com/DataDog/datadog-ci/pull/439

lraucy:
Hello @emilgelman! Quick question to make sure we understand your use-case properly:

The CLI already supports the env var `HTTPS_PROXY` to configure a proxy (like `HTTPS_PROXY=http://my-proxy-address:3128`).

Would this configuration work for your use case, rather than introducing a new env var `DATADOG_JUNIT_INTAKE_URL`? If not, could you elaborate?

The `DATADOG_SOURCEMAP_INTAKE_URL` env var was mostly designed for internal testing, not for proxy usage.

emilgelman:
hi @lraucy , apologies for the late response. I've tried using `HTTPS_PROXY` for this, yet I'm getting 
[attempt 1] Retrying jUnitXML upload [report.xml]: Socket is closed

I'm getting the same error trying to use `datadog-ci sourcemaps upload`, yet it works when I override with `DATADOG_SOURCEMAP_INTAKE_URL`. Anything else I can try?

lraucy:
Hi @emilgelman , sorry for the late response as well: what kind of proxy are you using? I have trouble to understand how `datadog-ci sourcemaps upload` could upload files to Datadog if you override `DATADOG_SOURCEMAP_INTAKE_URL` with your proxy URL: how does the proxy knows to which URL it should send sourcemaps to?

emilgelman:
Hey @lraucy 
I'll try to provide additional context. To enforce certain standards and perform custom audits, any request towards Datadog's API is being proxied by an internal service we have (we are using Spring Cloud Gateway).

For example, if we need to use the Monitors API, instead of calling `GET api.datadoghq.com/api/v1/monitor`, our users would call `GET api.mycompany.com/v1/monitors`, where certain business logic is executed and then the request is routed to Datadog's API as is, based on the endpoint.

[...]

PR #439, opened 2021-11-21
allow overriding junit intake url with env

emilgelman:
Allow setting a custom junit intake url with `DATADOG_JUNIT_INTAKE_URL` environment variable

### DIFF
--- a/src/commands/junit/README.md
+++ b/src/commands/junit/README.md
@@ -40,6 +40,7 @@ Additionally you might configure the `junit` command with environment variables:
 - `DD_TAGS`: set global tags applied to all spans. The format must be `key1:value1,key2:value2`.
   - The resulting dictionary will be merged with whatever is in the `--tags` parameter. If a `key` appears both in `--tags` and `DD_TAGS`, whatever value is in `DD_TAGS` will take precedence.
 - `DATADOG_SITE`: choose your Datadog site, e.g. datadoghq.com or datadoghq.eu.
+- `DATADOG_JUNIT_INTAKE_URL`: override the full URL for the intake endpoint.
 
 ### Optional dependencies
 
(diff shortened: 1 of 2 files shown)


src/commands/junit/README.md
```diff
@@ -40,6 +40,7 @@ Additionally you might configure the `junit` command with environment variables:
 - `DD_TAGS`: set global tags applied to all spans. The format must be `key1:value1,key2:value2`.
   - The resulting dictionary will be merged with whatever is in the `--tags` parameter. If a `key` appears both in `--tags` and `DD_TAGS`, whatever value is in `DD_TAGS` will take precedence.
 - `DATADOG_SITE`: choose your Datadog site, e.g. datadoghq.com or datadoghq.eu.
+- `DATADOG_JUNIT_INTAKE_URL`: override the full URL for the intake endpoint.
```
juan-fernandez, line 43: is this something we want to show to users? Are we doing it for sourcemaps?

juan-fernandez: it seems more like an internal thing

jybp: FYI in the sourcemaps readme we do have it mentioned [here](https://github.com/DataDog/datadog-ci/tree/master/src/commands/sourcemaps#setup):  
````
It is also possible to override the full URL for the intake endpoint by defining the DATADOG_SOURCEMAP_INTAKE_URL environment variable.
```

juan-fernandez: oh sry I missed that. In any case, is there any reason to show it to users?

jybp: I guess the only reason would be for the proxy case so customers can easily know what to configure.

juan-fernandez (approved):

juan-fernandez (changes requested):
I'm going to block this PR until we figure out if `HTTPS_PROXY` can be used for this use case

emilgelman:
@juan-fernandez any news about this? This is blocking our team from using this feature.

lraucy:
> @juan-fernandez any news about this? This is blocking our team from using this feature.

I just answered directly on the issue thread: https://github.com/DataDog/datadog-ci/issues/438

PR #440, opened 2021-11-22
[lambda] Remove Lambda state check

duncanista:
### What and why?

AWS Lambda state check has no longer a transition from Pending at all.
Therefore, the use of `isLambdaActive` is no longer needed.

Learn more about tracking the state of lambda [here](https://aws.amazon.com/blogs/compute/tracking-the-state-of-lambda-functions/).

### How?

Removing `isLambdaActive`, its references and tests.

### Review checklist


duncanista:
Misunderstood the lambda state check, it is still there. Check more [here](https://aws.amazon.com/blogs/compute/coming-soon-expansion-of-aws-lambda-states-to-all-functions/).

PR #443, opened 2021-11-23
[lambda] Coerce boolean for `mergeXrayTraces`, `flushMetricsToLogs`, and `tracing`

duncanista:
Currently, we were just accepting string booleans for `mergeXrayTraces`, `flushMetricsToLogs`, and `tracing`. If the user specified a boolean in the config file for such fields, it would crash because there is no `.toLowerCase()` method for booleans. 

### What and why?

Coerce booleans to avoid an error when user specifies a value of type `boolean|string` for `mergeXrayTraces`, `flushMetricsToLogs`, and `tracing`.

 
Screen Shot 2021-11-23 at 3 09 45 PM

### How?

Created a function to coerce booleans: `coerceBoolean`.
Add tests and fix current ones.

### Review checklist

### DIFF
--- a/src/commands/lambda/__tests__/functions/commons.test.ts
+++ b/src/commands/lambda/__tests__/functions/commons.test.ts
@@ -80,6 +80,11 @@ describe('commons', () => {
       expect(coerceBoolean(true, 'false', 'true')).toBe(false)
       expect(coerceBoolean(false, 'true', 'False')).toBe(true)
     })
+
+    test('return the first boolean when one of the values provided is boolean', () => {
+      expect(coerceBoolean(true, false, 'truE', true)).toBe(false)
+      expect(coerceBoolean(false, true, 'False', false)).toBe(true)
+    })
   })
   describe('collectFunctionsByRegion', () => {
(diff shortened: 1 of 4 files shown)


src/commands/lambda/__tests__/functions/commons.test.ts
```diff
@@ -68,6 +69,18 @@ describe('commons', () => {
...
   })
+
+  describe('coerceBoolean', () => {
+    test('return fallback when none of the values provided can be parsed to boolean', () => {
+      expect(coerceBoolean(true, 'NotBoolean', 123, [], {})).toBe(true)
+      expect(coerceBoolean(false, 'NotBooleanEither', 456, ['An array'], {booleanInObject: true})).toBe(false)
+    })
+
+    test('return the first boolean when one of the values provided can be parsed to boolean', () => {
+      expect(coerceBoolean(true, 'false', 'true')).toBe(false)
```
DarcyRaynerDD, line 80: Let's make sure we have a test case with an actual boolean value

DarcyRaynerDD (approved):

PR #450, opened 2021-12-02
[lambda] Simplify instrument for trial

duncanista:
### What and why?

Refer to [SLS-1712](https://datadoghq.atlassian.net/browse/SLS-1712).

### How?

- Adding prompts (if environment variables missing) for:
  - AWS Credentials (regex validation).
  - Datadog API Key (regex validation) and site to send data.

- Creating tests for interactive instrumentation/uninstrumentation

### Review checklist

  - [x] Tested interactive instrumentation
  - [x] Tested interactive uninstrumentation
- [ ] Documented the feature

### How to use?

The flag `-i` (`--interactive`) is required to be set in order for this feature to work. No function name, extension and library version are required. Since the user will be required to input them. Same goes for AWS credentials and Datadog Environment Variables.

Example:
```zsh
datadog-ci lambda {instrument|uninstrument} -i
```

After executing `instrument` the user will be prompted the following:

#### AWS Credentials
Screen Shot 2021-12-15 at 7 48 02 PM

#### Datadog Environment Variables 
Screen Shot 2021-12-15 at 7 48 23 PM
Screen Shot 2021-12-15 at 7 52 55 PM
Note: this API Key IS NOT REAL.
Screen Shot 2021-12-15 at 7 53 27 PM

Now the user will be required to select the functions to instrument.
Screen Shot 2021-12-15 at 7 56 01 PM

After selecting, the user needs to confirm the changes to be applied.
Screen Shot 2021-12-15 at 7 58 24 PM

After that, functions will be instrumented.

Note: for `uninstrument`, Datadog Environment Variables are not required and won't be prompted.

### What's next?
- Allow the user to type the region instead of selecting it (check `fuse.js` and `inquirer-checkbox-plus-prompt` packages, see example [here](https://github.com/SBoudrias/Inquirer.js#prompts).)
- Allow the user to type the name of the function they want to instrument (same as above).
- Tell the user which latests versions are being used.


src/commands/lambda/uninstrument.ts
```diff
@@ -71,7 +118,7 @@ export class UninstrumentCommand extends Command {
 
         configGroups.push({configs, lambda, cloudWatchLogs})
       } catch (err) {
-        this.context.stdout.write(`Couldn't fetch lambda functions. ${err}\n`)
+        this.context.stdout.write(`${red('[Error]')} Couldn't fetch lambda functions. ${err}\n`)
```
nhinsch, line 121: lambda -> Lambda

src/commands/lambda/uninstrument.ts
```diff
@@ -108,6 +155,16 @@ export class UninstrumentCommand extends Command {
       return 0
     }
 
+    const willUpdate = willUpdateFunctionConfigs(configList)
+    if (this.interactive && willUpdate) {
+      this.context.stdout.write(`${yellow('[!]')} Confirmation needed.\n`)
+      const isConfirmed = await requestChangesConfirmation('Do you wanna apply the changes?')
```
nhinsch, line 161: wanna -> want to

src/commands/lambda/uninstrument.ts
```diff
@@ -2,27 +2,74 @@ import {CloudWatchLogs, Lambda} from 'aws-sdk'
...
   private regExPattern?: string
   private region?: string
 
   public async execute() {
     const lambdaConfig = {lambda: this.config}
     this.config = (await parseConfigFile(lambdaConfig, this.configPath)).lambda
 
-    const hasSpecifiedFuntions = this.functions.length !== 0 || this.config.functions.length !== 0
+    let hasSpecifiedFuntions = this.functions.length !== 0 || this.config.functions.length !== 0
+    // Trial user experience
```
nhinsch, line 35: Nit: I don't think you need this "trial user experience" comment, the code should speak for itself as much as possible

src/commands/lambda/uninstrument.ts
```diff
@@ -2,27 +2,74 @@ import {CloudWatchLogs, Lambda} from 'aws-sdk'
...
+  private interactive = false
   private regExPattern?: string
   private region?: string
 
   public async execute() {
     const lambdaConfig = {lambda: this.config}
     this.config = (await parseConfigFile(lambdaConfig, this.configPath)).lambda
 
-    const hasSpecifiedFuntions = this.functions.length !== 0 || this.config.functions.length !== 0
+    let hasSpecifiedFuntions = this.functions.length !== 0 || this.config.functions.length !== 0
```
nhinsch, line 34: hasSpecifiedFuntions -> hasSpecifiedFun**c**tions

src/commands/lambda/uninstrument.ts
```diff
@@ -2,27 +2,74 @@ import {CloudWatchLogs, Lambda} from 'aws-sdk'
...
+      this.region = region
+
+      if (!hasSpecifiedFuntions) {
+        try {
+          const lambda = new Lambda({region})
+          this.context.stdout.write('Fetching lambda functions, this might take a while.\n')
+          const functionNames =
+            (await getAllLambdaFunctionConfigs(lambda)).map((config) => config.FunctionName!).sort() ?? []
+          if (functionNames.length === 0) {
+            this.context.stdout.write(`${red('[Error]')} You don't have any Lambda Functions in your AWS account.\n`)
```
nhinsch, line 58: I think this error could be misleading. For example, the customer could have Lambda functions in a different region of their account, but not this region. Maybe we should just say something like "Couldn't find any Lambda functions in the specified region."

src/commands/lambda/uninstrument.ts
```diff
@@ -2,27 +2,74 @@ import {CloudWatchLogs, Lambda} from 'aws-sdk'
...
+        return 1
+      }
+
+      const region = this.region ?? this.config.region ?? process.env[AWS_DEFAULT_REGION_ENV_VAR]
+      this.region = region
+
+      if (!hasSpecifiedFuntions) {
+        try {
+          const lambda = new Lambda({region})
+          this.context.stdout.write('Fetching lambda functions, this might take a while.\n')
```
nhinsch, line 54: Should we show the customer what region we are searching here?

src/commands/lambda/uninstrument.ts
```diff
@@ -2,27 +2,74 @@ import {CloudWatchLogs, Lambda} from 'aws-sdk'
...
+            (await getAllLambdaFunctionConfigs(lambda)).map((config) => config.FunctionName!).sort() ?? []
+          if (functionNames.length === 0) {
+            this.context.stdout.write(`${red('[Error]')} You don't have any Lambda Functions in your AWS account.\n`)
+
+            return 1
+          }
+          const functions = await requestFunctionSelection(functionNames)
+          this.functions = functions
+        } catch (err) {
+          this.context.stdout.write(`${red('[Error]')} Couldn't fetch lambda functions. ${err}\n`)
```
nhinsch, line 65: lambda -> Lambda

src/commands/lambda/prompt.ts
```diff
@@ -0,0 +1,195 @@
...
+    validate: (value) => {
+      if (!value || !sentenceMatchesRegEx(value, AWS_ACCESS_KEY_ID_REG_EXP)) {
+        return 'Enter a valid AWS Access Key ID.'
+      }
+
+      return true
+    },
+  },
+  {
+    // AWS_SCRET_ACCESS_KEY_ENV_VAR question
```
nhinsch, line 42: SCRET -> SECRET

src/commands/lambda/prompt.ts
```diff
@@ -0,0 +1,195 @@
...
+    {
+      name: `Plain text ${bold('API Key')} (Recommended for trial users) `,
+      value: {
+        envVar: CI_API_KEY_ENV_VAR,
+        message: 'API Key:',
+      },
+    },
+    new Separator(),
+    {
+      name: `AWS Key Management Service ${bold('(KMS) API Key')}`,
```
nhinsch, line 76: I would describe this as an "API key encrypted with AWS Key Management Service (KMS)".

src/commands/lambda/uninstrument.ts
nhinsch: (unless they ALWAYS need to specify a region beforehand. I am thinking about a scenario where we use a default region and don't tell them)

src/commands/lambda/instrument.ts
```diff
@@ -44,12 +55,55 @@ export class InstrumentCommand extends Command {
     const lambdaConfig = {lambda: this.config}
     this.config = (await parseConfigFile(lambdaConfig, this.configPath)).lambda
 
+    let hasSpecifiedFuntions = this.functions.length !== 0 || this.config.functions.length !== 0
+    // Trial user experience
```
nhinsch, line 59: Per my other comment, don't think we need this comment here

src/commands/lambda/instrument.ts
```diff
@@ -44,12 +55,55 @@ export class InstrumentCommand extends Command {
     const lambdaConfig = {lambda: this.config}
     this.config = (await parseConfigFile(lambdaConfig, this.configPath)).lambda
 
+    let hasSpecifiedFuntions = this.functions.length !== 0 || this.config.functions.length !== 0
```
nhinsch, line 58: Funtions -> Fun**c**tions

src/commands/lambda/functions/commons.ts
```diff
@@ -100,6 +107,88 @@ export const collectFunctionsByRegion = (
...
+        }
+      }
+    }
+  }
+
+  return latestVersion
+}
+
+export const isMissingAWSCredentials = () =>
+  process.env[AWS_ACCESS_KEY_ID_ENV_VAR] === undefined || process.env[AWS_SECRET_ACCESS_KEY_ENV_VAR] === undefined
```
nhinsch, line 173: Nit: I would add a blank line between each function in this file

nhinsch (approved):
Looks great! ✨ Just left a few small nits...

src/commands/lambda/uninstrument.ts
duncanista: That's right, this is misleading, I forgot to specify for the region.

src/commands/lambda/uninstrument.ts
duncanista: We probably could, but this only happens when the user has the environment variable `AWS_DEFAULT_REGION` set. Which normally doesn't happen since most of the users would directly pipe the role with the desired region. If no region is set and that environment variable is not exported, we always ask them to set it. 
Nonetheless, when instrumenting interactively, the user will be required to prompt the desired region (if not previously set by, lets say AWS_VAULT).

nhinsch: OK, sounds good!

src/commands/lambda/prompt.ts
```diff
@@ -0,0 +1,195 @@
...
+    return true
+  },
+})
+
+export const requestAWSCredentials = async () => {
+  try {
+    const awsCredentialsAnswers = await prompt(awsCredentialsQuestions)
+    process.env[AWS_ACCESS_KEY_ID_ENV_VAR] = awsCredentialsAnswers[AWS_ACCESS_KEY_ID_ENV_VAR]
+    process.env[AWS_SECRET_ACCESS_KEY_ENV_VAR] = awsCredentialsAnswers[AWS_SECRET_ACCESS_KEY_ENV_VAR]
+    process.env[AWS_DEFAULT_REGION_ENV_VAR] = awsCredentialsAnswers[AWS_DEFAULT_REGION_ENV_VAR]
```
tianchu, line 147: Let's also ask for `AWS_SESSION_TOKEN` as an **optional** env var, which is needed when using temporary credentials (fairly common) https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_use-resources.html#using-temp-creds-sdk-cli

duncanista: Added it!

PR #451, opened 2021-12-02
[lambda] Add `DD_CAPTURE_LAMBDA_PAYLOAD` support.

duncanista:
### What and why?

`DD_CAPTURE_LAMBDA_PAYLOAD` is the flag which controls whether we capture and transmit lambda payloads to the backend. 
It will be set with the option `--capture-lambda-payload` and with the config variable `captureLambdaPayload`. They need to be set with a boolean or a quoted boolean (string boolean, being a boolean between quotes such as `'false'|"FALSE"`, etc)

Refer to [SLS-1546](https://datadoghq.atlassian.net/browse/SLS-1546).

### How?

- Adding constants for the environment variable. 
- Updated tests.

### Review checklist


src/commands/lambda/__tests__/functions/instrument.test.ts
```diff
@@ -99,7 +99,7 @@ describe('instrument', () => {
             "Variables": Object {
               "DD_FLUSH_TO_LOG": "false",
               "DD_LAMBDA_HANDLER": "handler.hello",
-              "DD_MERGE_XRAY_TRACES": "false",
```
IvanTopolcic, line 102: Why the changes to these tests?

IvanTopolcic (commented):
Looks good to me, just a quick question!

src/commands/lambda/__tests__/functions/instrument.test.ts
duncanista: I reverted a change and forgot to revert the tests too. Will change them back, thank you for pointing it out.

IvanTopolcic (approved):
LGTM!

PR #455, opened 2021-12-07
[scripts] fix launch command

m-rousse:
### What and why?

This PR fixes the `launch` command by pointing it to `src/cli.ts` instead of `src/index.ts` following the PR to move all `index.ts` files to `cli.ts`: https://github.com/DataDog/datadog-ci/pull/432


etnbrd (approved):
LGTM, good catch, thanks 🙇

PR #463, opened 2022-01-11
[lambda] Standardize lambda-ci error messages/handling

thedavlee:
### What and why?

Standardized how errors are displayed and handled within the lambda-ci commands. 

[SLS-1907](https://datadoghq.atlassian.net/browse/SLS-1907)

### How?

- Previously, Errors were displayed in one of the following two ways:
    - [Error] Error Message (where the [Error] tag is displayed in red)
    - Error Message

- Standardized all error messages in the instrument and uninstrument commands to be a red [Error] tag, followed by the error message itself.

### Review checklist

### DIFF
--- a/src/commands/lambda/__tests__/instrument.test.ts
+++ b/src/commands/lambda/__tests__/instrument.test.ts
@@ -607,7 +607,9 @@ TagResource -> arn:aws:lambda:us-east-1:123456789012:function:lambda-hello-world
         const output = context.stdout.toString()
         expect(code).toBe(1)
         expect(output).toMatchInlineSnapshot(`
-                                                            "No functions specified for instrumentation.
+                                                            "${red(
+                                                              '[Error]'
+                                                            )} No functions specified for instrumentation.
                                                             "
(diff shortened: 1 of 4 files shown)


IvanTopolcic:
LGTM, just wondering, have you tested the error messages that aren't caught by unit tests? I recall some of them embed other errors and iirc we'd have something like `[Error] sample error [Error] embedded error` which looks a little weird.

thedavlee:
> LGTM, just wondering, have you tested the error messages that aren't caught by unit tests? I recall some of them embed other errors and iirc we'd have something like `[Error] sample error [Error] embedded error` which looks a little weird.

Yeah @IvanTopolcic, I noticed those when I was running through the tests. Errors like:
image

image

 I can do some processing on the nested errors to remove the embedded error tags, do you think we should?

IvanTopolcic:
I think this is fine as it is tbh, as long as the error message is clear, approved!

IvanTopolcic (approved):

PR #467, opened 2022-01-20
[synthetics] add possibility to pass variables through args

m-rousse:
### What and why?

This PR adds a new flag to the `synthetics run-tests` command to define variables from the CLI.
It allows customer to explicitely set variables used in synthetics tests from value of environment variables when running the `datadog-ci` command.

### How?

It adds a new option to the command defined in `src/commands/synthetics/command.ts` that gets passed an array of strings, which is later parsed to split it between the variable name and its value before adding it to the `config.global` object containing overrides applied to all tests that will be triggered.

These variables will be appended to the list of variables already defined in the config files.

### Review checklist


src/commands/synthetics/command.ts
```diff
@@ -275,14 +285,15 @@ export class RunTestCommand extends Command {
...
+RunTestCommand.addOption('jUnitReport', Command.String('-j,--jUnitReport'))
 RunTestCommand.addOption('publicIds', Command.Array('-p,--public-id'))
-RunTestCommand.addOption('testSearchQuery', Command.String('-s,--search'))
+RunTestCommand.addOption('runName', Command.String('-n,--runName'))
 RunTestCommand.addOption('subdomain', Command.Boolean('--subdomain'))
+RunTestCommand.addOption('testSearchQuery', Command.String('-s,--search'))
 RunTestCommand.addOption('tunnel', Command.Boolean('-t,--tunnel'))
-RunTestCommand.addOption('jUnitReport', Command.String('-j,--jUnitReport'))
-RunTestCommand.addOption('runName', Command.String('-n,--runName'))
+RunTestCommand.addOption('variableStrings', Command.Array('-v,--variables'))
```
lefebvree, line 299: nit: singular `variable` as I think they are only passed one-by-one (`... --variable A=1 --variable B=2`)
```suggestion
RunTestCommand.addOption('variableStrings', Command.Array('-v,--variable'))
```

src/commands/synthetics/utils.ts
```diff
@@ -534,3 +534,22 @@ export const retry = async <T, E extends Error>(
...
 }
+
+export const parseVariablesFromCli = (variableArguments: string[] = []): {[key: string]: string} | undefined => {
+  const variables: {[key: string]: string} = {}
+
+  for (const variableArgument of variableArguments) {
+    const separatorIndex = variableArgument.indexOf('=')
+
+    if (separatorIndex === -1) {
+      continue
```
lefebvree, line 545: Should we add a warning on ignored invalid variables?
Also small edge case I can think of is something like `=val` setting an empty variable name

lefebvree (approved):

PR #472, opened 2022-02-03
[ci] remove dependencies E2E test from ci

m-rousse:
### What and why?

The `snyk` feature is sunsetting, also the dependency graph upload is planned to be removed.

This PR removes the E2E test from the CI to unblock the CI and allow further merges.


piotr-oles (approved):
Thanks!

PR #474, opened 2022-02-03
[Serverless] Add additional runtimes to serverless CI

jcstorms1:
### What and why?

Adds support to instrument and uninstrument .NET, Go, Ruby and Java lambda functions via the Serverless CI

### How?

Adds `dotnecore3.1`, `provided.al2`, `Java8.al2`, `Java11`, `Ruby2.5` and `Ruby2.7` to the supported runtimes. 
- instrument correctly applies layers and adds necessary environment variables 
- uninstrument removes the layer and all env vars

### Review checklist


src/commands/lambda/constants.ts
```diff
@@ -96,6 +107,11 @@ export const VERSION_ENV_VAR = 'DD_VERSION'
 export const ENVIRONMENT_ENV_VAR = 'DD_ENV'
 export const EXTRA_TAGS_ENV_VAR = 'DD_TAGS'
 export const CAPTURE_LAMBDA_PAYLOAD_ENV_VAR = 'DD_CAPTURE_LAMBDA_PAYLOAD'
+export const ENABLE_PROFILING = 'CORECLR_ENABLE_PROFILING'
```
nhinsch, line 110: Nit: All of the other env var consts end in `_ENV_VAR`, I would follow that pattern to be consistent

nhinsch:
I added a "do not merge" tag just to make sure we don't merge this before we have a chance to update it with the final production layer.

src/commands/lambda/functions/instrument.ts
```diff
@@ -135,16 +146,21 @@ export const calculateUpdateRequest = async (
...
+      needsUpdate = true
+      updateRequest.Handler = HANDLER_LOCATION[runtime]
+    }
   }
 
   // Update Env Vars
-  if (oldEnvVars[LAMBDA_HANDLER_ENV_VAR] === undefined) {
-    needsUpdate = true
-    changedEnvVars[LAMBDA_HANDLER_ENV_VAR] = config.Handler ?? ''
+  // Don't add dotnet
```
nhinsch, line 158: What do you mean by this comment `Don't add dotnet`?

src/commands/lambda/functions/instrument.ts
```diff
@@ -220,6 +236,16 @@ export const calculateUpdateRequest = async (
     }
   }
 
+  // Add dotnet Env Vars
```
nhinsch, line 239: Nit: I don't think you need this comment, the code speaks for itself 👍

src/commands/lambda/functions/instrument.ts
jcstorms1: We don't need to add the handler env var for dotnet. I added more detail to the comment.

nhinsch:
We aren't going to support the ARM architecture for .NET functions (tracer does not support this architecture yet). Can we verify/add a test that the CLI won't allow the user to instrument a .NET ARM function?

src/commands/lambda/constants.ts
```diff
@@ -96,6 +107,11 @@ export const VERSION_ENV_VAR = 'DD_VERSION'
 export const ENVIRONMENT_ENV_VAR = 'DD_ENV'
 export const EXTRA_TAGS_ENV_VAR = 'DD_TAGS'
 export const CAPTURE_LAMBDA_PAYLOAD_ENV_VAR = 'DD_CAPTURE_LAMBDA_PAYLOAD'
+export const ENABLE_PROFILING_ENV_VAR = 'CORECLR_ENABLE_PROFILING'
+export const PROFILER_ENV_VAR = 'CORECLR_PROFILER'
+export const PROFILER_PATH_ENV_VAR = 'CORECLR_PROFILER_PATH'
+export const DOTNET_TRACER_HOME_ENV_VAR = 'DD_DOTNET_TRACER_HOME'
+export const INTEGRATIONS_ENV_VAR = 'DD_INTEGRATIONS'
```
nhinsch, line 114: Update from Max: We actually don't need to apply this DD_INTEGRATIONS env var, so we can remove it throughout this PR

src/commands/lambda/constants.ts
```diff
@@ -1,34 +1,50 @@
 export const DD_LAMBDA_EXTENSION_LAYER_NAME = 'Datadog-Extension'
 export const EXTENSION_LAYER_KEY = 'extension'
+export const DOTNET_RUNTIME = 'dotnetcore3.1'
 export const RUNTIME_LAYER_LOOKUP = {
   [EXTENSION_LAYER_KEY]: DD_LAMBDA_EXTENSION_LAYER_NAME,
+  'dotnetcore3.1': 'poc-maxday-dotnet',
```
nhinsch, line 6: We released the prod layer, the final name is `dd-trace-dotnet` so we can remove this placeholder name now

src/commands/lambda/constants.ts
```diff
@@ -1,34 +1,50 @@
...
   PYTHON,
+  RUBY,
 }
 
-export const RUNTIME_LOOKUP: {[key: string]: RuntimeType} = {
+export const RUNTIME_LOOKUP: { [key: string]: RuntimeType } = {
+  'dotnetcore3.1': RuntimeType.DOTNET,
+  'java11': RuntimeType.JAVA,
+  'java8.al2': RuntimeType.JAVA,
+  'provided.al2': RuntimeType.GO,
```
nhinsch, line 39: Technically `provided.al2` doesn't have to be Go. It's a custom runtime. It could be Go, but it could also be anything else.

src/commands/lambda/constants.ts
```diff
@@ -1,34 +1,50 @@
...
   'python3.9': 'Datadog-Python39',
+  'ruby2.5': 'Datadog-Ruby2-5',
+  'ruby2.7': 'Datadog-Ruby2-7',
 } as const
 // We exclude the Extension Layer Key in order for the runtime
 // to be used directly in HANDLER_LOCATION.
 export type Runtime = Exclude<keyof typeof RUNTIME_LAYER_LOOKUP, typeof EXTENSION_LAYER_KEY>
-
-export const ARM_RUNTIMES = [EXTENSION_LAYER_KEY, 'python3.8', 'python3.9']
+export const ARM_RUNTIMES = [EXTENSION_LAYER_KEY, 'java8.al2', 'java11', 'provided.al2', 'python3.8', 'python3.9']
```
nhinsch, line 22: What does this constant mean?

src/commands/lambda/functions/uninstrument.ts
```diff
@@ -1,28 +1,34 @@
...
   SERVICE_ENV_VAR,
   SITE_ENV_VAR,
   TRACE_ENABLED_ENV_VAR,
   VERSION_ENV_VAR,
 } from '../constants'
-import {FunctionConfiguration, LogGroupConfiguration, TagConfiguration} from '../interfaces'
-import {calculateLogGroupRemoveRequest} from '../loggroup'
-import {calculateTagRemoveRequest} from '../tags'
-import {getLambdaFunctionConfigs, getLambdaFunctionConfigsFromRegex, getLayers, isSupportedRuntime} from './commons'
+import { FunctionConfiguration, LogGroupConfiguration, TagConfiguration } from '../interfaces'
```
nhinsch, line 28: Nit: What is adding all of these extra spaces between `{}`? Is this something you're doing manually or do you have some kind of automated tool that is doing it?

src/commands/lambda/functions/instrument.ts
```diff
@@ -135,16 +147,20 @@ export const calculateUpdateRequest = async (
   let needsUpdate = false
 
   // Update Handler
-  const expectedHandler = HANDLER_LOCATION[runtime]
-  if (config.Handler !== expectedHandler) {
-    needsUpdate = true
-    updateRequest.Handler = HANDLER_LOCATION[runtime]
+  if (runtime !== DOTNET_RUNTIME && runtime !== 'java11' && runtime !== 'java8.al2' && runtime !== 'provided.al2' && runtime !== 'ruby2.5' && runtime !== 'ruby2.7') {
```
nhinsch, line 150: This seems like a rather long `if` statement. Is there a way to make this a bit shorter and easier to read?

src/commands/lambda/functions/uninstrument.ts
```diff
@@ -102,12 +108,14 @@ export const calculateUpdateRequest = (config: Lambda.FunctionConfiguration, run
...
   let needsUpdate = false
 
-  // Remove Handler
-  const expectedHandler = HANDLER_LOCATION[runtime]
-  if (config.Handler === expectedHandler) {
-    needsUpdate = true
-    updateRequest.Handler = oldEnvVars[LAMBDA_HANDLER_ENV_VAR]
-    delete oldEnvVars[LAMBDA_HANDLER_ENV_VAR]
+  // Remove Handler. Dotnet should not have a handler ENV_VAR
+  if (runtime !== DOTNET_RUNTIME && runtime !== 'java11' && runtime !== 'java8.al2' && runtime !== 'provided.al2' && runtime !== 'ruby2.5' && runtime !== 'ruby2.7') {
```
nhinsch, line 112: See other comment -- this `if` statement seems a bit long and complex

src/commands/lambda/functions/uninstrument.ts
jcstorms1: Must be when I run `yarn format` or it is eslint.

src/commands/lambda/constants.ts
```diff
@@ -81,6 +86,13 @@ export const GOVCLOUD_LAYER_AWS_ACCOUNT = '002406178527'
 export const SUBSCRIPTION_FILTER_NAME = 'datadog-ci-filter'
 export const TAG_VERSION_NAME = 'dd_sls_ci'
 
+// Export const values for .NET tracer
+export const CORECLR_ENABLE_PROFILING = '1'
+export const CORECLR_PROFILER = '{846F5F1C-F9AE-4B07-969E-05C26BC060D8}'
+export const CORECLR_PROFILER_PATH = '/opt/datadog/Datadog.Trace.ClrProfiler.Native.so'
+export const DD_DOTNET_TRACER_HOME = '/opt/datadog'
+export const DD_INTEGRATIONS = '/opt/datadog/integrations.json'
```
nhinsch, line 94: Per Max, we don't actually need to use this `DD_INTEGRATIONS` const anywhere

src/commands/lambda/constants.ts
```diff
@@ -96,6 +108,13 @@ export const VERSION_ENV_VAR = 'DD_VERSION'
 export const ENVIRONMENT_ENV_VAR = 'DD_ENV'
 export const EXTRA_TAGS_ENV_VAR = 'DD_TAGS'
 export const CAPTURE_LAMBDA_PAYLOAD_ENV_VAR = 'DD_CAPTURE_LAMBDA_PAYLOAD'
+export const ENABLE_PROFILING_ENV_VAR = 'CORECLR_ENABLE_PROFILING'
+export const PROFILER_ENV_VAR = 'CORECLR_PROFILER'
+export const PROFILER_PATH_ENV_VAR = 'CORECLR_PROFILER_PATH'
+export const DOTNET_TRACER_HOME_ENV_VAR = 'DD_DOTNET_TRACER_HOME'
+export const JAVA_TOOL_OPTIONS_ENV_VAR = 'JAVA_TOOL_OPTIONS'
+export const LOGS_INJECTION_ENV_VAR = 'DD_LOGS_INJECTION'
+export const JMXFETCH_ENV_VAR = 'DD_JMXFETCH_ENABLED'
```
nhinsch, line 117: Are we using these Java env vars anywhere?

src/commands/lambda/constants.ts
```diff
@@ -1,46 +1,51 @@
...
-  'python3.6': PYTHON_HANDLER_LOCATION,
-  'python3.7': PYTHON_HANDLER_LOCATION,
-  'python3.8': PYTHON_HANDLER_LOCATION,
-  'python3.9': PYTHON_HANDLER_LOCATION,
-}
+// We exclude the Extension Layer Key in order for the runtime
+// to be used directly in HANDLER_LOCATION.
+export type Runtime = keyof typeof RUNTIME_LOOKUP
+export type LayerRuntime = keyof typeof RUNTIME_LAYER_LOOKUP
+export const ARM_RUNTIMES = [EXTENSION_LAYER_KEY, 'python3.8', 'python3.9']
```
nhinsch, line 43: What does `ARM_RUNTIMES` mean here?

src/commands/lambda/constants.ts
```diff
@@ -1,46 +1,51 @@
 export const DD_LAMBDA_EXTENSION_LAYER_NAME = 'Datadog-Extension'
 export const EXTENSION_LAYER_KEY = 'extension'
+export const DOTNET_RUNTIME = 'dotnetcore3.1'
 export const RUNTIME_LAYER_LOOKUP = {
```
nhinsch, line 4: I know it was already there, but I'm not sure `RUNTIME_LAYER_LOOKUP` is the best name for that map. To me that would suggest that we can use this to get the layer for a specific runtime. That is mostly what this does, but it also includes the extension which is not a runtime per se. Maybe a better name would just be `LAYER_LOOKUP` or something like that.

src/commands/lambda/constants.ts
```diff
@@ -1,46 +1,51 @@
...
-  'nodejs14.x': NODE_HANDLER_LOCATION,
-  'python3.6': PYTHON_HANDLER_LOCATION,
-  'python3.7': PYTHON_HANDLER_LOCATION,
-  'python3.8': PYTHON_HANDLER_LOCATION,
-  'python3.9': PYTHON_HANDLER_LOCATION,
-}
+// We exclude the Extension Layer Key in order for the runtime
+// to be used directly in HANDLER_LOCATION.
+export type Runtime = keyof typeof RUNTIME_LOOKUP
+export type LayerRuntime = keyof typeof RUNTIME_LAYER_LOOKUP
```
nhinsch, line 42: What does `LayerRuntime` mean here? It looks like this also includes the extension?

src/commands/lambda/functions/instrument.ts
```diff
@@ -134,17 +146,49 @@ export const calculateUpdateRequest = async (
...
+        `Only the extension version should be set for the ${runtime} runtime. Please remove the layer version from the instrument command and only use the extension version.`
+      )
+    }
+  }
+  // Ruby requires handler redirection, which needs to be performed manually.
+  // We need to inform the customer to only set the extension and remove they layer version.
+  if (RUNTIME_LOOKUP[runtime] === RuntimeType.RUBY) {
+    if (settings.layerVersion !== undefined) {
+      throw new Error(
+        'The Ruby layer requires extensive manual instrumentation. Please only set the extension version with the instrument command.'
```
nhinsch, line 163: I would avoid the word "extensive" here -- that makes it sound scarier than it is 🙂

src/commands/lambda/functions/instrument.ts
```diff
@@ -134,17 +146,49 @@ export const calculateUpdateRequest = async (
...
-  const expectedHandler = HANDLER_LOCATION[runtime]
-  if (config.Handler !== expectedHandler) {
-    needsUpdate = true
-    updateRequest.Handler = HANDLER_LOCATION[runtime]
+  // If a layer version is set for Java or a custom runtime.
+  // We need to inform the customer to only set the extension argument.
+  if (RUNTIME_LOOKUP[runtime] === RuntimeType.JAVA || RUNTIME_LOOKUP[runtime] === RuntimeType.CUSTOM) {
+    if (settings.layerVersion !== undefined) {
+      throw new Error(
+        `Only the extension version should be set for the ${runtime} runtime. Please remove the layer version from the instrument command and only use the extension version.`
```
nhinsch, line 154: I think customers could get confused when we tell them to remove the "layer version". Technically the extension is a layer too. Maybe we should be clear that `-v` is a runtime-specific layer version, and `-e` is the extension layer version. It's the runtime-specific layer that we can't add in these runtimes using Datadog CI.

src/commands/lambda/constants.ts
```diff
@@ -1,46 +1,49 @@
...
+  JAVA,
   NODE,
   PYTHON,
+  RUBY,
 }
 
-export const RUNTIME_LOOKUP: {[key: string]: RuntimeType} = {
+export const RUNTIME_LOOKUP = {
+  'dotnetcore3.1': RuntimeType.DOTNET,
+  java11: RuntimeType.JAVA,
```
nhinsch, line 26: Super nit: I would put this in quotes to be consistent with all of the others.

src/commands/lambda/constants.ts
```diff
@@ -1,46 +1,49 @@
...
-export const HANDLER_LOCATION = {
-  'nodejs12.x': NODE_HANDLER_LOCATION,
-  'nodejs14.x': NODE_HANDLER_LOCATION,
-  'python3.6': PYTHON_HANDLER_LOCATION,
-  'python3.7': PYTHON_HANDLER_LOCATION,
-  'python3.8': PYTHON_HANDLER_LOCATION,
-  'python3.9': PYTHON_HANDLER_LOCATION,
-}
+export type Runtime = keyof typeof RUNTIME_LOOKUP
+export type LayerName = keyof typeof LAYER_LOOKUP
```
nhinsch, line 40: I would choose a different name here.

Right now `LayerName` is defined as the **keys** of `LAYER_LOOKUP` (`dotnetcore3.1`, `nodejs12.x`, etc.). 

However, it looks like the layer names are actually the **values** of `LAYER_LOOKUP` (`dd-trace-dotnet`, `Datadog-Node12-x`, etc.).

I think I would name this `LayerKeys`. That reflects that they are the keys that let us get the layer name, not the layer names themselves.

src/commands/lambda/constants.ts
jcstorms1: It won't let me. Lint is forcing to look like that for some reason.

src/commands/lambda/functions/commons.ts
```diff
@@ -258,12 +260,12 @@ export const getLambdaFunctionConfigs = (
...
 export const getLayerArn = (
   config: Lambda.FunctionConfiguration,
-  runtime: Runtime,
+  layer: LayerName,
   region: string,
   settings?: InstrumentationSettings
 ) => {
-  let layerName = RUNTIME_LAYER_LOOKUP[runtime]
-  if (ARM_RUNTIMES.includes(runtime) && config.Architectures?.includes(ARM64_ARCHITECTURE)) {
+  let layerName = LAYER_LOOKUP[layer]
```
nhinsch, line 267: Here is a good example of why I think it would be better to name the type `LayerName` as `LayerKey`.

Here `layer` is a `LayerName`. But, it's not actually the name of a layer. To get the name of the layer, we need to look it up with LAYER_LOOKUP.

If `layer` were a `LayerKey`, it would make more sense that we use it to GET the name of the layer.

src/commands/lambda/functions/instrument.ts
```diff
@@ -134,17 +147,58 @@ export const calculateUpdateRequest = async (
...
+        'The Ruby layer requires additional manual instrumentation. Please only set the extensionVersion argument with the instrument command.'
+      )
+    }
+  }
+
+  // We don't support ARM Architecture for .NET at this time. Abort instrumentation if the combination is detected.
+  if (RUNTIME_LOOKUP[runtime] === RuntimeType.DOTNET) {
+    if (config.Architectures?.includes(ARM64_ARCHITECTURE)) {
+      throw new Error(
+        'Instrumenting arm64 architecture is not currently supported for .NET. Please only instrument .NET functions using x86_64 architecture.'
```
nhinsch, line 173: 👌

src/commands/lambda/functions/instrument.ts
```diff
@@ -134,17 +147,58 @@ export const calculateUpdateRequest = async (
...
-  const expectedHandler = HANDLER_LOCATION[runtime]
-  if (config.Handler !== expectedHandler) {
-    needsUpdate = true
-    updateRequest.Handler = HANDLER_LOCATION[runtime]
+  // If a layer version is set for Java or a custom runtime.
+  // We need to inform the customer to only set the extension argument.
+  if (RUNTIME_LOOKUP[runtime] === RuntimeType.JAVA || RUNTIME_LOOKUP[runtime] === RuntimeType.CUSTOM) {
+    if (settings.layerVersion !== undefined) {
+      throw new Error(
+        `Only the extensionVersion argument should be set for the ${runtime} runtime. Please remove the layerVersion argument from the instrument command.`
```
nhinsch, line 155: According to the README, the arguments are `--extension-version` and `--layer-version` (i.e. kebab-case, not camelCase)

src/commands/lambda/functions/instrument.ts
```diff
@@ -134,17 +147,58 @@ export const calculateUpdateRequest = async (
...
+      )
+    }
+  }
+  // Ruby requires handler redirection, which needs to be performed manually.
+  // We need to inform the customer to only set the extension and remove they layer version.
+  if (RUNTIME_LOOKUP[runtime] === RuntimeType.RUBY) {
+    if (settings.layerVersion !== undefined) {
+      throw new Error(
+        'The Ruby layer requires additional manual instrumentation. Please only set the extensionVersion argument with the instrument command.'
```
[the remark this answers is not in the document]
jcstorms1: I was thinking the same thing.

src/commands/lambda/instrument.ts
```diff
@@ -207,6 +207,20 @@ export class InstrumentCommand extends Command {
     }
 
     const configList = configGroups.map((group) => group.configs).reduce((a, b) => a.concat(b))
+    // Returns false if not all runtimes are of the same RuntimeType across multiple functions
+    const runtimesAreUniform = configList
```
nhinsch, line 211: I would break this out into its own little function and write unit tests for it.

src/commands/lambda/constants.ts
nhinsch: Oh, I bet it's because it does not contain a dot, but the other ones do, which is why you have to quote them. Never mind then!

src/commands/lambda/functions/commons.ts
```diff
@@ -191,6 +193,14 @@ export const isMissingDatadogEnvVars = () => isMissingDatadogSiteEnvVar() || isM
 
 export const getAllLambdaFunctionConfigs = async (lambda: Lambda) => getLambdaFunctionConfigsFromRegex(lambda, '.')
 
+// Returns false if not all runtimes are of the same RuntimeType across multiple functions
```
nhinsch, line 196: I think you could eliminate this comment if you made the name of the function more specific. The function checks that the runtime **types** are uniform -- i.e. python3.8 and python3.9 are considered uniform. Also, I always like to name functions as verbs to make it more clear that they're functions. So, I would suggest the name `checkRuntimeTypesAreUniform()`.

nhinsch (approved):
Looks good! Just left a few last nits

PR #476, opened 2022-02-09
Add git-metadata to exports for datadog-ci

IvanTopolcic:
### What and why?

This PR adds a few custom functions for uploading git-metadata and allowing use in a library instead of as a CLI. Serverless needs this functionality to use as part of other deployment-related libraries, such as datadog-cdk-constructs, datadog-serverless-plugin, etc.

### How?

This PR duplicates and simplifies some functionality. I didn't want to refactor the datadog-ci code because I did not want to break datadog-ci for our users, so I re-used some of the code and made changes to it so that it could be used as a library instead of CLI. The majority of this work was making parameters arguments to functions instead of getting them from the environment. We also want it to silently fail in the case that `git` was not installed (as `git` is not a requirement for our deployment tools). In that case, it would just not install the source code integration.

### Review checklist

### DIFF
--- a/src/commands/git-metadata/__tests__/git.test.ts
+++ b/src/commands/git-metadata/__tests__/git.test.ts
@@ -1,4 +1,5 @@
-import {getCommitInfo, gitRemote, stripCredentials} from '../git'
+import * as simpleGit from 'simple-git'
+import {getCommitInfo, gitRemote, newSimpleGitOrFail, stripCredentials} from '../git'
 
 interface MockConfig {
   hash?: string
@@ -108,6 +109,33 @@ describe('git', () => {
       expect(commitInfo!.hash).toBe('abcd')
       expect(commitInfo!.trackedFiles).toStrictEqual(['myfile.js'])
       expect(commitInfo!.remote).toBe('https://overridden')
+    })
+  })
+
+  describe('newSimpleGitOrFail', () => {
+    test('should throw an error if git is not installed', async () => {
+      jest.spyOn(simpleGit, 'gitP').mockImplementation(() => {
+        throw Error('gitp error')
+      })
+      await expect(newSimpleGitOrFail()).rejects.toThrow('gitp error')
+    })
+
+    test('should throw an error if revparse throws an error', async () => {
+      const mock = createMockSimpleGit({}) as any
+      jest.spyOn(simpleGit, 'gitP').mockReturnValue(mock)
+      jest.spyOn(mock, 'revparse').mockImplementation(async () => {
+        throw Error('revparse error')
+      })
+
+      await expect(newSimpleGitOrFail()).rejects.toThrow('revparse error')
+    })
+
+    test('should not throw any errors', async () => {
+      const mock = createMockSimpleGit({}) as any
+      jest.spyOn(simpleGit, 'gitP').mockReturnValue(mock)
+      jest.spyOn(mock, 'revparse').mockResolvedValue('1234')
+
+      await expect(newSimpleGitOrFail()).resolves.not.toThrow()
     })
   })
 })
--- a/src/commands/git-metadata/__tests__/git.test.ts
+++ b/src/commands/git-metadata/__tests__/git.test.ts
@@ -1,6 +1,6 @@
 import * as simpleGit from 'simple-git'
 
-import {getCommitInfo, gitRemote, newSimpleGitOrFail, stripCredentials} from '../git'
+import {getCommitInfo, gitRemote, newSimpleGit, stripCredentials} from '../git'
 
 interface MockConfig {
   hash?: string
@@ -92,7 +92,7 @@ describe('git', () => {
         remotes: [{name: 'first', refs: {push: 'https://git-repo'}}],
         trackedFiles: ['myfile.js'],
       }) as any
-      const commitInfo = await getCommitInfo(mock, process.stdout)
+      const commitInfo = await getCommitInfo(mock)
 
       expect(commitInfo).toBeDefined()
       expect(commitInfo!.hash).toBe('abcd')
@@ -104,7 +104,7 @@ describe('git', () => {
         hash: 'abcd',
         trackedFiles: ['myfile.js'],
       }) as any
-      const commitInfo = await getCommitInfo(mock, process.stdout, 'https://overridden')
+      const commitInfo = await getCommitInfo(mock, 'https://overridden')
 
       expect(commitInfo).toBeDefined()
       expect(commitInfo!.hash).toBe('abcd')
@@ -113,12 +113,12 @@ describe('git', () => {
     })
   })
 
-  describe('newSimpleGitOrFail', () => {
+  describe('newSimpleGit', () => {
     test('should throw an error if git is not installed', async () => {
       jest.spyOn(simpleGit, 'gitP').mockImplementation(() => {
         throw Error('gitp error')
       })
-      await expect(newSimpleGitOrFail()).rejects.toThrow('gitp error')
+      await expect(newSimpleGit()).rejects.toThrow('gitp error')
     })
 
     test('should throw an error if revparse throws an error', async () => {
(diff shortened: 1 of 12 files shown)


src/commands/synthetics/crypto.ts
```diff
@@ -1,5 +1,4 @@
 import {generateKeyPairSync} from 'crypto'
-
```
m-rousse, line 2: Our convention is to separate imports from Node API, then 3rd party libraries, then project imports.
Could we keep this blank line? (same for other removed blank lines in imports of other files of this PR)

src/commands/git-metadata/git.ts
```diff
@@ -25,6 +24,25 @@ export const newSimpleGit = async (): Promise<simpleGit.SimpleGit> => {
   return simpleGit.gitP(options)
 }
 
+// Returns a configured SimpleGit or undefined
+// in the case that the user does not have git installed
+export const newSimpleGitOrFail = async (): Promise<simpleGit.SimpleGit> => {
```
m-rousse, line 29: Could this function be factorized with `newSimpleGit` as it shares most of its implementation?

src/commands/git-metadata/git.ts
```diff
@@ -93,3 +111,16 @@ export const getCommitInfo = async (
 
   return new CommitInfo(hash, remote, trackedFiles)
 }
+
+export const getCommitInfoBasic = async (git: simpleGit.SimpleGit): Promise<CommitInfo> => {
```
m-rousse, line 115: Could this function be factorized with `getCommitInfo`?
As the difference is whether the error is thrown of printed on the standard output, it could always be thrown and the display of the error to the stdout should be handled by the command itself.
The model we used in the synthetics command was to export as a library all functions that are "general" and implemented in `cli.ts` anything that is specific to the CLI usage/interaction of the command. With this model, handling the standard output of the function should be done in `cli.ts`.

yarn.lock
```diff
@@ -2,35 +2,29 @@
 # yarn lockfile v1
```
m-rousse: It seems the lockfile was almost completely rebuilt, though I do not see any `package.json` changes.
Are these changes necessary? If not, could you revert this file?

src/commands/git-metadata/library.ts
```diff
@@ -0,0 +1,85 @@
...
+        },
+        retries: 5,
+      })
+
+      if (status !== UploadStatus.Success) {
+        throw new Error('Error uploading commit information.')
+      }
+
+      return payload.hash
+    } catch (error) {
```
m-rousse, line 71: This catch does not change the default behavior, could it be removed?

src/commands/git-metadata/__tests__/git.test.ts
```diff
@@ -1,4 +1,5 @@
-import {getCommitInfo, gitRemote, stripCredentials} from '../git'
+import * as simpleGit from 'simple-git'
+import {getCommitInfo, gitRemote, newSimpleGitOrFail, stripCredentials} from '../git'
```
m-rousse, line 2:
```suggestion

import {getCommitInfo, gitRemote, newSimpleGitOrFail, stripCredentials} from '../git'
```

src/commands/git-metadata/__tests__/library.test.ts
```diff
@@ -0,0 +1,77 @@
+import {SimpleGit} from 'simple-git'
+import {SourceCodeIntegration} from '..'
```
m-rousse, line 2:
```suggestion

import {SourceCodeIntegration} from '..'
```

src/commands/git-metadata/library.ts
```diff
@@ -0,0 +1,85 @@
...
+      isRepo = simpleGit.checkIsRepo()
+    } catch {
+      return false
+    }
+
+    // Only enable if the system has `git` installed and we're in a git repo
+    return apiKey !== undefined && isRepo
+  }
+
+  public async addSourceCodeIntegration(): Promise<string> {
```
jybp, line 32: Should this be renamed to something like `uploadGitMetadata` since `add` doesn't really describe what's happening under-the-hood? Just want to make sure since it'll be imported elsewhere.

src/commands/git-metadata/git.ts
jybp: On the global model: In our case the cli specific logic is inside `upload.ts`. I agree that it'd be great to reduce code duplication if possible so that changes don't have to be applied twice in `upload.ts` and `library.ts`. As long as it's not a complete headache that is.

src/commands/synthetics/crypto.ts
IvanTopolcic: Ah missed that thanks, will do!

src/commands/git-metadata/git.ts
IvanTopolcic: I'm not really sure how to refactor this function because the original logical swallows the exceptions. newSimpleGit() will return a GitClient even if git is not installed, and subsequent calls to the gitclient will throw an error. In upload.ts, this is handled by checking if the return value of getCommitInfo is undefined, and getCommitInfo returns undefined by wrapping everything in a try catch and just returning undefined if an exception is thrown. To me, it makes more sense to fail early. Do you have any recommendations here?

src/commands/git-metadata/library.ts
IvanTopolcic: Good catch! I think `uploadGitCommitHash` is descriptive enough.

src/commands/git-metadata/git.ts
IvanTopolcic: I didn't want to touch the existing implementation because I wasn't sure if there were any specific reasons the code was written the way it was. I agree that it makes the most sense to me to add the base functionality to `library` or `git` or whatever else is exported, and leave the pretty-format and printing to the CLI command (`upload`). This way, we can just handle the errors in the functions that are calling the library/git commands.

 I didn't want to introduce breaking behavior and re-writing the code to throw an error instead of printing and returning undefined. If you'd like, I can go ahead and try to refactor this a little so that we're printing the error inside the calling code instead of inside the library. @jybp and @m-rousse, thoughts?

src/commands/git-metadata/git.ts
IvanTopolcic: After a discussion with @jybp I've refactored the Upload command to use the library.

src/commands/git-metadata/git.ts
IvanTopolcic: I've gone ahead and refactored `getCommitInfo` to just throw the errors. The onus should be on the caller to catch and handle these appropriately.

src/commands/git-metadata/library.ts
```diff
@@ -0,0 +1,72 @@
...
+  let isRepo
+  try {
+    simpleGit = await newSimpleGit()
+    isRepo = simpleGit.checkIsRepo()
+  } catch {
+    return false
+  }
+
+  // Only enable if the system has `git` installed and we're in a git repo
+  return apiKey !== undefined && isRepo
```
m-rousse, line 19: Nit: We try to avoid `let` as much as possible, you can use `return` in the `try` instead
```suggestion
  try {
    const simpleGit = await newSimpleGit()
    const isRepo = simpleGit.checkIsRepo()

    // Only enable if the system has `git` installed and we're in a git repo
    return apiKey !== undefined && isRepo
  } catch {
    return false
  }
```

m-rousse (approved):
This LGTM, thanks for bearing with the review! :+1:

(Also one CI check failed, once fixed this is OK in my opinion)

src/commands/git-metadata/library.ts
```diff
@@ -0,0 +1,70 @@
...
+  try {
+    const simpleGit = await newSimpleGit()
+    const isRepo = simpleGit.checkIsRepo()
+
+    // Only enable if the system has `git` installed and we're in a git repo
+    return apiKey !== undefined && isRepo
+  } catch {
+    return false
+  }
+}
```
jybp, line 18: nit: Just wondering if having `apiKey !== undefined` is really necessary in this exported func since it's passed as arg. Shouldn't it be checked by whoever call this function?  
Without it the function could be renamed to something like `isCwdInGitRepository` making what is being done explicit.

jybp (approved):
Thanks it looks great! I've run the command with the PR's code and it's uploading repository data properly. It seems safe to merge 👍

PR #483, opened 2022-02-21
Deprecate URL variables in favor of `startUrlSubstitutionRegex`

sdeprez:
### What and why?

We are deprecating the "URL variables" mechanism (using `PROTOCOL`, `DOMAIN`, etc.) in favor of `startUrlSubstitutionRegex` which is a more powerful mechanism and which is compatible with variables predefined in the start URL). Another alternative can also be to use variables in the original definition of a test and to override those variables when triggering those tests with `datadog-ci`.

### DIFF
--- a/src/commands/synthetics/README.md
+++ b/src/commands/synthetics/README.md
@@ -180,7 +180,7 @@ All options under the `config` key are optional and allow overriding the configu
   - `count`: (integer) number of attempts to perform in case of test failure.
   - `interval`: (integer) interval between the attempts (in milliseconds).
 - `startUrl`: (string) new start URL to provide to the test.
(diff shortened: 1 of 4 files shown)


jakepruitt (approved):
👍  looks good from my perspective

src/commands/synthetics/README.md
```diff
@@ -178,8 +180,14 @@ All options under the `config` key allow overriding the configuration of the tes
   - `count`: (integer) number of attempts to perform in case of test failure.
   - `interval`: (integer) interval between the attempts (in milliseconds).
 - `startUrl`: (string) new start URL to provide to the test.
+- `startUrlSubstitutionRegex`: (string) regex to modify the starting URL of the test (browser and HTTP tests only), whether it was given by the original test or by the configuration override `startUrl`. If the URL contains variables, this regex will be applied after the interpolation of the variables. The format is `s/your_regex/your_substitution/modifiers`, for instance `s/(https://www.)(.*)/$1extra-$2/` to transform `https://www.example.com` into `https://www.extra-example.com`.
```
lefebvree, line 183: We can add a quick mention that it should follow JS regex syntax, I think there are some differences between regex language flavors. Maybe we should also move the mention of variables below in the deprecated part.

lefebvree (approved):
LGTM

src/commands/synthetics/README.md
sdeprez: I'll clarify the flavor indeed but the field override `variables` itself is not deprecated, it's the "URL variables" that you could use in your `startingUrl`.

PR #485, opened 2022-02-22
Standalone binary for `datadog-ci`

juan-fernandez:
### What and why?

This introduces standalone binary releases. The goal is to remove barrier of entry for customers for which having to install a recent version of node in their CI is problematic. We have seen this, even internally for usage from datadog-agent.

### How?

- Build binary with https://github.com/nexe/nexe. Current target is `linux-x64-14.15.3`. It's the latest available version in nexe: https://github.com/nexe/nexe/releases. Another option would be to build node from source but it takes a while. 
- In `ci.yml` GHA, add a step to test the generated binary: https://github.com/DataDog/datadog-ci/pull/485/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR71
- In `release.yml` GHA, add a step to generate binary and upload it to the release assets: https://github.com/DataDog/datadog-ci/pull/485/files#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34R18
- Add a `datadog-ci version` for quick testing of the binary.
- Add e2e tests for the standalone binary in https://github.com/DataDog/datadog-ci/pull/485/files#diff-ba729a36573673d1867150b6a7782450c8492f2f3627aad4552213f278cbe4dc

### Review checklist


m-rousse:
When using `nexe`, access to files is changed (cf. [doc](https://github.com/nexe/nexe#resources)), this makes the glob search of synthetics fail (I tested this feature only with `./datadog-ci synthetics run-tests --config path/to/global.config.json` with a `files` configuration set to `"["{,!(node_modules)/**/}staging.synthetics.json"]`)

I think it may impact some other commands too (IIRC sourcemap upload uses globs too).

.github/workflows/ci.yml
```diff
@@ -68,6 +68,26 @@ jobs:
...
+        with:
+          node-version: '10.24.1'
+      - uses: actions/setup-python@v2
+        with:
+          python-version: '2.7'
+      - run: npm config set python python2.7
+      - run: yarn install --immutable
+      - run: yarn build
+      - name: Bundle standalone binary
+        run: yarn dist-standalone -t 'linux-x64-14.15.3'
```
m-rousse, line 87: Nexe does not have a `10.24.1` version, though I think we should bundle the package with the same version of node than the one used to install dependencies/build the package.
Either we should reconsider the node version we use for development or settle on using `14.15.3` for the bundled package only. (and continue testing the package with 10, 12, 14 and 16 versions)

.github/workflows/release.yml
```diff
@@ -14,3 +14,25 @@ jobs:
...
+      - name: Install node
+        uses: actions/setup-node@v1
+        with:
+          node-version: '10.24.1'
+      - name: Install project dependencies
+        run: yarn install --immutable
+      - name: Bundle library
+        run: yarn build
+      - name: Create standalone binary
+        run: yarn dist-standalone
```
m-rousse, line 31: I think this command will fail as it will try to use node 10.24.1 by default, which `nexe` has no release for.

Also this job shares multiple steps with the `standalone-binary-test` job, would it be possible to factorize them? (we could have "build", "test" and "release" jobs all relying on ci artifacts to pass the built bundle between each other if github action supports it)

standalone-e2e/standalone-binary.test.js
```diff
@@ -0,0 +1,80 @@
+const util = require('util')
```
m-rousse, line 1: Would that be OK to write this file in typescript to have all our code in the same language as much as possible?

juan-fernandez:
> When using `nexe`, access to files is changed (cf. [doc](https://github.com/nexe/nexe#resources)), this makes the glob search of synthetics fail (I tested this feature only with `./datadog-ci synthetics run-tests --config path/to/global.config.json` with a `files` configuration set to `"["{,!(node_modules)/**/}staging.synthetics.json"]`)
> 
> I think it may impact some other commands too (IIRC sourcemap upload uses globs too).

that's a good catch, thanks! I'll see what I can do

standalone-e2e/standalone-binary.test.js
```diff
@@ -0,0 +1,80 @@
...
+      const junitHelpText = sanitizeOutput(stdout)
+      expect(junitHelpText).toContain('Upload jUnit XML test reports files to Datadog.')
+    })
+  })
+  describe('lambda', () => {
+    it('can be called', async () => {
+      const {stdout} = await exec(`${STANDALONE_BINARY_PATH} lambda instrument --help`)
+      const lambdaHelpText = sanitizeOutput(stdout)
+      expect(lambdaHelpText).toContain(
+        'datadog-ci lambda instrument [-f,--function #0] [--functions-regex,--functionsRegex #0]'
```
m-rousse, line 53: I'm afraid it may break easily, could we only ensure all commands output the command/subcommand here? (eg. `datadog-ci lambda instrument`)

.github/workflows/release.yml
juan-fernandez: that's a good catch, I missed adding the target in here. 

As for the sharing of artifacts, I'd have to check how that works. It's not a slow step and it's not blocking the release so I think it's fine. We can look into speeding this up later, I think

.github/workflows/ci.yml
juan-fernandez: I'll use `14.15.3` for the bundled package for the moment. We can revisit the other options later

juan-fernandez:
> When using `nexe`, access to files is changed (cf. [doc](https://github.com/nexe/nexe#resources)), this makes the glob search of synthetics fail (I tested this feature only with `./datadog-ci synthetics run-tests --config path/to/global.config.json` with a `files` configuration set to `"["{,!(node_modules)/**/}staging.synthetics.json"]`)
> 
> I think it may impact some other commands too (IIRC sourcemap upload uses globs too).

I can't reproduce. Globs seem to work for `junit upload`. How could I reproduce this? Is there any `staging.synthetics.json` file I can use to test?

juan-fernandez:
> > When using `nexe`, access to files is changed (cf. [doc](https://github.com/nexe/nexe#resources)), this makes the glob search of synthetics fail (I tested this feature only with `./datadog-ci synthetics run-tests --config path/to/global.config.json` with a `files` configuration set to `"["{,!(node_modules)/**/}staging.synthetics.json"]`)
> > I think it may impact some other commands too (IIRC sourcemap upload uses globs too).
> 
> I can't reproduce. Globs seem to work for `junit upload`. How could I reproduce this? Is there any `staging.synthetics.json` file I can use to test?

ok I've managed to reproduce, thanks! I think it gets solved by passing `--no-fs`, as suggested in https://github.com/nexe/nexe/issues/571. Could you try `yarn dist-standalone` now? It seems to be working

m-rousse (approved):
This seems all good to me :+1:

PR #492, opened 2022-03-08
Standalone Binary: Use `pkg` instead of `nexe`

juan-fernandez:
### What and why?

It seems `nexe` requires the app to be packaged before bundling the executable. That means using `ncc` or `rollup` (more context on https://github.com/nexe/nexe/issues/717#issuecomment-724725113). The reason why we didn't catch this testing is that the tests were run in the project folder. The binary created by `nexe` required the javascript files of the project. This made it fail miserably when tested outside. 

`pkg` on the other hand seems to work out of the box without extra configuration or the need to use a bundler like `ncc` or `rollup`. 

Tested in https://github.com/DataDog/datadog-api-client-python/pull/869 and https://github.com/DataDog/dd-trace-js/pull/1894

### How?

* Use `pkg` instead of `nexe`.
* Make sure standalone binary tests *do not* require local files (by removing them before testing). 

### Review checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -87,6 +87,7 @@ jobs:
         run: yarn dist-standalone -t node14-linux-x64 -o datadog-ci-linux-x64
       - name: Remove dist and src folder to check that binary can stand alone
         run: |
+          ls
           rm -rf dist
           rm -rf src
           ls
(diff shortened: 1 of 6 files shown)


m-rousse (approved):
This LGTM, thanks for catching that!

PR #493, opened 2022-03-08
[lambda instrument] Always ask for region in interactive mode

tianchu:
### What?

1. Log the actual error when max retries exceeded when pulling the list of Lambda functions
2. Decouple the interactive mode prompt for AWS region and credentials into separate asks. Because it's possible that one is configured while the other not.
3. In the interactive mode, always ask the user to specify (or confirm the default) aws region.

Image 2022-03-08 at 4 18 07 PM

### Why?

The original issue is that when running the `lambda instrument -i` in interactive mode, the default AWS region configured by the users may not always be correctly detected by the ci. Even when region not detected, the ci continues and throws a max retries error (now it logs the actual error msg). We can fix this by improving the logic to detect the default region, but I realized that a better fix is to **always** ask the user to specify the AWS region or confirm the default:
1. it guarantees that the region is configured (fixing the original error)
2. the user may wants to use a region other than the default one (now it's possible)

### Review checklist

### DIFF
--- a/src/commands/lambda/constants.ts
+++ b/src/commands/lambda/constants.ts
@@ -53,32 +53,6 @@ export const SITES: string[] = [
   'ddog-gov.com',
 ]
 
-export const AWS_REGIONS: string[] = [
-  'us-east-1',
-  'us-east-2',
-  'us-west-1',
-  'us-west-2',
-  'af-south-1',
-  'ap-east-1',
-  'ap-south-1',
-  'ap-northeast-3',
-  'ap-northeast-2',
-  'ap-southeast-1',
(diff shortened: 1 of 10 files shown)


src/commands/lambda/constants.ts
```diff
@@ -53,32 +53,6 @@ export const SITES: string[] = [
   'ddog-gov.com',
 ]
 
-export const AWS_REGIONS: string[] = [
```
maxday, line 56: :100:

maxday (approved):
LGTM 🔥

PR #494, opened 2022-03-09
Support windows and macOS for standalone binary

juan-fernandez:
### What and why?

### How?

### Review checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -69,22 +69,58 @@ jobs:
           DATADOG_APP_KEY: ${{ secrets.datadog_app_key }}
 
   standalone-binary-test:
-    name: Test standalone binary
+    name: Test standalone binary in ubuntu
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
       - name: Install node
         uses: actions/setup-node@v1
         with:
-          node-version: '14.15.3'
-      - uses: actions/setup-python@v2
-        with:
-          python-version: '2.7'
-      - run: npm config set python python2.7
+          node-version: '14.18.3'
       - run: yarn install --immutable
       - run: yarn build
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -32,7 +32,9 @@
   },
   "scripts": {
     "build": "yarn clean; tsc",
+    "build:win": "yarn clean:win & tsc",
     "check-licenses": "node bin/check-licenses.js",
+    "clean:win": "rm dist -r",
```
juan-fernandez, line 37: `;` and `-rf` are not valid in windows' powershell

.github/workflows/ci.yml
```diff
@@ -69,22 +69,58 @@ jobs:
...
-      - uses: actions/setup-python@v2
+          node-version: '14.18.3'
+      - run: yarn install --immutable
+      - run: yarn build
+      - name: Bundle standalone binary
+        run: yarn dist-standalone -t node14-linux-x64 -o datadog-ci-linux
+      - name: Remove dist and src folder to check that binary can stand alone
+        run: |
+          rm -rf dist
+          rm -rf src
```
m-rousse, line 87: Should it also remove `node_modules/` to ensure modules are properly embedded in the binary too?

juan-fernandez: yeah, good idea

juan-fernandez: oops, no, we can't do that. We still use `jest` for running the tests

juan-fernandez: we could get rid of everything and then install `jest`, but we also need `ts-jest`, and maybe more. I'd keep it as is and think about it in the future, WDYT?

m-rousse: Yes, let's keep it as is then! :+1:

PR #496, opened 2022-03-14
[Synthetics] Update `failOnCriticalErrors`, Improve command output

lefebvree:
### What and why?

Improve Synthetics command output to make suite of action clearer, update `failOnCriticalErrors` to also fail on any client errors (config/connection errors, and HTTP response errors except "Test not found" which never result in a critical error).

### How?

- Change `CriticalError` scope.
  `CriticalError` are errors triggering a non-0 command exit when `failOnCriticalErrors` is set, they were scoped to `5xx` server responses only and now include all unexpected errors during tests fetch / trigger / result polling / tunnel setup (with the exception of `404` on tests not found, that will never fail the command).
- Add a message when `failOnCriticalErrors` is `false` and a critical error occured to exit command with 1, add option to the readme.
- Replace test fetched messages to make it clearer that the test has not been triggered yet.
- Display message when tests have been triggered with the number of tests waited for, this is done through a new exposed `testsWait` hook.

### Review checklist

### DIFF
--- a/src/commands/synthetics/__tests__/cli.test.ts
+++ b/src/commands/synthetics/__tests__/cli.test.ts
@@ -337,7 +337,7 @@ describe('run-test', () => {
             }),
           }
           jest.spyOn(runTests, 'getApiHelper').mockImplementation(() => apiHelper as any)
-          jest.spyOn(ciUtils, 'parseConfigFile').mockImplementation(async (config, _) => config)
+          jest.spyOn(ciUtils, 'parseConfigFile').mockImplementation(async (config, __) => config)
 
           expect(await command.execute()).toBe(expectedExit)
           expect(apiHelper.searchTests).toHaveBeenCalledTimes(1)
@@ -354,7 +354,7 @@ describe('run-test', () => {
             }),
           }
           jest.spyOn(runTests, 'getApiHelper').mockImplementation(() => apiHelper as any)
-          jest.spyOn(ciUtils, 'parseConfigFile').mockImplementation(async (config, _) => config)
+          jest.spyOn(ciUtils, 'parseConfigFile').mockImplementation(async (config, __) => config)
           jest.spyOn(utils, 'getSuites').mockImplementation((() => [getTestSuite()]) as any)
 
           expect(await command.execute()).toBe(expectedExit)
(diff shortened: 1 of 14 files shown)


src/commands/synthetics/command.ts
```diff
@@ -82,8 +82,17 @@ export class RunTestCommand extends Command {
...
-        if (error instanceof CriticalError && this.config.failOnCriticalErrors) {
-          return 1
+
+        if (error instanceof CriticalError) {
+          if (this.config.failOnCriticalErrors) {
+            return 1
+          } else {
+            this.reporter.error(
+              chalk.yellow(
+                'Because `failOnCriticalErrors` is not set, the command will exit with an error code 0. Use `failOnCriticalErrors: true` to exit with an error code 1.\n'
```
ruizb, line 92: `failOnCriticalErrors` may be set, but to `false`, and we would still exit with 0 code.

I guess this is nit picking: how about "is not set or disabled"?

src/commands/synthetics/__tests__/cli.test.ts
```diff
@@ -295,4 +300,109 @@ describe('run-test', () => {
       )
     })
   })
+
+  describe('exit code respects `failOnCriticalErrors`', () => {
+    test('404 test not found never exit with 1', async () => {
```
ruizb, line 305: Does this test case cover both non-critical errors "NO_TESTS_TO_RUN" and "NO_RESULTS_TO_POLL"? 🤔

src/commands/synthetics/reporters/default.ts
```diff
@@ -355,25 +368,23 @@ export class DefaultReporter implements MainReporter {
...
     }
 
     this.write(`${idDisplay} ${getMessage()}\n`)
   }
 
   public testWait(test: Test) {
-    const idDisplay = `[${chalk.bold.dim(test.public_id)}]`
-
-    this.write(`${idDisplay} Waiting results for "${chalk.green.bold(test.name)}"\n`)
+    return
```
ruizb, line 388: Is `testWait` not used anymore (and `testsWait` is used as a replacement)? If so, should we add a JS doc with a `@deprecated` flag?

ruizb (approved):

src/commands/synthetics/reporters/default.ts
m-rousse: It is not used in our default reporter, but still available for normal use, so no need to mark it as deprecated

m-rousse (approved):
LGTM

src/commands/synthetics/__tests__/cli.test.ts
lefebvree: Added a test for `NO_RESULTS_TO_POLL` as well 👍

PR #500, opened 2022-03-17
[dsyms] RUMM-1922 Slice fat dSYMs by their cpu architecture

### LINKED ISSUE #428: dSYM upload error due to file size
**Describe what happened**
While using the error tracking new feature for iOS mobile SDK we are getting this error:

Include any error messagFailed upload dSYM for [/var/folders/hr/bqjwwk1x06705z3hzqmdsc740000gp/T/datadog-ci/dsyms/Millennium.app.dSYM/1636131837536/Millennium.app.dSYM]: Request failed with status code 413 (Request Entity Too Large)e or stack trace if available.

According to the documentation there is no limit so this seems like a bug.

Average size sent is around 200MB.

**Additional context**
 - iOS
 - Error tracking

ncreated:
### What and why?

📦 This PR updates `datadog-ci dsyms upload`. Now, it will extract each CPU arch slice from a fat dSYM file into separate slimmed dSYM files before starting the upload. This is to mitigate Intake limit for uploading large files.

```
one fat dSYM: [arm64|armv7] --->  two slimmed dSYMs: [arm64] + [armv7]
```

The size contribution of each arch inside fat dSYM is comparable, meaning that by extracting each slice from N-archs dSYM we cut the size `N` times. Each arch slice has its own `uuid`, so this change has no other impact.

#### New dependency
The arch extraction is done with [lipo](https://www.unix.com/man-page/osx/1/LIPO/). This tool requires macOS environment (like [dwarfdump](https://www.unix.com/man-page/osx/1/dwarfdump/) which we already use for dSYM inspection), so it's behaviour is mocked for unit tests when running on CI (when running on macOS, it uses real `lipo`).

#### New logic
The `dsyms` command now operates in two temporary directories:
* `<tmp>/datadog-ci/dsyms/intermediate` - for extracting slimmed dSYMs;
* `<tmp>/datadog-ci/dsyms/upload` - for creating `<uuid>.zip` archives for each dSYM.

The total size of dSYMs can be significant (hundreds of MBs). To avoid leaving dirty state on user machine, I refactored the way tmp location is managed to ensure that we delete `<tmp>/datadog-ci/` upon completion. To further mitigate this impact, only fat dSYMs are extracted to `intermediate` location.

#### New test fixtures
I replaced previous `dsyms/__tests__/test files/` fixture with new `dsyms/__tests__/fixtures/`. It now includes one single-arch dSYM, one multi-arch dSYM and `all.zip` archive with both.

#### Updated console output
I updated the renderer for nicer output, not mentioning the upload path for each dSYM (only once at the beginning):
```bash
Starting upload with concurrency 20. 
Will look for dSYMs in /Users/maciek.grzybowski/Library/Developer/Xcode/DerivedData/Datadog-buqrwolycbazzjgupavdpyqwjzqb/Build/Products/Release-iphoneos
Will use temporary intermediate directory: /var/folders/4p/_mspg7ld7298z4qkrf_66my40000gn/T/645087472783701.9/datadog-ci/dsyms/intermediate
Will use temporary upload directory: /var/folders/4p/_mspg7ld7298z4qkrf_66my40000gn/T/645087472783701.9/datadog-ci/dsyms/upload
Uploading F76D9B3D-9F91-3F72-95F7-1E0192684EB2.zip (Datadog, arch: arm64, UUID: F76D9B3D-9F91-3F72-95F7-1E0192684EB2)
Uploading 7F54F24E-EB20-3930-90BB-EBEB5DB67C46.zip (DatadogCrashReporting, arch: arm64, UUID: 7F54F24E-EB20-3930-90BB-EBEB5DB67C46)
Uploading 9A24335C-4C1B-3367-8690-A487A93AAA70.zip (DatadogObjc, arch: arm64, UUID: 9A24335C-4C1B-3367-8690-A487A93AAA70)
Uploading 9D4CB426-DCF9-3118-B8A6-C95F10CFACEB.zip (Example, arch: arm64, UUID: 9D4CB426-DCF9-3118-B8A6-C95F10CFACEB)

Command summary:
✅ Uploaded 4 dSYMs in 5.915 seconds.
✨  Done in 9.90s.
```

### Review checklist

- [ ] ~\[_Synthetics_\] New releases of `datadog-ci` MUST be followed by a [synthetics-ci-github-action](https://github.com/DataDog/synthetics-ci-github-action) release~

Fixes #428


src/commands/dsyms/utils.ts
```diff
@@ -26,64 +17,37 @@ export const isZipFile = async (filepath: string) => {
...
-    const regexMatches = line.match(UUID_REGEX)
-    if (regexMatches && regexMatches.length > 0) {
-      uuids.push(regexMatches[0])
-    }
-  })
-
-  return uuids
+export const deleteDirectory = async (directoryPath: string) => {
+  await execute(`rm -rf '${directoryPath}'`)
 }
```
ncreated, line 30: @lraucy @m-rousse I'm not OK with this `rm -rf` as it might be dangerous. Unfortunately, I couldn't find other way for deleting `<tmp>/datadog-ci` recursively with our `node` setup. [Here](https://github.com/DataDog/datadog-ci/pull/300#discussion_r683321374) we mentioned [rimraf](https://www.npmjs.com/package/rimraf) - do you think this is sth I could look into?

Alternatively, as I know the intermediate directories structure I could iterate it recursively and use non-recursive `fs.unlink()` on each file starting from leaf elements. Not sure if that would be generic solution though.

Otherwise, as we only create directories inside `/tmp` we could orphan it and let the OS flush it when needed. I'm not a fan of this neither - now when receiving 250MB of fat dSYMs at input, we will be creating 250MB of slimmed dSYMs in `<tmp>/.../intermediate/` and ~70MB of compressed payloads in `<tmp>/.../upload/`.

m-rousse (approved):
Ok for the `.gitignore` change (the only one I am codeowner for)

src/commands/dsyms/utils.ts
m-rousse: I'd be in favor of introducing `rimraf` for this purpose :+1:

ncreated: Thanks @m-rousse. I did it (https://github.com/DataDog/datadog-ci/pull/500/commits/7b47146b0a34c8d780460a262b613331010b5df7). I had to add [`rimraf`](https://www.npmjs.com/package/rimraf) and [`@types/rimraf`](https://www.npmjs.com/package/@types/rimraf) for TS support. I hope that's correct (please confirm, I'm very new to `node` / `yarn` / `ts`).

m-rousse: This looks good to me :+1:

xgouchet (approved):

PR #508, opened 2022-03-24
Update ssh2 to a version whose dependencies build on M1

technicalpickles:
### What and why?

Our app has one optional dependency (cpu-features) that fails to build on M1, and I've been able to prod along fixed releases to eventually arrive here.

Previous efforts:
- https://github.com/mscdex/ssh2/issues/1083
- https://github.com/mscdex/cpu-features/issues/7
- https://github.com/google/cpu_features/issues/179

### How?

This updates the ssh2 version to 1.8.0. I was also considering using ^1.4.0 to allow all 1.x releases.

### Review checklist


m-rousse:
Hi @technicalpickles,
Thanks for your contribution!
You mention using version `1.18.0` in the description however the diff shows version `1.8.0` - just to make sure this is indeed version `1.8.0` that is expected?
Also, could you run `yarn` to update the lockfile and also commit the changes to `yarn.lock`?

technicalpickles:
@m-rousse thanks! It was supposed to be 1.8.0. I pushed up a change with the yarn.lock.

.node-version
```diff
@@ -1 +1 @@
-10.24.1
+16.13.2
```
m-rousse, line 1: Actually we prefer to keep the package compatible with node 10 as long as there is no express need to bump to a newer version. Could you revert this change?

m-rousse (approved):
:+1:

technicalpickles:
Thank you!

PR #510, opened 2022-04-01
[CIAPP-3830] Add internal tag to junit upload to enable logs collection

AdrianLC:
### What and why?
This is a new flag to opt in to send logs from junit XML imports. We need a way to opt in as logs have an additional cost.

### How?
There is a new internal tag `_dd.junitxml_logs` that is already being read in the backend.

### Review checklist

### DIFF
--- a/src/commands/junit/upload.ts
+++ b/src/commands/junit/upload.ts
@@ -77,10 +77,10 @@ export class UploadJUnitXMLCommand extends Command {
   }
   private dryRun = false
   private env?: string
+  private logsEnabled = false
   private maxConcurrency = 20
   private service?: string
   private tags?: string[]
-  private logsEnabled = false
 
   public async execute() {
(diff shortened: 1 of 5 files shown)


src/commands/junit/upload.ts
```diff
@@ -97,6 +102,14 @@ export class UploadJUnitXMLCommand extends Command {
       this.config.env = this.env
     }
 
+    if (
+      !this.logs &&
```
juan-fernandez, line 106: why do you need to check `!this.logs` ?

juan-fernandez (approved):

src/commands/junit/upload.ts
AdrianLC: To make --logs override the env var value so that it can be true even if the env var is not defined

PR #511, opened 2022-04-01
[Synthetics] Add `testResult` reporting, Allow passing suites to `executeTests`

lefebvree:
### What and why?

This PR adds non-breaking features to extends datadog-ci-as-a-library capabilities:
- Add a new reporter function `testResult()` that will be called every time we receive the result of a waited test. This allows app using datadog-ci to get real-time events for results instead of waiting for all results of a test to be received.
- Add a new `executeTests()` optional `suites` argument to provide already parsed lists of tests for which tests overrides will be computed without relying on internal [`glob`](https://www.npmjs.com/package/glob) suite files read (which can fail in environment with no / non standard filesystem or permissions like a VSCode extension).

### How?

- Add a new reporter function `testResult()` (noop for default reported, optional for others). This requires passing the reporter through `waitForResults` and changing its definition but although it is exported, this function is only used internally in `executeTests`.
- Extends `getTestsList` returned tests with any optional `suites` provided in `executeTests`.

### Review checklist


src/commands/synthetics/README.md
```diff
@@ -255,12 +255,13 @@ Reporters can hook themselves into the `MainReporter` of the command.
...
 | `testTrigger` | `(test: Test, testId: string, executionRule: ExecutionRule, config: ConfigOverride)`    | called when a test is triggered.                                |
 | `testWait`    | `(test: Test)`                                                                          | called when a test is waiting to receive its results.           |
 | `testsWait`   | `(tests: Test[])`                                                                       | called when all tests are waiting to receive their results.     |
+| `testResult`  | `(trigger: TriggerResponse, result: PollResult)`                                        | called when a test result is received.                          |
+| `testEnd`     | `(test: Test, results: PollResult[], baseUrl: string, locationNames: LocationsMapping)` | called when all results for a test are received.                |
+| `runEnd`      | `(summary: Summary)`                                                                    | called at the end of the run.                                   |
```
lefebvree, line 267: Reordered by the hooks call order in a complete test run timeline, added `testResult` and rephrased `testEnd` to mention it triggers when _all_ results are received

src/commands/synthetics/index.ts
```diff
@@ -1,5 +1,5 @@
 export {CiError, CriticalError} from './errors'
 export * from './interfaces'
 export {DefaultReporter} from './reporters/default'
-export {executeTests} from './run-test'
+export {executeTests, getApiHelper} from './run-test'
```
m-rousse, line 4: Why is it necessary to export `getApiHelper`? from a library point of view, it is rather internal

ruizb (approved):

src/commands/synthetics/index.ts
lefebvree: Stopped exporting `getApiHelper()` in favor of utils `fetchTest()` function

m-rousse (approved):

PR #512, opened 2022-04-01
Check for aws credentials in ~/.aws/credentials as well as environment variables in interactive mode

zARODz11z:
### What and why?
https://datadoghq.atlassian.net/browse/SLS-2109
`datadog-ci lambda instrument -i` didnt check for aws credentials in `~/.aws/credentials` and only looked for [environment variables](https://docs.datadoghq.com/serverless/installation/python/?tab=datadogcli#:~:text=gov.com%0Aexport-,AWS_ACCESS_KEY_ID,-%3D%22%3CACCESS%20KEY%20ID) before prompting the user to enter their credentials.
### How?
Add some functionality to check for aws creds in `~/.aws/credentials` by calling aws-sdk's config.credentials. One downside is config.credentials returns an object like: 
Screen Shot 2022-04-01 at 3 04 27 PM
As you can see it exposes aws_access_key_id but the secret key is never returned in an object. I don't think this is a security concern since the secret key isn't the one being returned to that object but I figured I would bring it up if there are any suggestions

### Review checklist

### DIFF
--- a/src/commands/lambda/__tests__/functions/commons.test.ts
+++ b/src/commands/lambda/__tests__/functions/commons.test.ts
@@ -1,6 +1,6 @@
 /* tslint:disable:no-string-literal */
-import {config, Credentials} from "aws-sdk"
-jest.mock("aws-sdk")
+import {config as aws_sdk_config, Credentials} from 'aws-sdk'
+jest.mock('aws-sdk')
 import {Lambda} from 'aws-sdk'
 import {
   AWS_ACCESS_KEY_ID_ENV_VAR,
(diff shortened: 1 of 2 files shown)


zARODz11z:
In this scenario I did not have the env vars set but I did have a default profile in my `.aws/credentials` so the ci tool skipped asking me for my creds. This should make the ux smoother.

Screen Shot 2022-04-01 at 3 30 46 PM


However if their is no default profile in `.aws/credentials` and the env vars are unset then the tool will still prompt for creds as it should. Screen Shot 2022-04-01 at 3 34 09 PM

src/commands/lambda/__tests__/functions/commons.test.ts
```diff
@@ -287,20 +288,32 @@ describe('commons', () => {
...
 
+    test('returns true when only AWS_ACCESS_KEY_ID environment variable is set and `~/.aws/credentials` are missing', () => {
       process.env[AWS_ACCESS_KEY_ID_ENV_VAR] = 'SOME-AWS-ACCESS-KEY-ID'
-      expect(isMissingAWSCredentials()).toBe(true)
+      aws_sdk_config.credentials = undefined
+      expect(isMissingAWSCredentials()).toBe(true) // We return true since AWS_SECRET_ACCESS_KEY_ENV_VAR is missing
     })
 
-    test('returns false when AWS credentials are set', () => {
+    test('returns false when AWS credentials via environment variabiables are set', () => {
```
hghotra, line 303:
```suggestion
    test('returns false when AWS credentials via environment variables are set', () => {
```

src/commands/lambda/functions/commons.ts
```diff
@@ -171,9 +171,21 @@ export const findLatestLayerVersion = async (layer: LayerKey, region: string) =>
...
+  ) {
+    if (aws_sdk_config.credentials) {
+      return false
+    } else {
+      return true
+    }
+  } else {
+    return false
+  }
+}
```
hghotra, line 188: This can be simplified to:

```suggestion
export const isMissingAWSCredentials = () => {
  // If env vars and aws_sdk_config.credentials are not set return true otherwise return false
  return
    (process.env[AWS_ACCESS_KEY_ID_ENV_VAR] === undefined ||
    process.env[AWS_SECRET_ACCESS_KEY_ENV_VAR] === undefined) &&
    !aws_sdk_config.credentials
  ) 
}
```

zARODz11z: ooooo this is way cleaner. Seeing this now makes sense but what sort of tips do you have on simplifying boolean logic like the one I had set earlier?

src/commands/lambda/__tests__/functions/commons.test.ts
zARODz11z: https://github.com/DataDog/datadog-ci/pull/512/commits/80712c3a03f9d87117b62761f770c0c7baf8bc32

src/commands/lambda/functions/commons.ts
zARODz11z: https://github.com/DataDog/datadog-ci/pull/512/commits/80712c3a03f9d87117b62761f770c0c7baf8bc32

hghotra: You should create a truth table with all the inputs and the outputs and then try to simplify the expression into its simplest form that still produces the same outputs for the truth table as the original expression.

PR #513, opened 2022-04-05
Make select target function easier with searchable inquirer prompt

zARODz11z:
### What and How?

Uses [inquirer-checkbox-plus-prompt](https://github.com/mokkabonna/inquirer-autocomplete-prompt) which is listed as one of the supported inquirer plugins on the inquier [readme](https://github.com/SBoudrias/Inquirer.js#:~:text=Checkbox%20list%20with%20autocomplete%20and%20other%20additions). This makes it so the checkbox is searchable by the user so that they wont have to manually scroll through the functions to instrument the one they wish. They can get there faster by searching. see the gif https://a.cl.ly/YEurPzpN

### Why?
Manually scrolling through a bunch of functions is tedious
https://datadoghq.atlassian.net/browse/SLS-2111

### Review checklist


src/commands/lambda/prompt.ts
```diff
@@ -1,13 +1,6 @@
...
-  CheckboxQuestion,
-  ConfirmQuestion,
-  InputQuestion,
-  ListQuestion,
-  prompt,
-  QuestionCollection,
-  Separator,
-} from 'inquirer'
+import {ConfirmQuestion, InputQuestion, ListQuestion, prompt, QuestionCollection, Separator} from 'inquirer'
+import inquirer from 'inquirer'
```
zARODz11z, line 3: I know these should be grouped into one import but for some reason I get an error when trying to import `inquirer` on line 2. https://a.cl.ly/7KuQkdkx
not really sure what is the issue with that

src/commands/lambda/prompt.ts
```diff
@@ -24,6 +17,10 @@ import {
 } from './constants'
 import {sentenceMatchesRegEx} from './functions/commons'
 
+// @ts-ignore
```
zARODz11z, line 20: I know using the ignore is kinda hacky but I this is the way I was able to get it to work.
Without the ignore I get
```
Could not find a declaration file for module 'inquirer-checkbox-plus-prompt'. '/Users/andrew.rodriguez/go/src/github.com/DataDog/datadog-ci/node_modules/inquirer-checkbox-plus-prompt/index.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/inquirer-checkbox-plus-prompt` if it exists or add a new declaration (.d.ts) file containing `declare module 'inquirer-checkbox-plus-prompt';`ts(7016)
```

Where would I put this .d.ts file?

package.json
```diff
@@ -61,8 +61,10 @@
     "deep-extend": "0.6.0",
     "fast-xml-parser": "3.19.0",
     "form-data": "3.0.0",
+    "fuzzy": "^0.1.3",
     "glob": "7.1.4",
     "inquirer": "8.2.0",
+    "inquirer-checkbox-plus-prompt": "^1.0.1",
```
zARODz11z, line 67: another concern Tian pointed out is how do we know this package is reliable? It solves the issue but i'm not sure how to decide if it's reliable or not

src/commands/lambda/prompt.ts
zARODz11z: hmmm I think I understand why I received the error, is it because I was sort of mixing the two import styles you described above?. The problem is that inquirer isn't some available property, instead what I did on line 3 was import the whole module and all its properties/methods as `inquirer` and that is no problem. Does this sound right?

I think to resolve this issue I should use the 2nd style of importing and prefix all the previously used inquirer stuff with `inquirer.`

zARODz11z: made the change in https://github.com/DataDog/datadog-ci/pull/513/commits/9fdd2b761d574c6020bfab0fb831d639df74e941

astuyve: Great answer, you're correct. You can prefix the rest of the methods with `inquirer`, or you could simply add the specific methods you used (registerPrompt, QuestionCollection, InputQuestion, ListQuestion, Separator, etc) to the existing import.

package.json
astuyve: There are a few key things I look for:
1. Is the API stable? If it's > 1.x, the answer is usually `yes`.
2. Is this core to my functionality? If not, can we do it ourself?
3. Is the project active? If it's not, are we comfortable owning our use of it in our application or library?
4. Does it have any dependencies? If so, are we comfortable taking them on?
5. What's the license? Is it compatible with the license we use?

In this case, the `inquirer-checkbox-plus-prompt` is pretty simple, around 500 lines of code. It's not actively updated, but it might literally just be `done` and not need any updates.

I'd investigate if we could take just the methods you need and add them (with attribution, of course, if the license allows).

src/commands/lambda/prompt.ts
sdeprez: You can use `require` to avoid the `ts-ignore` (it may require to disable `eslint` though) but if the type is simple, you should indeed add your `d.ts` (next to that file for instance).

src/commands/lambda/prompt.ts
```diff
@@ -125,18 +121,31 @@ export const datadogEnvVarsQuestions = (datadogApiKeyType: Record<string, any>):
...
-  message: 'Select the functions to modify',
+  highlight: true,
+  message:
+    'Select the functions to modify (Press <space> to select, p.s. start typing the name instead of manually scrolling)',
   name: 'functions',
-  type: 'checkbox',
-  validate: (selectedFunctions) => {
+  pageSize: 10,
+  searchable: true,
+  source: (answersSoFar: any, input: any) => {
```
sdeprez, line 138: Can we type those? If not, prefer `unknown`

src/commands/lambda/prompt.ts
```diff
@@ -125,18 +121,31 @@ export const datadogEnvVarsQuestions = (datadogApiKeyType: Record<string, any>):
...
+    input = input || ''
+
+    return new Promise((resolve) => {
+      const fuzzyResult = filter(input, functionNames)
+      const data = fuzzyResult.map((element) => element.original)
+      resolve(data)
+    })
+  },
+  type: 'checkbox-plus',
+  validate: (selectedFunctions: string | any[]) => {
```
sdeprez, line 148: Same remark

sdeprez (approved):
Let's improve the typings if possible, otherwise LGTM

src/commands/lambda/prompt.ts
zARODz11z: updated in latest commit

src/commands/lambda/prompt.ts
zARODz11z: updated in latest commit

PR #516, opened 2022-04-13
Ar/ask for service env version tags in interactive

zARODz11z:
### What and why?
`datadog-ci lambda instrument -i` (interactive mode) does not allow the user to add env, service, and version environment variables (DD_ENV, DD_SERVICE, DD_VERSION). This PR adds some extra prompts to allow users in interactive mode to add those environment variables. See screenshots

Screen Shot 2022-04-13 at 2 59 00 PM
Screen Shot 2022-04-13 at 3 03 41 PM

### How?

Added 3 more inquirer prompts.

### Review checklist


src/commands/lambda/__tests__/instrument.test.ts
```diff
@@ -1216,6 +1223,60 @@ ${red('[Error]')} Unexpected error
...
+          process.env[SERVICE_ENV_VAR] = mockDatadogService
+          process.env[VERSION_ENV_VAR] = mockDatadogVersion
+        })
+
+        const cli = makeCli()
+        const context = createMockContext() as any
+        const code = await cli.run(['lambda', 'instrument', '-i'], context)
+        const output = context.stdout.toString()
+        expect(code).toBe(0)
+        expect(output).toMatchInlineSnapshot()
```
zARODz11z, line 1276: this should have a snapshot but I cant get it to generate

src/commands/lambda/instrument.ts
```diff
@@ -483,6 +501,30 @@ export class InstrumentCommand extends Command {
...
   }
 
+  private setEnvServiceVersion() {
+    const environment = process.env[ENVIRONMENT_ENV_VAR]
+    if (environment === '') {
+      // If a user doesn't enter something then inquirer will set the answer to an empty string and in that case we don't want to set these environment variables
+      this.environment = undefined
+    } else {
+      this.environment = environment
+    }
```
sdeprez, line 511: Nit but checking against the empty string is also a bit brittle IMO, why if the library sets that to `undefined` or `null` in the future? I would treat everything the same by simply checking `if (!environment)`. Also in that case you can simply shorten to `this.environment = process.env[ENVIRONMENT_ENV_VAR]  || undefined`.

Same remarks for the two checks below.

src/commands/lambda/prompt.ts
```diff
@@ -189,6 +213,26 @@ export const requestDatadogEnvVars = async () => {
...
+
+    const serviceQuestionAnswer = await prompt(serviceQuestion)
+    const inputedServiceQuestionAnswer = serviceQuestionAnswer[SERVICE_ENV_VAR]
+    process.env[SERVICE_ENV_VAR] = inputedServiceQuestionAnswer
+
+    const versionQuestionAnswer = await prompt(versionQuestion)
+    const inputedVersionQuestionAnswer = versionQuestionAnswer[VERSION_ENV_VAR]
+    process.env[VERSION_ENV_VAR] = inputedVersionQuestionAnswer
+  } catch (e) {
+    if (e instanceof Error) {
```
sdeprez, line 230: Why that check?

tsconfig.json
```diff
@@ -3,7 +3,7 @@
     "target": "es6",
     "module": "commonjs",
     "declaration" : true,
-    "sourceMap": false,
+    "sourceMap": true,
```
sdeprez, line 6: Why this change?

sdeprez (approved):
A few minor comments otherwise LGTM

src/commands/lambda/prompt.ts
```diff
@@ -110,6 +113,27 @@ const datadogSiteQuestion: ListQuestion = {
   type: 'list',
 }
 
+const envQuestion: InputQuestion = {
+  default: '',
```
lizapressman, line 117: Can we default to undefined to avoid having to do those checks for the empty string in `setEnvServiceVersion`?

lizapressman (approved):
Looks overall good to me with one question: 

Can a user instrument with interactive mode and set the forwarder instead of the extension or do we just default to setting the extension? 
If we do allow for the forwarder instead of the extension, we want to set tags instead of env vars

tsconfig.json
zARODz11z: was having an issue producing jest inline snapshots https://github.com/facebook/jest/issues/6744#issuecomment-861722016

src/commands/lambda/prompt.ts
zARODz11z: honestly I just copied the convention from the other try catch statements such as  https://github.com/DataDog/datadog-ci/blob/4b50c3b591ffcef1d3aca4876d6dc2e813b5bcf2/src/commands/lambda/prompt.ts#L181

src/commands/lambda/instrument.ts
zARODz11z: interesting, I didn't know this could do the same thing and I see how this is more robust, Thanks

src/commands/lambda/prompt.ts
zARODz11z: > Looks overall good to me with one question:
> 
> Can a user instrument with interactive mode and set the forwarder instead of the extension or do we just default to setting the extension? If we do allow for the forwarder instead of the extension, we want to set tags instead of env vars

interactive mode defaults to the extension and automatically picks up the right lambda library layer depending on the runtime. If we wanted to set the forwarder in interactive mode I think we would need to add some more logic but I think that should be in a seperate pr

PR #519, opened 2022-04-20
Release v1.3.0

juan-fernandez:
### What and why?

Release `v1.3.0`: 

* #510
* #512 
* #515 
* #516 
* #513

### DIFF
--- a/src/commands/lambda/__tests__/instrument.test.ts
+++ b/src/commands/lambda/__tests__/instrument.test.ts
@@ -1318,7 +1318,7 @@ UpdateFunctionConfiguration -> arn:aws:lambda:sa-east-1:123456789012:function:la
 }
 TagResource -> arn:aws:lambda:sa-east-1:123456789012:function:lambda-hello-world
 {
-  \\"dd_sls_ci\\": \\"v1.2.0\\"
+  \\"dd_sls_ci\\": \\"v1.3.0\\"
 }
 ${yellow('[!]')} Confirmation needed.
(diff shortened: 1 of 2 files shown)


src/commands/lambda/__tests__/instrument.test.ts
```diff
@@ -1318,7 +1318,7 @@ UpdateFunctionConfiguration -> arn:aws:lambda:sa-east-1:123456789012:function:la
 }
 TagResource -> arn:aws:lambda:sa-east-1:123456789012:function:lambda-hello-world
 {
-  \\"dd_sls_ci\\": \\"v1.2.0\\"
+  \\"dd_sls_ci\\": \\"v1.3.0\\"
```
lefebvree, line 1321: A bit tedious to update every time, we should look for a way to mock the version there in the future

lefebvree (approved):

src/commands/lambda/__tests__/instrument.test.ts
juan-fernandez: agreed. Would it be as simple as reading `package.json` from the root?

lefebvree: Yes, it seems it was already done in some tests. Created a small PR to update it: https://github.com/DataDog/datadog-ci/pull/520

juan-fernandez: nice, thanks

PR #525, opened 2022-04-29
[synthetics] Support multiple config overrides for the same test

ruizb:
### What and why?

This PR introduces the possibility to trigger multiple tests that have the same public ID, but different config overrides.

### How?

* Make sure to not display duplicated test results output.
* Adapt the output when fetching the tests to mention the number of config overrides.

### Review checklist


src/commands/synthetics/__tests__/reporters/default.test.ts
```diff
@@ -32,4 +32,101 @@ describe('Default reporter', () => {
...
+      name: 'Request on example.org',
+    }
+    const testId = 'aaa-bbb-ccc'
+
+    const cases: [string, ExecutionRule, ConfigOverride, RegExp][] = [
+      [
+        'Blocking test, without config overwrite',
+        ExecutionRule.BLOCKING,
+        {},
+        /\[.*aaa-bbb-ccc.*\].*Found test.*Request on example\.org.*\n?$/,
```
ruizb, line 51: I had to rely on regular expressions here as the output contains ANSI characters included by the Chalk library.

We could use the [strip-ansi](https://github.com/chalk/strip-ansi) library to remove these characters, and test the text in a stricter manner.

I didn't want to import another library in this project, but it's doable. Feel free to suggest it if you think it would be better 🙂

src/commands/synthetics/command.ts
```diff
@@ -119,12 +119,17 @@ export class RunTestCommand extends Command {
...
+
     const locationNames = triggers.locations.reduce((mapping, location) => {
       mapping[location.id] = location.display_name
 
       return mapping
     }, {} as LocationsMapping)
+
     let hasSucceeded = true // Determine if all the tests have succeeded
+
+    const testResultsAlreadyDisplayed = new Set<string>()
```
lefebvree, line 131: If I understand correctly the issue is that we iterate on `tests` which will contain multiple times the same test. 

But here we'll be iterating multiple times on the same results (`const testResults = results[test.public_id]`) for each override and we'll be updating the summary multiple times for the same results. One way we could fix that is just by iterating on `results`.

However with this interface no way to properly retrieve the `executionRule` of the override associated with the test run and fail only when we are supposed to 🤔 
- 2 overrides, one blocking other non-blocking
- only the non-blocking result fail
- `hasSucceeded` would be set to false I believe?

One way we could fix this in a non-breaking way could be to add a reference to the `Test` override associated with each `PollResult` (`PollResult.test: Test`) and start ignoring `tests` in this reporter, wdyt?

src/commands/synthetics/reporters/default.ts
```diff
@@ -381,7 +381,16 @@ export class DefaultReporter implements MainReporter {
...
     }
 
-    this.write(`${idDisplay} ${getMessage()}\n`)
+    const getConfigOverridesPart = () => {
+      const nbConfigsOverridden = Object.keys(config).length
+      if (nbConfigsOverridden === 0) {
+        return ''
+      }
+
+      return ' ' + chalk.gray(`(${nbConfigsOverridden} config${nbConfigsOverridden !== 1 ? 's' : ''} overridden)`)
```
lefebvree, line 390: nit: maybe `x config overrides` instead of `x configs overridden`?

src/commands/synthetics/reporters/default.ts
```diff
@@ -360,7 +360,7 @@ export class DefaultReporter implements MainReporter {
     )
   }
 
-  public testTrigger(test: Test, testId: string, executionRule: ExecutionRule, config: ConfigOverride) {
+  public testTrigger(test: Pick<Test, 'name'>, testId: string, executionRule: ExecutionRule, config: ConfigOverride) {
```
lefebvree, line 363: Why this change?

src/commands/synthetics/__tests__/reporters/default.test.ts
lefebvree: Using Regex here doesn't bother me personally, or we could also use `chalk.` in the expected output maybe

src/commands/synthetics/reporters/default.ts
ruizb: So I could call `defaultReporter.testTrigger` from a Jest test without building a whole `Test` object. In this method, we only use the name of the test, so requiring a whole object seems unnecessary IMO 🙂 (I believe this is called the "principle of least power"?)

I didn't change the interface definition, so custom reporters can still use the whole object, if need be.

src/commands/synthetics/command.ts
ruizb: I did not anticipate the amount of changes we had to make in order to take into account this case. I took this opportunity to write some Jest tests (~70% of the changes). I will try to create separate pull-requests to ease the review, or group the set of changes in commits, on the same PR.

ruizb:
Closing in favor of:
* https://github.com/DataDog/datadog-ci/pull/533
* https://github.com/DataDog/datadog-ci/pull/534

PR #526, opened 2022-05-03
[CIAPP] Add tag command

cg14823:
### What and why?

Introduces a new command `tag` that can be used in CI pipelines for `[GitLab, CirlceCi, GitHub, Buildkite]` to add arbitrary tags to the CI Visibility traces for the given pipeline.

### How?

The command sends the user defined tags to an endpoint together with the runner environmental variables that will be later used to match CI Visibility traces with the corresponding tags.

### Review checklist


src/commands/tag/tag.ts
```diff
@@ -0,0 +1,139 @@
...
+import {getCIEnv} from '../../helpers/ci'
+import {retryRequest} from '../../helpers/retry'
+import {parseTags} from '../../helpers/tags'
+import {getApiHostForSite, getRequestBuilder} from '../../helpers/utils'
+
+export class TagCommand extends Command {
+  public static usage = Command.Usage({
+    description: 'Add tags to a CI Pipeline trace pipeline or job span in Datadog.',
+    details: `
+            This command when run from a supported CI provider sends an arbitrary st of key:value
```
juan-fernandez, line 12: did you mean string? 
```suggestion
            This command when run from a supported CI provider sends an arbitrary string of key:value
```

src/commands/tag/tag.ts
```diff
@@ -0,0 +1,139 @@
...
+    if (!this.config.apiKey) {
+      this.context.stdout.write(
+        `Neither ${chalk.red.bold('DATADOG_API_KEY')} nor ${chalk.red.bold('DD_API_KEY')} is in your environment.\n`
+      )
+      throw new Error('API key is missing')
+    }
+
+    const site = process.env.DATADOG_SITE || process.env.DD_SITE || 'datadoghq.com'
+    const baseAPIURL = `https://${getApiHostForSite(site)}`
+    const instance = getRequestBuilder({baseUrl: baseAPIURL, apiKey: this.config.apiKey})
```
juan-fernandez, line 95: maybe it's clear if you use
```suggestion
    const request = getRequestBuilder({baseUrl: baseAPIURL, apiKey: this.config.apiKey})
```

src/commands/tag/tag.ts
```diff
@@ -0,0 +1,139 @@
...
+            },
+            type: 'ci_custom_tag',
+          },
+        },
+        method: 'post',
+        url: 'api/v2/ci/pipeline/tags',
+      })
+
+      return resp
+    }
```
juan-fernandez, line 115: I think this could simply be
```suggestion
    const doRequest = () => {
      return request({
        data: {
          data: {
            attributes: {
              ci_env: ciEnv,
              ci_level: level,
              provider,
              tags,
            },
            type: 'ci_custom_tag',
          },
        },
        method: 'post',
        url: 'api/v2/ci/pipeline/tags',
      })
    }
```

src/helpers/ci.ts
```diff
@@ -546,3 +546,40 @@ const parsePipelineNumber = (pipelineNumberStr: string | undefined): number | un
     return isFinite(pipelineNumber) ? pipelineNumber : undefined
   }
 }
+
+export const getCIEnv = (): {ciEnv: Record<string, string>; provider: string} => {
```
juan-fernandez, line 550: why do you need a new util to extract the environment? Couldn't you reuse the ones we use for `junit`?

juan-fernandez (commented):
this looks great! Amazing that it's your first PR here 😄 . 

Only left minor comments, but worth looking before merging

src/commands/tag/README.md
```diff
@@ -0,0 +1,54 @@
...
+Additionally you might configure the `tag` command with environment variables:
+
+- `DATADOG_API_KEY` or `DD_API_KEY` (**required**): API key used to authenticate the requests.
+- `DD_TAGS`: set global tags applied to all spans. The format must be `key1:value1,key2:value2`.
+- `DATADOG_SITE`: choose your Datadog site, e.g. datadoghq.com or datadoghq.eu.
+
+### Supported providers
+
+The tag command only works for the following providers: [Buildkite, CircleCI, GitHub, GitLab]. If used in
+any other provider it will fail. Note that for GitHub actions only the level `pipeline` is supported.
```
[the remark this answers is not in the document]
cg14823: It will fail an return an error message such as:

```
Cannot use level "job" for GitHub Actions.
```

I'll clarify it in the Readme

src/commands/tag/tag.ts
cg14823: I think I meant to write `set` and probably missed the `e`

src/helpers/ci.ts
cg14823: The junit one remaps things to the tags we use so things like `ci.pipeline.name`. For this to work we need the runner env variables including some that we do not map to any of our specific ci tags.

src/helpers/ci.ts
```diff
@@ -546,3 +546,40 @@ const parsePipelineNumber = (pipelineNumberStr: string | undefined): number | un
...
+    return {
+      ciEnv: getEnvs('BUILDKITE_'),
+      provider: 'buildkite',
+    }
+  }
+
+  throw new Error('Only providers [GitHub, GitLab, CircleCI, Buildkite] are supported')
+}
+
+const getEnvs = (prefix: string): Record<string, string> =>
```
juan-fernandez, line 582: I'd rename this to `getEnvVars`

src/helpers/ci.ts
```diff
@@ -546,3 +546,40 @@ const parsePipelineNumber = (pipelineNumberStr: string | undefined): number | un
...
+  Object.entries(process.env)
+    .filter(([key, value]) => key.startsWith(prefix) && !/(PASSWORD)|(TOKEN)|(SECRET)|(KEY)/i.test(key))
```
juan-fernandez, line 584: maybe this is safer? 
```suggestion
    .filter(([key, value]) => key.startsWith(prefix) && !/(PASS)|(TOKEN)|(SECRET)|(KEY)/i.test(key))
```

juan-fernandez (commented):
a couple of nits

juan-fernandez (approved):
great job!

PR #530, opened 2022-05-05
[CIAPP] Add custom metrics command

cg14823:
### What and why?

Very similar to https://github.com/DataDog/datadog-ci/pull/526.

Introduces a new command `metric` that can be used in CI pipelines for [GitLab, CirlceCi, GitHub, Buildkite] to add arbitrary numerical tags to the CI Visibility traces for the given pipeline.

### How?

The command sends the user defined metrics to an endpoint together with the runner environmental variables that will be later used to match CI Visibility traces with the corresponding tags.

### Review checklist


src/commands/metric/README.md
```diff
@@ -0,0 +1,53 @@
+# metric command
+
+Add numeric tags to CI Visibility pipeline and job spans.
+
+## Usage
+
+```bash
+datadog-ci metric [--no-fail] [--level <pipeline|job>] [--tags]
```
juan-fernandez, line 8:
```suggestion
datadog-ci metric [--no-fail] [--level <pipeline|job>] [--metrics]
```

src/commands/metric/README.md
```diff
@@ -0,0 +1,53 @@
...
+```
+
+For example:
+
+```bash
+datadog-ci metric --level job --tags binary.size:1024
```
juan-fernandez, line 14:
```suggestion
datadog-ci metric --level job --metrics binary.size:1024
```

src/commands/metric/README.md
```diff
@@ -0,0 +1,53 @@
...
+### End-to-end testing process
+
+To verify this command works as expected, you can tag a mock pipeline and validate the command returns 0:
+
+```bash
+export DATADOG_API_KEY='<API key>'
+export BUILDKITE=true
+export BUILDKITE_BUILD_ID=uuid
+
+yarn launch metric --level pipeline --tags foo:1
```
juan-fernandez, line 45:
```suggestion
yarn launch metric --level pipeline --metrics foo:1
```

juan-fernandez (commented):
minor typos

README.md
```diff
@@ -85,6 +86,7 @@ Further documentation for each command can be found in its folder, ie:
 - [JUnit XML](src/commands/junit)
 - [Trace](src/commands/trace)
 - [Tag](src/commands/tag)
+- [Metric] (src/command/tag)
```
AdrianLC, line 89:
```suggestion
- [Metric] (src/command/metric)
```

AdrianLC (changes requested):
Blocking for the typo mounting docs.

But a couple suggestions for the CLI interface:
- If `--level` is required and has no default value, it should not be an optional style flag with `--` but a mandatory positional parameter.
- Same for `--tags`, `--metrics`, if they are required it should be positional parameters.

I'd do something like this:
```
datadog-ci metric pipeline -- binary.size:500 command.runtime:67.1
```
Another option would be to default `--level=job` and require it only to set the level to pipeline. That would make sense too.

Anyway I have no idea how to actually make clipanion do this

cg14823:
Regarding flags versus positional arguments. This is my personal opinion, but I generally find it easier to user and understand CLIs that work with flags rather than positional arguments are they are more self-documenting

e.g:

```
datadog-ci tag --level pipeline --tags a:b --tags c:d
```

vs

```
datadog-ci tag pipeline a:b c:d
```

Although longer I find the flag one easier to understand.

src/commands/metric/metric.ts
```diff
@@ -0,0 +1,143 @@
...
+
+export class MetricCommand extends Command {
+  public static usage = Command.Usage({
+    description: 'Add metrics to a CI Pipeline trace pipeline or job span in Datadog.',
+    details: `
+            This command when run from a supported CI provider sends an arbitrary set of key:value
+            numeric tags to Datadog to include in the CI Visibility traces.
+    `,
+    examples: [
+      ['Add a binary size to the current pipeline', 'datadog-ci metric --level pipeline --tags binary.size:500'],
```
juan-fernandez, line 33:
```suggestion
      ['Add a binary size to the current pipeline', 'datadog-ci metric --level pipeline --metrics binary.size:500'],
```

src/commands/metric/metric.ts
```diff
@@ -0,0 +1,143 @@
...
+      ['Tag the current CI job with a command runtime', 'datadog-ci metric --level job --tags command.runtime:67.1'],
```
juan-fernandez, line 34:
```suggestion
      ['Tag the current CI job with a command runtime', 'datadog-ci metric --level job --metrics command.runtime:67.1'],
```

juan-fernandez (approved):

AdrianLC:
> Regarding flags versus positional arguments. This is my personal opinion, but I generally find it easier to user and understand CLIs that work with flags rather than positional arguments are they are more self-documenting
> 
> e.g:
> 
> ```
> datadog-ci tag --level pipeline --tags a:b --tags c:d
> ```
> 
> vs
> 
> ```
> datadog-ci tag pipeline a:b c:d
> ```
> 
> Although longer I find the flag one easier to understand.

The level could still be optional though

AdrianLC (approved):

PR #533, opened 2022-05-09
[synthetics] Support multiple config overrides for same test

ruizb:
### What and why?

This PR introduces the possibility to trigger multiple tests that have the same public ID, but different config overrides.

### How?

* Introduce a `ResultOutcome` enum, in the `utils` module
* Add `failedNonBlocking` in the `Summary`
* Update the command's `renderResults` method
* Update the following methods of the default reporter: `testTrigger`,  `testEnd` and `runEnd`
  * `testTrigger` now includes extra info regarding the number of config overrides for each test run
  * `testEnd` is called for each result
  * `runEnd` contains more information in the summary
* Add Jest tests (on a second PR):
  * https://github.com/DataDog/datadog-ci/pull/534

---

When rendering the command results, we are now iterating over the list of sorted results, and not the over the tests anymore. As a consequence, we call `reporter.testEnd` function with a list of a single result:

```ts
reporter.testEnd(test, [pollResult], ...)
```

This hook should actually be called `resultEnd`, but I didn't want to rename it, or deprecate.`testEnd` in favor of `resultEnd`. Let me know if we should do it!

### Review checklist


src/commands/synthetics/command.ts
```diff
@@ -114,53 +113,66 @@ export class RunTestCommand extends Command {
     triggers: Trigger,
     startTime: number
   ) {
-    // Sort tests to show success first then non blocking failures and finally blocking failures.
-    tests.sort(this.sortTestsByOutcome(results))
+    const getTest = (publicId: string) => tests.find((t) => t.public_id === publicId) as Test
+
+    const sortedPollResults = Object.entries(results)
```
lefebvree, line 118: What do you think about having a new interface including `Test`, `PollResult`, `ExecutionRule`, and `ResultOutcome` that we would pass along everywhere, it might simplify things a bit overall and avoid recomputing things multiple times?

src/commands/synthetics/command.ts
```diff
@@ -114,53 +113,66 @@ export class RunTestCommand extends Command {
...
+        summary.failedNonBlocking++
       } else {
         summary.failed++
-        if (test.options.ci?.executionRule !== ExecutionRule.NON_BLOCKING) {
-          hasSucceeded = false
-        }
+        hasSucceeded = false
       }
 
       this.reporter?.testEnd(
```
lefebvree, line 173: So now `testEnd()` would be called for every result instead of once per test when all its results are received?
If so it mimics the `testResult()` (in latest `master`), [doc](https://github.com/DataDog/datadog-ci/blob/8659032811231c6f8918c633d5f703e659963f32/src/commands/synthetics/README.md#L263) might need to be updated. 
Also it feels a bit strange to always invoke reporters with an array of a single item, if necessary we can redefine the reporter interfaces (maybe to see with @sdeprez now what make the most sense to also work with batches later since they'll probably need to be updated anyway).

src/commands/synthetics/interfaces.ts
```diff
@@ -91,10 +92,16 @@ export interface MultiStepsTestResult extends TestResult {
...
+interface Enrichment {
+  batch_id: string
+  config_override: ConfigOverride & {executionRule: ExecutionRule}
+}
+
 export interface PollResult {
   check?: Test
   check_id?: string
   dc_id: number
+  enrichment?: Partial<Enrichment>
```
lefebvree, line 104: What are the cases where `enrichment` would not be defined? Could that cause incorrect `executionRule` evaluation?

src/commands/synthetics/reporters/default.ts
```diff
@@ -293,58 +291,57 @@ export class DefaultReporter implements MainReporter {
...
-      chalk.green(`${chalk.bold(summary.passed)} passed`),
-      chalk.red(`${chalk.bold(summary.failed)} failed`),
+    const {bold: b, gray, green, red, yellow} = chalk
+
+    const lines: string[] = []
+
+    const runSummary = [
+      green(`${b(summary.passed)} passed`),
+      red(`${b(summary.failed)} failed`),
+      yellow(`${b(summary.failedNonBlocking)} failed (non-blocking)`),
```
lefebvree, line 301: Maybe this can be hidden unless there are any failed non blocking result?

src/commands/synthetics/interfaces.ts
ruizb: As far as I can tell, it should always be defined. Though to minimise as much as possible the changes on this PR, I set it as optional. I can make it a required property, but with a bit of "noise" for the review (i.e. adding `enrichment: {},` in some places) 👀

src/commands/synthetics/utils.ts
```diff
@@ -205,12 +206,51 @@ export const hasResultPassed = (result: Result, failOnCriticalErrors: boolean, f
...
+const isTestNonBlocking = (test: Pick<Test, 'options'>): boolean =>
+  test.options.ci?.executionRule === ExecutionRule.NON_BLOCKING
+
+/**
+ * The test's execution rule takes precedence over the result's
+ * execution rule, when the test's rule is non-blocking.
+ */
+export const getResultExecutionRule = (test: Test, result: PollResult): ExecutionRule => {
+  if (isTestNonBlocking(test)) {
+    return ExecutionRule.NON_BLOCKING
```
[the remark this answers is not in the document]
ruizb: I think you are right, and we should remove `getResultExecutionRule` in favour of the existing `getExecutionRule` function, which has the order to describe there 👍 

(initially I thought that the test's execution rule took precedence over the config override's rule, but only if the test's rule was non-blocking. But I guess I got confused on this part 🤔)

src/commands/synthetics/command.ts
ruizb: The `testResult()` hook is called whenever a _finished_ result has been polled. Here, we need to wait for all the results to be polled, because we have to sort them by outcome (passed > failed non blocking > failed) before displaying them in the report.

The `testEnd()` name now feels weird indeed, but I didn't want to change the `MainReporter` interface. As you mentioned, I would be in favour of re-designing a better interface, most likely with breaking changes 🙂

src/commands/synthetics/command.ts
ruizb: I would love to have such interface, for example

```ts
interface ResultToRender {
  test: Test,
  result: PollResult,
  executionRule: ExecutionRule,
  resultOutcome: ResultOutcome
}
```

However, wouldn't we need to have a new method for the `MainReporter`? Something that takes a `ResultToRender` as one of its arguments... 🤔

lefebvree: Yes either that or we deconstruct it because reporter calls, maybe not call it `ResultToRender` to leave `rendering` to the actual reporter though

src/commands/synthetics/interfaces.ts
lefebvree: Good to me to leave as optional then 👍

src/commands/synthetics/command.ts
ruizb: So the `renderResults` method of the command would take a "result package" as a parameter instead, and this "result package" would be built earlier in the process, not here in this method, right?

Because if we build this package only here, and deconstruct it to pass the arguments to the reporter methods a bit later, I don't think it will be very useful ^^

lefebvree (approved):
LGTM, we can update interfaces in follow-up PRs for batches

PR #534, opened 2022-05-09
[synthetics] [tests] Support multiple config overrides for same test

ruizb:
### What and why?

This PR adds Jest tests for the one introducing the possibility to trigger multiple tests that have the same public ID, but different config overrides.

### How?

* Add tests for the new functions exported from `utils`
* Add tests for the `renderResults` method of the command
* Add tests for the `testTrigger`, `testEnd` and `runEnd` methods of the default reporter

### Review checklist


src/commands/synthetics/__tests__/fixtures.ts
```diff
@@ -275,3 +291,121 @@ export const getSyntheticsProxy = () => {
...
+  resultPassed: boolean
+}
+
+interface RenderResultsTestFixtures {
+  polledResultsByPublicId: Record<string, PollResult[]>
+  test: Test
+  triggerResult: TriggerResponse
+}
+
+export class RenderResultsHelper {
```
lefebvree, line 326: nit: Why use a class here, only `createFixtures()` seems to be called?  We can just use a global functions and a variable for `resultIdCounter` if there is no object state

src/commands/synthetics/__tests__/reporters/default.test.ts
```diff
@@ -32,4 +37,324 @@ describe('Default reporter', () => {
...
+          testsNotFound: new Set(['bbb-bbb-bbb']),
+        },
+      },
+    ]
+
+    test.each(cases)('$description', (testCase) => {
+      reporter.runEnd(testCase.summary)
+      expect(writeMock).toHaveBeenCalledWith(testCase.expectedOutput)
+    })
+  })
```
lefebvree, line 359: [Jest snapshots](https://jestjs.io/docs/snapshot-testing) might prove useful for those tests, we don't use them in the synthetics reporter test suite but I think they could simplify things if you want to take a look at it

lefebvree (approved):
LGTM, quite complex fixtures but a very nice coverage 👍

src/commands/synthetics/__tests__/fixtures.ts
ruizb: Initially that's what I did, but I thought it would be nicer to have some kind of "namespace" for this particular type of fixtures (hence the class as the namespace).

lefebvree (approved):

PR #535, opened 2022-05-11
[s8s] Fix critical error message display

Drarig29:
### What and why?

This PR improves the display of critical errors by adding the `error.message` into the `CriticalError`.

image

### How?

A new optional message property is added to the `CriticalError` constructor. It is optional because we don't always have an `error.message` to pass to the constructor, and for such errors we [directly log a message in the reporter](https://github.com/DataDog/datadog-ci/blob/67e99c912f525bbd1556fd8ca3d6411674e171a9/src/commands/synthetics/command.ts#L186-L214).

### Review checklist

### DIFF
--- a/src/commands/synthetics/__tests__/run-test.test.ts
+++ b/src/commands/synthetics/__tests__/run-test.test.ts
@@ -184,7 +184,7 @@ describe('run-test', () => {
         jest.spyOn(runTests, 'getApiHelper').mockImplementation(() => apiHelper as any)
         await expect(
           runTests.executeTests(mockReporter, {...ciConfig, testSearchQuery: 'a-search-query', tunnel: true})
-        ).rejects.toMatchError(new CriticalError(error))
+        ).rejects.toMatchError(new CriticalError(error, 'Server Error'))
       })
 
       test(`getTestsToTrigger throws - ${status}`, async () => {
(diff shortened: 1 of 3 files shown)


src/commands/synthetics/__tests__/run-test.test.ts
```diff
@@ -254,7 +254,12 @@ describe('run-test', () => {
       jest.spyOn(runTests, 'getApiHelper').mockImplementation(() => apiHelper as any)
       await expect(
         runTests.executeTests(mockReporter, {...ciConfig, publicIds: ['public-id-1', 'public-id-2'], tunnel: true})
-      ).rejects.toMatchError(new CriticalError('TRIGGER_TESTS_FAILED'))
+      ).rejects.toMatchError(
+        new CriticalError(
+          'TRIGGER_TESTS_FAILED',
+          '[] Failed to trigger tests: query on baseURLurl returned: "Bad Gateway"\n'
```
Drarig29, line 260: I need to fix this "baseURLurl". This issue comes from `formatBackendErrors()`, which directly concatenates the `baseURL` with the `url` from the Axios error. Axios [handles this internally](https://github.com/axios/axios/blob/1f13dd7e26124a27c373c83eff0a8614acc1a04f/lib/core/Axios.js#L123) and we may want to do the same as they do (cf. their [test file](https://github.com/axios/axios/blob/de5e006142cb428e712d00c05bda17c3c4c83ef1/test/specs/core/buildFullPath.spec.js)).

m-rousse:
Could you post a screenshot of what the error now looks like?

m-rousse (approved):
All good, many thanks!

PR #543, opened 2022-05-31
Code refactoring to prepare batch endpoint

sdeprez:
### What and why?

Preliminary PR to prepare the batch endpoint. A review commit by commit is advised.

### How?

This PR is split into 4 commits:
1. Stop typing `APIHelper` and infer from actual typings
2. Stop passing `failOnCriticalErrors` and `failOnTimeout` everywhere to get the "status" of a result. This is done by adding a property `.passed` at the root level of the result object, that takes into account those two parameters.
3. Simplify `waitForResults` output and handling of results afterwards:
    - return a flat list of results
    - clearly separate `PollResult` (what's returned by the API) from a new object `Result` which the object we built internally that contains all the fields necessary from the start. Also type `Result.result` as `ServerResult` to indicate the part of `Result` that is directly given by the server.
    - Note: the signature of `waitForResults` is  now even more complex, this could be simplified but this PR was already bigger than expected, so more cleaning will be done in another PR.
4. Fix the tests

More generally, this PR tries to introduce two principles:
- separate typings of the API which should only strictly type what's returned from the typings of our internal objects, with the format that we want
- build complete objects from the start, as much as possible, and avoids mutating objects along the way

### DIFF
--- a/src/commands/synthetics/__tests__/reporters/__snapshots__/default.test.ts.snap
+++ b/src/commands/synthetics/__tests__/reporters/__snapshots__/default.test.ts.snap
@@ -21,9 +21,9 @@ exports[`Default reporter runEnd Simple case with 1 test with 1 result (passed) 
 `;
 
 exports[`Default reporter testEnd 1 API test (blocking), 1 location, 3 results: success, failed non-blocking, failed 1`] = `
-"[1m[31m✖[39m[22m [[1m[31mblocking[39m[22m] [[1m[2maaa-aaa-aaa[22m[22m] [1mTest name[22m - [1m[31mlocation: [1mFrankfurt (AWS)[22m[1m[39m[22m
+"[1m[32m✓[39m[22m [[1m[2maaa-aaa-aaa[22m[22m] [1mTest name[22m - [1m[32mlocation: [1mFrankfurt (AWS)[22m[1m[39m[22m
   ⎋ Total duration: 123 ms - Result URL: [2m[36mhttps://app.datadoghq.com/synthetics/details/aaa-aaa-aaa?resultId=1&from_ci=true[39m[22m 
-  [1m[31m✖[39m[22m [1m[31m[1mGET[22m[1m - http://fake.url[39m[22m
+  [1m[32m✓[39m[22m [1m[32m[1mGET[22m[1m - http://fake.url[39m[22m
 
 [1m[33m✖[39m[22m [[1m[33mnon-blocking[39m[22m] [[1m[2maaa-aaa-aaa[22m[22m] [1mTest name[22m - [1m[33mlocation: [1mFrankfurt (AWS)[22m[1m[39m[22m
   ⎋ Total duration: 123 ms - Result URL: [2m[36mhttps://app.datadoghq.com/synthetics/details/aaa-aaa-aaa?resultId=2&from_ci=true[39m[22m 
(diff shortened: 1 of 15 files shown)


src/commands/synthetics/utils.ts
```diff
@@ -283,30 +291,40 @@ export const waitForResults = async (
...
 
-  while (triggerResults.filter((tr) => !tr.result).length) {
+  const getTest = (id: string): Test => tests.find((t) => t.public_id === id)!
+
+  while (inProgressTriggers.size) {
     const pollingDuration = new Date().getTime() - pollingStartDate
 
     // Remove test which exceeded their pollingTimeout
-    for (const triggerResult of triggerResults.filter((tr) => !tr.result)) {
+    for (const triggerResult of inProgressTriggers) {
```
lefebvree, line 300: Unrelated to this PR but just realised the order of events in this loop makes us wait 5s (`POLLING_INTERVAL`) for nothing, and probably mark some tests as timed out that are received just after 🤔 
```
+ compute pollingDuration
- fail timed out tests
- fail tunnel tests
- stop if max timeout reached  <- extra 5s unnecessary wait if we're not even fetching anything
- fetch results                <- might receive results of tests marked as timed out just before
- wait 5s
```
Should probably be:
```
- fetch results
+ compute pollingDuration
- fail timed out tests
- fail tunnel tests
- stop if max timeout reached
- wait 5s
```

src/commands/synthetics/utils.ts
```diff
@@ -261,18 +267,20 @@ export const waitForResults = async (
...
     failOnCriticalErrors?: boolean
     failOnTimeout?: boolean
   },
   reporter: MainReporter,
   tunnel?: Tunnel
-) => {
-  const triggerResultMap = createTriggerResultMap(triggerResponses, options.defaultTimeout, triggerConfigs)
-  const triggerResults = [...triggerResultMap.values()]
+): Promise<Result[]> => {
+  const inProgressTriggers = createInProgressTriggers(triggerResponses, options.defaultTimeout, triggerConfigs)
```
lefebvree, line 279: Having a depleting set for that is 👌

src/commands/synthetics/__tests__/reporters/__snapshots__/default.test.ts.snap
```diff
@@ -61,9 +61,9 @@ exports[`Default reporter testEnd 1 API test (non-blocking), 1 location, 3 resul
...
 
 exports[`Default reporter testEnd 1 API test, 1 location, 1 result: success 1`] = `
-"[1m[32m✓[39m[22m [[1m[2maaa-aaa-aaa[22m[22m] [1mTest name[22m - [1m[32mlocation: [1mFrankfurt (AWS)[22m[1m[39m[22m
+"[1m[31m✖[39m[22m [[1m[31mblocking[39m[22m] [[1m[2maaa-aaa-aaa[22m[22m] [1mTest name[22m - [1m[31mlocation: [1mFrankfurt (AWS)[22m[1m[39m[22m
   ⎋ Total duration: 123 ms - Result URL: [2m[36mhttps://app.datadoghq.com/synthetics/details/aaa-aaa-aaa?resultId=1&from_ci=true[39m[22m 
-  [1m[32m✓[39m[22m [1m[32m[1mGET[22m[1m - http://fake.url[39m[22m
+  [1m[31m✖[39m[22m [1m[31m[1mGET[22m[1m - http://fake.url[39m[22m
 
 "
 `;
```
lefebvree, line 69: I think we have some unwanted changes here where some tests are shown as failing because `getApiResult()` sets `passed: false` and it doesn't get updated?

lefebvree (approved):
LGTM, great simplifications and the interface renaming makes things way clearer! 
We should just double check the default reporter test snapshots

src/commands/synthetics/interfaces.ts
```diff
@@ -9,15 +9,8 @@ export interface MainReporter {
...
-    test: Test,
-    results: PollResult[],
-    baseUrl: string,
-    locationNames: LocationsMapping,
-    failOnCriticalErrors: boolean,
-    failOnTimeout: boolean
-  ): void
-  testResult(triggerResponse: TriggerResponse, result: PollResult): void
+  testEnd(test: Test, results: Result[], baseUrl: string, locationNames: LocationsMapping): void
+  testResult(triggerResponse: TriggerResponse, result: Result): void
```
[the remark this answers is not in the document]
sdeprez: Good idea, I'll do it in the next PR!

src/commands/synthetics/utils.ts
sdeprez: Good catch, I'll create a card for that.

src/commands/synthetics/__tests__/reporters/__snapshots__/default.test.ts.snap
sdeprez: Ah good catch, I'll fix that (those snapshots were not easy to read!)

sdeprez: Actually I have the feeling it was the test that was broken before 🤔 . Take this one for instance: https://github.com/DataDog/datadog-ci/blob/master/src/commands/synthetics/__tests__/reporters/default.test.ts#L155 it says 3 success but then it passes `passed: false` later on to `createApiPollResult`, so the red cross `x` is legit, and it was actually broken before? @ruizb any idea?

ruizb: It says 3 results, 1 success + 1 failed + 1 failed non-blocking, hence the `passed: false` for some of them 👀 

I think Etienne got it right 🤔

sdeprez: It was that 👍 https://github.com/DataDog/datadog-ci/pull/543/commits/a30d61c2aa0012c179637342660455b4261b77df#diff-f8505416f94fe5414df0ab9e792e8fd3a8ebd956950ac9574ce67b9a52ac9cdaR142

src/commands/synthetics/interfaces.ts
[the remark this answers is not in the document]
sdeprez: https://github.com/DataDog/datadog-ci/pull/545

PR #547, opened 2022-06-14
Add react native source maps upload command

louiszawadzki:
# Important note

This PR is ready to receive feedback and has been tested with the current staging API.
However, changes can still be made to this API so it is **not yet ready to be merged**.

### What and why?

Add react native source maps upload command.
I choose not to change the existing `sourcemaps` command because they don't share much logic in common. There are only a few functions that today are the same (mainly in `git.ts` and `validation.ts`), but these can still change independently.

### How?

- I duplicated the command from sourcemaps
- I changed the arguments
- Made the changes function by function for the upload
- Made the changes function by function for the renderer
- Removed unused code
- Added tests

### Review checklist


louiszawadzki:
Waiting for the API to be available to try out the command

src/commands/react-native/git.ts
```diff
@@ -0,0 +1,197 @@
...
+      if (list) {
+        list.push(f)
+      } else {
+        this.trackedFilenames.set(filename, new Array<string>(f))
+      }
+    }
+  }
+
+  // Looks up the sources declared in the sourcemap and return a list of related tracked files.
+  public matchRNSourcemap(stdout: Writable, srcmapPath: string): string[] | undefined {
```
louiszawadzki, line 134: This bit should be extracted (can differ for RN and browser)

src/commands/react-native/git.ts
```diff
@@ -0,0 +1,197 @@
+import fs from 'fs'
```
louiszawadzki, line 1: This file should be put in common between RN and browser

src/commands/react-native/interfaces.ts
```diff
@@ -0,0 +1,80 @@
...
+          options: {
+            contentType: 'application/json',
+            filename: 'repository',
+          },
+          value: this.gitData!.gitRepositoryPayload,
+        })
+      }
+      content.set('git_repository_url', {value: this.gitData!.gitRepositoryURL})
+      content.set('git_commit_sha', {value: this.gitData!.gitCommitSha})
+    }
```
louiszawadzki, line 53: This part could be put in common

src/commands/react-native/upload.ts
```diff
@@ -0,0 +1,319 @@
...
+
+    await this.addRepositoryDataToPayloads(payloads)
+
+    return payloads
+  }
+
+  // GetRepositoryPayload generates the repository payload for a specific sourcemap.
+  // It specifically looks for the list of tracked files that are associated to the source paths
+  // declared inside the sourcemap.
+  private getRepositoryPayload = (repositoryData: RepositoryData, sourcemapPath: string): string | undefined => {
```
louiszawadzki, line 208: This function can be put in common

src/commands/react-native/utils.ts
```diff
@@ -0,0 +1,17 @@
+export const getBaseIntakeUrl = () => {
```
louiszawadzki, line 1: this file can be put in common

src/commands/react-native/validation.ts
```diff
@@ -0,0 +1,51 @@
...
+export class InvalidPayload extends Error {
+  public reason: string
+
+  constructor(reason: string, message?: string) {
+    super(message)
+    this.reason = reason
+  }
+}
+
+const checkFile: (path: string) => {empty: boolean; exists: boolean} = (path: string) => {
```
louiszawadzki, line 13: this function can be put in common

src/commands/react-native/interfaces.ts
```diff
@@ -0,0 +1,80 @@
...
+    version: string,
+    projectPath: string,
+    platform: RNPlatform,
+    build: string
+  ): MultipartPayload {
+    const content = new Map<string, MultipartValue>([
+      ['cli_version', {value: cliVersion}],
+      ['service', {value: service}],
+      ['version', {value: version}],
+      ['build_number', {value: build}],
```
jybp, line 33: I don't think the backend is expecting any `build_number`, but `bundle_version` instead cc @rpelliard

src/commands/react-native/README.md
```diff
@@ -0,0 +1,120 @@
...
+By default the React Native SDK uses your app's bundle identifier as service.
+
+* `--bundle` (required) should be set to the path to your generated js bundle file, `main.jsbundle` for iOS or `index.android.bundle` for Android.
+
+* `--sourcemap` (required) should be set to the path to your generated sourcemap file, `main.jsbundle.map` for iOS or `index.android.bundle.map` for Android.
+
+* `--release-version` (required) will be used to match the `version` tag set on the RUM SDK.
+This should be the "Version" or "MARKETING_VERSION" in XCode for iOS and the "versionName" in your `android/app/build.gradle` for Android.
+
+* `--build-version` (required) is used to avoid overwriting your sourcemaps by accident.
```
jybp, line 48: Does this refers to `bundle_version` in the RFC?

src/commands/react-native/README.md
```diff
@@ -0,0 +1,120 @@
...
+- from the first remote otherwise
+
+The value can be overriden with `--repository-url`.
+
+Example: With a remote `git@github.com:Datadog/example.git`, links pointing to `https://github.com/Datadog/example` are generated.
+This behavior can be overriden with links to `https://gitlab.com/Datadog/example` with the flag `--repository-url=https://gitlab.com/Datadog/example`.
+
+#### Setting the project path
+
+TODO
```
jybp, line 84: The section from the sourcemaps command:
https://github.com/DataDog/datadog-ci/blob/louiszawadzki/add-react-native-source-maps-upload/src/commands/sourcemaps/README.md?plain=1#L72-L78

If I understand correctly, after unminification there could be a "project path" to remove in react native stack trace frames. So we could just reuse the same section wording. cc @rpelliard

louiszawadzki: Good question, I need to understand what this is doing. I could not make it work so far. I will discuss with @rpelliard next week.

src/commands/react-native/README.md
louiszawadzki: As mentioned in the previous comment, these 2 are different. @rpelliard can you update the RFC?

src/helpers/git-sourcemaps-repository-data.ts
```diff
@@ -2,7 +2,7 @@ import fs from 'fs'
 import * as simpleGit from 'simple-git'
 import {Writable} from 'stream'
 import {URL} from 'url'
-import {renderGitWarning, renderSourcesNotFoundWarning} from './renderer'
+import {renderGitWarning, renderSourcesNotFoundWarning} from '../commands/sourcemaps/renderer'
```
m-rousse, line 5: As much as possible, common helpers should include functions from commands, nor should they write to standard output as it is specific to each command.

Could you remove these rendering functions?
I expect `getRepositoryData` could just not catch the exception and have it handled by the calling function (within the command, which has access to rendering functions) and `matchSourcemap` could return an empty array and the array content should be handled in `upload.ts`.

src/helpers/git-sourcemaps-repository-data.ts
```diff
@@ -2,7 +2,7 @@ import fs from 'fs'
 import * as simpleGit from 'simple-git'
```
m-rousse, line 2: The `TrackedFilesMatcher` class does not seem tied to git particularly, could it be moved to a `sourcemap.ts` helpers file (to be shared amongst the two sourcemaps commands)?
That way this file could be merged with `git-span-metadata.ts` and be only `git.ts` with only git specific functions.

src/commands/react-native/upload.ts
```diff
@@ -0,0 +1,319 @@
...
+
+    if (!this.platform) {
+      this.context.stderr.write('Missing platform\n')
+
+      return 1
+    }
+
+    if (!RN_SUPPORTED_PLATFORMS.includes(this.platform)) {
+      this.context.stderr.write(
+        `Platform ${this.platform} is not supported.\nSupported platforms are ios and android.\n`
```
lraucy, line 99: Could we make this documentation future proof by using the list instead of manually encoding the possible values in the string?
Also even if that's outside of the scope of the PR I feel we could have a quick win by providing shared helpers functions to validate (but I don't want to super-charge this PR too much)
```typescript
try {
    validateRequiredField("--bundle-path", this.bundle)
    validateOptionField("--platform", this.platform, RN_SUPPORTED_PLATFORMS)
} catch (e) {
    this.context.stderr.write(e)

    return 1
}
```

src/commands/react-native/upload.ts
```diff
@@ -0,0 +1,319 @@
...
+      datadogSite: this.config.datadogSite,
+      metricsLogger: metricsLogger.logger,
+    })
+    const useGit = this.disableGit === undefined || !this.disableGit
+    const initialTime = Date.now()
+    const payloads = await this.getPayloadsToUpload(useGit)
+    const requestBuilder = this.getRequestBuilder()
+    const uploadMultipart = this.upload(requestBuilder, metricsLogger, apiKeyValidator)
+    try {
+      const results = await asyncPool(this.maxConcurrency, payloads, uploadMultipart)
```
lraucy, line 147: I don't see a need for the asyncPool in this command since there is only a single upload.

src/commands/react-native/upload.ts
```diff
@@ -0,0 +1,319 @@
...
+      payloads.map(async (payload) => {
+        const repositoryPayload = this.getRepositoryPayload(repositoryData!, payload.sourcemapPath)
+        payload.addRepositoryData({
+          gitCommitSha: repositoryData.hash,
+          gitRepositoryPayload: repositoryPayload,
+          gitRepositoryURL: repositoryData.remote,
+        })
+      })
+    )
+  }
```
lraucy, line 186: Same, here we can remove the promise and do it with a blocking call.

src/commands/react-native/upload.ts
```diff
@@ -0,0 +1,319 @@
...
+
+  // Looks for the sourcemaps and minified files on disk and returns
+  // the associated payloads.
+  private getMatchingRNSourcemapFiles = async (): Promise<RNSourcemap[]> => [
+    new RNSourcemap(this.bundle!, this.sourcemap!),
+  ]
```
lraucy, line 192: This function and the one below should return the type `RNSourcemap`, not `Promise<RNSourcemap[]>`

src/commands/react-native/upload.ts
```diff
@@ -0,0 +1,319 @@
...
+      ]),
+      overrideUrl: `v1/input/${this.config.apiKey}`,
+    })
+  }
+
+  private upload(
+    requestBuilder: RequestBuilder,
+    metricsLogger: MetricsLogger,
+    apiKeyValidator: ApiKeyValidator
+  ): (sourcemap: RNSourcemap) => Promise<UploadStatus> {
```
lraucy, line 254: Same, this function can probably be simplified.

src/commands/react-native/upload.ts
louiszawadzki: I agree it'd be a lot easier to understand what `execute` does like this.
In the meantime I would like to have this code merged soon to avoid conflicts if someone makes changes to the `sourcemaps` command or starts a global refactoring.

Is it ok for you if I open a new PR to do the change once this one is merged?

src/commands/react-native/upload.ts
louiszawadzki: I agree on this. However we are going to support multi-file upload later (in 1-2 months).
I believed it would be easier to keep this code until we need it, rather than temporarily changing it because we don't need such functionality - anyway it was much quicker for me to make that command work.

I don't have a very big opinion on this, I'm ok to make the change as well, I should be able to revert it quickly in a few weeks when we do multi-files upload.

src/helpers/git-sourcemaps-repository-data.ts
louiszawadzki: @lraucy what is your opinion on this?  
@m-rousse what would be the reason to merge both files?

I believe merging all the git-related stuff in one file is decreasing the [cohesion](https://en.wikipedia.org/wiki/Cohesion_(computer_science)) of the code, which is not a good thing: this file is going to be harder to maintain, if we couple both the span metadata and sourcemaps repository data it's likely to create bugs and since both are used by separate team it's going to take a lot of resources to solve this.
We should split files based on their responsibilities, not on the tools they used. It's not likely that we'd have to make a change to both the span metadata and sourcemaps data at the same time.

m-rousse: Goal is to have a simpler grouping of functions, having `functions related to git` and `sourcemap helpers` would make re-using git related functions in other commands easier than having `function to detect git context` (the "span metadata" one) and `functions to get git metadata formatted for sourcemap`.
The abstraction I suggest is data oriented and not command oriented to make it easier to navigate the code when implementing commands that will rely on these helpers.

src/commands/react-native/README.md
```diff
@@ -0,0 +1,120 @@
...
+You cannot upload new sourcemaps for a specific `build-version` and `service` combination.
+This should be the "Build" or "CURRENT_PROJECT_VERSION" in XCode for iOS and the "versionCode" in your `android/app/build.gradle` for Android.
+
+In addition, some optional parameters are available:
+
+* `--disable-git` (default: false): prevents the command from invoking git in the current working directory and sending repository related data to Datadog (hash, remote URL and the paths within the repository of the sources referenced in the sourcemap).
+* `--dry-run` (default: `false`): it will run the command without the final step of upload. All other checks are performed.
+* `--repository-url` (default: empty): overrides the repository remote with a custom URL. For example: https://github.com/my-company/my-project
+
+### Link errors with your source code
```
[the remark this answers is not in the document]
jybp: That's true but I'm not sure where it would live. Explaining under each command's readme what it's collecting on the local repository does make sense.

src/helpers/git-sourcemaps-repository-data.ts
louiszawadzki: As discussed, I've created a `git` folder with 3 files:
- `get-git-data` that contains git commands 
- `format-git-sourcemaps-data` and `format-git-span-data` that contains the logic to format the data for these cases

Upon reflexion, I am not sure an `index` file is needed, with auto-import it is still easy to import the files and the module is still small

src/commands/react-native/renderer.ts
```diff
@@ -0,0 +1,125 @@
...
+  return chalk.red(`${ICONS.FAILED} Failed upload sourcemap for ${sourcemapPathBold}: ${errorMessage}\n`)
+}
+
+export const renderRetriedUpload = (payload: RNSourcemap, errorMessage: string, attempt: number) => {
+  const sourcemapPathBold = `[${chalk.bold.dim(payload.sourcemapPath)}]`
+
+  return chalk.yellow(`[attempt ${attempt}] Retrying sourcemap upload ${sourcemapPathBold}: ${errorMessage}\n`)
+}
+
+export const renderSuccessfulCommand = (statuses: UploadStatus[], duration: number, dryRun: boolean) => {
```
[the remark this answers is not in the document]
louiszawadzki: I think it is the same case as with all the things related to multiple file uploads, I find it easier to reuse working code that will be fully used soon than changing it temporarily.
Is it ok for you?

src/commands/react-native/upload.ts
louiszawadzki: For the one below, it is actually async because we execute a git command in `this.addRepositoryDataToPayloads` -> `getRepositoryData` -> git commands

src/commands/react-native/upload.ts
louiszawadzki: @lraucy to me that's not a big issue as we probably won't need to edit this function. It is an async function since the last `return` returns a Promise. We could break it in multiple chunks,  but that would need a complete refactoring of the `execute` function to be done properly. 
I'd rather do that in a separate PR.

src/commands/react-native/upload.ts
lraucy: yes 👍

src/commands/react-native/README.md
```diff
@@ -0,0 +1,122 @@
...
+- You have installed the `git` executable
+- You can run `datadog-ci` in the git repository
+
+When these requirements are met, the upload command reports git information such as:
+
+- the current commit hash
+- the repository URL
+- the list of file paths that are tracked in the repository. Only tracked file paths related to a source map are gathered.
+
+<!-- Check this part -->
```
alai97, line 66: bump

src/commands/react-native/README.md
```diff
@@ -0,0 +1,122 @@
...
+
+You can override this behavior with links to `https://gitlab.com/Datadog/example` with the `--repository-url=https://gitlab.com/Datadog/example` flag.
+
+#### Set the project path
+
+By default, paths inside React Native source maps are the absolute paths of files on the machine where they were bundled (for example: `/Users/user/MyProject/App.ts`).
+
+If you are not running the `react-native upload` command from your React Native project root, you need to specify the `--project-path` argument with the absolute path to your React Native project root.
+
+<!-- TODO: test this once unminification is on prod -->
```
alai97, line 83: bump

alai97 (approved):
Looks good!

PR #557, opened 2022-06-30
[Synthetics] Update command output: Add batch URL, ignore skipped step

lefebvree:
### What and why?

Improve datadog-ci Synthetics command output.

### How?

- Don't display skipped steps following critical failed step in Browser test results (replace with `⇢ | X skipped steps`)
- Display batch URL above summary
- Stop displaying override count on skipped tests

### Review checklist

### DIFF
--- a/src/commands/synthetics/__tests__/reporters/default.test.ts
+++ b/src/commands/synthetics/__tests__/reporters/default.test.ts
@@ -3,7 +3,7 @@ import {BaseContext} from 'clipanion/lib/advanced'
 import {ConfigOverride, ExecutionRule, MainReporter, Result, Summary, Test} from '../../interfaces'
 import {DefaultReporter} from '../../reporters/default'
 import {createSummary} from '../../utils'
-import {getApiResult, getApiTest} from '../fixtures'
+import {getApiResult, getApiTest, getFailedBrowserResult} from '../fixtures'
 
 /**
  * A good amount of these tests rely on Jest snapshot assertions.
@@ -14,6 +14,7 @@ import {getApiResult, getApiTest} from '../fixtures'
  */
 
 describe('Default reporter', () => {
+  const baseUrlFixture = 'https://app.datadoghq.com/'
   const writeMock = jest.fn()
   const mockContext: unknown = {
     context: {
@@ -32,7 +33,7 @@ describe('Default reporter', () => {
       ['log', ['log']],
       ['reportStart', [{startTime: 0}]],
       ['resultEnd', [getApiResult('1', getApiTest()), '']],
-      ['runEnd', [createSummary()]],
+      ['runEnd', [createSummary(), '']],
       ['testTrigger', [{}, '', '', {}]],
       ['testsWait', [[{}]]],
     ]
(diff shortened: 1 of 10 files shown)


src/commands/synthetics/reporters/default.ts
```diff
@@ -139,9 +150,14 @@ const renderResultOutcome = (
...
   if (test.type === 'browser') {
+    // We render the step only if the test hasn't passed to avoid cluttering the output.
     if (!result.passed && 'stepDetails' in result) {
-      // We render the step only if the test hasn't passed to avoid cluttering the output.
-      return result.stepDetails.map(renderStep).join('\n')
+      const criticalFailedStepIndex = result.stepDetails.findIndex((s) => s.error && !s.allowFailure) + 1
+
+      return [
+        ...result.stepDetails.slice(0, criticalFailedStepIndex).map(renderStep),
+        renderSkippedSteps(result.stepDetails.slice(criticalFailedStepIndex)),
```
ruizb, line 159: Nit? This will render a break line character even if there are no skipped steps to render:

```ts
console.log(['foo', undefined].join('\n'))
// "foo\n"
```

src/commands/synthetics/reporters/default.ts
```diff
@@ -63,6 +63,17 @@ const renderStep = (step: Step) => {
   return `    ${icon} | ${duration} - ${step.description}${value}${error}`
 }
 
+const renderSkippedSteps = (steps: Step[]): string | undefined => {
+  if (!steps.length) {
+    return
+  }
+  if (steps.length === 1) {
+    return renderStep(steps[0])
```
ruizb, line 71: What is the purpose of rendering a full "skipped step" when there is (I think?) no information available anyway (no duration, no error...)? 🤔

src/commands/synthetics/__tests__/reporters/default.test.ts
```diff
@@ -160,6 +166,7 @@ describe('Default reporter', () => {
     const baseSummary: Summary = createSummary()
 
     const complexSummary: Summary = {
+      batchId: 'batch-id',
```
ruizb, line 169: We should set the `batchId` to the `baseSummary` instead:
```ts
const baseSummary = {...createSummary(), batchId: 'batch-id'}

const complexSummary: Summary = {
  ...baseSummary,
  criticalErrors: 2,
  ...
}
```
 Right now, the 2 other cases ("simple case" + "case where [...] empty") do not render the `Results URL: ` line 🤔

ruizb (commented):
Some minor comments/questions, LGTM otherwise 🐰

alai97 (approved):

src/commands/synthetics/reporters/default.ts
lefebvree: It contains the name of the step, and since it take as much space might as well show it I guess

src/commands/synthetics/__tests__/reporters/default.test.ts
lefebvree: That was done willingly to see if both cases are handled without issue

src/commands/synthetics/reporters/default.ts
lefebvree: Nice catch!

ruizb (approved):

PR #562, opened 2022-07-05
[ci-app] Add test runs URL on `datadog-ci upload` successful finish.

drodriguezhdez:
### What and why?

In the CI Visibility context, when the `datadog-ci upload` command finishes, there is no actionable for the user to jump into Datadog. This PR adds a link to the `test-runs` page filtering by `ci.pipeline.url` and `ci.job.url`.

### How?

### Review checklist


src/commands/junit/renderer.ts
```diff
@@ -27,8 +29,26 @@ export const renderRetriedUpload = (payload: Payload, errorMessage: string, atte
...
+  let fullStr = ''
+  fullStr += chalk.green(`${ICONS.SUCCESS} Uploaded ${fileCount} files in ${duration} seconds.\n`)
+
+  const testRunsUrl = getTestRunsUrl(spanTags)
+  if (testRunsUrl) {
+    fullStr += chalk.green(
+      '=================================================================================================\n'
+    )
+    fullStr += chalk.green(
+      '* View test runs and detailed time reports on Datadog (can take a few minutes to become available)\n'
```
juan-fernandez, line 42: what do you mean by detailed time reports?

src/commands/junit/utils.ts
```diff
@@ -1,7 +1,41 @@
...
+
+  return 'https://app.datadoghq.com'
+}
+
+export const getTestRunsUrl = (spanTags: SpanTags): string => {
+  if (!spanTags[CI_PIPELINE_URL] && !spanTags[CI_JOB_URL]) {
+    return ''
+  }
+
+  let query = '@test.status:fail'
```
juan-fernandez, line 32: why fixed to failed tests?

src/commands/junit/utils.ts
```diff
@@ -1,7 +1,41 @@
...
+  if (spanTags[CI_PIPELINE_URL]) {
+    query += ` @ci.pipeline.url:"${spanTags[CI_PIPELINE_URL]}"`
+  }
+  if (spanTags[CI_JOB_URL]) {
+    query += ` @ci.job.url:"${spanTags[CI_JOB_URL]}"`
+  }
+
+  return `${getBaseUrl()}/ci/test-runs?query=${encodeURIComponent(query)}&index=citest`
```
juan-fernandez, line 40: is the index necessary?

juan-fernandez (commented):
this LGTM and I think it's a good change, but I'd involve product. Could this be a bit noisy?

src/commands/junit/renderer.ts
drodriguezhdez: I copied what we're showing for `dogweb` jobs. Totally opened to change it if there is a better copy 👌🏼

src/commands/junit/utils.ts
drodriguezhdez: Based on my experience, the majority of the time the user wants to know which tests failed. But again, we can remove it if we want and show every test.

juan-fernandez: yeah the problem, I'd say, is that we'd redirect them to an empty list most of the times

src/commands/junit/renderer.ts
KickingTheTV: Could we say, "View the detailed test report for this commit on Datadog (this may take a few minutes to become available)"

And this is where the link takes them? To the Commit Overview page? Or does it take them somewhere else?

src/commands/junit/utils.ts
```diff
@@ -1,7 +1,59 @@
...
+  if (!spanTags[CI_PIPELINE_URL] && !spanTags[CI_JOB_URL]) {
+    return ''
+  }
+
+  let query = ''
+  if (spanTags[CI_PIPELINE_URL]) {
+    query += ` @ci.pipeline.url:"${spanTags[CI_PIPELINE_URL]}"`
+  }
+  if (spanTags[CI_JOB_URL]) {
+    query += ` @ci.job.url:"${spanTags[CI_JOB_URL]}"`
```
juan-fernandez, line 37: do we need both query params if job url is available?

juan-fernandez (approved):
lgtm

src/commands/junit/__tests__/upload.test.ts
```diff
@@ -139,22 +158,14 @@ describe('upload', () => {
       process.env.DD_ENV = 'ci'
       const context = createMockContext()
       const command = new UploadJUnitXMLCommand()
-      const [firstFile, secondFile] = await command['getMatchingJUnitXMLFiles'].call({
-        basePaths: ['./src/commands/junit/__tests__/fixtures'],
+      const spanTags: SpanTags = await command['getSpanTags'].call({
```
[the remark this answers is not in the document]
drodriguezhdez: The code only asserted the `spanTags` dict. As we have a dedicated method for this, the test keeps doing the same, but with the `getSpanTags` method.

src/commands/junit/utils.ts
drodriguezhdez: It's not required, but I added it to quickly filter results by pipeline via removing that. I'll ask for more opinions here. If none considers it useful, I can remove it.

juan-fernandez: got it. Yeah I'm fine keeping it

drodriguezhdez: changed!

PR #564, opened 2022-07-06
[synthetics] Extract renderResults function from command

ruizb:
### What and why?

In order to bump the datadog-ci version of https://github.com/DataDog/synthetics-ci-github-action, we have to extract the `renderResults` function out of the Command class, to make it available for the GH action.

### How?

Remove any reference to `this`, providing additional parameters to the `renderResults` function, and move it to the `utils` module.

### Review checklist


src/commands/synthetics/command.ts
```diff
@@ -87,59 +87,7 @@ export class RunTestCommand extends Command {
...
-        hasSucceeded = false
-      }
-
-      this.reporter?.resultEnd(result, this.getAppBaseURL())
-    }
-
-    this.reporter?.runEnd(summary, this.getAppBaseURL())
-
-    return hasSucceeded ? 0 : 1
+    return renderResults(results, summary, this.config, startTime, this.reporter)
```
jakepruitt, line 90: And now for an aesthetic nit: Could we switch up the order here ? Either go with `startTime` before `this.config` or `this.config` and `this.reporter` before the `results, summary, startTime` ?

I'd also be cool with this switching to an object with alphabetical ordering, since five parameters is a lot.

jakepruitt (approved):

src/commands/synthetics/command.ts
ruizb: I went with the same order as the GitHub action: https://github.com/DataDog/synthetics-ci-github-action/blob/447c7c3f611e23f78381a8ad15d4664c7df3ce78/src/process-results.ts#L3-L11

Although, since we'll have to update the GH action anyway, we might as well follow your suggestion here 🙂 I'll push a commit with the changes!

PR #569, opened 2022-07-06
[junit] Send the hostname

albertvaka:
### What and why?

Adds a new tag with the hostname that uploads the xml. This will be used if the XML itself doesn't contain an hostname field and the user doesn't override it manually by setting a `hostname` tag.

Also bumps `cireport_version`.

### How?

Reads `os.hostname()` in `upload.ts` and passes it in the payload that is sent in `api.ts`.

An alternative would be to call `os.hostname()` directly from `api.ts`, but I find it cleaner this way (happy to discuss though).

### Review checklist

### DIFF
--- a/src/commands/junit/__tests__/upload.test.ts
+++ b/src/commands/junit/__tests__/upload.test.ts
@@ -160,6 +160,19 @@ describe('upload', () => {
         key2: 'value2',
       })
     })
+    test('should set hostname', async () => {
+      const context = createMockContext()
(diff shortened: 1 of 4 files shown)


albertvaka:
Thanks for asking 😄 They were not, and they didn't expect versions > 2 😅

We will fix the backend to accept versions > 2, but for now I think we don't need to bump it here.

juan-fernandez (approved):

PR #560, opened 2022-07-04
Use intake v2 format for sourcemap, git-metadata and dsym uploads

rpelliard:
This migrates the `sourcemaps`, `dsyms` and `git-metadata` commands to the new format of the API endpoint used to upload files.

The new format must contain all JSON-encoded metadata as a single part of the request named `event`, and each attachment in a separate part, with an appropriate `filename`.

PR #574, opened 2022-07-19
Use intake v2 for react-native sourcemaps uploads

rpelliard:
This migrates the React Native sourcemaps upload command to the v2 API endpoint. (other commands have been migrated in https://github.com/DataDog/datadog-ci/pull/560).

### DIFF
--- a/src/commands/react-native/interfaces.ts
+++ b/src/commands/react-native/interfaces.ts
@@ -27,29 +27,18 @@ export class RNSourcemap {
     build: string
   ): MultipartPayload {
     const content = new Map<string, MultipartValue>([
-      ['cli_version', {value: cliVersion}],
-      ['service', {value: service}],
-      ['version', {value: version}],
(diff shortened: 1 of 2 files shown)


src/commands/react-native/interfaces.ts
```diff
@@ -67,6 +56,38 @@ export class RNSourcemap {
...
+      cli_version: cliVersion,
+      platform,
+      project_path: projectPath,
+      service,
+      type: 'react_native_sourcemap',
+      version,
+    }
+    if (this.gitData !== undefined) {
+      metadata.git_repository_url = this.gitData!.gitRepositoryURL
+      metadata.git_commit_sha = this.gitData!.gitCommitSha
```
louiszawadzki, line 80: nitpick: `!` don't do anything here since we already checked that `this.gitData` is not undefined, but they were there before so not a blocker for me.

PR #548, opened 2022-06-16
Add xcode command to upload sourcemaps automatically

louiszawadzki:
### What and why?

Add a script to be run as a XCode build phase to automatically upload sourcemaps on every build

### How?

This commands runs the RN bundle script, then uploads the sourcemaps with automated detection of parameters.

### Review checklist

- [x] Feature or bugfix MUST have appropriate tests (unit, integration)
- [x] \[_Synthetics_\] New releases of `datadog-ci` MUST be followed by a [synthetics-ci-github-action](https://github.com/DataDog/synthetics-ci-github-action) release

PR #580, opened 2022-07-22
Enable specifying API key and site from datadog-ci.json for dsyms command

louiszawadzki:
### What and why?

To make it easier to upload all kinds of debug files for cross platform applications, we want to unify the configuration of commands that can be used.

`react-native upload` command already gets to read the API key and site from a JSON file in https://github.com/DataDog/datadog-ci/pull/548

I did not document the behavior since this is mainly for automated cross-platform uploads, we don't want to add too many options to iOS developers.

### How?

I extracted the function to set the env variables.
I think the error handling can still diverge for `react-native` and `dsyms` since this is not documented for `dsyms`

### Review checklist


src/commands/dsyms/upload.ts
```diff
@@ -190,6 +201,17 @@ export class UploadCommand extends Command {
...
+  private resolveConfig = async () => {
+    try {
+      this.config = await parseConfigFile(this.config, this.configPath || DEFAULT_CONFIG_PATH)
+      setApiKeyAndSiteEnvVariablesFromConfig(this.config)
+    } catch (error) {
+      if (this.configPath) {
+        throw error
+      }
+    }
+  }
```
ncreated, line 213: This control flow looks quite complex: no matter of command parameters, we always try to parse either user config (`this.configPath`) or a default one (`DEFAULT_CONFIG_PATH`) and then only capture errors when user config is configured (otherwise, ignore errors).

Took me a while, but I understand what's happening here and the final behaviour seems correct. However, given that we do similar logic in callee `parseConfigFile`, it doesn't seem very straightforward:
```ts
export const parseConfigFile = async <T>(baseConfig: T, configPath?: string): Promise<T> => {
  try {
    const resolvedConfigPath = configPath ?? 'datadog-ci.json'
    const parsedConfig = await getConfig(resolvedConfigPath)

    return deepExtend(baseConfig, parsedConfig)
  } catch (e) {
    if (e.code === 'ENOENT' && configPath) {
      throw new Error('Config file not found')
    }

    if (e instanceof SyntaxError) {
      throw new Error('Config file is not correct JSON')
    }
  }

  return baseConfig
}
```

Is it JS/TS convention to use Errors to control the program flow and ignore them conditionally? Coming from static languages background, it looks very fragile to me.  I'd find it much more readable (from top to bottom) if we check in the upstream logic if config resolution is needed:
```ts
if (this.configPath) {
   // EVERY error thrown from here would indicate user error, unconditionally:
   resolveConfig(this.configPath)

} else if (pathExists(DEFAULT_CONFIG_PATH)) {
   // EVERY error thrown from here would indicate user error, unconditionally:
   resolveConfig(DEFAULT_CONFIG_PATH)

} // OTHERWISE, user clearly doesn't want to use config file
```

ncreated (approved):
Looks good 👍. I left one non-blocking question on error handling.

About documenting `configPath` for `upload dsyms` - I'd rather keep it aside. We didn't have such feature requests for `upload dsyms` and it's something we only add for RN needs because we share the implementation accidentally.

src/commands/dsyms/upload.ts
fuzzybinary: It might also be nice if more of this logic was pulled into `utils.ts`. The default config location won't change from command to command and the result should always be to call `setApiKeyAndSiteEnvVariablesFromConfig` anyway yes?

louiszawadzki: It totally makes sense to move this logic into `utils.ts`. I think we can improve the error flow indeed, but it is common to use errors to control the flow in functional programming, it is sometimes referred as [railway programming](https://fsharpforfunandprofit.com/rop/)

src/commands/dsyms/__tests__/upload.test.ts
```diff
@@ -318,4 +329,29 @@ describe('execute', () => {
...
+    expect(output[5]).toContain(
+      'Uploading C8469F85-B060-3085-B69D-E46C645560EA.zip (DDTest, arch: armv7, UUID: C8469F85-B060-3085-B69D-E46C645560EA)'
+    )
+    expect(output[6]).toContain(
+      'Uploading 06EE3D68-D605-3E92-B92D-2F48C02A505E.zip (DDTest, arch: arm64, UUID: 06EE3D68-D605-3E92-B92D-2F48C02A505E)'
+    )
+    expect(output[7]).toContain(
+      'Uploading 3BC12422-63CC-30E8-B916-E5006CE3286C.zip (DDTest, arch: arm64, UUID: 3BC12422-63CC-30E8-B916-E5006CE3286C)'
+    )
+    expect(output[10]).toContain('Handled 3 dSYMs with success')
```
m-rousse, line 355: You may want to use [Jest snapshots](https://jestjs.io/docs/snapshot-testing) to ease testing of the output of the command.
It is being used for the synthetics command, cf. https://github.com/DataDog/datadog-ci/blob/74a41e248d1c21985129eb07a0ad817ba035c874/src/commands/synthetics/__tests__/reporters/default.test.ts#L216

src/helpers/utils.ts
```diff
@@ -24,6 +24,19 @@ export const getConfig = async (configPath: string) => {
   return JSON.parse(configFile)
 }
 
+/**
+ * Setting the API Key and site as environment variables is mandatory to call the metrics
+ * logger. Call this function if you read their values from somewhere else (e.g. `datadog-ci.json`).
```
m-rousse, line 29: According to `datadog-metrics` documentation it is not mandatory to pass the API key through an environment variable, it is possible to set it in the `.init` function of datadog metrics.
cf. https://github.com/dbader/node-datadog-metrics#initialization

Reading the API key from a file, then setting it an environment variable to later re-read it from a dependency seems quite indirect, could we init `datadog-metrics` with the API key directly instead?

src/commands/dsyms/upload.ts
ncreated: Good idea to put extract most of it into utils (that would solve the problem of duplicating constant for default config path) 👍.

While I agree that conditioning Errors makes total sense in FP, here we use it in imperative flow with side effects, which doesn't make the concept easier.

src/commands/dsyms/__tests__/upload.test.ts
louiszawadzki: Good idea, but I am not sure it is worth doing it here for 3 tests.
As we have some random values and formatting in the output, it would require a bit of work to mock the randomness providers and make the snapshot readable enough so that it's easy to understand what broke in them.

I think snapshots are great for many things, they make it a lot easier to test big values and update them.
But that's also their drawback, that it's really easy to update, so there is less involvement in the process of updating them.
When you rewrite a line, you have more chances of thinking "Wait a minute, I should double check that this is what I want, doesn't seem right", whereas with snapshots you can enter a list of unrelated commands, but as you're not performing the change in itself you can more easily skip a bug.

So for a command that is not getting a lot of modifications, I am ok with these checks :)

m-rousse: Yes, all good for me too!

m-rousse (approved):
:+1:

maxep (approved):

maycmlee (commented):
Nothing for docs to review.

DarcyRaynerDD (approved):
Approving for Serverless

rpelliard (approved):

PR #581, opened 2022-07-22
Better back of for tags rate limit

cg14823:
### What and why?

Increase the backoff when we do retries to the custom tags endpoint as some customers that send a lot of them may be hitting the rate limit and the to fast retries is making it worse as it not gives enough time for the rate limit threshold to refresh.

### How?

Increasing the initial backoff time

### Review checklist


drodriguezhdez (approved):
LGTM

PR #586, opened 2022-07-25
[SW-2245] Add batchUrl to summary

Drarig29:
### What and why?

This PR adds the batch URL in the summary of a test run. Thanks to this, we don’t need to build the URL in all the integrations of datadog-ci. Otherwise, should our endpoint change, we’d need to update all of the integrations.

### How?

Instead of building the URL in the default reporter, we do it (with the same helper function), at the same time we set `summary.batchId`.

### Review checklist


src/commands/synthetics/interfaces.ts
```diff
@@ -328,6 +328,7 @@ export interface Suite {
 
 export interface Summary {
   batchId?: string
+  batchUrl?: string
```
lefebvree, line 331: Passing a base URL is definitely not ideal but I feel like we're overloading the summary a bit beyond its purpose. Since we also have the same issue with `resultEnd` how about exposing more straightforward utility functions like `getBatchUrl(config, batchId)` and `getResultUrl(config, result)` that can be use by library users

lefebvree (approved):

PR #595, opened 2022-08-05
Add test with network mocking

louiszawadzki:
### What and why?

This PR goal is to show an example of how we can test commands at network level.


louiszawadzki:
Closing this as this is too big of a change to be done right now.

PR #596, opened 2022-08-08
[sourcemaps] Require minified URL to have a host, or leading slash

rpelliard:
### What and why?

This improves the validation of the `--minified-path-prefix` parameter.
Today, passing something like `warn: unrelated log line\nhttps://example.com/statix/js` passes the validation, while it shouldn't.
This can be an issue if the output of a command (with unexpected logs in stdout) is passed to this parameter.

### How?

Instead of requiring the parsed URL to have a `protocol` (we have a test weirdly testing the opposite, which I suspect is accidentally passing because of the leading slash), require it to have a host.

`warn: unrelated log line\nhttps://example.com/statix/js` is being parsed without error by [`new URL`](https://nodejs.org/api/url.html#new-urlinput-base), but gives an almost empty URL with only a protocol.
Checking that the `host` is present should filter out this case, and should still validate correctly the two options we want to offer to the user, i.e. either:
- Pass a full URL with a host (and optional URL path) (and optional URL schema, which is ignored for unminification)
- Pass a path only (starting with a leading `/`)

### Review checklist

### DIFF
--- a/src/commands/sourcemaps/__tests__/upload.test.ts
+++ b/src/commands/sourcemaps/__tests__/upload.test.ts
@@ -37,7 +37,7 @@ describe('upload', () => {
   })
 
   describe('isMinifiedPathPrefixValid: full URL', () => {
-    test('should return false', () => {
+    test('should return true', () => {
       const command = new UploadCommand()
       command['minifiedPathPrefix'] = 'http://datadog.com/js'
 
@@ -46,7 +46,7 @@ describe('upload', () => {
   })
 
   describe('isMinifiedPathPrefixValid: URL without protocol', () => {
-    test('should return false', () => {
+    test('should return true', () => {
       const command = new UploadCommand()
       command['minifiedPathPrefix'] = '//datadog.com/js'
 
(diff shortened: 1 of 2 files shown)


src/commands/sourcemaps/__tests__/upload.test.ts
```diff
@@ -37,7 +37,7 @@ describe('upload', () => {
   })
 
   describe('isMinifiedPathPrefixValid: full URL', () => {
-    test('should return false', () => {
+    test('should return true', () => {
```
rpelliard, line 40: updated messages to be consistent with the check

src/commands/sourcemaps/__tests__/upload.test.ts
```diff
@@ -46,7 +46,7 @@ describe('upload', () => {
   })
 
   describe('isMinifiedPathPrefixValid: URL without protocol', () => {
-    test('should return false', () => {
+    test('should return true', () => {
```
rpelliard, line 49: I think this test was "accidentally" passing, because of the leading slash. It should now pass because we check the `host` instead of the `protocol`.

jybp (approved):

PR #602, opened 2022-08-22
[sourcemaps] Detect --minified-path-prefix misusage

jybp:
### What and why?

The PR adds a warning in the output when the `--minified-path-prefix` flag may have been misused.
We've had several tickets already where unminification wasn't working because customers uploaded sourcemaps with the wrong flag value which might not be easy to spot at first glance. The outputs were as such:

```
Uploading sourcemap build/static/js/abc.chunk.js.map for JS file available at https://my-website.com/static/js/static/js/abc.chunk.js
``` 

After this PR, launching the following command:
```
$ yarn launch sourcemaps upload ./src/commands/sourcemaps/__tests__ --service my-service --minified-path-prefix https://static.datadog.com/fixtures/basic --release-version 1.2
```

will now output the following warning:

```
⚠️ The --minified-path-prefix flag value "https://static.datadog.com/fixtures/basic" seems to repeat "fixtures/basic" which is already present in the path "/fixtures/basic/common.min.js"
```

### How?

We try to detect if the suffix of `--minified-path-prefix` is already present as a prefix in the path inferred from the first positional argument (directory in which sourcemaps are located).

### Review checklist

### DIFF
--- a/src/commands/sourcemaps/__tests__/upload.test.ts
+++ b/src/commands/sourcemaps/__tests__/upload.test.ts
@@ -101,7 +101,8 @@ describe('upload', () => {
         new Sourcemap(
           'src/commands/sourcemaps/__tests__/fixtures/sourcemap-with-no-files/empty.min.js',
           'http://example/empty.min.js',
-          'src/commands/sourcemaps/__tests__/fixtures/sourcemap-with-no-files/empty.min.js.map'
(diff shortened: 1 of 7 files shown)


rpelliard (approved):
Nice ! What are your thoughts about trying to automatically remediate this (i.e. dropping the duplicate part automatically) ?
I guess this would be a breaking change for anyone with a legitimate duplication in the path, so we would need an override mechanism - maybe it's best to keep it simple and just log a warning as implemented here ?

PR #604, opened 2022-08-23
[sourcemaps] Log service and version when starting upload

rpelliard:
Logs the `--service-name` and `--release-version` parameters in the sourcemaps upload command output.

Common misuse of the CLI is to pass an incorrect service or version; since those are often passed through environment variables, the command itself or the command output don't contain these parameters, which makes it hard to debug, both for users and for us, when we only have access to the command output.


rpelliard:
It's already logged.

PR #605, opened 2022-08-24
Update container image instructions

fermayo:
### What and why?

Changes the instructions on how to build a run the CLI using a container. The current instructions do not take into account the fact that the CLI needs `git` for some commands and access to the local filesystem.

### How?

- New `Dockerfile` that is `alpine` based (to reduce size) and that installs `git` (required by several commands) and the CLI using `npm` (as the static binary is still experimental). It also installs the latest version of the CLI.
- Adds mounting the current directory into the container to have access to git metadata (required by several commands).

### Review checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,28 +59,26 @@ datadog-ci version
 
 ### Container image (**experimental**)
 
-In order to run `datadog-ci` from a container, you can use the following `Dockerfile:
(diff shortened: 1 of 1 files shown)


juan-fernandez (approved):

buraizu (approved):

m-rousse (approved):
This LGTM, thanks @fermayo !

PR #609, opened 2022-08-25
synthetics: simplify getTestsToTrigger

KevinYeramian:
### What and why?

- Rename `getTest` to `getTestWithPublicId`
- Extract code in `getTest` and `getAndOverrideTest` to simplify `getTestsToTrigger`
- Rename handleConfig to getOverriddenConfig

### DIFF
--- a/src/commands/synthetics/utils.ts
+++ b/src/commands/synthetics/utils.ts
@@ -55,17 +55,17 @@ export const getOverriddenConfig = (
   config?: ConfigOverride
 ): TestPayload => {
   const executionRule = getExecutionRule(test, config)
-  let handledConfig: TestPayload = {
+  let overriddenConfig: TestPayload = {
     executionRule,
     public_id: publicId,
   }
 
   if (!config || !Object.keys(config).length) {
-    return handledConfig
+    return overriddenConfig
   }
 
-  handledConfig = {
-    ...handledConfig,
+  overriddenConfig = {
+    ...overriddenConfig,
     ...pick(config, [
       'allowInsecureCertificates',
       'basicAuth',
@@ -90,10 +90,10 @@ export const getOverriddenConfig = (
     if (URL_VARIABLES.some((v) => config.startUrl?.includes(v))) {
       reporter.error('[DEPRECATION] The usage of URL variables is deprecated, see explanation in the README\n\n')
     }
-    handledConfig.startUrl = template(config.startUrl, context)
+    overriddenConfig.startUrl = template(config.startUrl, context)
   }
 
-  return handledConfig
+  return overriddenConfig
 }
 
 export const setCiTriggerApp = (source: string): void => {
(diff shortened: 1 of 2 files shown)


src/commands/synthetics/utils.ts
```diff
@@ -293,7 +293,7 @@ export const waitForResults = async (
     return hasTunnel ? 'Tunneled' : locationNames[dcId] || dcId
   }
 
-  const getTest = (id: string): Test => tests.find((t) => t.public_id === id)!
+  const getTestWithPublicId = (id: string): Test => tests.find((t) => t.public_id === id)!
```
ruizb, line 296: Nit: rename parameter `id` to `publicId`

src/commands/synthetics/utils.ts
```diff
@@ -48,7 +48,7 @@ const template = (st: string, context: any): string =>
 
 export let ciTriggerApp = process.env.DATADOG_SYNTHETICS_CI_TRIGGER_APP || 'npm_package'
 
-export const handleConfig = (
+export const getOverriddenConfig = (
```
ruizb, line 51: Nit: rename `handledConfig` to `overriddenConfig` inside this function

src/commands/synthetics/utils.ts
```diff
@@ -455,16 +455,62 @@ export const getReporter = (reporters: Reporter[]): MainReporter => ({
...
+  }
+}
+
+const getTestAndOverrideConfig = async (
+  api: APIHelper,
+  {config, id, suite}: TriggerConfig,
+  reporter: MainReporter,
+  summary: Summary
+) => {
+  id = PUBLIC_ID_REGEX.test(id) ? id : id.substr(id.lastIndexOf('/') + 1)
```
ruizb, line 483: I'd be in favor of creating a new variable, instead of mutating the one extracted from the second parameter 🤔  Something such as:

```ts
// or `parsedId`, or `sanitizedId`, or `formattedId`...
const normalizedId = PUBLIC_ID_REGEX.test(id) ? id : id.substr(id.lastIndexOf('/') + 1)
```

src/commands/synthetics/utils.ts
```diff
@@ -455,16 +455,62 @@ export const getReporter = (reporters: Reporter[]): MainReporter => ({
...
+
+  const getTestResult = await getTest(api, {config, id, suite})
```
ruizb, line 485: Kind of nit? I know you meant as "result of getTest" but the name could suggest that `getTestResult` is a function that we can call. How about `testResult`? And use `const {test} = testResult` a few lines below 👀

src/commands/synthetics/utils.ts
```diff
@@ -475,40 +521,26 @@ export const getTestsToTrigger = async (
...
-      }
+  const overriddenTestsToTrigger: TestPayload[] = []
+  const waitedTests: Test[] = []
+  testsAndConfigsOverride.forEach(({test, errorMessage, overriddenConfig}) => {
+    if (errorMessage) {
+      errorMessages.push(errorMessage)
+    }
 
-      const overriddenConfig = handleConfig(test, id, reporter, config)
+    if (overriddenConfig) {
```
ruizb, line 536: Is there a reason why we check if `overriddenConfig` is truthy? (whereas before we didn't care)

ruizb (approved):
Mostly nits and 1-2 questions, LGTM 🐰

src/commands/synthetics/utils.ts
KevinYeramian: Previously everything was done in the same function, and thee push was done only if we successfully get the config https://github.com/DataDog/datadog-ci/pull/609/files#diff-741a985551714cec0c6f7a6423d704525619e8f57db6c4f7c685d95d88a8d0cfL499. Now, now we return the overwritten config instead of directly adding it to the array.

PR #610, opened 2022-08-26
Fix React Native path in xcode command

louiszawadzki:
### What and why?

Using `${indexFilePath}/..` would return ENOTDIR when trying to spawn the bundle script, so I removed last part of the path instead.
I could have also used `.replace('/index.js', '')` but I think this solution is more robust if the name of the "main" file changes.

### How?

Require.resolve cannot be mocked by jest, and I don't think it is worth adding `react-native` as a dependency to this project.
I could not find a way to unit testing this in a way that prevents potential bugs, so I added a comment instead.

### Review checklist

### DIFF
--- a/src/commands/react-native/xcode.ts
+++ b/src/commands/react-native/xcode.ts
@@ -2,13 +2,20 @@
 import {spawn} from 'child_process'
 import {Cli, Command} from 'clipanion'
 import {existsSync} from 'fs'
+import {sep} from 'path'
 import {UploadCommand} from './upload'
 
+/**
+ * Because jest cannot mock require.resolve, reactNativePath cannot
+ * be unit tested. If you make any change to it, make sure to test
+ * it with a real project.
+ */
 const reactNativePath = (() => {
   try {
     const reactNativeIndexFile = require.resolve('react-native')
 
-    return `${reactNativeIndexFile}/..`
+    // We need to remove the trailing "/index.js" at the end of the path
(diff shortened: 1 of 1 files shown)


src/commands/react-native/xcode.ts
```diff
@@ -4,11 +4,17 @@ import {Cli, Command} from 'clipanion'
...
+ * be unit tested. If you make any change to it, make sure to test
+ * it with a real project.
+ */
 const reactNativePath = (() => {
   try {
     const reactNativeIndexFile = require.resolve('react-native')
 
-    return `${reactNativeIndexFile}/..`
+    // We need to remove the trailing "/index.js" at the end of the path
+    return reactNativeIndexFile.split('/').slice(0, -1).join('/')
```
0xnm, line 17: will it work on Windows where delimiter is `\`?

louiszawadzki: This script will only be ran from Xcode, which is not available on windows

0xnm: makes sense. Ideally it is better to use something like `Os.Separator` (pseudo-symbol).

0xnm (approved):

src/commands/react-native/xcode.ts
louiszawadzki: I changed it to `path.sep` so that it can be more easily reused on other scripts if needed 👍

PR #612, opened 2022-08-29
Synthetics: Extract code to simplify waitForResults

KevinYeramian:
### What and why?
Why: readability.

- Extract code in `getBatch`, `getTestByPublicId` , `getPollResultMap` and `waitForBatchToFinish`

### DIFF
--- a/src/commands/synthetics/utils.ts
+++ b/src/commands/synthetics/utils.ts
@@ -337,6 +337,7 @@ const getResultFromBatch = (
   }
 
   const test = getTestByPublicId(resultInBatch.test_public_id, tests)
+
   return {
     executionRule: resultInBatch.execution_rule,
     location: getLocation(resultInBatch.location, test),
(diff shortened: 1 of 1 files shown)


src/commands/synthetics/utils.ts
```diff
@@ -293,50 +362,8 @@ export const waitForResults = async (
...
-
-  const pollResultMap: {[key: string]: PollResult} = {}
-  try {
-    const pollResults = await api.pollResults(batch.results.map((r) => r.result_id))
-    pollResults.forEach((r) => (pollResultMap[r.resultID] = r))
-  } catch (e) {
-    throw new EndpointError(`Failed to poll results: ${formatBackendErrors(e)}\n`, e.response?.status)
-  }
-
   for (const resultInBatch of batch.results) {
```
lefebvree, line 367: I think we could also extract this

lefebvree (approved):
LGTM, nice changes!
I think we can get some of that out of `utils.ts` in new `batch.ts`+`result.ts` files, wdyt? (feel free to ignore for this PR)

lefebvree (approved):

PR #616, opened 2022-08-31
Add flutter-symbol upload command

fuzzybinary:
### What and why?

This adds a command - `flutter-symbols upload` - that locates symbol  and mapping files for Flutter projects and uploads them with the proper information to Datadog intake.

This command handles upload iOS dSYMS, Android Proguard Mapping Files, and Dart Symbol Files, depending on the parameters specified. The command looks for the various files in their default locations post a Flutter build.

Note that Dart Symbol Files do not have a default location so that parameter is required.

### How?

The command farms out to the `dsyms upload` command for the iOS portion, but implements uploaders for Android and Flutter. The Android implementation duplicates the functionality of the Datadog gradle plugin, but I think it's worth it to keep uploading simple for our users.

One thing of note is how this is tested -- I created an upload helper method that I can easily mock, that ensures that I pass the proper Multipart data to each upload request.  I also created a helper method for performing a sub-command, which is mocked to test calling the dSYM upload.

### Review checklist


src/commands/flutter-symbols/__tests__/fixtures/pubspecs/missingVersionPubspec.yaml
```diff
@@ -0,0 +1,20 @@
+name: valid_pubspec_example
+description: A pubspec that's valid
```
louiszawadzki, line 2: Is this pubspec still valid even if the version is missing?

src/commands/flutter-symbols/__tests__/fixtures/pubspecs/invalidPubspec.yaml
```diff
@@ -0,0 +1,22 @@
+name: invalid_pubspec_example
+description: A pubspec that's just slightly off for testing
```
louiszawadzki, line 2: Maybe we could specify what's off with this pubspec so that it makes it easier to understand what's supposed to be detected

src/commands/flutter-symbols/renderer.ts
```diff
@@ -0,0 +1,127 @@
...
+  chalk.yellow(`${ICONS.WARNING} An error occured while invoking git: ${errorMessage}
+Make sure the command is running within your git repository to fully leverage Datadog's git integration.
+To ignore this warning use the --disable-git flag.\n`)
+
+export const renderArgumentMissingError = (argumentName: string) =>
+  chalk.red(`${ICONS.FAILED} Error: parameter "${argumentName}" is required.\n`)
+
+export const renderMissingPubspecError = (pubspecLocation: string) =>
+  chalk.red(
+    `${ICONS.FAILED} Could not find pubspec at '${pubspecLocation}'. A pubspec.yaml is required or the --version argument must be sepcified.\n`
```
louiszawadzki, line 88:
```suggestion
    `${ICONS.FAILED} Could not find pubspec at '${pubspecLocation}'. A pubspec.yaml is required or the --version argument must be specified.\n`
```

src/commands/flutter-symbols/upload.ts
```diff
@@ -0,0 +1,444 @@
...
+  VALUE_NAME_DART_MAPPING,
+  VALUE_NAME_JVM_MAPPING,
+} from './interfaces'
+
+export class UploadCommand extends Command {
+  public static usage = Command.Usage({
+    description: '',
+    details: `
+        `,
+    examples: [],
```
louiszawadzki, line 52: Is there a reason to keep these empty for now? I believe they help for understanding the command.

src/commands/flutter-symbols/upload.ts
```diff
@@ -0,0 +1,444 @@
...
+      git_repository_url: this.gitData?.remote,
+      platform,
+      service: this.serviceName,
+      type,
+      variant: this.flavor,
+      version: this.version!,
+    }
+  }
+
+  private async parsePubspec(pubspecLocation: string): Promise<number> {
```
louiszawadzki, line 231: Minor comment: this function's goal is to read the version from the pubspec and set the `version` of the command, this could be reflected in its name, like `parseVersionFromPubspec`. If you'd like to keep this one's name more generic, maybe this can also be achieved in splitting it into a part that parses the pubspec, and one part that reads and sets the version.

src/commands/flutter-symbols/__tests__/upload.test.ts
```diff
@@ -0,0 +1,582 @@
...
+      ...(gitRespositoryUrl && {git_repository_url: gitRespositoryUrl}),
+      platform,
+      service: 'fake.service',
+      type: 'flutter_symbol_file',
+      variant: 'release',
+      version: '1.0.0',
+    })
+
+    test('uploads correct multipart payloads without repository', async () => {
+      ;(uploadMultipartHelper as jest.Mock).mockResolvedValue('')
```
louiszawadzki, line 471: nitpick: I think this `;` can be removed

src/commands/flutter-symbols/upload.ts
fuzzybinary: 👍 Absolutely right, I'll add those in.

src/commands/flutter-symbols/__tests__/fixtures/pubspecs/missingVersionPubspec.yaml
fuzzybinary: Flutter doesn't complain about it, though to be honest I have no idea what it does in that case? Likely it just doesn't modify the versions set in platform specific files.

src/commands/flutter-symbols/upload.ts
```diff
@@ -0,0 +1,472 @@
+// Disabling no-unused-variable / no-empty temporarily so certain private methods
+// can remain in while this feature is in progress
+// tslint:disable: no-unused-variable
```
louiszawadzki, line 3: Can we remove `tslint:disable: no-unused-variable` now? If not maybe edit the comment

fuzzybinary: Yes they can! Good catch!

louiszawadzki (approved):

src/commands/flutter-symbols/README.md
```diff
@@ -0,0 +1,53 @@
+## Overview
+
+To deobfuscate and symbolicate errors and crashes, upload your Flutter Symbols, iOS dSYMs, and Android mapping Files to Datadog
```
alai97, line 3:
```suggestion
To deobfuscate and symbolicate errors and crashes, upload your Flutter Symbols, iOS dSYMs, and Android mapping files to Datadog.
```

src/commands/flutter-symbols/README.md
```diff
@@ -0,0 +1,53 @@
...
+## Setup
+
+You need to have `DATADOG_API_KEY` in your environment.
+
+```bash
+# Environment setup
+export DATADOG_API_KEY="<API KEY>"
+```
+
+You can configure the tool to use Datadog EU by defining the `DATADOG_SITE` environment variable as `datadoghq.eu`. By default, the requests are sent to Datadog US.
```
alai97, line 14:
```suggestion
In order to configure the tool to use Datadog EU, define the `DATADOG_SITE` environment variable as `datadoghq.eu`. By default, requests are sent to Datadog US.
```

src/commands/flutter-symbols/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+After running `flutter build --split-debug-info=./debug-info --obfuscate`, you can upload all files to Datadog by running this command: 
+
+```bash
+datadog-ci flutter-symbols upload --dart-symbols-location ./debug-info --service-name com.companyname.application --ios-dyms --android-mapping
+```
+
+| Parameter | Condition | Description |
+|-----------|-----------|-------------|
+| `--service-name` | Required | Set as the service name you are uploading files for. Datadog uses this service name to find corresponding files based on the `service` options set during configuration of the the RUM Flutter SDK.<br>By default, the RUM Flutter SDK uses your application's bundle identifier as the service. |
```
alai97, line 41:
```suggestion
| `--service-name` | Required | Set as the service name you are uploading files for. Datadog uses this service name to find corresponding files based on the `service` options set when you configured the Datadog Flutter SDK for RUM.<br>By default, the Datadog Flutter SDK for RUM uses your application's bundle identifier as the service. |
```

src/commands/flutter-symbols/README.md
```diff
@@ -0,0 +1,53 @@
...
+| `--flavor` | Optional | The build flavor that was built. Defaults to `release` |
```
alai97, line 42:
```suggestion
| `--flavor` | Optional | The build flavor that was built. Defaults to `release`. |
```

src/commands/flutter-symbols/README.md
```diff
@@ -0,0 +1,53 @@
...
+| `--dart-symbols-location`  | Optional  | The locaiton of your Dart symbol files. This should be the same path specified to `--split-debug-info`. |
+| `--ios-dysms` | Optional | Upload iOS dSYM files to Datadog |
```
alai97, line 44:
```suggestion
| `--ios-dysms` | Optional | Upload iOS dSYM files to Datadog. |
```

src/commands/flutter-symbols/README.md
```diff
@@ -0,0 +1,53 @@
...
+| `--ios-dsyms-location` | Optional | Specify the location of iOS dSYM files. By default, these are located at `./build/ios/archive/Runner.xcarchive/dSYMs`. Adding this parameter automatically sets `--ios-dsyms`. |
+| `--android-mapping` | Optional | Upload Android Proguard mapping file. This is usually only necessary if you specify `--obfuscate` as a parameter to `flutter build` |
```
alai97, line 46:
```suggestion
| `--android-mapping` | Optional | Upload Android Proguard mapping file. This is only necessary if you specify `--obfuscate` as a parameter to `flutter build`. |
```

src/commands/flutter-symbols/README.md
```diff
@@ -0,0 +1,53 @@
...
+| `--android-mapping-location` | Optional | Specify the location of your Android Proguard mapping file. By default, this is located at `./build/app/outputs/mapping/[flavor]/mapping.txt`. Adding this parameter automatically sets `--android-mapping` |
```
alai97, line 47:
```suggestion
| `--android-mapping-location` | Optional | Specify the location of your Android Proguard mapping file. By default, this is located at `./build/app/outputs/mapping/[flavor]/mapping.txt`. Adding this parameter automatically sets `--android-mapping`. |
```

src/commands/flutter-symbols/README.md
```diff
@@ -0,0 +1,53 @@
...
+| `--pubspec-location` | Optional | The location of the pubspec for this application. The pubspec is used to automatically determine the version number of the application. Defaults to the current directory. |
+| `--version` | Optional | The version of your application. This defaults to the version specified in your pubspec. Note that only one upload is needed for a specific `version` and `service` combination. Subsequent uploads are ignored until the `version` changes. |
```
alai97, line 49:
```suggestion
| `--version` | Optional | The version of your application. This defaults to the version specified in your pubspec. Only one upload is needed for a specific `version` and `service` combination. Subsequent uploads are ignored until the `version` changes. |
```

alai97 (approved):
Looks good, some copy suggestions!

src/commands/flutter-symbols/upload.ts
```diff
@@ -0,0 +1,468 @@
...
+import {getArchInfoFromFilename, getFlutterRequestBuilder, uploadMultipartHelper} from './helpers'
+import {
+  DART_SYMBOL_FILE_NAME,
+  JVM_MAPPING_FILE_NAME,
+  MappingMetadata,
+  TYPE_DART_SYMBOLS,
+  TYPE_JVM_MAPPING,
+  VALUE_NAME_DART_MAPPING,
+  VALUE_NAME_JVM_MAPPING,
+} from './interfaces'
```
m-rousse, line 42: Convention in this repo is to group imports by:
- first party imports (only `fs` in that case)
- third party imports (`clipanion`, `js-yaml`, `glob`, `tiny-async-pool`)
- project imports with pathes with the most `../` first

Unfortunately `tslint` does not support checking for that now, I've created a card to move `datadog-ci` to `eslint` which will also support this rule.

src/commands/flutter-symbols/upload.ts
```diff
@@ -0,0 +1,468 @@
...
+
+export class UploadCommand extends Command {
+  public static usage = Command.Usage({
+    description: 'Upload symbol files for Flutter.',
+    details: `
+            This command will upload all symbol files for Flutter applications in order to symbolicate errors and
+            crash reports recieved by Datadog. This includes uploading iOS dSYMs, Proguard mapping files, and Dart
```
m-rousse, line 49:
```suggestion
            crash reports received by Datadog. This includes uploading iOS dSYMs, Proguard mapping files, and Dart
```

src/commands/flutter-symbols/upload.ts
```diff
@@ -0,0 +1,468 @@
...
+
+    const metricsLogger = getMetricsLogger({
+      apiKey: this.config.apiKey,
+      datadogSite: this.config.datadogSite,
+      defaultTags: [
+        `version:${this.version}`,
+        `service:${this.serviceName}`,
+        `cli_version:${this.cliVersion}`,
+        'flutter:true',
+        'platform:android',
```
m-rousse, line 133: I am not to fully understand the different usage cases of the command, but I was under the impression that it may also target iOS.
Is this tag always valid?

src/commands/flutter-symbols/upload.ts
```diff
@@ -0,0 +1,468 @@
...
+  private getFlutterSymbolFiles(dartSymbolLocation: string): string[] {
+    const symbolPaths = glob.sync(buildPath(dartSymbolLocation, '*.symbols'))
+
+    return symbolPaths
+  }
+
+  private getGitDataPayload(): MultipartValue | undefined {
+    if (this.gitData === undefined) {
+      return undefined
+    }
```
m-rousse, line 192: This check is already performed before each call of `getGitDataPayload`.
You could pass `this.gitData` as an argument, avoiding to check for its definition here and prevent the function from possibly returning `undefined` (which in turn allows to remove the `!` when [setting the payload](https://github.com/DataDog/datadog-ci/pull/616/files#diff-3459254cea55c76190454bf3617bfea34521699313c555379c3519d3c3328732R303)).

More generally: we try to avoid using `!` as much as possible as it means we are forcing the type checking of typescript to believe something that "should" be obvious for typescript when parsing the code.

src/commands/flutter-symbols/upload.ts
```diff
@@ -0,0 +1,468 @@
...
+
+      return 1
+    }
+
+    try {
+      const doc = yaml.load(fs.readFileSync(pubspecLocation, 'utf8')) as any
+      if (doc.version) {
+        this.version = doc.version
+
+        const buildIndex = this.version!.indexOf('+')
```
m-rousse, line 252: The parsing logic of semantic versions may have edge cases that are not obvious to cover.

We may want to add the `semver` package to parse the version in a more robust fashion? (it is already present in the node_modules as it is imported by `dd-trace` at least, so it should not increase the size of the project)

src/commands/flutter-symbols/upload.ts
```diff
@@ -0,0 +1,468 @@
...
+    const exitCode = await performSubCommand(dsyms.UploadCommand, dsymUploadCommand, this.context)
+    if (exitCode && exitCode !== 0) {
+      return UploadStatus.Failure
+    }
+
+    return UploadStatus.Success
+  }
+
+  private async verifyParameters(): Promise<boolean> {
+    if (!this.serviceName) {
```
m-rousse, line 410: To improve the UX it would be nice to display all errors at once (instead of requiring the user to re-run the command after having fixed one issue at a time).

src/commands/flutter-symbols/helpers.ts
```diff
@@ -0,0 +1,45 @@
+import * as path from 'path'
+import {upload} from '../..//helpers/upload'
+import {UploadOptions} from '../..//helpers/upload'
+import {getRequestBuilder} from '../..//helpers/utils'
+import {getBaseSourcemapIntakeUrl} from '../../helpers/base-intake-url'
+import {RequestBuilder} from '../../helpers/interfaces'
+import {MultipartPayload} from '../../helpers/upload'
```
m-rousse, line 7:
```suggestion
import {getBaseSourcemapIntakeUrl} from '../../helpers/base-intake-url'
import {RequestBuilder} from '../../helpers/interfaces'
import {MultipartPayload , UploadOptions, upload} from '../../helpers/upload'
import {getRequestBuilder} from '../../helpers/utils'
```

m-rousse:
Great PR! As this new command calls the `dsyms` one for iOS symbols, should the `dsyms` command be deprecated so there is only one way to upload DSYMs?

fuzzybinary:
> Great PR! 

Thanks 🤗

> As this new command also supports iOS symbols, should the `dsyms` command be deprecated so there is only one way to upload DSYMs?

No, we should keep it so there's only one way to upload dSYMs, if that makes sense. 

The idea is that iOS users will use the dSYM command and not have to worry about any Flutter stuff, but Flutter users have to upload the same files with different required parameters, we just combined them for simplicity on the Flutter side, and to make sure the same codepaths are used in both cases.

m-rousse:
All good then, thanks for the explanation :+1:

src/commands/flutter-symbols/README.md
fuzzybinary: Yeah, I was wondering about this. @louiszawadzki we may want to make this change in the React Native readme as well.

src/commands/flutter-symbols/README.md
fuzzybinary: I think there may be situations where this would be necessary even without the `--obfuscate paraemter`, so I'm going to leave the "usually" in.

src/commands/flutter-symbols/upload.ts
fuzzybinary: 👍 I'll move around imports.

src/commands/flutter-symbols/upload.ts
fuzzybinary: That's a good catch.

The dSYMs sub command should create its own metrics logger for iOS, but even so, this tag probably still isn't 100% correct.

I'll actually probably replace `flutter:true` with `platform:flutter`, and add `platform:android` only during the Mapping upload.

src/commands/flutter-symbols/upload.ts
fuzzybinary: Yeah, that's a good call, and should be fairly easy to implement.

fuzzybinary:
@m-rousse I think I addressed everything. Let me know if you see anything I missed!

m-rousse (approved):
This looks good to me, thanks @fuzzybinary :+1:

src/helpers/git/format-git-sourcemaps-data.ts
```diff
@@ -123,6 +123,16 @@ export class TrackedFilesMatcher {
...
   }
 
+  // Return a list of all tracked files
+  public rawTrackedFilesList() {
+    let rawList: string[] = []
+    this.trackedFilenames.forEach((value) => {
+      rawList = rawList.concat(value)
+    })
+
+    return rawList
```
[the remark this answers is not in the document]
fuzzybinary: @m-rousse I had to revert this because `flat` is not available in Node 10 (which is what CI is running)

rpelliard (approved):

PR #617, opened 2022-09-01
Synthetics: Add `uploadMobileApplication` endpoint

KevinYeramian:
### What and why?

- Add `uploadMobileApplication` endpoint
- Update mobile pre-signed test

### DIFF
--- a/src/commands/synthetics/__tests__/api.test.ts
+++ b/src/commands/synthetics/__tests__/api.test.ts
@@ -144,8 +144,8 @@ describe('dd-api', () => {
     const spy = jest.spyOn(axios, 'create').mockImplementation((() => () => ({data: PRESIGNED_URL_PAYLOAD})) as any)
     const api = apiConstructor(apiConfiguration)
     const {getMobileApplicationPresignedURL} = api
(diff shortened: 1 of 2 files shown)


src/commands/synthetics/__tests__/api.test.ts
```diff
@@ -154,6 +158,15 @@ describe('dd-api', () => {
     spy.mockRestore()
   })
 
+  test('should upload a mobile application with a presigned URL', async () => {
+    const spy = jest.spyOn(axios, 'create').mockImplementation((() => () => ({data: PRESIGNED_URL_PAYLOAD})) as any)
+    const api = apiConstructor(apiConfiguration)
+    const {uploadMobileApplication} = api
+    await uploadMobileApplication(Buffer.from(''), MOBILE_PRESIGNED_URL_PAYLOAD.presigned_url_params)
+
```
lefebvree, line 166: Should add a spy check for the empty buffer passer to axios, and we can remove the `({data: PRESIGNED_URL_PAYLOAD})) as any)` return is the response body is not used

lefebvree (approved):

PR #621, opened 2022-09-02
Synthetics-mobile: Upload new applications

KevinYeramian:
### What and why?

- Upload new applications `executeTests`
- Add `requestUnstable` for mobile endpoints
- Upload mobile endpoints


src/commands/synthetics/utils.ts
```diff
@@ -778,8 +778,8 @@ export const renderResults = ({
   return hasSucceeded ? 0 : 1
 }
 
-export const getDatadogHost = (useIntake = false, config: SyntheticsCIConfig) => {
-  const apiPath = 'api/v1'
+export const getDatadogHost = (useIntake = false, isUnstableApi = false, config: SyntheticsCIConfig) => {
+  const apiPath = isUnstableApi ? 'api/unstable' : 'api/v1'
```
nmizoguchi, line 782: Should we use `version = 'v1' | 'unstable'` here instead?

src/commands/synthetics/utils.ts
```diff
@@ -778,8 +778,8 @@ export const renderResults = ({
   return hasSucceeded ? 0 : 1
 }
 
-export const getDatadogHost = (useIntake = false, config: SyntheticsCIConfig) => {
-  const apiPath = 'api/v1'
+export const getDatadogHost = (useIntake = false, isUnstableApi = false, config: SyntheticsCIConfig) => {
```
m-rousse, line 781: Arguably I think it would be easier to reason about if this function was taking an object so we have named fields. (`{useIntake: boolean, apiVersion: 'v1' | 'unstable', config: SyntheticsCIConfig}`)
If it has too deep implications, feel free to create a card in SW board

nmizoguchi (approved):

src/commands/synthetics/api.ts
```diff
@@ -161,13 +160,22 @@ const getMobileApplicationPresignedURL = (
...
+  })
+  form.append('file', fileBuffer)
+
   await retryRequest(
     {
-      data: fileBuffer,
+      data: form,
+      headers: {...form.getHeaders(), 'Content-Length': form.getLengthSync()},
+      maxBodyLength: Infinity,
+      maxContentLength: Infinity,
```
m-rousse, line 176: Is there a limit in the backend? (the backend is a cloud provider's object storage? eg. s3)

m-rousse (approved):
Looks good, feel free to merge if the upload backend is indeed a cloud provider's cloud storage solution.

src/commands/synthetics/api.ts
KevinYeramian: As discussed IRL, this option is internal to Axios and doesn't impact the request.

src/commands/synthetics/mobile.ts
```diff
@@ -83,24 +83,28 @@ export const uploadApplicationAndOverrideConfig = async (
   overriddenTestsToTrigger: TestPayload,
   uploadedApplicationByPath: {[applicationFilePath: string]: {applicationId: string; fileName: string}[]}
 ): Promise<void> => {
-  if (!userConfigOverride.mobileApplicationVersionFilePath) {
+  if (!(userConfigOverride.mobileApplicationVersionFilePath || userConfigOverride.mobileApplicationVersion)) {
```
nmizoguchi, line 86: very small nit, but I would avoid doing the `not(A || B)`, and do `not A && not B` instead, as it feels easier to  understand!

src/commands/synthetics/mobile.ts
```diff
@@ -83,24 +83,28 @@ export const uploadApplicationAndOverrideConfig = async (
...
   }
 
   const testApplicationId = test.options.mobileApplication!.applicationId
-  await uploadApplicationIfNeeded(
-    api,
-    userConfigOverride.mobileApplicationVersionFilePath,
-    testApplicationId,
-    uploadedApplicationByPath
-  )
+  if (userConfigOverride.mobileApplicationVersionFilePath) {
```
nmizoguchi, line 91: Since the method called below is `uploadApplicationIfNeeded()`, I would say that the check if should upload or not should be done inside it! Can you move this check inside the method?

Maybe make [uploadedApplicationByPath](https://github.com/DataDog/datadog-ci/blob/5a36bce2391a544367e1b22a11834ce66680475c/src/commands/synthetics/mobile.ts#L34) as optional, and check if it's not set.

nmizoguchi (approved):
A few suggestions in the code, but logic LGTM overall

src/commands/synthetics/mobile.ts
KevinYeramian: IMO, this check shouldn't be done in `uploadApplicationIfNeeded`. `uploadApplicationIfNeeded` uploads the application if it is not cached. We can enhance the naming with `uploadApplicationIfNotCached` ?

src/commands/synthetics/mobile.ts
KevinYeramian: Actually this condition is not needed, I will remove it.

src/commands/synthetics/mobile.ts
nmizoguchi: Humm IMO `ifNeeded` gives the idea of checking the edge cases, and only doing the upload "if needed".

IMHO we can just say then `uploadApplication()`, and skip if already cached. No need to keep it in the method name!

Another option which would be great for unit testing is having a `shouldUploadApplication()` which checks for all these constraints (if there is an override, if the file is already cached, etc), and let `uploadApplication()` only perform the upload 👍

PR #630, opened 2022-09-13
[git-metadata|lambda] Normalize repository URL

jybp:
Add a new `normalizeRemote` function to prevent colons from ending up in `git.repository_url` tags and being improperly parsed by tracers. 

-----

`normalizeRemote` is added to:
-  the `lambda` command before tagging
- the exported `uploadGitCommitHash` in library.ts used by `serverless-plugin-datadog`

### DIFF
--- a/src/commands/git-metadata/git.ts
+++ b/src/commands/git-metadata/git.ts
@@ -29,12 +29,12 @@ export const gitRemote = async (git: simpleGit.SimpleGit): Promise<string> => {
   for (const remote of remotes) {
     // We're trying to pick the remote called with the default git name 'origin'.
     if (remote.name === 'origin') {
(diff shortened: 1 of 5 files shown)


src/commands/git-metadata/git.ts
```diff
@@ -29,12 +29,12 @@ export const gitRemote = async (git: simpleGit.SimpleGit): Promise<string> => {
   for (const remote of remotes) {
     // We're trying to pick the remote called with the default git name 'origin'.
     if (remote.name === 'origin') {
-      return stripCredentials(remote.refs.push)
+      return remote.refs.push
```
jybp, line 32: Moved in `getCommitInfo` directly to also apply it when an optional repositoryURL is passed as arg

PR #633, opened 2022-09-15
[lambda] Support AWS named profiles

duncanista:
### What and why?

The `aws-cli` allows users to have multiple credentials in a file, commonly separated by named profiles. Those credentials are stored by default in the path `~/.aws/credentials`. AWS also allows users to have those credentials store in a specific path, which the user can provide by setting the environment variable `AWS_SHARED_CREDENTIALS_FILE`. These credentials are mainly used to use AWS services through the CLI.

The main goal of this PR is to allow the user to provide those credentials by using named profiles in the same manner AWS handles it. `-p,--profile` is a feature which takes a `string` and grabs the credentials from the expected file path. 

### How?

- Adding a the `-p,--profile` option in the `lambda` commands.
- Handling the credentials using the `aws-sdk` given a profile.

=> **This feature could be considered as a breaking change if used in CI with credentials that have MFA enabled** <=

### Review checklist


kayayarai (approved):

IvanTopolcic (approved):
LGTM, nice work 👍

PR #640, opened 2022-09-23
[Synthetics] Add tests result wait spinner to default reporter

lefebvree:
### What and why?

Add a fancy spinner to the `synthetics` command when waiting for test results.

image

### How?

Use [ora](https://github.com/sindresorhus/ora) added from https://github.com/DataDog/datadog-ci/pull/638 to display `Waiting for...` text.

### Review checklist

### DIFF
--- a/src/commands/synthetics/reporters/default.ts
+++ b/src/commands/synthetics/reporters/default.ts
@@ -247,12 +247,12 @@ const getResultIconAndColor = (resultOutcome: ResultOutcome): [string, chalk.Cha
 }
 
 export class DefaultReporter implements MainReporter {
+  private context: BaseContext
   private testWaitSpinner?: ora.Ora
-  private writable: Writable
   private write: Writable['write']
 
   constructor({context}: {context: BaseContext}) {
-    this.writable = context.stdout
+    this.context = context
     this.write = context.stdout.write.bind(context.stdout)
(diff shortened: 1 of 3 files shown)


src/commands/synthetics/reporters/default.ts
```diff
@@ -255,9 +257,12 @@ const getResultIconAndColor = (resultOutcome: ResultOutcome): [string, chalk.Cha
 }
 
 export class DefaultReporter implements MainReporter {
+  private testWaitSpinner?: ora.Ora
+  private writable: Writable
   private write: Writable['write']
 
   constructor({context}: {context: BaseContext}) {
+    this.writable = context.stdout
```
ruizb, line 265: Nit? Should we replace `this.write(...)` with `this.writable.write(...)` to avoid the potential confusion between `write` and `writable` in this module, and simplify slightly by removing a property? 🤔

ruizb (approved):
Nit comment, LGTM 🐰

src/commands/synthetics/reporters/default.ts
lefebvree: Since `write` is widely used I think it's nice to have it as a property directly. But agreed the two of them is a bit confusing, we could store the context instead

PR #642, opened 2022-09-26
[synthetics] `result.device` should be optional

Drarig29:
### What and why?

This PR makes the `device` property optional in a browser test's `result` because it will be undefined if a test times out when waited for.

Displaying this `device` property is also refactored into a `formatDeviceId()` utility function.

### Review checklist

### DIFF
--- a/src/commands/synthetics/reporters/junit.ts
+++ b/src/commands/synthetics/reporters/junit.ts
@@ -372,9 +372,9 @@ export class JUnitReporter implements Reporter {
           {$: {name: 'check_id', value: result.test.public_id}},
           ...(isDeviceIdSet(result.result)
             ? [
-                {$: {name: 'device', value: result.result.device?.id}},
-                {$: {name: 'width', value: result.result.device?.width}},
-                {$: {name: 'height', value: result.result.device?.height}},
+                {$: {name: 'device', value: result.result.device.id}},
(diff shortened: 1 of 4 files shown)


src/commands/synthetics/reporters/junit.ts
```diff
@@ -370,11 +370,11 @@ export class JUnitReporter implements Reporter {
...
           {$: {name: 'check_id', value: result.test.public_id}},
-          ...('device' in result.result
+          ...(isDeviceIdSet(result.result)
             ? [
-                {$: {name: 'device', value: result.result.device.id}},
-                {$: {name: 'width', value: result.result.device.width}},
-                {$: {name: 'height', value: result.result.device.height}},
+                {$: {name: 'device', value: result.result.device?.id}},
+                {$: {name: 'width', value: result.result.device?.width}},
+                {$: {name: 'height', value: result.result.device?.height}},
```
lefebvree, line 377: I believe we don't need the optional chaining  here since `result.result` should be considered a `Required<BrowserServerResult>`?

lefebvree (approved):

src/commands/synthetics/reporters/junit.ts
Drarig29: Exactly! That's a git conflict mistake... I purposely used type guards to remove those optional chaining 😅
Thanks for spotting it!

PR #643, opened 2022-09-26
[synthetics] `Summary.batchId` should be required

Drarig29:
### What and why?

In our interface, the `batchId` property of a run summary was optional because it's not initially set.
But seen from the outside, the `batchId` will always be defined.

### How?

Internally we use an `InitialSummary` but we expose a `Summary`.

### Review checklist


lefebvree (approved):
LGTM!

PR #647, opened 2022-09-26
[synthetics] Make jUnit reports compatible with GitLab

Drarig29:
### What and why?

This PR makes the jUnit reports generated for Synthetic tests compatible with the test report tab in the results of a GitLab pipeline.

image

### How?

- Use the 3 possible status icons:
  - ⏩ for skipped tests
  - ❗️for failed non-blocking tests
  - ❌ for failed blocking tests

- Decouple test suites' stats and test cases' step stats:
  - Test suites' stats now **only include**: `errors`, `failures`, `skipped` and `tests`. The `allowfailures` and `warnings` stats were removed because they were too specific to the steps inside of a test case.
  - Test cases' stats are now **all prefixed with `steps_`: `steps_allowfailures`, `steps_count`, `steps_errors`, `steps_failures`, `steps_skipped`, `steps_warnings`. It's clearer what those stats are about **and** `skipped` was conflicting with the `<skipped>reason</skipped>` of a skipped test.

### Useful links

- Official XML schema: https://github.com/windyroad/JUnit-Schema/blob/master/JUnit.xsd
- Simplified XML schema: https://github.com/behave/behave/blob/main/etc/junit.xml/junit-4.xsd
- GitLab's JUnit report parser: https://gitlab.com/gitlab-org/gitlab/-/blob/master/lib/gitlab/ci/parsers/test/junit.rb

### Review checklist

### DIFF
--- a/src/commands/synthetics/__tests__/reporters/junit.test.ts
+++ b/src/commands/synthetics/__tests__/reporters/junit.test.ts
@@ -146,11 +146,17 @@ describe('Junit reporter', () => {
     })
 
     it('should use the same report for tests from same suite', () => {
-      const result = {...globalResultMock, test: {...globalTestMock, suite: 'same suite'}}
-      reporter.resultEnd(result, '')
-      reporter.resultEnd(result, '')
+      const results = [
+        {...globalResultMock, test: {...globalTestMock, suite: 'same suite'}},
+        {...globalResultMock, test: {...globalTestMock, suite: 'same suite'}},
+      ]
 
-      expect(reporter['json'].testsuites.testsuite.length).not.toBe(2)
+      results.forEach((result) => reporter.resultEnd(result, ''))
+
+      // We should have 1 unique report. Not 2 different ones.
+      expect(reporter['json'].testsuites.testsuite.length).toBe(1)
+
+      // And this unique report should include the 2 tests.
       expect(reporter['json'].testsuites.testsuite[0].$).toMatchObject({
         ...getDefaultSuiteStats(),
         tests: 2,
@@ -516,7 +522,7 @@ describe('GitLab test report compatibility', () => {
     expect(testCase.failure).toStrictEqual([
       {
         $: {step: 'Test name', type: 'INCORRECT_ASSERTION'},
-        _: '- Assertion(s) failed:\n    ▶ responseTime should be less than 1000. Actual: 1234',
+        _: '- Assertion failed:\n    ▶ responseTime should be less than 1000. Actual: 1234',
       },
     ])
   })
--- a/src/commands/synthetics/__tests__/reporters/junit.test.ts
+++ b/src/commands/synthetics/__tests__/reporters/junit.test.ts
@@ -4,15 +4,20 @@ import {Writable} from 'stream'
 
 import {BaseContext} from 'clipanion/lib/advanced'
 
-import {Result} from '../../interfaces'
+import {Device, ExecutionRule, Result, Test} from '../../interfaces'
 
 import {RunTestCommand} from '../../command'
-import {Args, getDefaultStats, JUnitReporter, XMLTestCase} from '../../reporters/junit'
+import {Args, getDefaultSuiteStats, getDefaultTestCaseStats, JUnitReporter, XMLTestCase} from '../../reporters/junit'
 import {
   BATCH_ID,
+  getApiResult,
+  getApiServerResult,
   getApiTest,
   getBrowserResult,
   getBrowserServerResult,
+  getBrowserTest,
+  getFailedBrowserResult,
+  getFailedMultiStepsServerResult,
   getMultiStep,
   getMultiStepsServerResult,
   getStep,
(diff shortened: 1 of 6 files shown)


src/commands/synthetics/__tests__/reporters/junit.test.ts
```diff
@@ -253,14 +322,202 @@ describe('Junit reporter', () => {
...
+    expect(testCase.$).toHaveProperty('file', 'tests.json') // Filename
+    expect(testCase.failure).toStrictEqual([failure]) // Status
+    expect(testCase.$).toHaveProperty('time', 20) // Duration
+  })
+
+  it('the icon in the Status column is correct (blocking vs. non-blocking)', () => {
+    // Mimics how GitLab chooses the Status icon.
+    const getStatusIcon = (testCase: XMLTestCase) => {
+      if (testCase.failure.length > 0) {
+        return '❌'
```
ruizb, line 467: We could use plain text for these ones (e.g. `'error'` and `'skipped'`) but emojis add some colors to these tests 🌈

src/commands/synthetics/__tests__/reporters/junit.test.ts
```diff
@@ -253,14 +322,202 @@ describe('Junit reporter', () => {
...
+    runName: 'Test Suite name',
+  }
+
+  let reporter: JUnitReporter
+
+  beforeEach(() => {
+    reporter = new JUnitReporter(commandMock as RunTestCommand)
+  })
+
+  it('all test case names are unique', () => {
```
ruizb, line 351: Nit: we could use `test` instead of `it` (`it` is generally used for "it should ..." sentences, though it doesn't really matter in the end)

ruizb (commented):
I've got a few more comments for you 😁

src/commands/synthetics/reporters/junit.ts
```diff
@@ -318,37 +405,70 @@ export class JUnitReporter implements Reporter {
...
+          {$: {name: 'type', value: test.type}},
+        ].filter((prop) => prop.$.value),
+      },
+      skipped: [],
+      warning: [],
+    }
+  }
+
+  // Update the current stats of a suite based on the processed test case.
+  private getSuiteStats(testCase: XMLTestCase, current: SuiteStats = getDefaultSuiteStats()): SuiteStats {
```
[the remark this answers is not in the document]
Drarig29: True, I renamed it at some point, but it means we would also rename `getTestCaseStats()` to `updateTestCaseStats()`, which we only need without any mutation here:

https://github.com/DataDog/datadog-ci/blob/52087bc7f23aaa744b1aa84c07de49be72124a6f/src/commands/synthetics/reporters/junit.ts#L495

src/commands/synthetics/reporters/junit.ts
```diff
@@ -392,8 +523,42 @@ export class JUnitReporter implements Reporter {
...
         ].filter((prop) => prop.$.value),
       },
-      testcase: [],
+      skipped: [],
       warning: [],
     }
   }
+
+  // Update the current stats of a test case based on the processed result.
+  private getTestCaseStats(result: Result, current: TestCaseStats = getDefaultTestCaseStats()): TestCaseStats {
```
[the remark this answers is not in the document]
Drarig29: I do agree with this 😉

src/commands/synthetics/__tests__/reporters/junit.test.ts
Drarig29: Yes, I like it! It's more visual than "error"/"failure" and it really shows what GitLab would display in the UI.

src/commands/synthetics/reporters/junit.ts
```diff
@@ -146,50 +198,62 @@ export class JUnitReporter implements Reporter {
...
     }
 
-    // Update stats for the suite.
-    suiteRun.$ = {
-      ...suiteRun.$,
-      ...this.getResultStats(result, getStats(suiteRun.$)),
-    }
+    const testCase = this.getTestCase(result, baseUrl)
+
+    // Errors and failures cannot co-exist: GitLab will always choose failures over errors.
```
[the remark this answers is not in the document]
Drarig29: "Errors and failures" here is not to confuse with what we call "errors and failures" in Synthetics generally. Here, it's just the difference between blocking and non-blocking: a test cannot be both at the same time.

src/commands/synthetics/__tests__/reporters/junit.test.ts
ruizb: I know it's a _nit_ but you reacted with a thumb-up so I guess you forgot to update these `it` to `test`? 😛

src/commands/synthetics/reporters/junit.ts
```diff
@@ -392,8 +499,43 @@ export class JUnitReporter implements Reporter {
...
+      // It's a browser test.
+      stepsStats = result.result.stepDetails
+        .map((step) => {
+          if (!step.subTestStepDetails) {
+            return [step]
+          }
+
+          return [step, ...step.subTestStepDetails]
+        })
+        .reduce((acc, val) => acc.concat(val), [])
```
ruizb, line 519: I was about to suggest using `flatMap` (or `map(...).flatten()`) as it's been available since Node 11, but I just saw we are still on Node 10.21 on this repository 😞

src/commands/synthetics/reporters/junit.ts
```diff
@@ -392,8 +499,43 @@ export class JUnitReporter implements Reporter {
...
+    for (const stepStats of stepsStats) {
+      stats.steps_count += stepStats.steps_count
+      stats.steps_errors += stepStats.steps_errors
+      stats.steps_failures += stepStats.steps_failures
+      stats.steps_skipped += stepStats.steps_skipped
+      stats.steps_allowfailures += stepStats.steps_allowfailures
+      stats.steps_warnings += stepStats.steps_warnings
+    }
+
+    return stats
```
ruizb, line 539: Not a requested change, but an example using `reduce` to avoid using a for loop and some mutation (just for the sake of sharing this kind of knowledge):

```ts
return stepsStats.reduce(
  (stats, stepStats) => ({
    ...stats,
    steps_count: stats.steps_count + stepStats.steps_count,
    // etc.
  }), getDefaultTestCaseStats())
```

ruizb (approved):
LGTM 🐰

src/commands/synthetics/__tests__/reporters/junit.test.ts
Drarig29: Oh I got lost and made the commit in the [subsequent PR](https://github.com/DataDog/datadog-ci/pull/648) 😅

src/commands/synthetics/reporters/junit.ts
Drarig29: Yes, but it's on purpose because of Azure DevOps sadly.

PR #649, opened 2022-09-27
[lambda] Improve devex for testing output with colors

duncanista:
### What and why?

Showing colors in our CI is amazing, until we have to get snapshots to verify the output.
There are multiple approaches for this:
1. Raw printing the shell output and save it as a snapshot, then compare it.
2. Mocking `chalk` module to ignore colors and just return the text, that way snapshots can be copied directly.

In this repo, both approaches are being used.

### How?

Creating a mock for `chalk` and updating current tests for `lambda`, while unmocking the module for other commands which have their own tests.

### Review checklist


astuyve (approved):

src/commands/lambda/__tests__/instrument.test.ts
```diff
@@ -1,9 +1,9 @@
 // tslint:disable: no-string-literal
 jest.mock('fs')
+jest.mock('chalk')
```
m-rousse, line 3: As it is mocked globally I do not think it is required to mock it here anymore

```suggestion
```

m-rousse (approved):
LGTM :+1:

src/commands/lambda/__tests__/instrument.test.ts
duncanista: Thanks!

PR #659, opened 2022-10-07
[CI Visibility] Improve handling of CI env for tag and metric cmd

cg14823:
### What and why?
This PR changes how we get the CI env for the tag and metrics command so that we only retrieve what is strictly needed as well as return an error if some of the env needed is missing.

### How?

Hardcode the required env

### Review checklist

### DIFF
--- a/src/helpers/ci.ts
+++ b/src/helpers/ci.ts
@@ -592,28 +592,28 @@ const parsePipelineNumber = (pipelineNumberStr: string | undefined): number | un
 export const getCIEnv = (): {ciEnv: Record<string, string>; provider: string} => {
   if (process.env.CIRCLECI) {
     return {
-      ciEnv: filterEnv(['CIRCLE_WORKFLOW_ID', 'CIRCLE_BUILD_NUM'], true),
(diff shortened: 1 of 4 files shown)


juan-fernandez (approved):
LGTM! Only a tiny nitpick

src/helpers/ci.ts
```diff
@@ -592,36 +592,52 @@ const parsePipelineNumber = (pipelineNumberStr: string | undefined): number | un
...
   }
 
   throw new Error('Only providers [GitHub, GitLab, CircleCI, Buildkite] are supported')
 }
 
-const getEnvVars = (prefix: string): Record<string, string> =>
-  Object.entries(process.env)
-    .filter(([key, value]) => key.startsWith(prefix) && !/(PASS)|(TOKEN)|(SECRET)|(KEY)/i.test(key))
-    .reduce((accum, [key, value]) => ({...accum, [key]: value}), {})
+const filterEnv = (values: string[], allValues: boolean): Record<string, string> => {
```
juan-fernandez, line 624: do we need `allValues`? It seems all our usages are setting `true`

cg14823: It was to use it for other things in the future but, fair we can add it in the future if needed. I'll change it.

PR #660, opened 2022-10-07
[synthetics] Validate `datadogSite` on config resolved

Drarig29:
### What and why?

This PR introduces a new `INVALID_CONFIG` critical error. For now, it's only used if the `datadogSite` is invalid to show a nice error message. It could later be used to do additional validation on the config.

This doesn't change the behavior in case of an invalid `datadogSite` since the command would have failed when querying the test configs anyway.

### Review checklist

### DIFF
--- a/src/commands/synthetics/utils.ts
+++ b/src/commands/synthetics/utils.ts
@@ -754,6 +754,18 @@ export const parseVariablesFromCli = (
   return Object.keys(variables).length > 0 ? variables : undefined
 }
 
+export const validateDatadogSite = (datadogSite: string) => {
+  const datadogSiteParts = datadogSite.split('.')
+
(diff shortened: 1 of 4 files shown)


src/commands/synthetics/utils.ts
```diff
@@ -754,6 +754,18 @@ export const parseVariablesFromCli = (
   return Object.keys(variables).length > 0 ? variables : undefined
 }
 
+export const validateDatadogSite = (datadogSite: string) => {
```
m-rousse, line 757: Could you add tests for this function?

Drarig29: This function achieves basic validation over the general format of a datadog site, so I just added basic testing. I'm not sure we want to do proper validation to reject `dqtqdqghq.com` for instance.

m-rousse (approved):

PR #666, opened 2022-10-12
[lint] Migrate to `eslint`

Drarig29:
### What and why?

This PR replaces `tslint` (deprecated since 2019) with `eslint`.

We had a lot of `any` and `require()`, so some rules were set as warnings. This can be incrementally changed later.

In terms of **code changes**, this PR takes care of the files that are only owned by the @DataDog/datadog-ci team.

### How?

The `eslint` config was first auto-generated with [tslint-to-eslint-config](https://github.com/typescript-eslint/tslint-to-eslint-config) and then changed a bit to be closer to our current codebase.

The code changes are a result of `eslint --fix`, then a few changes to make the linter pass.

### Review checklist


Drarig29:
The "Build and test" CI job will only pass **once all of the child PRs are merged** into this PR.

package.json
```diff
@@ -38,17 +38,17 @@
     "clean": "rm -rf dist/*",
     "dist-standalone": "pkg --compress GZip .",
     "dist-standalone:test": "jest --config ./jest.config-standalone.js",
-    "format": "yarn tslint --fix && yarn prettier --write",
```
m-rousse, line 41: Could we keep prettier here and in `lint` as some rules are specifically checked by prettier?
(cf. https://github.com/DataDog/datadog-ci/pull/666/files#diff-e2954b558f2aa82baff0e30964490d12942e0e251c1aa56c3294de6ec67b7cf5R171)

package.json
```diff
@@ -38,17 +38,17 @@
...
-    "lint": "(yarn tslint && yarn prettier --check) || (echo \"\nYou can fix this by running ==> yarn format <==\n\" && false)",
+    "lint": "yarn eslint || (echo \"\nYou can fix this by running ==> yarn format <==\n\" && false)",
     "no-only-in-tests": "grep -R \"\\.only[(\\.]\" $(find src -name '*.ts' | grep '__tests__'); test $? -eq 1 || (echo '.only was found in the tests, please remove it.' && false)",
     "prepack": "yarn build && node ./bin/make-it-executable.js",
     "prettier": "prettier \"src/**/*.{ts,js,json,yml}\" --ignore-path .gitignore",
     "test": "jest --colors",
     "test:debug": "node --inspect-brk `which jest` --runInBand",
     "test:ci-vis:itr": "jest --colors --config ./jest.config-ci-vis-itr.js --passWithNoTests",
-    "tslint": "tslint -p tsconfig.json -c tslint.json",
+    "eslint": "eslint $(git diff --name-only 6dba118d780aa19939edf0f52a4a48e6e5e17477..HEAD) --ignore-pattern package.json",
```
m-rousse, line 51: As this PR is already broken, having this fix for CI seems unnecessary

src/commands/dsyms/__tests__/upload.test.ts
```diff
@@ -1,9 +1,8 @@
-// tslint:disable: no-string-literal
-import {Cli} from 'clipanion/lib/advanced'
 import {existsSync, promises} from 'fs'
-import glob from 'glob'
 import {EOL, platform} from 'os'
 import path from 'path'
+import glob from 'glob'
+import {Cli} from 'clipanion/lib/advanced'
```
m-rousse, line 5: Could the rules be changed to have a blank line between import groups?
Else it is particularly strange to have linter force us to have `glob` right after `path`

package.json
Drarig29: We use a plugin for eslint which runs prettier as part of the `lint` process. (cf. https://github.com/DataDog/datadog-ci/pull/666/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R111).

https://github.com/prettier/eslint-plugin-prettier/blob/master/README.md?plain=1#L79-L91

src/commands/dsyms/__tests__/upload.test.ts
lefebvree: You can try this rule configuration to mirror what we have in other repos:
```js
'import/order': [
  'error',
  {
    alphabetize: {order: 'asc', caseInsensitive: true},
    'newlines-between': 'always',
    groups: [['builtin', 'object'], 'type', 'external', 'internal', 'parent', ['sibling', 'index']],
    // set different groups for 5 different import levels ('../../../../**', '../../../**', ...)
    pathGroups: [
      ...Array.from({length: 5}).map((_, index) => ({
        pattern: `${'../'.repeat(5 + 1 - index)}**`,
        group: 'parent',
        position: 'before',
      })),
    ],
  },
],
```

src/commands/react-native/errors.ts
```diff
@@ -13,19 +12,21 @@ export class CodepushHistoryParseError extends Error {
 
 export class CodepushHistoryCommandError extends Error {
   constructor(message: string, command: string) {
+    let errorMessage: string
     try {
-      const commandMessage: string = JSON.parse(message).errorMessage
+      errorMessage = JSON.parse(message).errorMessage
```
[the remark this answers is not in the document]
Drarig29: Because the [`constructor-super`](https://eslint.org/docs/latest/rules/constructor-super) eslint rule enforces to have a single `super()` call. The error says "Unexpected duplicate 'super()'".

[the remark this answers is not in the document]
m-rousse: Ohh ack, thanks

m-rousse (approved):

m-rousse (approved):

m-rousse:
Feel free to merge bypassing the branch protections!

PR #667, opened 2022-10-12
[lint][synthetics-worker] Migrate to `eslint`

Drarig29:
### What and why?

Related to https://github.com/DataDog/datadog-ci/pull/666.

Mostly import rearrangements, and non-null assertion operator deletions since TypeScript type inference improved over the time.

### Review checklist

You can **ignore the change in `package.json`**: it is **temporary** and here to make the CI pass on this PR. It will be removed once merged in #666.

- [ ] Ensure nothing is broken in your team scope

### DIFF
--- a/src/commands/synthetics/tunnel.ts
+++ b/src/commands/synthetics/tunnel.ts
@@ -72,8 +72,7 @@ export class Tunnel {
    * keepAlive will return a promise that tracks the state of the tunnel (and reject in case of error)
    */
   public async keepAlive() {
-    // eslint-disable-next-line @typescript-eslint/no-misused-promises
-    if (!this.ws || !this.ws.keepAlive()) {
+    if (!this.ws) {
       throw new Error('No WebSocket connection')
     }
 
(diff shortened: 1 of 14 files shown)


src/commands/synthetics/tunnel.ts
```diff
@@ -71,6 +72,7 @@ export class Tunnel {
    * keepAlive will return a promise that tracks the state of the tunnel (and reject in case of error)
    */
   public async keepAlive() {
+    // eslint-disable-next-line @typescript-eslint/no-misused-promises
```
m-rousse, line 75: Why ignoring the error?

If I read the code correctly, the assertion `!this.ws.keepAlive()` can never be false, so I'd suggest we remove it as it will be awaited (and the potential error caught) after.

Drarig29: I forgot to add a review comment about this. That was a question I wanted to ask, I was afraid to not understand a hidden functionality in this check. I'll remove it.

m-rousse (approved):

PR #674, opened 2022-10-14
Move datadog site validation in shared helpers

Drarig29:
### What and why?

We didn't want to maintain a list of datadog sites in #660, but the `lambda` command already does it. So, let's move this in the shared helpers so that all commands can use it.

### How?

Move `SITES` in a new `src/constants.ts` file and create a new `isValidDatadogSite()` validation helper.

### Review checklist


m-rousse (approved):

duncanista (approved):
LGTM! 🚀 
Thanks for the update.

PR #675, opened 2022-10-14
[ci] Detect wrong license csv format

Drarig29:
### What and why?

Tools that analyze license listing files can break because of a wrong CSV format.

It's the case for GitHub for instance:
image

### How?

Check for a wrong format in the `check-licenses.js` script.

A failing pipeline would look like this:

```
Look for dependencies in:
 [ './package.json' ] 

LICENSE-3rdparty.csv: Line 22 has 5 columns, but header has 4.
LICENSE-3rdparty.csv: Line 28 has 5 columns, but header has 4.
LICENSE-3rdparty.csv: Line 35 has 5 columns, but header has 4.
LICENSE-3rdparty.csv: Line 36 has 5 columns, but header has 4.
LICENSE-3rdparty.csv: Line 40 has 5 columns, but header has 4.
LICENSE-3rdparty.csv: Line 49 has 5 columns, but header has 4.
LICENSE-3rdparty.csv: Line 53 has 5 columns, but header has 4.

✅ All dependencies listed in LICENSE-3rdparty.csv

Stacktrace:
 Error: The CSV format is incorrect. You may have an extra comma in a copyright.
    at main (/Users/corentin.girard/repos/datadog-ci/bin/check-licenses.js:40:11)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
```

### Review checklist

### DIFF
--- a/bin/check-licenses.js
+++ b/bin/check-licenses.js
@@ -11,12 +11,13 @@ const LICENSE_FILE = 'LICENSE-3rdparty.csv'
 async function main() {
   const packageJsonPaths = await findPackageJsonPaths()
 
-  console.log(`Look for dependencies in:\n`, packageJsonPaths)
+  console.log(`Look for dependencies in:\n`, packageJsonPaths, '\n')
   const declaredDependencies = withoutDuplicates(
     packageJsonPaths.reduce((acc, packageJsonPath) => acc.concat(retrievePackageJsonDependencies(packageJsonPath)), [])
   ).sort()
 
-  const declaredLicenses = (await retrieveLicenses()).sort()
+  const { licenses, hasColumnCountMismatch } = await retrieveLicenses()
+  const declaredLicenses = licenses.sort()
 
   if (JSON.stringify(declaredDependencies) !== JSON.stringify(declaredLicenses)) {
     console.error(JSON.stringify(declaredDependencies, null, 2))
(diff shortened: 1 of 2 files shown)


bin/check-licenses.js
```diff
@@ -32,11 +33,16 @@ async function main() {
...
   console.log(`\n✅ All dependencies listed in ${LICENSE_FILE}`)
+
+  if (hasColumnCountMismatch) {
+    throw Error('The CSV format is incorrect. You may have an extra comma in a copyright.')
+  }
 }
 
 async function findPackageJsonPaths() {
-  const {stdout} = await exec('find . -path "*/node_modules/*" -prune -o -name "package.json" -print')
+  const { stdout } = await exec('find . -path "*/node_modules/*" -prune -o -name "package.json" -print')
```
m-rousse, line 45: Why this change?
Lint config we are using usually does not have spaces (it would be rule `object-curly-spacing` to `never` with eslint) neither does it seem set in the ongoing eslint migration.
```suggestion
  const {stdout} = await exec('find . -path "*/node_modules/*" -prune -o -name "package.json" -print')
```

bin/check-licenses.js
```diff
@@ -52,19 +58,38 @@ function withoutDuplicates(a) {
   return [...new Set(a)]
 }
 
+function splitColumns(s) {
```
m-rousse, line 61: It would be better to explicitly refer to CSV, what do you think of `parseCSVLine`?

bin/check-licenses.js
```diff
@@ -52,19 +58,38 @@ function withoutDuplicates(a) {
...
+  for await (const line of lines) {
+    const csvColumns = splitColumns(line)
+    if (header) {
+      headerColumnCount = csvColumns.length
+    }
     if (!header && csvColumns[0] !== 'file') {
       licenses.push(csvColumns[0])
+
+      if (csvColumns.length !== headerColumnCount) {
+        console.log(`${LICENSE_FILE}: Line ${lineNumber} has ${csvColumns.length} columns, but header has ${headerColumnCount}.`)
```
m-rousse, line 84:
```suggestion
        console.warn(`${LICENSE_FILE}: Line ${lineNumber} has ${csvColumns.length} columns, but header has ${headerColumnCount}.`)
```

bin/check-licenses.js
Drarig29: Oops, I wasn't using Prettier for JS files.

m-rousse (approved):

PR #676, opened 2022-10-19
Add support for buddy.works in junit upload

cg14823:
### What and why?

Change the datadog-ci so that it recognizes when it is running in [buddy.works](https://buddy.works)

### How?

Adding to the helper function the mappings for the buddy environment.

### Review checklist


juan-fernandez (approved):
🎉

PR #677, opened 2022-10-19
Add custom tags and metrics support for Buddy

cg14823:
### What and why?
Adds support for creating custom tags and metrics for Buddy. Note that only pipeline level is supported as we do not have enough information to identify a job.

⚠️  Do not merge until the backend is deployed

### Review checklist

### DIFF
--- a/src/commands/metric/metric.ts
+++ b/src/commands/metric/metric.ts
@@ -59,9 +59,9 @@ export class MetricCommand extends Command {
     try {
       const metrics = parseMetrics(this.metrics)
       const {provider, ciEnv} = getCIEnv()
-      // For GitHub only the pipeline level is supported as there is no way to identify the job from the runner.
-      if (provider === 'github' && this.level === 'job') {
-        this.context.stderr.write(`${chalk.red.bold('[ERROR]')} Cannot use level "job" for GitHub Actions.`)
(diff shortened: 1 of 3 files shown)


src/commands/metric/metric.ts
```diff
@@ -59,9 +59,9 @@ export class MetricCommand extends Command {
     try {
       const metrics = parseMetrics(this.metrics)
       const {provider, ciEnv} = getCIEnv()
-      // For GitHub only the pipeline level is supported as there is no way to identify the job from the runner.
-      if (provider === 'github' && this.level === 'job') {
-        this.context.stderr.write(`${chalk.red.bold('[ERROR]')} Cannot use level "job" for GitHub Actions.`)
+      // For GitHub and Buddy only the pipeline level is supported as there is no way to identify the job from the runner.
+      if ((provider === 'github' || provider === 'buddy') && this.level === 'job') {
```
juan-fernandez, line 63: why not keep a dictionary provider:display_name so we can keep the snapshot? I think `GitHub Actions` looks better than `CI provider github`. 
Something like: 
```javascript
const PROVIDER_TO_DISPLAY_NAME = {
  github: 'GitHub Actions'
}
`Cannot use level "job" for ${PROVIDER_TO_DISPLAY_NAME[provider]}.`
```

juan-fernandez (approved):
minor feedback. LGTM otherwise

PR #679, opened 2022-10-19
[dev] Add `DD_CI_BYPASS_SITE_VALIDATION` environment flag support

lefebvree:
### What and why?

Add support for development environment variable `DD_CI_BYPASS_SITE_VALIDATION` disabling site validation.

### How?

Ignore the `datadogSite` validation when the `DD_CI_BYPASS_SITE_VALIDATION` environment var value is truthy.

### Review checklist

### DIFF
--- a/src/helpers/__tests__/validation.test.ts
+++ b/src/helpers/__tests__/validation.test.ts
@@ -11,5 +11,9 @@ describe('validation', () => {
     expect(isValidDatadogSite('datadoghq.com')).toBe(true)
     expect(isValidDatadogSite('us3.datadoghq.com')).toBe(true)
     expect(isValidDatadogSite('US3.datadoghq.com')).toBe(true)
+
+    process.env.DD_CI_BYPASS_SITE_VALIDATION = 'true'
+    expect(isValidDatadogSite('')).toBe(true)
+    expect(isValidDatadogSite('random')).toBe(true)
   })
 })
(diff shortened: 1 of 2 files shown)


m-rousse (approved):

src/helpers/__tests__/validation.test.ts
```diff
@@ -11,5 +11,9 @@ describe('validation', () => {
     expect(isValidDatadogSite('datadoghq.com')).toBe(true)
     expect(isValidDatadogSite('us3.datadoghq.com')).toBe(true)
     expect(isValidDatadogSite('US3.datadoghq.com')).toBe(true)
+
+    process.env.DD_CI_BYPASS_SITE_VALIDATION = 'true'
```
ruizb, line 15: For the record, setting `= 'false'` would also bypass site validation, which may be unexpected.

A bit late as it's already merged, but we could:
* Either use any value instead of "true" in the test, to show that we are just checking if the env var is _truthy_ or not.
* Or, use a function or library that converts a string into a boolean in a smart way (checking for "true" and "false", "yes" and "no", etc. I'm pretty sure such library already exists).

lefebvree: We can add a simple `process.env.DD_CI_BYPASS_SITE_VALIDATION && !['false', '0'].includes(process.env.DD_CI_BYPASS_SITE_VALIDATION.toLowerCase())` if that's bothersome but it's a minor development utility, I think adding a library for that is a bit overkill

PR #680, opened 2022-10-19
[Synthetics] Extract Tunnel client API

lefebvree:
### What and why?

Extract the Synthetics `Tunnel` client to its own directory to be used as a standalone module free of external datadog-ci imports.

### How?

Update the `Tunnel` API:
- The optional proxy to use for the tunnel must now be instantiated before and the agent passed to the tunnel constructor.
  `getProxyAgent()` is called before initializing the tunnel in `executeTests()`.
- A `TunnelReporter` is now used which will emit the raw message logs, the default reporter expose a `getTunnelReporter()` utility function to format the logs with chalk.
- Move all tunnel files to the `tunnel/` directory and export the main class and interfaces in its `index.ts`.

### Review checklist


src/commands/synthetics/tunnel/tunnel.ts
```diff
@@ -30,24 +28,30 @@ export interface TunnelInfo {
...
-    this.logWarning = (message: string) => reporter.log(`[${chalk.bold.yellow('Tunnel')}] ${message}\n`)
+  private connected = false
+  private ws: WebSocket
+  private multiplexer?: Multiplexer
+  private forwardedSockets: Set<Socket> = new Set()
 
+  constructor(
+    private url: string,
+    private testIDs: string[],
+    proxyAgent?: ReturnType<typeof ProxyAgent>,
```
jakepruitt, line 52: Why is `ReturnType<typeof ...>` needed here ?

src/commands/synthetics/tunnel/tunnel.ts
```diff
@@ -30,24 +28,30 @@ export interface TunnelInfo {
...
+    private reporter?: TunnelReporter
```
jakepruitt, line 53: Why make the reporter optional ?

jakepruitt (approved):
Small nits, otherwise looks good to me!

src/commands/synthetics/tunnel/tunnel.ts
lefebvree: `ProxyAgent` is actually the constructor and the `proxy-agent` doesn't seem to expose the actual `ProxyAgent` type sadly

src/commands/synthetics/tunnel/tunnel.ts
lefebvree: I was hesitant making it optional, some integrations may not need it but maybe we can instead make its methods optional 🤔

jakepruitt: Ah gotcha, this was with integrations in mind. Then yeah, from a core functionality standpoint the reporter is not necessarily "required".

PR #681, opened 2022-10-20
Release v1.17.0

louiszawadzki:
### What and why?

Adds `--remove-sources-content` argument to react-native commands.


lefebvree (approved):
Feel free to ignore the Docker publishing jobs for the merge, they are dependant on the Github release creation we should just re-run them once the new version is published

PR #682, opened 2022-10-20
[major] Bump minimum supported node version to node 14 and bump other dependencies

juan-fernandez:
### What and why?

Support for Node 10 was dropped more than 1 year ago (https://endoflife.date/nodejs). Bumping the minimum required version allows us to use new language features and bump some internal dependencies (like `dd-trace`). 

### How?

* Bump minimum supported node version to 14.
* Bump `dd-trace`, `jest` and related dependencies. **Note**: not the latest version has been used since that breaks a lot of tests. This work will be separated. 
* Do not setup a datadog agent in CI, as the test reporting is now using the agentless intake. 

### Review checklist


package.json
```diff
@@ -102,22 +102,21 @@
...
     "eslint": "^7.32.0",
     "eslint-config-prettier": "^8.5.0",
     "eslint-import-resolver-typescript": "2.7.0",
     "eslint-plugin-import": "^2.26.0",
     "eslint-plugin-no-null": "^1.0.2",
     "eslint-plugin-prefer-arrow": "^1.2.3",
     "eslint-plugin-prettier": "4.0.0",
-    "jest": "27.0.5",
-    "jest-environment-node": "27.0.5",
+    "jest": "28.1.3",
```
juan-fernandez, line 113: I couldn't bump to the latest major (29) because it requires a bump to `ts-jest`, which requires a bump to `typescript` which would break our linter right now

m-rousse (approved):
🎉

albertvaka (approved):
🍰

Drarig29:
Could you also update the Node version badge in the README?

badge

juan-fernandez:
> Could you also update the Node version badge in the README?
> 
> badge

that's a good catch. I also updated the `node-version` file in https://github.com/DataDog/datadog-ci/pull/682/commits/28690b94a8b4bb7732b5a8f7ec01149e6829192b

PR #683, opened 2022-10-21
[CI Visibility] (innovation week) junitxml custom tags with xpath

AdrianLC:
### What and why?

Adding a new CLI flag to junit upload command to customize how tags are added to tests.

### How?

It's quite simple we only split the flags provided in the first `=` and forward it to the backend request

### Review checklist

### DIFF
--- a/src/commands/junit/__tests__/upload.test.ts
+++ b/src/commands/junit/__tests__/upload.test.ts
@@ -24,6 +24,12 @@ const createMockContext = () => {
         data += input
       },
     },
+    stderr: {
+      toString: () => data,
+      write: (input: string) => {
+        data += input
+      },
+    },
   }
 }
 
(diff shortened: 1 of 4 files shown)


src/commands/junit/api.ts
```diff
@@ -36,26 +36,30 @@ export const uploadJUnitXML = (request: (args: AxiosRequestConfig) => AxiosPromi
...
+  const metadata: Record<string, any> = {
     service: jUnitXML.service,
     ...jUnitXML.spanTags,
     '_dd.cireport_version': '2',
     '_dd.hostname': jUnitXML.hostname,
   }
 
   if (jUnitXML.logsEnabled) {
-    spanTags['_dd.junitxml_logs'] = 'true'
+    metadata['_dd.junitxml_logs'] = true
```
AdrianLC, line 47: This is on purpose.

While implementing this I changed how we decode in the backend so any JSON type is allowed now

src/commands/junit/__tests__/upload.test.ts
```diff
@@ -154,6 +160,41 @@ describe('upload', () => {
...
+          logs: true,
+          service: 'service',
+        },
+        {}
+      )
+      expect(firstFile.logsEnabled).toBe(true)
+      expect(secondFile.logsEnabled).toBe(true)
+    })
+  })
+  describe('getSpanTags', () => {
```
AdrianLC, line 197: just moved a couple of tests to a new describe

src/commands/junit/api.ts
```diff
@@ -36,26 +36,30 @@ export const uploadJUnitXML = (request: (args: AxiosRequestConfig) => AxiosPromi
     fileName = 'default_file_name'
   }
 
-  const spanTags: Record<string, string | undefined> = {
+  const metadata: Record<string, any> = {
```
juan-fernandez, line 39: why `any`? Could we scope it down? Like `string | boolean | undefined | {[key:string]: string }`?

juan-fernandez (approved):

src/commands/junit/api.ts
AdrianLC: This matches how the backend handles it now. Any type is allowed in the value really so I'm not sure I see the point but  I don't know typescript so if you think it's still better we can change it

juan-fernandez: yeah ok, let's leave it as is

PR #685, opened 2022-10-21
Run tests in different node versions

juan-fernandez:
### What and why?

Run tests in different node versions to improve reliability.

At some point we'd want to test different operating systems, but windows tests are broken right now, so that will require a different PR.

### How?

* Run unit tests in node@14, node@16 and node@18.

### Review checklist


lefebvree (approved):
LGTM!

PR #686, opened 2022-10-24
Support Expo managed application without expo dev client

louiszawadzki:
### What and why?

For expo applications without the expo dev client, the `MARKETING_VERSION` environment variable is not present and `CURRENT_PROJECT_VERSION` is always `1` - not the value that's actually used for the build. 
We therefore have to rely on the info.plist by default to get release version and build version.

If we can't find the plist of can't read the version (there is no known reason for that to be the case), we'll use the env variables as fallback values

### How?

- Parse the info.plist
- Get the 2 values we're interested in

### Review checklist


src/commands/react-native/xcode.ts
```diff
@@ -280,6 +295,41 @@ export class XCodeCommand extends Command {
...
+    if (!this.infoPlistPath) {
+      throw new Error('Could not find plist path')
+    }
+
+    return parsePlist(this.infoPlistPath).getPropertyValue(propertyName)
+  }
+
+  private getReleaseVersion = (): string | null => {
+    try {
+      const releaseVersion = this.getPlistValue('CFBundleShortVersionString')
```
plousada, line 308: I believe the iOS SDK falls back to `CFBundleVersion` if `CFBundleShortVersionString` isn't found. 
Should we apply the same logic here?

plousada (approved):

src/commands/react-native/xcode.ts
louiszawadzki: As discussed, we can assume that if it’s not defined in the Info.plist and in MARKETING_VERSION it’s likely a mistake.

m-rousse (approved):

AdrianLC (approved):

nachoBonafonte (approved):

PR #693, opened 2022-11-03
Set USE_HERMES env variable for RN 0.70

louiszawadzki:
### What and why?

For RN 0.70 users using Hermes, sourcemaps are not being generated due to a RN bug.
To solve this, we need to set the `USE_HERMES` environment variable: see https://github.com/DataDog/datadog-ci/issues/690

### How?

Logic and checks are based on the `react-native-xcode.sh` script from React Native.

### Review checklist


0xnm (approved):

src/commands/react-native/xcode.ts
```diff
@@ -354,15 +369,38 @@ export class XCodeCommand extends Command {
...
     }
 
-    // Check if hermes pod is present
-    return existsSync(`${process.env.PODS_ROOT}/hermes-engine/destroot/bin/hermesc`)
+    /**
+     * Checks if Hermes is in the Podfile.lock.
+     * This is the most recent check for detecting Hermes in the `react-native-xcode.sh` script:
+     * https://github.com/facebook/react-native/commit/8745a148b6d8358702b5300d73f4686c3aedb413
+     *
+     * If the Podfile.lock cannot be found, we assume Hermes is not enabled
```
fuzzybinary, line 395: Is there a situation where you could not have a Podfile.lock but still have a valid build, even in older versions of React Native? I'm wondering if this should at least give a warning if not an outright error.

PR #700, opened 2022-11-15
[SW-2489] Improve logs output

kalinichenko:
### What and why?

Improved wordings of references to a single test result and a batch results, as they were a bit fuzzy


Drarig29 (approved):
LGTM

PR #704, opened 2022-11-18
[CI App] Add support for custom tags & metrics in Jenkins

albertvaka:
### What and why?

Detect when running on Jenkins and forward the right env vars so we can add custom tags and metrics to Jenkins pipelines.

Also updates the error message to say that Jenkins and Buddy are supported (buddy was added recently but the message not updated).

### Review checklist

### DIFF
--- a/src/helpers/__tests__/ci.test.ts
+++ b/src/helpers/__tests__/ci.test.ts
@@ -224,7 +224,7 @@ describe('getCIEnv', () => {
     })
   })
 
-  test('gitlab', () => {
+  test('jenkins', () => {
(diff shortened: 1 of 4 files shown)


src/helpers/__tests__/ci.test.ts
```diff
@@ -223,4 +223,17 @@ describe('getCIEnv', () => {
       provider: 'gitlab',
     })
   })
+
+  test('gitlab', () => {
```
juan-fernandez, line 227:
```suggestion
  test('jenkins', () => {
```

juan-fernandez (approved):

PR #718, opened 2022-12-08
[dep] Update `simple-git` dependency & `@types/node`, Fix typing issues

lefebvree:
### What and why?

### How?

### Review checklist


src/commands/react-native/xcode.ts
```diff
@@ -187,7 +187,7 @@ export class XCodeCommand extends Command {
   }
 
   private bundleReactNativeCodeAndImages = async () => {
-    const bundleJSChildProcess = spawn(this.scriptPath, undefined, {
+    const bundleJSChildProcess = spawn(this.scriptPath, {
```
lefebvree, line 190: The only `spawn` overloads that matches `undefined` args prevents the usage of `inherit` in the `stdio` option (`SpawnOptionsWithoutStdio `):
```ts
function spawn(command: string, args?: ReadonlyArray<string>, options?: SpawnOptionsWithoutStdio): ChildProcessWithoutNullStreams;
```

I think it may be a shortcoming of `@types/node`, but otherwise `options` can be used as the second argument of spawn if there are no args (or we can use `[]` as args).

m-rousse (approved):

src/commands/react-native/xcode.ts
louiszawadzki: I'd be a bit more comfortable with the `[]` as args for the sake of consistency within this file, but that's a nitpick. If you want to merge as it is, fine by me 👍 

```suggestion
    const bundleJSChildProcess = spawn(this.scriptPath, [], {
```

louiszawadzki (approved):

PR #736, opened 2023-01-11
Remove git metadata uploading + add git repo url tag for SCI

thedavl:
### What and why?

Remove git metadata uploading when running the `lambda instrument` command with flag `-s` (`--source-code-integration`). Instead, add the repo url as a tag (alongside the commit sha that's already added) to the `DD_TAGS` env var. Fix tests.

This PR does not touch the datadog-ci `git-metadata` (`git-metadata upload`) command

### How?

Remove unneeded git upload function, get repo url from the git commit info (same place we get the commit hash).

These tags are added to `DD_TAGS` separated by commas, no spaces. The repo url is formatted in the following way:
`git.repository_url:github.com/thedavl/python_cdk_app.git`

- In order to not break backwards compatibility with tracer versions that don't support colons in tags, logic has been added to format git remotes turning `git@github.com:datadog/test.git` and `https://github.com/datadog/test.git` into `github.com/datadog/test.git`

- Previously, we would give an error if source code integration was enabled and the ddog api key wasn't provided. Removed this check (and an associated unit test), as this is no longer needed. The api key was previously used to upload the git metadata to datadog.

### Testing

Manually tested, by creating a new personal git repo with a sample app (which includes a forced exception in the lambda code), instrumenting it using my local modified `datadog-ci` with the `-s` flag, and checking for SCI functionality.

### Review checklist

### DIFF
--- a/src/commands/lambda/__tests__/instrument.test.ts
+++ b/src/commands/lambda/__tests__/instrument.test.ts
@@ -476,6 +476,75 @@ TagResource -> arn:aws:lambda:us-east-1:123456789012:function:lambda-hello-world
         expect(output).toMatch('Error: Local git repository is dirty')
       })
 
+      test ('ensure source code integration git remotes get formatted correctly', async () => {
+        ;(fs.readFile as any).mockImplementation((a: any, b: any, callback: any) => callback({code: 'ENOENT'}))
+        const lambda = makeMockLambda({
+          'arn:aws:lambda:us-east-1:123456789012:function:lambda-hello-world': {
+            FunctionArn: 'arn:aws:lambda:us-east-1:123456789012:function:lambda-hello-world',
+            Handler: 'index.handler',
+            Runtime: 'nodejs12.x',
+          },
+        })
+        ;(Lambda as any).mockImplementation(() => lambda)
+        process.env.DATADOG_API_KEY = '1234'
+
+        const context = createMockContext() as any
+        const instrumentCommand = InstrumentCommand
+        const instrumentCommandSpy = jest.spyOn(instrumentCommand.prototype as any, 'filterAndFormatGitRemote')
+
+        const mockGitStatus = jest.spyOn(instrumentCommand.prototype as any, 'getCurrentGitStatus')
+        mockGitStatus.mockImplementation(() => ({
+          ahead: 0,
+          hash: '1be168ff837f043bde17c0314341c84271047b31',
+          remote: 'https://github.com/datadog/test.git',
+          isClean: true,
+        }))
+
+        const cli = new Cli()
+        cli.register(instrumentCommand)
+
+        const cliInput = [
+          'lambda',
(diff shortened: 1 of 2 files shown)


src/commands/lambda/__tests__/instrument.test.ts
```diff
@@ -493,12 +493,9 @@ TagResource -> arn:aws:lambda:us-east-1:123456789012:function:lambda-hello-world
         mockGitStatus.mockImplementation(() => ({
           ahead: 0,
           hash: '1be168ff837f043bde17c0314341c84271047b31',
+          remote: 'git.repository_url:git@github.com:datadog/test.git',
```
IvanTopolcic, line 496: Can we confirm two things here?

1. All tracers that we use support colon values in their tag values. I think just a quick message to the source code integration team would work here. 

2. That this change is backwards compatible with functions running older versions of a layer? I'm thinking of a situation where a customer has an old version of the lambda tracer that doesn't support the colon in the tag value, and this adds the tag which could cause errors in the application. If datadog-ci also installs the latest layers then I don't think this is an issue.

IvanTopolcic (commented):
The PR looks good but I think we should check for the aforementioned tagging issue just to he 100% safe.

src/commands/lambda/__tests__/instrument.test.ts
IvanTopolcic: I'd actually also check to make sure that datadog-cdk-constructs and serverless-plugin-datadog use pinned versions of this dependency because I know that was an issue for us before and it actually broke some of the deployments for our customers.

thedavl: Writing what we discussed here:

- All tracer versions definitely do not support colons, this would break backwards compatibility with those tracers. Add logic to format the git remote to turn `git@github.com:datadog/test.git` and `https://github.com/datadog/test.git` into `github.com/datadog/test.git`
- datadog-cdk-contructs and serverless-plugin-datadog both use `uploadGitCommitHash` from the `git-metadata` library, which this Pr doesn't touch. So we're good for that

src/commands/lambda/instrument.ts
```diff
@@ -316,10 +315,26 @@ export class InstrumentCommand extends Command {
...
+      remote: gitCommitInfo?.remote,
+    }
+  }
+
+  private filterAndFormatGitRemote(rawRemote: string | undefined): string | undefined {
+    rawRemote = filterSensitiveInfoFromRepository(rawRemote)
+    if (!rawRemote) {
+      return rawRemote
+    }
+    rawRemote = rawRemote.replace(/git@github.com:|https:\/\/github.com\//, 'github.com/')
```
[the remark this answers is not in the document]
IvanTopolcic: awesome work!

IvanTopolcic (approved):

PR #738, opened 2023-01-12
Add getGitCommitInfo to git-metadata library

thedavl:
### What and why?

Expose a new `getGitCommitInfo` function in the git-metadata library, which will be imported and used by `datadog-cdk-constructs` and `serverless-plugin-datadog` to update the source code integration implementation. 

Specifically, the function will be used to get the `git.repository_url`which will be tagged to lambda functions 
- The legacy solution, uploading git metadata to datadog, will remain for backwards compatibility, but the new way of just tagging functions with `git.repository_url` and `git.commit.sha` will be promoted.

### How?

Expose a new `getGitCommitInfo` function in the git-metadata library, which returns the git repo url and the git commit sha. Optionally takes a `filterAndFormatGitRepoUrl ` param, which when passed in as false will tell the function to forgo filtering and formatting the git repo url.

- Otherwise (by default) the function will run the repo url through the `filterSensitiveInfoFromRepository` function, as well as normalize the url prefix.
  - "git@github.com:" and "https://github.com/" prefixes will be normalized into "github.com/", removing colons for older tracers that don't support them

### Testing

Updated unit tests to cover both the exposed library function, as well as a filter and format util helper.

Manually tested using an in-progress version of the `datadog-cdk-constructs` (which supports repo tagging)

### Review checklist


src/commands/lambda/__tests__/instrument.test.ts
```diff
@@ -442,75 +442,6 @@ TagResource -> arn:aws:lambda:us-east-1:123456789012:function:lambda-hello-world
         expect(output).toMatch('Error: Local git repository is dirty')
       })
 
-      test('ensure source code integration git remotes get formatted correctly', async () => {
```
thedavl, line 445: moved this unit test into the test file for helpers/utils

src/commands/git-metadata/library.ts
```diff
@@ -17,6 +17,21 @@ export const isGitRepo = async (): Promise<boolean> => {
...
+// ("git@github.com:" and "https://github.com/" prefixes will be normalized into "github.com/")
+export const getGitCommitInfo = async (filterAndFormatGitRepoUrl?: boolean): Promise<[string, string]> => {
+  const simpleGit = await newSimpleGit()
+  const payload = await getCommitInfo(simpleGit)
+
+  const gitRemote = filterAndFormatGitRepoUrl ? filterAndFormatGitRemote(payload.remote) : payload.remote
+
+  // gitRemote will never be undefined, as filterAndFormatGitRemote will ONLY return undefined if it's
+  // parameter value is also undefined. Added the " gitRemote ?? '' " to make the typechecker happy.
+  return [gitRemote ?? '', payload.hash]
```
thedavl, line 32: couldn't find another way around this, let me know if there's a better way

src/commands/git-metadata/library.ts
```diff
@@ -17,6 +17,21 @@ export const isGitRepo = async (): Promise<boolean> => {
...
+// filterAndFormatGitRepoUrl == true, the repositoryURL will have sensitive information filtered and
+// git prefix normalized.
+// ("git@github.com:" and "https://github.com/" prefixes will be normalized into "github.com/")
+export const getGitCommitInfo = async (filterAndFormatGitRepoUrl?: boolean): Promise<[string, string]> => {
+  const simpleGit = await newSimpleGit()
+  const payload = await getCommitInfo(simpleGit)
+
+  const gitRemote = filterAndFormatGitRepoUrl ? filterAndFormatGitRemote(payload.remote) : payload.remote
+
+  // gitRemote will never be undefined, as filterAndFormatGitRemote will ONLY return undefined if it's
```
thedavl, line 30: couldn't find another way around this, let me know if there's a better way

src/commands/git-metadata/library.ts
```diff
@@ -17,6 +17,21 @@ export const isGitRepo = async (): Promise<boolean> => {
   }
 }
 
+// getGitCommitInfo returns the current [repositoryURL, commitHash]. If parameter
+// filterAndFormatGitRepoUrl == true, the repositoryURL will have sensitive information filtered and
+// git prefix normalized.
+// ("git@github.com:" and "https://github.com/" prefixes will be normalized into "github.com/")
+export const getGitCommitInfo = async (filterAndFormatGitRepoUrl?: boolean): Promise<[string, string]> => {
```
thedavl, line 24: will be used in datadog-cdk and the serverless plugin to get the commit sha and remote repo url

src/commands/git-metadata/library.ts
```diff
@@ -17,6 +17,21 @@ export const isGitRepo = async (): Promise<boolean> => {
...
+// ("git@github.com:" and "https://github.com/" prefixes will be normalized into "github.com/")
+export const getGitCommitInfo = async (filterAndFormatGitRepoUrl = true): Promise<[string, string]> => {
+  const simpleGit = await newSimpleGit()
+  const payload = await getCommitInfo(simpleGit)
+
+  const gitRemote = filterAndFormatGitRepoUrl ? filterAndFormatGitRemote(payload.remote) : payload.remote
+
+  // gitRemote will never be undefined, as filterAndFormatGitRemote will ONLY return undefined if it's
+  // parameter value is also undefined. Added the " gitRemote ?? '' " to make the typechecker happy.
+  return [gitRemote ?? '', payload.hash]
```
thedavl, line 32: couldn't find another way around the type err, let me know if there's a better way

IvanTopolcic: It seems to me like this happens because `filterAndFormatGitRemote` returns a `string | undefined`. Maybe we could just change that to `string` if you're confident that it cannot return undefined?

IvanTopolcic (commented):
LGTM! Just one small question

src/commands/git-metadata/library.ts
thedavl: the problem is with the `filterSensitiveInfoFromRepository` function that `filterAndFormatGitRemote` calls, which returns `string | undefined`. So it seems like the `?? ''` has to either live here, or within the `filterAndFormatGitRemote` func

IvanTopolcic (approved):

src/helpers/utils.ts
```diff
@@ -278,3 +278,15 @@ export const filterSensitiveInfoFromRepository = (repositoryUrl: string | undefi
     return repositoryUrl
   }
 }
+
+// Removes sensitive info from the given git remote url and normalizes the url prefix.
+// "git@github.com:" and "https://github.com/" prefixes will be normalized into "github.com/"
+export const filterAndFormatGitRemote = (rawRemote: string | undefined): string | undefined => {
```
lefebvree, line 284: nit (not related to this PR directly): Since this function only formats Github URLs it should maybe convey that in its name (`filterAndFormatGithubRemote`), or should it be made generic to support any domain?

lefebvree (approved):
LGTM for datadog-ci

sashacmc (approved):
LGTM!

src/helpers/utils.ts
thedavl: I think it makes sense to rename it to `filterAndFormatGithubRemote` for now, I'll make that change. Thanks

PR #747, opened 2023-01-17
[Synthetics] Tunnel: Downgrade connection error logs to warnings

lefebvree:
### What and why?

Change tunnel connections error to warnings.

### DIFF
--- a/src/commands/synthetics/tunnel/tunnel.ts
+++ b/src/commands/synthetics/tunnel/tunnel.ts
@@ -259,7 +259,7 @@ export class Tunnel {
     }
     this.multiplexer = new Multiplexer((stream) => {
       stream.on('error', (error) => {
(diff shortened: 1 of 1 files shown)


ruizb (commented):
LGTM 🐰 

Would the logs still be displayed by default when running a "conventional" datadog-ci synthetics command? If so, I thought we didn't want to pollute the output with hard-to-understand error messages 🤔

ruizb (approved):

lefebvree:
> Would the logs still be displayed by default when running a "conventional" datadog-ci synthetics command? If so, I thought we didn't want to pollute the output with hard-to-understand error messages 🤔

They are only occurring if there are connection errors so nothing is displayed on nominal cases, if they occur when closing all connections they won't be the cause of test failure, hence why we downgrade them to warning.
However if they happen during the test run they are likely to cause request errors they're useful for us for debugging/support purposes.

PR #749, opened 2023-01-18
[s8s] Add lib wrapper to use from outside

yoannmoinet:
### What and why?

Using `executeTests` from outside is quite complex and requires a somewhat big bootstrap we should not have to worry about when using this as a library.

### How?

Create a wrapper around `executeTests` called `execute` (open on the naming).
This wrapper will handle the reporters and configuration defaults.

Also tested locally with 
- https://github.com/DataDog/web-ui/pull/77588

### Review checklist

### DIFF
--- a/src/commands/synthetics/run-test.ts
+++ b/src/commands/synthetics/run-test.ts
@@ -244,7 +244,7 @@ export const execute = async (
     runId?: string
     suites?: Suite[]
   }
-) => {
+): Promise<0 | 1> => {
   const startTime = Date.now()
   const localConfig = {
     ...DEFAULT_COMMAND_CONFIG,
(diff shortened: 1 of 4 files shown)


etnbrd (approved):
LGTM 👍

etnbrd:
I was about to leave a comment about the naming of `execute` on the account that it's quite broad, and already called like that in the command. But actually, it kind of makes sense to have the same naming between the command and the lib, as they do roughly the same thing (except for some config handling and reporting) on the same level (wrapping `executeTests` and `renderResults`).

src/commands/synthetics/run-test.ts
```diff
@@ -218,3 +230,61 @@ export const getTestsList = async (
...
+    reporters,
+    runId,
+    suites,
+  }: {
+    jUnitReport?: string
+    reporters?: (SupportedReporter | Reporter)[]
+    runId?: string
+    suites?: Suite[]
+  }
+) => {
```
Drarig29, line 247: nit: could you add an explicit return type here? (we usually have a lint rule for that but this repo doesn't have one)

Drarig29 (commented):
Good idea, I like the `(SupportedReporter | Reporter)[]` part 👍 

Only nits on my side

Drarig29 (approved):

PR #754, opened 2023-01-23
Display warning message when SCI fails, rather than exiting.

thedavl:
### What and why?

Adds a warning if source code integration fails when instrumenting lambda functions. 

Previously, the CI would exit completely if SCI fails.

Example warning printed:
image

### How?

Rather than returning, prints "[Warning] Couldn't add source code integration. Error: {SCI error encountered}"

### Review checklist


astuyve (approved):
Not sure of we want to add a note to the warning which specifies that the instrumentation process will continue, but otherwise this looks good to me.

thedavl:
Updated warning to specify the intrumentation process will continue without source code integration.

image

PR #761, opened 2023-01-26
Fix spec tests to include tag and branch if both present

albertvaka:
### What and why?

Updates the json fixtures generated from the spec and fixes the tests.


src/helpers/ci.ts
```diff
@@ -529,13 +513,24 @@ export const getCISpanTags = (): SpanTags | undefined => {
...
-    tags[GIT_BRANCH] = normalizeRef(tags[GIT_BRANCH])
-  }
+
   if (tags[GIT_TAG]) {
     tags[GIT_TAG] = normalizeRef(tags[GIT_TAG])
   }
 
+  if (tags[GIT_BRANCH]) {
+    // Here we handle the case where GIT_BRANCH actually contains a tag
+    let branch = tags[GIT_BRANCH] || ''
```
juan-fernandez, line 523:
```suggestion
    const branch = tags[GIT_BRANCH] || ''
```

juan-fernandez (approved):

src/helpers/ci.ts
```diff
@@ -278,7 +270,8 @@ export const getCISpanTags = (): SpanTags | undefined => {
       if (jobNameAndParams.length > 1 && jobNameAndParams[1].includes('=')) {
         finalPipelineName = jobNameAndParams[0]
       } else {
-        finalPipelineName = JOB_NAME.replace(`/${ref}`, '')
+        let normalizedBranch = normalizeRef(JENKINS_GIT_BRANCH)
```
albertvaka, line 273:
```suggestion
        const normalizedBranch = normalizeRef(JENKINS_GIT_BRANCH)
```

Issue #753, opened 2023-01-23
Allow override commit hash when `sourcemaps upload` .


takaya1992:
**Is your feature request related to a problem? Please describe:**

Our build environment doesn't have a `.git/` directory when building, so we can't associate sourcemaps and commit hash.

`sourcemaps upload` has an option to override Git's Repository URL.
But there is no option to override commit hash.
Is there any reason for that?

**Describe the solution you'd like:**

Add `--commit-hash` option in `sourcemaps upload` command.
I want to override the commit-hash as well as the `--repository-url`.

**Describe alternatives you've considered:**

A usecase for the `--commit-hash` option is when building the code inside a Docker image as an example.
Use when you don't want to include `.git/` in your image, but you can pass a "commit hash" as an argument or environment variable.

Hunga1:
+1

knaperek:
+1

PR #764, opened 2023-01-28
Add commit-hash option in `sourcemaps upload` command

takaya1992:
### What and why?

https://github.com/DataDog/datadog-ci/issues/753 override commit hash when `sourcemaps upload`.

### How?

add `--commit-hash` option in `sourcemaps upload` command.

### Review checklist

### DIFF
--- a/src/commands/flutter-symbols/upload.ts
+++ b/src/commands/flutter-symbols/upload.ts
@@ -194,7 +194,7 @@ export class UploadCommand extends Command {
 
   private async getGitMetadata(): Promise<RepositoryData | undefined> {
     try {
-      return await getRepositoryData(await newSimpleGit(), this.repositoryUrl)
+      return await getRepositoryData(await newSimpleGit(), this.repositoryUrl, undefined)
     } catch (e) {
       this.context.stdout.write(renderGitWarning(e))
     }
--- a/src/commands/sourcemaps/README.md
+++ b/src/commands/sourcemaps/README.md
@@ -73,7 +73,7 @@ This behavior can be overridden with links to `https://gitlab.com/Datadog/exampl
 
 The value can be overridden with `--commit-hash`.
 
-Example: Pass commit hash when there is no `.git/` directory, such as build in Docker.
+Example: Pass commit hash when there is no `.git/` directory, such as a build in Docker.
 
 This behavior can be overridden with links to `36485fb1c5cba61e95bbbf7b30c673f2b96e415d` with the flag `--commit-hash=36485fb1c5cba61e95bbbf7b30c673f2b96e415d`.
 
--- a/src/commands/sourcemaps/README.md
+++ b/src/commands/sourcemaps/README.md
@@ -69,6 +69,14 @@ The value can be overridden with `--repository-url`.
 Example: With a remote `git@github.com:Datadog/example.git`, links pointing to `https://github.com/Datadog/example` are generated.
 This behavior can be overridden with links to `https://gitlab.com/Datadog/example` with the flag `--repository-url=https://gitlab.com/Datadog/example`.
 
+#### Override commit hash
+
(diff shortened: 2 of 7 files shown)


src/commands/sourcemaps/README.md
```diff
@@ -69,6 +69,14 @@ The value can be overridden with `--repository-url`.
 Example: With a remote `git@github.com:Datadog/example.git`, links pointing to `https://github.com/Datadog/example` are generated.
 This behavior can be overridden with links to `https://gitlab.com/Datadog/example` with the flag `--repository-url=https://gitlab.com/Datadog/example`.
 
+#### Override commit hash
+
+The value can be overridden with `--commit-hash`.
+
+Example: Pass commit hash when there is no `.git/` directory, such as build in Docker.
```
maycmlee, line 76:
```suggestion
Example: Pass commit hash when there is no `.git/` directory, such as a build in Docker.
```

maycmlee (commented):
Just a small edit, let me know if it changes the meaning.

src/commands/sourcemaps/README.md
takaya1992: thanks

maycmlee (approved):
👍 for docs

louiszawadzki (approved):
All good for me 👍

takaya1992:
@maycmlee @louiszawadzki 
Thank you for your code review.
It seems that it is necessary to get other members to do a code review as well.
Do you need me to take any action?

src/commands/flutter-symbols/upload.ts
```diff
@@ -194,7 +194,7 @@ export class UploadCommand extends Command {
 
   private async getGitMetadata(): Promise<RepositoryData | undefined> {
     try {
-      return await getRepositoryData(await newSimpleGit(), this.repositoryUrl)
+      return await getRepositoryData(await newSimpleGit(), this.repositoryUrl, undefined)
```
juan-fernandez, line 197: nitpick: couldn't we have an optional parameter to the function instead of passing `undefined`?

juan-fernandez (approved):
looks good for CI Visibility. Left a nitpick comment

rpelliard:
Hello @takaya1992,

As I understand it from https://github.com/DataDog/datadog-ci/issues/753, you're trying to get links from the stack trace to your repository, without having access to the `.git` directory during sourcemaps upload, right ?

In order for this feature to work, the sourcemaps upload command needs to collect:
- the repository URL and commit hash
- the list of tracked files in the repository (i.e. output of `git ls-files`, fetched by [`gitTrackedFiles`](https://github.com/DataDog/datadog-ci/blob/5c22d31e4fb6167db93432ebaa4d9dc72dc5c493/src/commands/git-metadata/git.ts#L58))

Without all these, our backend cannot build the links to your repository. This PR looks like a nice way to address the missing commit hash, but the command will still try to retrieve the list of tracked files from the `.git` directory, and will fail.

I'm afraid that these changes will not be enough to solve your problem, and I don't see a simple way to make the metadata upload work without the `.git` directory.

dlmr:
Would it be an alternative to run `datadog-ci git-metadata upload` separately from the `datadog-ci sourcemaps upload` with this flag added? Do you see that as an alternative for getting the tracked files @rpelliard?

I have a similar situation as @takaya1992 and tried to run `git-metadata` separate from the build process where I have the full git context, outside of the Docker container, but it did no replace the need within `sourcemaps`.

rpelliard:
Decoupling the git metadata upload from the sourcemaps upload is the direction we want to go in, but I'm afraid that it won't work "as-is" with the two existing commands (given the way our back-end is made at the moment), and unfortunately there is no simple workaround for now.
I will make sure that this is followed up on internally, and I will close this PR since it won't make the feature work for the use-cases exposed here.

PR #767, opened 2023-02-01
[lambda] Update tests from inline snapshots to snapshots

duncanista:
### What and why?

Updates long inline snapshots to snapshot files.
Added the command `test:lambda` easy access to `jest src/commands/lambda --colors`

It's worth noting that only really large pieces of inline snapshots are being moved to snapshots. This so our test files are smaller and those snapshots are generated by running a command.

### How?

Replacing `.toMatchInlineSnapshot(...)` to `toMatchSnapshot()`.
Then run `--updateSnapshot`

### Review checklist

- [x] Ensured generated tests were the same as the previous ones.


lefebvree (approved):
LGTM for datadog-ci 👍

sfirrin (approved):

PR #769, opened 2023-02-03
[lambda] Improve error handling when instrumenting multiple functions.

duncanista:
### What and why?

Improve error handling when instrumenting (or un-instrumenting) multiple functions. 

Currently, when a user is instrumenting multiple functions, let's say 10. If one of this functions comes to fail, the CI will state that there's been an error while instrumenting functions, e.g. the 6th function. Nonetheless, it will not indicate if some functions were updated properly (from the 1st to 5th). This PR allows us to know exactly which functions fail, and the reason why. While not interrupting the process and allowing it to continue. 

1. At least one function has to be updated to consider the process succesful.
2. All functions failing is consider an error while instrumenting.

### How?

Changing how we update functions to be all awaited in two single promises to be granulized into individual ones.

### Review checklist


src/commands/lambda/__tests__/functions/commons.test.ts
```diff
@@ -629,7 +629,9 @@ describe('commons', () => {
       ]
       const cloudWatch = makeMockCloudWatchLogs({})
 
-      await updateLambdaFunctionConfigs(lambda as any, cloudWatch as any, configs)
+      for (const config of configs) {
+        await updateLambdaFunctionConfig(lambda as any, cloudWatch as any, config)
```
astuyve, line 633: if we expect a number of configs, we should use `await Promise.all` or `await Promise.allSettled`

src/commands/lambda/functions/commons.ts
```diff
@@ -418,23 +418,26 @@ export const isLayerRuntime = (runtime: string): runtime is LayerKey => LAYER_LO
...
     }
-    if (c.tagConfiguration !== undefined) {
-      await applyTagConfig(lambda, c.tagConfiguration)
+    if (config.tagConfiguration !== undefined) {
+      await applyTagConfig(lambda, config.tagConfiguration)
     }
-  })
-  await Promise.all(results)
+  } catch (e) {
+    if (e instanceof Error) {
```
astuyve, line 437: Hmmm, is there a case where `e` would not be an instance of an `Error`? I'm not sure of one.

Also either way, we don't need to create a new Error here, and likely don't need the try-catch block at all if we're not handling any error.

sfirrin (approved):
Looks good :+1:

src/commands/lambda/instrument.ts
```diff
@@ -286,20 +286,54 @@ export class InstrumentCommand extends Command {
...
+        for (const config of group.configs) {
+          try {
+            await updateLambdaFunctionConfig(group.lambda, group.cloudWatchLogs, config)
+          } catch (err) {
+            failedUpdates.push({functionARN: config.functionARN, error: err})
+            totalFailedUpdates += 1
+          }
+        }
+
+        if (failedUpdates.length === group.configs.length) {
```
astuyve, line 304: Seems we're processing failed updates both in the inner `group.configs` loop, but also in the outer `configGroups` loop, where we `renderFail` on lines 322 and 323. Is there an opportunity to combine and simplify the logic here?

I'm a bit unclear on the difference in this block than the lower block.

src/commands/lambda/uninstrument.ts
```diff
@@ -198,20 +199,54 @@ export class UninstrumentCommand extends Command {
...
-      spinner.start()
-      try {
-        await Promise.all(promises)
-        spinner.succeed(renderer.renderUpdatedLambdaFunctions(promises.length))
-      } catch (err) {
-        this.context.stdout.write(renderer.renderFailureDuringUpdateError(err))
-        spinner.fail(renderer.renderFailedUpdatingLambdaFunctions())
+      const totalFunctions = Object.values(configGroups).reduce((c, group) => (c += group.configs.length), 0)
+      let totalFailedUpdates = 0
+      for (const group of configGroups) {
```
astuyve, line 204: The loop construct here and for the `instrument` method. Is it worthwhile to extract and combine this logic? Usually I wait until the 3rd time I need something to extract it, but I worry given the complexity here that someone may update this method or the `instrument` method and forget to mirror it here.

astuyve (approved):
Couple notes but nothing blocking except perhaps that `error` discussion. Good job!!

src/commands/lambda/functions/commons.ts
duncanista: I was doing it to type guard errors, since every error is of type `unknown`. But thank you for catching it! I don't know why I was raising a `new` error. I think the current state needs to be changes to address those comments. Thank you AJ.

src/commands/lambda/instrument.ts
duncanista: One is to state errors for the current region, and the other one for all the instrumented functions overall. 

For example here. Where we expect 3 functions to be updated but one fails.
Screenshot 2023-02-03 at 2 57 23 PM

It might be a little verbose if you are instrumenting for only one region.

src/commands/lambda/uninstrument.ts
duncanista: You are right, previously it was mirrored but the logic seemed to be really simple, with the added complexity, it might cause an issue. Let me think how I can abstract it given the nature of the output writing.

duncanista: Solved, please let me know what you think of this new alternative.

PR #777, opened 2023-02-14
[CIAPP] Add `SARIF` upload command

drodriguezhdez:
### What and why?

This PR adds a new command to upload SARIF reports.

### How?

The PR implementation follows the same pattern we're using for other upload commands like `junit upload`.

### Review checklist


src/commands/sarif/upload.ts
```diff
@@ -0,0 +1,199 @@
...
+
+const errorCodesStopUpload = [400, 403]
+
+const ajv = new Ajv()
+addFormats(ajv)
+
+const validateSarif = (sarifReportPath: string) => {
+  const sarifJsonSchemaValidate = ajv.compile(sarifJsonSchema)
+  try {
+    const sarifReportContent = JSON.parse(String(fs.readFileSync(sarifReportPath)))
```
juan-fernandez, line 40: I think this should work. Let me know if it doesn't:
```suggestion
    const sarifReportContent = require(sarifReportPath)
```

src/commands/sarif/upload.ts
```diff
@@ -0,0 +1,199 @@
...
+    const valid = sarifJsonSchemaValidate(sarifReportContent)
+    if (!valid) {
+      return sarifJsonSchemaValidate.errors
+    }
+  } catch (error) {
+    return error.message
+  }
```
juan-fernandez, line 47: you're returning `.errors` or `error.message`: are they of the same type?

juan-fernandez (commented):
looking pretty good! Left some comments

src/commands/sarif/__tests__/upload.test.ts
```diff
@@ -0,0 +1,263 @@
...
+      )
+
+      expect(files.length).toEqual(1)
+
+      expect(files[0]).toMatchObject({
+        service: 'service',
+        reportPath: './src/commands/sarif/__tests__/fixtures/valid-results.sarif',
+      })
+    })
+    test('should not fail for invalid single files', async () => {
```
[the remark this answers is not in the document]
drodriguezhdez: Yes, it's similar to the test we have for the `junit` command [(link)](https://github.com/DataDog/datadog-ci/blob/master/src/commands/junit/__tests__/upload.test.ts#L102).

I'd say it has to do with "do not fail" the command if a file does not exist

src/commands/sarif/upload.ts
```diff
@@ -0,0 +1,199 @@
+import fs from 'fs'
+import path from 'path'
+
+import Ajv from 'ajv'
+import addFormats from 'ajv-formats'
```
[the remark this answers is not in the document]
drodriguezhdez: Apparently, the library does not contain custom formats (such as `uri`) https://github.com/oasis-tcs/sarif-spec/blob/main/Documents/CommitteeSpecifications/2.1.0/sarif-schema-2.1.0.json#L10-L14

```
"$schema": {
      "description": "The URI of the JSON schema corresponding to the version.",
      "type": "string",
      "format": "uri"
    },
```

These formats are a separated packaged called [ajv-formats](https://github.com/ajv-validator/ajv-formats)
which includes a bunch of formats (including `uri`).

src/commands/sarif/upload.ts
drodriguezhdez: After the change, the tests are failing :(, so I guess it's not working

src/commands/sarif/upload.ts
```diff
@@ -0,0 +1,199 @@
...
+  renderRetriedUpload,
+  renderFailedUpload,
+  renderInvalidFile,
+} from './renderer'
+import {getBaseIntakeUrl} from './utils'
+
+const errorCodesStopUpload = [400, 403]
+
+const ajv = new Ajv()
+addFormats(ajv)
```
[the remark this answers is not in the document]
drodriguezhdez: Changed 👍🏼

src/commands/sarif/upload.ts
drodriguezhdez: Changed. Now it should return a `string`. However, not sure if that's the correct way to represent the array as `string`.

src/commands/sarif/upload.ts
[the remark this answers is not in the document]
juan-fernandez: got it, thanks

src/commands/sarif/upload.ts
juan-fernandez: yeah nvm, I thought `sarifReportPath` had a `.json` extension, but it's `.sarif`. The format should be json compatible but `require` won't work.

src/commands/sarif/api.ts
```diff
@@ -0,0 +1,75 @@
...
+  let uniqueFileName = `${fileName}-${sarifReport.service}-${metadata[GIT_SHA]}`
+
+  if (metadata[CI_PIPELINE_URL]) {
+    uniqueFileName = `${uniqueFileName}-${metadata[CI_PIPELINE_URL]}`
+  }
+  if (metadata[CI_JOB_URL]) {
+    uniqueFileName = `${uniqueFileName}-${metadata[CI_JOB_URL]}`
+  }
+
+  form.append('junit_xml_report_file', fs.createReadStream(sarifReport.reportPath).pipe(createGzip()), {
```
juan-fernandez, line 56: would it be this? 
```suggestion
  form.append('sarif_report_file', fs.createReadStream(sarifReport.reportPath).pipe(createGzip()), {
```

src/commands/sarif/api.ts
```diff
@@ -0,0 +1,75 @@
...
+    filename: `${getSafeFileName(uniqueFileName)}.xml.gz`,
+  })
+
+  return request({
+    data: form,
+    headers: form.getHeaders(),
+    maxBodyLength,
+    method: 'POST',
+    url: 'api/v2/cicodescan',
```
juan-fernandez, line 65: is this endpoint ready? If it isn't, we probably shouldn't publish this yet

src/commands/sarif/utils.ts
```diff
@@ -0,0 +1,7 @@
+export const getBaseIntakeUrl = () => {
+  if (process.env.DATADOG_SITE || process.env.DD_SITE) {
+    return `https://cicodescan-intake.${process.env.DATADOG_SITE || process.env.DD_SITE}`
```
juan-fernandez, line 3: is this intake ready?

juan-fernandez (commented):
Looks mostly ready 😄 . Left some minor comments

src/commands/sarif/__tests__/api.test.ts
```diff
@@ -0,0 +1,7 @@
+import {getSafeFileName} from '../api'
```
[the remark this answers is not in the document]
drodriguezhdez: Yes, refactored!

src/commands/sarif/api.ts
drodriguezhdez: good catch. Changed

src/commands/sarif/utils.ts
drodriguezhdez: The intake should be ready for only-internal usage

src/commands/sarif/api.ts
drodriguezhdez: The endpoint should be ready for only-internal usage

src/commands/junit/__tests__/api.test.ts
```diff
@@ -1,7 +0,0 @@
-import {getSafeFileName} from '../api'
-
-describe('getSafeFileName', () => {
-  test('filters unsafe characters out', () => {
-    expect(getSafeFileName('http://gitlab.com/-/pipelines/12345')).toEqual('http___gitlab_com___pipelines_12345')
```
juan-fernandez, line 5: thanks for this!

juan-fernandez (approved):
LGTM!

src/commands/sarif/README.md
```diff
@@ -0,0 +1,66 @@
...
+For example:
+
+```bash
+datadog-ci sarif upload --service my-service --tags key1:value1 --tags key2:value2 sarif-reports/go-reports sarif-reports/java-reports sarif-report/single-report.sarif
+```
+
+- The positional arguments are the directories or file paths in which the SARIF reports are located. If you pass a folder, the CLI will look for all `.sarif` files in it.
+- `--service` (default: `DD_SERVICE` env var) should be set as the name of the service you're uploading SARIF reports for.
+- `--tags` is a array of key value pairs of the shape `key:value`. This will set global tags applied to all results.
+  - The resulting dictionary will be merged with whatever is in the `DD_TAGS` environment variable. If a `key` appears both in `--tags` and `DD_TAGS`, whatever value is in `DD_TAGS` will take precedence.
```
urseberry, line 26:
```suggestion
- `--tags` is a array of key value pairs of the form `key:value`. This parameter sets global tags applied to all results. The upload process merges the tags passed on the command line with the tags in the `DD_TAGS` environment variable. If a key appears in both `--tags` and `DD_TAGS`, the value in `DD_TAGS` takes precedence.
```

src/commands/sarif/README.md
```diff
@@ -0,0 +1,66 @@
...
+- `--env` (default: `DD_ENV` env var) is a string that represents the environment where you want your tests to appear in.
```
urseberry, line 27:
```suggestion
- `--env` (default: `DD_ENV` env var) is a string that represents the environment in which you want your tests to appear.
```

src/commands/sarif/README.md
```diff
@@ -0,0 +1,66 @@
...
+- `--max-concurrency` (default: `20`): number of concurrent uploads to the API.
+- `--dry-run` (default: `false`): it will run the command without the final upload step. All other checks are performed.
+- `--no-verify` (default: `false`): it will run the command without performing reports validation on the CLI.
```
urseberry, line 30:
```suggestion
- `--dry-run` (default: `false`): runs the command without the final upload step. All other checks are performed.
- `--no-verify` (default: `false`): runs the command without performing report validation on the CLI.
```

src/commands/sarif/README.md
```diff
@@ -0,0 +1,66 @@
...
+
+#### Environment variables
+
+Additionally you might configure the `sarif` command with environment variables:
```
urseberry, line 34:
```suggestion
Additionally, you may configure the `sarif` command with environment variables:
```
Similarly to above, this list would also be easier to read if it were formatted as a table or description list.

urseberry (approved):

PR #781, opened 2023-02-15
Add new default values for commands ran from XCode

louiszawadzki:
### What and why?

We can now run the `xcode` command with `../node_modules/.bin/datadog-ci react-native xcode`, instead of `yarn datadog-ci react-native xcode`.
This changes the current working directory to be the `ios/` folder of the RN app instead of the root of the RN app.
We have a similar change for the dsyms command.

The default value for the `react-native` path should be in this case `../node_modules/react-native` instead of `node_modules/react-native`.
Similarly, the default path for the `datadog-ci.json` file should be `../datadog-ci.json` instead of `./datadog-ci.json`.

### How?

I've left the legacy values to be picked up first to avoid regressions.
I've only changed the default path for `datadog-ci.json` for the react native and dsyms upload commands.

### Review checklist

### DIFF
--- a/src/commands/flutter-symbols/upload.ts
+++ b/src/commands/flutter-symbols/upload.ts
@@ -10,7 +10,7 @@ import {newApiKeyValidator} from '../../helpers/apikey'
 import {getRepositoryData, RepositoryData} from '../../helpers/git/format-git-sourcemaps-data'
 import {getMetricsLogger, MetricsLogger} from '../../helpers/metrics'
 import {MultipartValue, UploadStatus} from '../../helpers/upload'
-import {buildPath, DEFAULT_CONFIG_PATH, performSubCommand, resolveConfigFromFile} from '../../helpers/utils'
+import {buildPath, DEFAULT_CONFIG_PATHS, performSubCommand, resolveConfigFromFile} from '../../helpers/utils'
 
 import * as dsyms from '../dsyms/upload'
 import {newSimpleGit} from '../git-metadata/git'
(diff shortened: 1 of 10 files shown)


0xnm (approved):

src/commands/flutter-symbols/upload.ts
```diff
@@ -119,7 +119,7 @@ export class UploadCommand extends Command {
 
     this.config = await resolveConfigFromFile(this.config, {
       configPath: this.configPath,
-      defaultConfigPath: DEFAULT_CONFIG_PATH,
+      defaultConfigPaths: ['datadog-ci.json'],
```
fuzzybinary, line 122: Why not use `DEFAULT_CONFIG_PATHS` here?

fuzzybinary (approved):

src/commands/flutter-symbols/upload.ts
louiszawadzki: I don't want to create any regression on other commands - if you're fine with `'datadog-ci.json'`, having `'../datadog-ci.json'` will only create regressions.
If you think that `DEFAULT_CONFIG_PATHS` should rather be `['datadog-ci.json']`, that also makes sense to me and I can change it

cswatt (approved):
👍 for docs

src/commands/flutter-symbols/upload.ts
lefebvree: I think it'd makes sense to keep `DEFAULT_CONFIG_PATHS = ['datadog-ci.json']`, use it everywhere and have `react-native` be the exception with `['datadog-ci.json', '../datadog-ci.json']`

lefebvree (approved):

duncanista (approved):
LGTM 👍🏽

PR #784, opened 2023-02-17
SYNTH-8640-adapt-report-summary

kalinichenko:
### What and why?

Amend the synthetics-test report so it shows up the possibility to speed up tests execution through parallelisation.

### DIFF
--- a/src/commands/synthetics/__tests__/reporters/default.test.ts
+++ b/src/commands/synthetics/__tests__/reporters/default.test.ts
@@ -191,6 +191,7 @@ describe('Default reporter', () => {
   describe('runEnd', () => {
     beforeEach(() => {
       writeMock.mockClear()
+      jest.useFakeTimers()
       reporter.reportStart({startTime: Date.now() - 567890}) // 9m 28s
     })
 
--- a/src/commands/synthetics/__tests__/reporters/default.test.ts
+++ b/src/commands/synthetics/__tests__/reporters/default.test.ts
@@ -191,6 +191,7 @@ describe('Default reporter', () => {
   describe('runEnd', () => {
     beforeEach(() => {
       writeMock.mockClear()
+      reporter.reportStart({startTime: Date.now() - 567890}) // 9m 28s
     })
 
     const baseSummary: Summary = getSummary()
@@ -229,6 +230,18 @@ describe('Default reporter', () => {
 
     test.each(cases)('$description', (testCase) => {
       reporter.runEnd(testCase.summary, MOCK_BASE_URL)
+      const mostRecentOutput = writeMock.mock.calls[writeMock.mock.calls.length - 1][0]
+      expect(mostRecentOutput).toMatchSnapshot()
+    })
+
+    const orgMaxConcurrencyCaps: {description: string; cap: number}[] = [
+      {cap: 0, description: 'communicates 0 parallelization'},
+      {cap: 1, description: 'communicates no (1 test at a time) parallelization'},
+      {cap: 2, description: 'communicates 2 tests parallelization'},
+    ]
+
+    test.each(orgMaxConcurrencyCaps)('$description', (testCase) => {
+      reporter.runEnd({...baseSummary, passed: 1}, MOCK_BASE_URL, {orgMaxConcurrencyCap: testCase.cap})
       const mostRecentOutput = writeMock.mock.calls[writeMock.mock.calls.length - 1][0]
       expect(mostRecentOutput).toMatchSnapshot()
     })
--- a/src/commands/synthetics/api.ts
+++ b/src/commands/synthetics/api.ts
@@ -16,6 +16,7 @@ import {
   ServerBatch,
   ServerTest,
   SyntheticsCIConfig,
+  SyntheticsSettings,
   TestSearchResult,
   Trigger,
 } from './interfaces'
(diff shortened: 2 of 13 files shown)


src/commands/synthetics/run-test.ts
```diff
@@ -155,8 +157,11 @@ export const executeTests = async (
       tunnel
     )
 
+    const {orgMaxConcurrencyCap} = await getSettings(api)
```
lefebvree, line 160: Why fetch it in `executeTests` if we're not using it? The function is already quite complex, we could move that to `execute` directly IMO and get rid of the mutable as well

src/commands/synthetics/utils.ts
```diff
@@ -796,6 +807,14 @@ export const renderResults = ({
 
   reporter.runEnd(summary, getAppBaseURL(config))
 
+  reporter.log(`Max parallelization configured: ${orgMaxConcurrencyCap} tests running at the same time\n\n`)
```
lefebvree, line 810: I think this would be better fitted in the reporters `runEnd` instead of calling `.log` directly from utils. Maybe adding a new `SyntheticsSettings` parameter to it for a non-breaking api change.

We should add a safety condition that `orgMaxConcurrencyCap` is defined and >0, especially since 0 is unlimited I believe and just in case something changes there.
Also may be we don't want to fail the whole run because we didn't manage to retrieve the settings, wdyt? 🤔

nit:
```suggestion
  reporter.log(`Max parallelization configured: ${orgMaxConcurrencyCap} test${orgMaxConcurrencyCap > 1 ? 's' : ''} running at the same time\n\n`)
```

src/commands/synthetics/reporters/default.ts
```diff
@@ -323,9 +323,10 @@ export class DefaultReporter implements MainReporter {
 
     if (summary.batchId) {
       const batchUrl = getBatchUrl(baseUrl, summary.batchId)
-      lines.push('View test batch in Continuous Testing Explorer: ' + chalk.dim.cyan(batchUrl))
+      lines.push('View full summary in Datadog: ' + chalk.dim.cyan(batchUrl))
     }
-    lines.push(`${b('Run summary:')} ${runSummary.join(', ')}${extraInfoStr}\n\n`)
+    lines.push(`\n${b('Continuous Testing Summary:')}`)
+    lines.push(`Test Results: ${runSummary.join(', ')}${extraInfoStr}\n`)
```
lefebvree, line 329: I think we also want to add the `Total Duration: ...` here, even if it is already displayed before results it would be useful to have it at the output end, next to the parallelization message.

src/commands/synthetics/api.ts
```diff
@@ -96,6 +97,17 @@ const searchTests = (request: (args: AxiosRequestConfig) => AxiosPromise<TestSea
   return resp.data
 }
 
+const getSettings = (request: (args: AxiosRequestConfig) => AxiosPromise<SyntheticsSettings>) => async () => {
```
lefebvree, line 100: nit: `getOrgSettings` or `getSyntheticsOrgSettings`

src/commands/synthetics/utils.ts
```diff
@@ -793,8 +808,8 @@ export const renderResults = ({
 
     reporter.resultEnd(result, getAppBaseURL(config))
   }
-
-  reporter.runEnd(summary, getAppBaseURL(config))
+  const settings = await getOrgSettings(getApiHelper(config), reporter)
```
lefebvree, line 811: I think we should move that to `execute` in `run-test.ts` so that `renderResults` only handles rendering

src/commands/synthetics/utils.ts
```diff
@@ -267,6 +268,20 @@ const getPollResultMap = async (api: APIHelper, batch: Batch) => {
...
+  reporter: MainReporter
+): Promise<SyntheticsOrgSettings | undefined> => {
+  let settings
+  try {
+    settings = await api.getSyntheticsOrgSettings()
+  } catch (e) {
+    reporter.error(`Failed to get settings: ${formatBackendErrors(e)}`)
+  }
+
+  return settings
```
lefebvree, line 282: nit
```suggestion
  try {
    return api.getSyntheticsOrgSettings()
  } catch (e) {
    reporter.error(`Failed to get Synthetics organisation settings: ${formatBackendErrors(e)}`)
  }
```

src/commands/synthetics/reporters/default.ts
```diff
@@ -323,9 +327,38 @@ export class DefaultReporter implements MainReporter {
...
+      const sec = Math.round((this.totalDuration % (60 * 1000)) / 1000)
+      lines.push(
+        `Total Duration:${min > 0 ? ' ' + min.toString() + 'm' : ''}${sec > 0 ? ' ' + sec.toString() + 's' : ''}`
+      )
+    }
+
+    if (
+      orgSettings &&
+      typeof orgSettings.orgMaxConcurrencyCap !== 'undefined' &&
+      orgSettings.orgMaxConcurrencyCap !== 1
```
lefebvree, line 354: To confirm with @peculiarism when we want to show this message but I think everytime `orgMaxConcurrencyCap > 0` ?

src/commands/synthetics/__tests__/reporters/default.test.ts
```diff
@@ -191,6 +191,7 @@ describe('Default reporter', () => {
   describe('runEnd', () => {
     beforeEach(() => {
       writeMock.mockClear()
+      reporter.reportStart({startTime: Date.now() - 567890}) // 9m 28s
```
lefebvree, line 194: Just to be safe we can add a `jest.useFakeTimers()` (nitpicky though, I think the test will alway evaluate under a second)
```suggestion
      jest.useFakeTimers()
      reporter.reportStart({startTime: Date.now() - 567890}) // 9m 28s
```

lefebvree (approved):
LGTM, just nits!

kayayarai (approved):

src/commands/synthetics/reporters/default.ts
kalinichenko: @peculiarism confirmed the message should also take place for the case with a cap of 1

PR #785, opened 2023-02-17
[package] Remove test files when published

Drarig29:
### What and why?

I noticed we were not excluding the test files when creating the `.tgz` file to publish on NPM. So they end up in the `node_modules` when datadog-ci is installed somewhere.

You can explore the files [here](https://www.npmjs.com/package/@datadog/datadog-ci/v/2.4.1?activeTab=explore).

### How?

Sadly, Yarn 1 (that we're using in this project) [doesn't fully support](https://github.com/yarnpkg/yarn/issues/685#issuecomment-484304385) `.npmignore` or blocklist rules in the `files` property of `package.json`.

So I had to use `rimraf` in the `prepack` script.

- Before:
  -  `yarn prepack && du -sh dist`: **3.2 MB**
- After:
  -  `yarn prepack && du -sh dist`: **1.8 MB**

### Review checklist


lefebvree (approved):
Nice find!

PR #787, opened 2023-02-23
[ci-visibility] Add `--metrics` and `DD_METRICS` options

juan-fernandez:
### What and why?

Some customers want to add numerical tags (metrics) to their JUnit uploads. They can't do so through `DD_TAGS` because every tag is assumed a string. 

By adding `--metrics` and `DD_METRICS`, customers are able to pass numerical tags to their tests. 

### How?

* Add `--metrics` option to `junit upload` command. 
* Add `DD_METRICS` env var.

### Review checklist


AdrianLC (approved):

drichards-87 (approved):

Drarig29 (approved):
nit: just a typo, LGMT otherwise

PR #790, opened 2023-02-24
Adapt git-metadata upload to synchronize local git repository.

lraucy:
### What and why?

This PR changes the command `datadog-ci git-metadata upload` to synchronize local git repository with the backend, when enabled.
The logic is the same as existing in the tracers libraries (see this initial [PR](https://github.com/DataDog/dd-trace-js/pull/2245) introducing it in dd-trace-js as well as [the current version of the code](https://github.com/DataDog/dd-trace-js/blob/9c54395/packages/dd-trace/src/ci-visibility/exporters/git/git_metadata.js)).

Before this PR, datadog-ci git-metadata upload sends the commit sha of HEAD as well as the list of corresponding tracked files.
This PR adds an option --git-sync to also synchronize the local git repository with the backend in a more extensive way, which is:
- check what are the local commits
- ask the backend which commits are known
- send packfiles containing the commits new to the backend (excluding file content, so the actual source code is NOT sent to the backend)

### How?

Rather than re-using the code from dd-trace-js, I decided to re-implement, despite the dd-trace-js compatibility. Many reasons led me to that decision:
- importing dd-trace-js from datadog-ci wasn't necessarily a good option
- Copying the code from dd-trace-js would have mean copying some dd-trace-js libraries code as well, and different libraries are available and used between dd-trace-js and datadog-ci. It seemed too much code and libraries to import in datadog-ci, while we already have everything we need in datadog-ci to implement the same behavior
- dd-trace-js logic is sometimes more complex than what can be done in datadog-ci, as it has to deal with the agent as a proxy. With datadog-ci we can expect to always have direct access to the backend API

### Review checklist


src/commands/git-metadata/gitdb.ts
```diff
@@ -0,0 +1,281 @@
...
+  const localCommitData = JSON.stringify({
+    meta: {
+      repository_url: repoURL,
+    },
+    data: latestCommits.map((commit) => ({
+      id: commit,
+      type: 'commit',
+    })),
+  })
+  const response = await runRequest(log, 'search_commits', () =>
```
edrevo, line 117: Should we check the status code of the response before trying to read the body?

src/commands/git-metadata/gitdb.ts
```diff
@@ -0,0 +1,281 @@
...
+    throw new Error(`Invalid commit format: ${sha} (different from sanitized ${sanitizedCommit})`)
+  }
+  if (!isValidSha(sanitizedCommit)) {
+    throw new Error(`Invalid commit format: ${sanitizedCommit}`)
+  }
+
+  return sanitizedCommit
+}
+
+const getCommitsToUpload = async (git: simpleGit.SimpleGit, commitsToExclude: string[]) => {
```
edrevo, line 156: This function will return ids for commits and trees, so I would use another name.

```suggestion
const getObjectsToUpload = async (git: simpleGit.SimpleGit, commitsToExclude: string[]) => {
```

src/commands/git-metadata/gitdb.ts
```diff
@@ -0,0 +1,281 @@
...
+  const rawResponse = await git.raw(
+    ['rev-list', '--objects', '--no-object-names', '--filter=blob:none', '--since="1 month ago"', 'HEAD'].concat(
+      commitsToExclude.map((sha) => '^' + sha)
+    )
+  )
+  const commitsToInclude = rawResponse.split('\n').filter((c) => c !== '')
+
+  return commitsToInclude
```
edrevo, line 164: You are bringing into memory all of the ids of all objects that need to be uploaded. That can potentially be megabytes worth of data. I'm not sure if the tracers are bringing this info to memory or if they are pipeline the `git rev-list` command to the `git pack-objects` command.

I am fine with this as a first approach, but I thought it would be good to mention that this can be memory intensive.

src/commands/git-metadata/gitdb.ts
```diff
@@ -0,0 +1,281 @@
...
+  form.append('pushedSha', pushedSha, {contentType: 'application/json'})
+  const packFileContent = fs.readFileSync(packfilePath)
+  // The original filename includes a random prefix, so we remove it here
+  const [, filename] = path.basename(packfilePath).split('-')
+  form.append('packfile', packFileContent, {
+    filename,
+    contentType: 'application/octet-stream',
+  })
+
+  return runRequest(log, 'pakfile', () =>
```
edrevo, line 256: Will typescript do the right thing here? Or will your function now return a `Promise<Promise<AxiosResponse<T>>>`?

edrevo (commented):
Did you take a look at https://github.com/DataDog/dd-trace-js/pull/2803 ? I think it would be good to have that in datadog-ci too, since many CI providers will only checkout the latest commit (without fetching any of the historial git metadata)

src/commands/git-metadata/gitdb.ts
lraucy: This is done by axios, which will throw an exception if the response is not in the 2xx range ([see here the default behavior](https://github.com/axios/axios/blob/v0.21.4/README.md?plain=1#L410) which is what is configured since we don't override this option)

src/commands/git-metadata/gitdb.ts
lraucy: My understanding is that at least the dd-trace-js implementation is loading it in memory as well, see [here](https://github.com/DataDog/dd-trace-js/blob/6763246af16adc2e00b292410241df6be14817db/packages/dd-trace/src/plugins/util/git.js#L50).

AFAIK this should take 40 bytes * (number of commits to upload + number of tree objects to upload). I expect it to be a few MBs max, since we're only sending data from the past month, but I might be wrong? If my expectations are right I'd expect it to be fine for most use cases?

Anyway it's indeed worth adding a comment.

edrevo: Yeah, I think it is fine leaving as is.

src/commands/git-metadata/gitdb.ts
lraucy: Yes, if what you meant is you'd expect this line to be `return await runRequest...`, eslint displays an error `Redundant use of await on a return value.`. The response here will be `Promise<AxiosResponse<T>>`. 

(actually by inspecting the type my editor says it's `Promise<AxiosResponse<AxiosResponse>>` due to a "mistake" in the definition of the type `RequestBuilder`, I just fixed than in another commit).

edrevo: Magic!

lraucy:
> Did you take a look at [DataDog/dd-trace-js#2803](https://github.com/DataDog/dd-trace-js/pull/2803) ? I think it would be good to have that in datadog-ci too, since many CI providers will only checkout the latest commit (without fetching any of the historial git metadata)

Ah I didn't see that, thanks for pointing that out, I will take a look.

src/commands/git-metadata/gitdb.ts
```diff
@@ -0,0 +1,284 @@
...
+    ['rev-list', '--objects', '--no-object-names', '--filter=blob:none', '--since="1 month ago"', 'HEAD'].concat(
+      commitsToExclude.map((sha) => '^' + sha)
+    )
+  )
+  const commitsToInclude = rawResponse.split('\n').filter((c) => c !== '')
+
+  return commitsToInclude
+}
+
+// need to transform into async here.
```
juan-fernandez, line 169: is this something you were planning to do this PR or later on?

juan-fernandez (approved):
I only have a couple of nitpicks. Otherwise LGTM :). Great job!

src/commands/git-metadata/gitdb.ts
lraucy: Discussed offline, I wanted to have this command async but in the end that wouldn't change anything since all steps are sequentials here. I just removed the comment.

src/commands/git-metadata/gitdb.ts
```diff
@@ -0,0 +1,297 @@
...
+}
+
+const unshallowRepositoryWhenNeeded = async (log: Logger, git: simpleGit.SimpleGit) => {
+  const isShallow = (await git.revparse('--is-shallow-repository')) === 'true'
+  if (isShallow) {
+    log.info('[unshallow] Git repository is a shallow clone., unshallowing it...')
+    log.info('[unshallow] Setting remote.origin.partialclonefilter to "blob:none" to avoid fetching file content')
+    await git.addConfig('remote.origin.partialclonefilter', 'blob:none')
+    log.info('[unshallow] Running git fetch --shallow-since="1 month ago" --update-shallow --refetch')
+    await git.fetch(['--shallow-since="1 month ago"', '--update-shallow', '--refetch'])
```
rpelliard, line 108: Do you know if `1 month ago` is always 30d, or if we're in March we could end up with only 28d of data ? Would it be safer to put `30 days ago` here ?

src/commands/git-metadata/gitdb.ts
```diff
@@ -0,0 +1,297 @@
...
+  }
+
+  const generatePackfiles = (packfilePath: string) => {
+    const packObjectResults = child_process
+      .execSync(`git pack-objects --compression=9 --max-pack-size=3m ${packfilePath}`, {
+        input: commits.join('\n'),
+      })
+      .toString()
+      .split('\n')
+      .filter((sha) => sha)
```
sashacmc, line 195: To be honest not a clear what this filter does? Removes empty lines?

rpelliard (approved):

src/commands/git-metadata/gitdb.ts
```diff
@@ -0,0 +1,297 @@
...
+      .map((sha) => `${packfilePath}-${sha}.pack`)
+
+    return packObjectResults
+  }
+
+  // Try using tmp folder first:
+  try {
+    const tmpFolder = os.tmpdir()
```
sashacmc, line 203: Why we generate some random prefix (and don't check possible collisions) instead of using something like mkdtemp?

src/commands/git-metadata/gitdb.ts
```diff
@@ -0,0 +1,297 @@
...
+  try {
+    await uploadPackfiles(log, request, repoURL, latestCommits[0], packfiles)
+    log.debug('Successfully uploaded packfiles.')
+  } catch (err) {
+    log.warn(`Failed to upload packfiles: ${err}`)
+    throw err
+  }
+}
+
+const getRepoURL = gitRemote
```
sashacmc, line 93: Just curious, why we need it?

src/commands/git-metadata/gitdb.ts
```diff
@@ -0,0 +1,297 @@
...
+
+const getLatestLocalCommits = async (git: simpleGit.SimpleGit) => {
+  const logResult = await git.log(['-n 1000', '--since="1 month ago"'])
```
sashacmc, line 96: Why we have this magic numbers there?

sashacmc (changes requested):

src/commands/git-metadata/gitdb.ts
```diff
@@ -0,0 +1,281 @@
...
+    if (c.type !== 'commit' || c.id === undefined) {
+      throw new Error('Invalid commit type response')
+    }
+
+    return sanitizeCommit(c.id)
+  })
+}
+
+const sanitizeCommit = (sha: string) => {
+  const isValidSha = (s: string) => /[0-9a-f]{40}/.test(s)
```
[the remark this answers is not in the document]
lraucy: I don't think we need it soon, but anyway that's not a big change to also support 64 chars to be future-proof, I'll make the change.

src/commands/git-metadata/gitdb.ts
lraucy: I'm pretty sure the answer is we'll have only 28d of data. I don't think that's too bad as we expect this limit to be useful only when uploading the repo for the first time, and having 28d or 30d doesn't really change much. I will keep the current behavior as this is also what is implemented in the tracer. cc @edrevo FYI

src/commands/git-metadata/gitdb.ts
```diff
@@ -0,0 +1,297 @@
...
+
+  const sanitizedCommit = sha.replace(/[^0-9a-f]+/g, '')
+  if (sanitizedCommit !== sha) {
```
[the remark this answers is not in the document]
lraucy: I took that from the dd-trace-js implementation, but looking at the code I don't think there is any good reason.
For simplification I'll do two changes:
- replace the regex `/[0-9a-f]{40}/` by `/^[0-9a-f]{40}$/` to test exact match
- remove the sanitization part and only test against the regex above

cc @juan-fernandez fyi in case I'm missing something

src/commands/git-metadata/gitdb.ts
lraucy: will replace with:
```
import {gitRemote as getRepoURL} from './git'
```

src/commands/git-metadata/gitdb.ts
lraucy: Moved to a constant.

src/commands/git-metadata/gitdb.ts
lraucy: This is based on the specs written for all tracers. mkdtemp might be more relevant indeed but the existing implem seems enough for our use case: collisions are still not very likely and wouldn't be a big issue. Most of the time these files will be in a tmpdir anyway, and for the cases where it's in cwd: I don't expect collision to have an impact except for making troubleshooting harder. We're talking about an edge case of an edge case here.

src/commands/git-metadata/gitdb.ts
lraucy: Yes. I changed the filter to make it more clear.

sashacmc: Ok, but what about other lines?
I tried to run `git pack-objects --compression=9 --max-pack-size=3m`
and I got something like
```
Enumerating objects: 1, done.
Counting objects: 100% (1/1), done.
ec27d75473692ce16b631211cf50995703274499
Writing objects: 100% (1/1), done.
```
maybe it depends on the git setting, but I don't see how other lines filtered there.

src/commands/git-metadata/gitdb.ts
sashacmc: Could you please give a link to this specs

src/commands/git-metadata/gitdb.ts
lraucy: Lines 1, 2 and 4 are written to stderr while line 3 is written to stdout. The command `execSync` only returns stdout.

sashacmc: Ok! Thanks!

src/commands/git-metadata/gitdb.ts
lraucy: @sashacmc I reworked this to use `mkdtemp` as well as adding a manual deletion step of the temp folder after the command ran (redundant with the OS in case we're in tmpdir but useful when we are in cwddir).

src/commands/git-metadata/gitdb.ts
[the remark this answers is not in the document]
juan-fernandez: yeah good catch. I don't remember why I did it like this, but it's wrong 😅 . Not only because of the unnecessary replacement and test, but because the regex should indeed be `/^[0-9a-f]{40}$/`. I'll update it in `dd-trace-js`, thanks

src/commands/git-metadata/gitdb.ts
```diff
@@ -0,0 +1,308 @@
...
+}
+
+const generatePackFilesForCommits = async (log: Logger, commits: string[]): Promise<[string[], string | undefined]> => {
+  if (commits.length === 0) {
+    return [[], undefined]
+  }
+
+  const generatePackfiles = async (baseTmpPath: string): Promise<[string[], string | undefined]> => {
+    const randomPrefix = String(Math.floor(Math.random() * 10000))
+    const tmpPath = await mkdtemp(path.join(baseTmpPath, 'dd-packfiles-'))
```
sashacmc, line 204: Ok, but please note that mkdtemp already generate random prefix, so we don't need do it manually

sashacmc (approved):
See note about mkdtemp, everything also LGTM

src/commands/git-metadata/gitdb.ts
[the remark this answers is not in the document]
juan-fernandez: ok I now remember why I had this 😅 . It's because of this: 
https://github.com/DataDog/dd-trace-js/pull/2876/checks?check_run_id=11845280052 

```
Uncontrolled command line
Critical
This command line depends on a user provided value
```

It's kind of silly because if it's anything other than a SHA it will not go through, but by sanitizing the alert goes away.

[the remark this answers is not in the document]
juan-fernandez: so I'll keep the sanitize in `dd-trace-js` but I'll update the regex anyway 👍

src/commands/git-metadata/gitdb.ts
lraucy: mkdtemp generates a suffix in the directory name, while the randomPrefix here is used in each file name. At this point I'm not sure the random prefix in each individual filename is really necessary, but I'm blindly following the spec implemented in all tracers here to avoid too much rework, as it doesn't seem to hurt and we can easily revisit later.

PR #791, opened 2023-02-27
Upgrade to Yarn 3

NoortsDD:
https://datadoghq.atlassian.net/browse/SYNTH-3299

### What and why?

- Upgrade Yarn from v1 to v3.4.1

### How?

- Following the migration guide: https://yarnpkg.com/getting-started/migration

### Review checklist

- [ ] TODO...


.github/workflows/ci.yml
```diff
@@ -146,4 +146,7 @@ jobs:
         uses: actions/setup-node@v3
         with:
           node-version: '14'
+      # The install step has been added here such that the `.yarn/install-state.gz` file is generated. This file is used
+      # by the script `check-licenses` below.
+      - run: yarn install --immutable
       - run: yarn check-licenses
```
NoortsDD, line 152: Running `yarn check-licenses` locally will succeed because it generates the state file by itself. This however does not seem to happen in the CI, this is why this `yarn install` was added.

bin/make-it-executable.js
```diff
@@ -3,10 +3,20 @@ const path = require('path');
...
+/**
+ * The forEach loop below requires an object with script paths as properties to loop over. When the `bin` property in
+ * the package.json only has a single property, then the object is converted to a string upon a `yarn install` call
+ * (see: https://docs.npmjs.com/cli/v9/configuring-npm/package-json#bin). The string to object transformation here thus
+ * allows us to support 1+ properties whilst keeping the loop below.
+ */
+if (typeof scriptPaths === 'string') {
+  scriptPaths = { "datadog-ci": scriptPaths }
+}
+
```
NoortsDD, line 19: This change was "required" because of the adjustment made to the `bin` property in the `package.json`. I'm curious to know if a different approach would be preferred (e.g.: ignoring the change made to the `package.json` after a `yarn install`).

This approach at least allows us to expand the `scriptPaths` defined in the `bin` property later, from 1 to 2+ paths, as the loop to handle multiple items will still be there.

.github/workflows/ci.yml
NoortsDD: This could potentially be replaced by a Python version of the license checker (e.g. in the worker repo).

lefebvree (approved):
LGTM!

PR #796, opened 2023-03-02
CI env vars: add support for TeamCity

albertvaka:
### What and why?

When running in TeamCity CI, read the job info from the env vars.

### Review checklist


juan-fernandez (commented):
code looks good but it looks as though the generated tags include `"_dd.ci.env_vars": "{}",`, which the fixtures do not include. I guess we could just not add `"_dd.ci.env_vars": "{}",` for certain providers

albertvaka:
We don't need it for teamcity since we won't support correlating tests and pipelines, so I've removed it from the code. This should now be ready to review.

lefebvree (approved):

juan-fernandez (approved):

PR #803, opened 2023-03-21
[React Native] [Flutter] [dSYMs] Improve API key resolution

louiszawadzki:
### What and why?

The current behaviour for resolving API keys is counter-intuitive for flutter/react-native/dsyms command, as we overwrite the environment API key by the `datadog-ci.json` file.

This PR inverts this resolution logic, to make it easier to temporarily override the API key and match other commands.

We also display a message in API key is defined in 2 ways, including the last 4 digits of the key to make debug in these cases easier.

### How?

Added a function to apply the env after the config file.

### Review checklist

### DIFF
--- a/src/commands/flutter-symbols/upload.ts
+++ b/src/commands/flutter-symbols/upload.ts
@@ -124,12 +124,12 @@ export class UploadCommand extends Command {
     this.config = await resolveConfigFromFileAndEnvironment(
       this.config,
       {
-        configPath: this.configPath,
-        defaultConfigPaths: DEFAULT_CONFIG_PATHS,
-      },
-      {
         apiKey: process.env.DATADOG_API_KEY,
         datadogSite: process.env.DATADOG_SITE,
+      },
+      {
+        configPath: this.configPath,
+        defaultConfigPaths: DEFAULT_CONFIG_PATHS,
       }
     )
 
--- a/src/commands/flutter-symbols/upload.ts
+++ b/src/commands/flutter-symbols/upload.ts
@@ -16,6 +16,7 @@ import {
   performSubCommand,
   resolveConfigFromFileAndEnvironment,
 } from '../../helpers/utils'
+import {checkAPIKeyOverride} from '../../helpers/validation'
 
 import * as dsyms from '../dsyms/upload'
 import {newSimpleGit} from '../git-metadata/git'
@@ -34,7 +35,6 @@ import {
   renderArgumentMissingError,
   renderCommandInfo,
   renderCommandSummary,
-  renderDuplicateAPIKey,
   renderFailedUpload,
   renderGeneralizedError,
   renderGitWarning,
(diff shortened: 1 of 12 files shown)


src/commands/react-native/__tests__/upload.test.ts
```diff
@@ -144,6 +156,32 @@ describe('execute', () => {
...
       version: '1.23.4',
     })
+    expect(apiKeyValidatorSpy).toHaveBeenCalledWith({
+      apiKey: '12345678900987654321aabbccddeeff',
+      datadogSite: expect.anything(),
+      metricsLogger: expect.anything(),
+    })
+  })
+
+  test('uses API Key from env over config from JSON file', async () => {
```
fuzzybinary, line 166: Should this test maybe be added to the other two platforms?

src/commands/flutter-symbols/upload.ts
```diff
@@ -117,10 +122,26 @@ export class UploadCommand extends Command {
...
+    this.config = await resolveConfigFromFileAndEnvironment(
+      this.config,
+      {
+        apiKey: process.env.DATADOG_API_KEY,
+        datadogSite: process.env.DATADOG_SITE,
+      },
+      {
+        configPath: this.configPath,
+        defaultConfigPaths: DEFAULT_CONFIG_PATHS,
+        configFromFileCallback: (configFromFile: any) => {
```
fuzzybinary, line 134: I feel weird having this callback duplicated in 3 places, as well as the error message. Is there any clean way to consolidate them into the main function?

louiszawadzki: I'm open to merging these 3 occurences, but I would need some help on how to keep the current separation of concerns.

For each command, the output messages are formatted by a file called `renderer.ts`, and the reference to the `stdout` is kept in the `execute` method. 
I'm not really into passing either the `renderDuplicateAPIKey` or `stdout` to the `resolveConfigFromFileAndEnvironment` util function.

Speaking of the `resolveConfigFromFileAndEnvironment` function, I did not feel right about adding some logic related to api keys into it. If we want to change the env variable name or the config file key for some reason, we might easily forget to change it in this function.

So the best solution I found was to pass a callback that receives the resolved config from the config file, and let each command handle the duplication detection and print to stdout.

It can be argued that we can put the duplication detection inside `resolveConfigFromFileAndEnvironment` though, but I'm still afraid this will be hidden and not reflected in the name. 

What would you recommend?

louiszawadzki: I've put something in the `validation` helper that seemed adequate, let me know what you think

fuzzybinary: Yeah, I think that's adequate for now. At some point we may want to think about seeing if we can take parameter and environment resolution higher up in the tool, so it's always consistent across all of the commands, but I think that would be major version worthy....

fuzzybinary (approved):

lefebvree (approved):
LGTM, agreed with the comments, we should aim to unify that across all commands for consistency at some point

Issue #798, opened 2023-03-08
AWS v3 Upgrade


robfig:
### Feature request description

datadog-ci is the last remaining dependency of ours that is on AWS v2 SDK. That dependency alone pulls down 86MB of code into node_modules and prints a deprecation warning about reaching EOL in a couple months.

### Solution

Upgrade to AWS SDK v3. They have codemods to help and copious upgrade instructions.

### Command

None

PR #807, opened 2023-03-24
[lambda] upgrade `aws-sdk` to v3

duncanista:
### What and why?

Upgrades the `aws-sdk` to v3.

Users are slowly migrating to the v3, since it provides more benefits, like package decrease. The v2 contained all the packages, even for services we're not using. 

#798 

### How?

Several changes were made to the `lambda` commands, since this package is now modularized. A 1:1 update was not done, since it would haven't improved performance at all.

Upgraded fixtures and tests too.

### Review checklist


sfirrin (approved):
Looks great :+1:

PR #808, opened 2023-03-24
[synthetics] Do incremental MD5 calculation in `uploadMobileApplications`

kevin-pietsch:
### What and why?

Avoid calculating MD5 hash on entire file at once as this caused issues with large files in similar front-end code.

### How?
Use `fs.ReadStream` to incrementally calculate MD5 and avoid doing it on the entire file at once.

However, the upload itself *still* reads the entire file into memory using `fs.readFile` in order to set `Content-Length` header in the upload request. This will need to be revisited to handle uploads larger than 2GB which are disallowed by
the `fs.readFile` API. One solution may be to use multi-part upload in combination with `fs.ReadStream` when doing
the upload.

### Review checklist

- [ X] Feature or bugfix MUST have appropriate tests (unit, integration)
  - Associated unit test modified and passing
  - Also manually tested with sample mini app upload

### DIFF
--- a/src/commands/synthetics/mobile.ts
+++ b/src/commands/synthetics/mobile.ts
@@ -4,13 +4,6 @@ import fs from 'fs'
 import {APIHelper} from './api'
 import {Test, TestPayload, UserConfigOverride} from './interfaces'
 
-/**
- * Create an md5 hash of a file at a given path. This is done incrementally so
- * entire file does not need to fit in memory.
-
- * @param file path to the file that will be hashed
- * @returns md5 hash of file as string
- */
 export const getMD5HashFromFile = async (file: string): Promise<string> => {
   const hash = crypto.createHash('md5')
   const input = fs.createReadStream(file)
--- a/src/commands/synthetics/mobile.ts
+++ b/src/commands/synthetics/mobile.ts
@@ -1,13 +1,24 @@
 import * as crypto from 'crypto'
-import {promises as fs} from 'fs'
+import fs from 'fs'
 
 import {APIHelper} from './api'
 import {Test, TestPayload, UserConfigOverride} from './interfaces'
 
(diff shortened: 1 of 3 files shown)


src/commands/synthetics/mobile.ts
```diff
@@ -1,27 +1,31 @@
...
 ): Promise<string> => {
-  const fileBuffer = await fs.readFile(applicationPathToUpload)
-  const md5 = await getMD5HashFromFileBuffer(fileBuffer)
+  const md5 = await getMD5HashFromFile(applicationPathToUpload)
   const {presigned_url_params: presignedUrl, file_name: fileName} = await api.getMobileApplicationPresignedURL(
     mobileApplicationId,
     md5
   )
 
+  const fileBuffer = await fs.promises.readFile(applicationPathToUpload)
```
KevinYeramian, line 28: Let's use the stream from `fs.createReadStream` in `uploadMobileApplication` instead of the `Buffer` to not manipulate the file with two different formats. It should work out of the box according to the [axios doc](https://axios-http.com/docs/req_config#:~:text=only%3A%20Stream%2C%20Buffer-,data,-%3A%20%7B%0A%20%20%20%20firstName).

xiaoPierre (approved):
LGTM 👍

src/commands/synthetics/mobile.ts
nmizoguchi: Wdyt we create a follow-up card for this? We can move forward with the Buffer implementation for now!

kevin-pietsch: Created new issue here:
https://datadoghq.atlassian.net/browse/SYNTH-9246

nmizoguchi: Perfect! Thank you!

nmizoguchi (approved):

PR #818, opened 2023-04-03
Add `--mobileApplicationVersionFilePath` option

Drarig29:
💡This PR can be reviewed commit by commit

---

### What and why?

Currently, we need to use a test file to configure this option:
```json
// myTest.synthetics.json
{
  "tests": [
    {
      "id": "aaa-aaa-aaa",
      "config": {
        "mobileApplicationVersionFilePath": "path/to/application.apk"
      }
    }
  ]
}
```

To ease this, this PR adds a `--mobileApplicationVersionFilePath` option which will be a global override. This can still be overridden for specific tests thanks to a test file.

### How?

Because this argument is passed as a CLI option, it should affect all tests (unless there is a more precise override for specific tests).

So the option is added in `config.global` (which is a `UserConfigOverride`), and applied to all tests by:

https://github.com/DataDog/datadog-ci/blob/ed341f6b18b872f1d45d8a597e9bff708b824e2e/src/commands/synthetics/run-test.ts#L214-L220

### Review checklist

### DIFF
--- a/src/commands/synthetics/__tests__/cli.test.ts
+++ b/src/commands/synthetics/__tests__/cli.test.ts
@@ -1,4 +1,4 @@
-import {Cli} from 'clipanion/lib/advanced'
+import {BaseContext, Cli} from 'clipanion/lib/advanced'
 
 import * as ciUtils from '../../../helpers/utils'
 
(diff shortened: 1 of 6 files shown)


src/commands/synthetics/command.ts
```diff
@@ -265,6 +267,7 @@ RunTestCommand.addOption('failOnMissingTests', Command.Boolean('--failOnMissingT
 RunTestCommand.addOption('failOnTimeout', Command.Boolean('--failOnTimeout'))
 RunTestCommand.addOption('files', Command.Array('-f,--files'))
 RunTestCommand.addOption('jUnitReport', Command.String('-j,--jUnitReport'))
+RunTestCommand.addOption('mobileApplicationVersionFilePath', Command.String('--mobileApplicationVersionFilePath'))
```
Drarig29, line 270: Should we add an option shortcut? All the single letters switches I can think of are already used, so it could be just a shorter option name...

What about `--mobileApplication`?

nmizoguchi: Maybe even shorter like `--mobileApp`

src/commands/synthetics/utils.ts
```diff
@@ -573,9 +577,18 @@ export const getTestsToTrigger = async (
     )
   )
 
+  // Keep track of uploaded applications to avoid uploading them twice.
```
etnbrd, line 580: How all of these changes are related to the added flag?
Why remove the `&& overridenConfig` here?

src/commands/synthetics/__tests__/cli.test.ts
```diff
@@ -113,8 +122,10 @@ describe('run-test', () => {
       command['configPath'] = overrideCLI.configPath
       command['datadogSite'] = overrideCLI.datadogSite
       command['failOnCriticalErrors'] = overrideCLI.failOnCriticalErrors
+      command['failOnMissingTests'] = overrideCLI.failOnMissingTests
```
etnbrd, line 125: Should this be part of the same PR? Isn't it part of the ongoing refactoring for error handling?

src/commands/synthetics/__tests__/cli.test.ts
```diff
@@ -164,6 +179,91 @@ describe('run-test', () => {
       })
     })
 
+    test('override from global < ENV < test file', async () => {
```
etnbrd, line 182: I'm not sure to follow how this test works, so maybe we could add a few more comments?
Also the test just above seems to have a title going the other way: `override from config file < ENV < CLI`

src/commands/synthetics/__tests__/cli.test.ts
Drarig29: No it's not. It's just that this test was incomplete, and I didn't want to create a PR just to update tests.

src/commands/synthetics/__tests__/cli.test.ts
Drarig29: The test just above doesn't test the same code path.

- `test('override from config file < ENV < CLI')` is testing that CLI arguments have precedence over what's configured in the "config file" (e.g. `datadog-ci.json`)
- `test('override from global < ENV < test file')` is testing that what's configured in "test files" (e.g. tests.synthetics.json) has precedence over what's configured in the [`global` property](https://github.com/DataDog/datadog-ci/blob/bd0a8536ab8ab0bbf0bbd2d0136c944f3eba898b/.github/workflows/e2e/global.config.json#L5) of the "config file".

etnbrd: After discussing offline, we concluded the best course of action was to:
- explain clearly the 6 different cases of overrides from config file to ENV to CLI to test file
- refactor the testing of these overrides in the test below (`override locations with ENV variable`) ...
- ... to later split it into smaller tests.

src/commands/synthetics/__tests__/cli.test.ts
```diff
@@ -153,127 +186,119 @@ describe('run-test', () => {
...
-        content: {tests: [{config: {}, id: 'publicId'}]},
-        name: 'Suite 1',
-      }
-
-      jest.spyOn(ciUtils, 'resolveConfigFromFile').mockImplementation(async (config, _) => config)
-      jest.spyOn(utils, 'getSuites').mockImplementation((() => [conf]) as any)
-
-      // Throw to stop the test
-      const triggerTests = jest.fn(() => {
-        throw getAxiosHttpError(502, {message: 'Bad Gateway'})
```
etnbrd, line 190: My understanding is that this was used to prevent actually triggering the tests. What's implemented to prevent triggering the tests? Switching from `triggerTests` to `getTestsToTrigger`?

src/commands/synthetics/__tests__/cli.test.ts
```diff
@@ -153,127 +186,119 @@ describe('run-test', () => {
...
 
+      const expectTestWithConfig = (
+        config: Partial<UserConfigOverride>
+      ): Parameters<typeof utils['getTestsToTrigger']> => [
+        (apiHelper as unknown) as api.APIHelper,
+        [{suite: 'Suite 1', id: 'publicId', config}],
+        expect.anything(),
+        false,
+        false,
+        false,
```
etnbrd, line 217: Maybe these lines are worth a small comment?

etnbrd (approved):
Thanks for taking into account and refactoring these tests 🙇 
I left a bunch of comments, but going to approve that anyway.

Issue #663, opened 2022-10-11
Unable to override `failOnMissingTests` through config file when using Github Action


kevcao-certik:
**Describe what happened**
Using the Datadog Synthetics CI GitHub Action, I am unable to set `failOnMissingTests` to `false` and the workflow still fails when no tests are found.

**Steps to reproduce the issue:**
- Create a simple GitHub `.yml` workflow file with the following contents:
```yml
name: Run Datadog Synthetics Tests
  run-tests:
    runs-on: ubuntu-latest
    steps:
      - name: Generate Synthetics Config
        run: |
          echo '{ 
            "failOnMissingTests": false,
            "global": {,
              "startUrl": "<START-URL-HERE>",
              "locations": ["aws:us-east-1"]
            }
          }' > synthetics.json
      - name: Run Datadog Synthetics Tests
        uses: DataDog/synthetics-ci-github-action@v0.6.0
        with:
          api_key: ${{ secrets.DD_API_KEY }}
          app_key: ${{ secrets.DD_APP_KEY }}
          test_search_query: 'tag:<non-existent-tag>'
          config_path: "./synthetics.json"
```

**Expected behaviour:**
No tests should be run, but workflow should not fail.

**Actual behaviour:**
Workflow fails with error:
```
| No test to run.
[Run Datadog Synthetics Tests/run-tests]   ❗  ::error::Running Datadog Synthetics tests failed.
[Run Datadog Synthetics Tests/run-tests]   ❌  Failure - Main Run Datadog Synthetics Tests
[Run Datadog Synthetics Tests/run-tests] exitcode '1': failure
[Run Datadog Synthetics Tests/run-tests] 🏁  Job failed
```

Drarig29:
Hey @kevcao-certik!

I'm closing this issue as it should be fixed in [version 0.11.0](https://github.com/DataDog/synthetics-ci-github-action/releases/tag/v0.11.0) 😄

PR #822, opened 2023-04-06
[synthetics] Export new utils for error handling

Drarig29:
### What and why?

Currently, the error handling in our CI integrations isn't aligned with what datadog-ci does (#663).

This PR refactors datadog-ci to extract and export new utils that will be used in our CI integrations.

Draft PRs on the CI integrations to use this new API:
- https://github.com/DataDog/synthetics-ci-github-action/pull/122
- https://github.com/DataDog/datadog-ci-azure-devops/pull/46

### How?

This PR creates 4 new utils:
- `reportCiError()` will allow us to remove files like [src/report-ci-error.ts (github action)](https://github.com/DataDog/synthetics-ci-github-action/blob/main/src/report-ci-error.ts).
- `reportExitLogs()` reports all logs that are common between failed and successful runs.
- `getExitReason()` replaces code that does manual error handling, like in [here](https://github.com/DataDog/synthetics-ci-github-action/blob/345aa9092cc2b009d51650d6b6ea59b0ed30c76c/src/main.ts#L28-L33) or [here](https://github.com/DataDog/synthetics-ci-github-action/blob/345aa9092cc2b009d51650d6b6ea59b0ed30c76c/src/main.ts#L16-L26).
- `toExitCode()` is a helper that is used for CLI usages only.

In the CI integrations, we had code like:

```ts
if (
  resultSummary.criticalErrors > 0 ||
  resultSummary.failed > 0 ||                                  
  (resultSummary.timedOut > 0 && config.failOnTimeout) ||      
  resultSummary.testsNotFound.size > 0                         
) {
  core.setFailed(`Datadog Synthetics tests failed: ${printSummary(resultSummary, config)}`)
}
```

But `getExitReason()` [uses `getResultOutcome()`](https://github.com/DataDog/datadog-ci/blob/b8116d049ffe4a2d1c1025eec7911c13b9d2283b/src/commands/synthetics/utils.ts#L813), which uses `result.passed`:

https://github.com/DataDog/datadog-ci/blob/b8116d049ffe4a2d1c1025eec7911c13b9d2283b/src/commands/synthetics/interfaces.ts#L91-L92

So, the following conditions will be checked implicitly by `getExitReason()`:
- `resultSummary.criticalErrors > 0` (which **wasn't aligned** with `datadog-ci` because it was ignoring `failOnCriticalErrors`)
- `resultSummary.timedOut > 0 && config.failOnTimeout`
- `resultSummary.failed > 0`

And `getExitReason()` can replace `resultSummary.testsNotFound.size > 0` because `executeTests()` will throw `MISSING_TESTS` if `failOnMissingTests` is enabled.

With this PR, the **non critical error** `NO_TESTS_TO_RUN` will never make the CI fail. There is **no configuration option** that changes this behavior and that's expected (we have a [unit test for that](https://github.com/DataDog/datadog-ci/blob/b8116d049ffe4a2d1c1025eec7911c13b9d2283b/src/commands/synthetics/__tests__/cli.test.ts#L280-L296)). This will solve #663.

### Review checklist

### DIFF
--- a/src/commands/synthetics/utils.ts
+++ b/src/commands/synthetics/utils.ts
@@ -829,55 +829,61 @@ export const renderResults = ({
   reporter.runEnd(summary, getAppBaseURL(config), orgSettings)
 }
 
-export const handleExit = (
+export const reportExitLogs = (
   reporter: MainReporter,
-  config: Pick<CommandConfig, 'failOnTimeout'>,
-  results: Result[]
-): 0 | 1 => {
-  if (!config.failOnTimeout && results.some((result) => result.timedOut)) {
+  config: Pick<CommandConfig, 'failOnTimeout' | 'failOnCriticalErrors'>,
+  {results, error}: {results?: Result[]; error?: unknown}
+) => {
+  if (!config.failOnTimeout && results?.some((result) => result.timedOut)) {
     reporter.error(
       chalk.yellow(
-        'Because `failOnTimeout` is disabled, the command will exit with an error code 0. ' +
-          'Use `failOnTimeout: true` to exit with an error code 1.\n'
+        'Because `failOnTimeout` is disabled, the command will succeed. ' +
+          'Use `failOnTimeout: true` to make it fail instead.\n'
       )
     )
   }
 
-  const hasFailedTests = results.some((result) => getResultOutcome(result) === ResultOutcome.Failed)
-  if (hasFailedTests) {
-    return 1
+  if (!config.failOnCriticalErrors && error instanceof CriticalError) {
+    reporter.error(
+      chalk.yellow(
+        'Because `failOnCriticalErrors` is not set or disabled, the command will succeed. ' +
+          'Use `failOnCriticalErrors: true` to make it fail instead.\n'
+      )
+    )
   }
 
-  return 0
-}
-
-export const handleExitOnError = (
-  reporter: MainReporter,
-  config: Pick<CommandConfig, 'failOnCriticalErrors' | 'failOnMissingTests'>,
-  error: unknown
-): 0 | 1 => {
   if (error instanceof CiError) {
     reportCiError(error, reporter)
+  }
+}
 
+export const getExitReason = (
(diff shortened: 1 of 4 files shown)


src/commands/synthetics/utils.ts
```diff
@@ -802,14 +800,63 @@ export const renderResults = ({
       summary.failedNonBlocking++
     } else {
       summary.failed++
-      hasSucceeded = false
```
Drarig29, line 805: This `else` translates to `if (resultOutcome === ResultOutcome.Failed)`.

This "hasSucceeded = false" was moved to `handleExit()` [below](https://github.com/DataDog/datadog-ci/pull/822/files#diff-741a985551714cec0c6f7a6423d704525619e8f57db6c4f7c685d95d88a8d0cfR825-R828).

src/commands/synthetics/utils.ts
```diff
@@ -269,11 +269,13 @@ const getPollResultMap = async (api: APIHelper, batch: Batch) => {
 }
 
 export const getOrgSettings = async (
-  api: APIHelper,
```
Drarig29, line 272: We don't export the `api` module so it's not possible to create an instance `APIHelper` externally.
So we need to create the instance _in the helper_, as we already do in `fetchTest()`.

src/commands/synthetics/utils.ts
```diff
@@ -331,6 +333,7 @@ const getResultFromBatch = (
   }
 }
 
+// XXX: We shouldn't export functions that take an `APIHelper` because the `utils` module is exported while `api` is not.
```
Drarig29, line 336: I'll address the `XXX`'s in another PR.

src/commands/synthetics/utils.ts
etnbrd: Is this comment still valid? I don't see the `hasFailedTests` in the diff.

etnbrd (approved):
LGTM 👍

Drarig29 (commented):
Just a few notes to confirm there are no breaking changes.

PR #836, opened 2023-04-13
Junit upload: also upload git metadata

albertvaka:
### What and why?

Upload git metadata to gitdb, like done in the `git-metadata upload` command, when using the `junit upload` command.

Screenshot 2023-04-16 at 10 36 09

### How?

Reuses the `uploadToGitDB` function. Adds a couple arguments to control the new behavior.

### Review checklist

### DIFF
--- a/src/commands/junit/__tests__/upload.test.ts
+++ b/src/commands/junit/__tests__/upload.test.ts
@@ -326,7 +326,10 @@ describe('execute', () => {
   })
 
   test('without git metadata', async () => {
-    const {context, code} = await runCLI(['--skip-git-metadata-upload', process.cwd() + '/src/commands/junit/__tests__/fixtures/single_file.xml'])
+    const {context, code} = await runCLI([
+      '--skip-git-metadata-upload',
+      process.cwd() + '/src/commands/junit/__tests__/fixtures/single_file.xml',
+    ])
     const output = context.stdout.toString().split(os.EOL)
     expect(code).toBe(0)
     expect(output[5]).not.toContain('Syncing git metadata')


src/commands/junit/upload.ts
```diff
@@ -264,3 +290,5 @@ UploadJUnitXMLCommand.addOption('basePaths', Command.Rest({required: 1}))
 UploadJUnitXMLCommand.addOption('maxConcurrency', Command.String('--max-concurrency'))
 UploadJUnitXMLCommand.addOption('logs', Command.Boolean('--logs'))
 UploadJUnitXMLCommand.addOption('rawXPathTags', Command.Array('--xpath-tag'))
+UploadJUnitXMLCommand.addOption('skipGitMetadataUpload', Command.Boolean('--skip-git-metadata-upload'))
+UploadJUnitXMLCommand.addOption('gitRepositoryURL', Command.String('--git-repository-url'))
```
juan-fernandez, line 294: is this field now mandatory? Can't we read it from the git metadata extraction step?

juan-fernandez (approved):
nice! Do we need to update any public documentation? At least to add the new fields

src/commands/junit/upload.ts
juan-fernandez: is that logic in this PR? As far as I can tell, `this.gitRepositoryURL` is always being used when calling `uploadToGitDB`. I might be missing something though

src/commands/junit/api.ts
```diff
@@ -18,6 +18,10 @@ import {renderUpload} from './renderer'
 // We don't want any hard limit enforced by the CLI, the backend will enforce a max size by returning 413 errors.
 const maxBodyLength = Infinity
 
+export const datadogSite = process.env.DATADOG_SITE || process.env.DD_SITE || 'datadoghq.com'
+export const intakeUrl = `https://cireport-intake.${datadogSite}`
+export const apiUrl = 'https://api.' + datadogSite
```
juan-fernandez, line 23: for consistency 
```suggestion
export const apiUrl = `https://api.${datadogSite}`
```

juan-fernandez (commented):
LGTM. I only have a doubt about `gitRepositoryURL`

juan-fernandez:
also, linter is failing: https://github.com/DataDog/datadog-ci/actions/runs/4715427011/jobs/8362411019?pr=836. It looks like a code formatting issue

src/commands/junit/upload.ts
juan-fernandez: oh got it, thanks

juan-fernandez (approved):

PR #837, opened 2023-04-13
[ci] Improve discoverability of the lint SARIF report

Drarig29:
### What and why?

We can't have 2 eslint [formatters](https://eslint.org/docs/latest/use/formatters) at the same time, so we can't have both eslint's usual text output _and_ a SARIF report (that we use to create automated comments on PRs).

### How?

We already had a CI job to upload the SARIF report as an artifact, but it was not run when the lint command was failing.
So this PR changes the `upload-artifact` job to always run.

And since the SARIF report is only used as part of the CI, I moved the report creation in a separate yarn script named `lint:ci`. Now when running `yarn lint` locally, we should get eslint's usual text output.

Since knowing where to find artifacts (and that they exist) is not obvious, the lint job will now show you a link to easily find the pipeline artifact. (see example in [this purposely failing job](https://github.com/DataDog/datadog-ci/actions/runs/4689177258/jobs/8310600228))

### Review checklist

- [ ] ~~Feature or bugfix MUST have appropriate tests (unit, integration)~~


albertvaka (commented):
Could we print the contents of the sarif file to the console after generating it, so we don't have to download it manually?

albertvaka (approved):
In any case, approving because this is already better than what we had before.

Drarig29:
@albertvaka awesome idea!

Here is the result ([see job](https://github.com/DataDog/datadog-ci/actions/runs/4690102902/jobs/8312750590?pr=837)):

image

PR #840, opened 2023-04-13
[synthetics] Reword the description of `startUrlSubstitutionRegex`

Drarig29:
### What and why?

The description was a bit off, since the `s/your_regex/your_substitution/modifiers` syntax doesn't really support more complex transformations.

And the name "The JavaScript regex syntax" is a bit off as well since this substitution syntax is custom. It's not supported by JavaScript.

### Review checklist

- [ ] ~~Feature or bugfix MUST have appropriate tests (unit, integration)~~


hestonhoffman (approved):
LGTM!

PR #841, opened 2023-04-13
[lambda] update how AWS credentials are handled

duncanista:
### What and why?

AWS CloudShell already provides credentials. These are not being used by `datadog-ci` since we only check for environment variables. Therefore, an update of how we get AWS credentials from environment is required.

### How?

Instead of checking just for credentials in environment variables, we are using the function `fromNodeProviderChain` from `@aws-sdk/credential-provider` which is a chain that checks in multiple places for AWS credentials.

### Review checklist


joeyzhao2018 (approved):
LGTM

PR #842, opened 2023-04-14
[React Native] Fix react native upload command by adding --bundle option back

louiszawadzki:
### What and why?

Re-add bundle name metadata in react native sourcemaps upload.
Without bundle name, the upload fails.

### How?

We have to keep the option for bundle not to be defined to keep retro-compatibility. If it is not defined, we use the default value for bundle and print a warning.

### Review checklist

Works again for iOS: 
image

And for Android (with this additional PR https://github.com/DataDog/dd-sdk-reactnative/pull/450):
image


src/commands/react-native/README.md
```diff
@@ -46,6 +46,7 @@ datadog-ci react-native upload --platform android --service com.company.app --bu
...
 | `--platform`        | Required  | Identifies if you are uploading iOS or Android source maps.                                                                                                                                                                                                                                                                                                     |
 | `--service`         | Required  | Set as the service name you are uploading source maps for. Datadog uses this service name to find corresponding source maps based on the `service` tag set on the RUM React Native SDK.<br>By default, the RUM React Native SDK uses your application's bundle identifier as the service.                                                                       |
+| `--bundle`          | Required  | Should be set as the path to your generated JS bundle file, `main.jsbundle` for iOS and `index.android.bundle` for Android.                                                                                                                                                                                                                                     |
```
buraizu, line 49: Could you confirm if this should be `Should` or `Must` as outlined in the RFC followed by the docs team?
https://github.com/DataDog/documentation/blob/master/CONTRIBUTING.md#rfc-2119
https://datatracker.ietf.org/doc/html/rfc2119

buraizu, line 49: Could you confirm if this should be `Should` or `Must` as outlined in the RFC followed by the docs team?
https://github.com/DataDog/documentation/blob/master/CONTRIBUTING.md#rfc-2119
https://datatracker.ietf.org/doc/html/rfc2119

src/commands/react-native/README.md
```diff
@@ -116,6 +117,7 @@ This command calls the `upload` command, setting the release version to `{releas
...
 | `--platform`   | Required  | Identifies if you are uploading iOS or Android source maps.                                                                                                                                                                                                                               |
 | `--service`    | Required  | Set as the service name you are uploading source maps for. Datadog uses this service name to find corresponding source maps based on the `service` tag set on the RUM React Native SDK.<br>By default, the RUM React Native SDK uses your application's bundle identifier as the service. |
+| `--bundle`     | Required  | Should be set as the path to your generated JS bundle file, `main.jsbundle` for iOS and `index.android.bundle` for Android.                                                                                                                                                               |
```
buraizu, line 120: As above, could you confirm if this should be `Must`?

buraizu (approved):
LGTM overall, but flagged a couple of places to confirm if `should` is the correct term per [RFC2119](https://datatracker.ietf.org/doc/html/rfc2119)

0xnm (approved):

src/commands/react-native/README.md
louiszawadzki: 👍 Good point, I updated the docs :)

PR #849, opened 2023-04-20
[React Native] Support environment variables in Info.plist for react-native xcode

louiszawadzki:
### What and why?

On the default template for React Native, the Info.plist file uses environment variables for the version and build number.

### How?

We look if the value in the Info.plist is an env variable. 
If that's the case, we look for the environment variable or use the default values.

Screenshot of a working example:
image

image

### Review checklist

### DIFF
--- a/src/helpers/__tests__/plist.test.ts
+++ b/src/helpers/__tests__/plist.test.ts
@@ -25,7 +25,7 @@ describe('plist util', () => {
     it('returns an empty string for an env variable that is not declared', () => {
       const plist = parsePlist('src/helpers/__tests__/plist-fixtures/Info.plist')
       expect(() => plist.getPropertyValue('CFBundleExecutable')).toThrowError(
-        "Environment variable EXECUTABLE_NAME for property value $(EXECUTABLE_NAME) wasn't found."
+        "Environment variable $(EXECUTABLE_NAME) for key CFBundleExecutable wasn't found."
       )
     })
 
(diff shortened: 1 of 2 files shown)


src/helpers/__tests__/plist.test.ts
```diff
@@ -13,6 +16,19 @@ describe('plist util', () => {
...
+    it('returns the value for an env variable', () => {
+      process.env.EXECUTABLE_NAME = 'executable name'
+      const plist = parsePlist('src/helpers/__tests__/plist-fixtures/Info.plist')
+      expect(plist.getPropertyValue('CFBundleExecutable')).toBe('executable name')
+    })
+
+    it('returns an empty string for an env variable that is not declared', () => {
+      const plist = parsePlist('src/helpers/__tests__/plist-fixtures/Info.plist')
+      expect(() => plist.getPropertyValue('CFBundleExecutable')).toThrowError(
+        "Environment variable EXECUTABLE_NAME for property value $(EXECUTABLE_NAME) wasn't found."
```
Drarig29, line 28: nit: Wouldn't it be more helpful to say something like this instead?

> Environment variable `$(EXECUTABLE_NAME)` for property `CFBundleExecutable` wasn't found.

IMO, saying both "EXECUTABLE_NAME" and "$(EXECUTABLE_NAME)" in the message is redundant, since you [`return propertyValue`](https://github.com/DataDog/datadog-ci/pull/849/files#diff-7d620898cbcf1cd44bc614f8927b751be084e36d4cb1bf8dd8dd1867e793a497R15) if the regex doesn't have a match.

Drarig29 (approved):
Just one nit, LGTM otherwise!

src/helpers/__tests__/plist.test.ts
louiszawadzki: Totally agree, that's much better 👍

PR #852, opened 2023-04-25
[git-metadata] Make git sync as default behavior.

lraucy:
### What and why?

Git sync has been introduced in a previous version as an optional step. After more thorough testing it seems to work fine so the idea is to now release that as a default. There is still the option to disable it through --no-git-sync, just in case this new sync would fail the command in some setups.

Additionally I kept the `--git-sync` option for backward compatibility, and ignore any error for the gov DC as git sync is not available there yet.

### How?

### Review checklist

### DIFF
--- a/src/commands/git-metadata/upload.ts
+++ b/src/commands/git-metadata/upload.ts
@@ -42,7 +42,7 @@ export class UploadCommand extends Command {
   }
   private dryRun = false
   private verbose = false
-  private gitSync = false
+  private noGitSync = false
   private directory = ''
   private logger: Logger = new Logger((s: string) => {
     this.context.stdout.write(s)
@@ -108,7 +108,7 @@ export class UploadCommand extends Command {
       inError = true
     }
 
-    if (this.gitSync) {
+    if (!this.noGitSync) {
       try {
         this.logger.info('Syncing GitDB...')
         const elapsed = await timedExecAsync(this.uploadToGitDB.bind(this), {
@@ -213,5 +213,5 @@ UploadCommand.addPath('git-metadata', 'upload')
 UploadCommand.addOption('dryRun', Command.Boolean('--dry-run'))
 UploadCommand.addOption('verbose', Command.Boolean('--verbose'))
 UploadCommand.addOption('directory', Command.String('--directory'))
-UploadCommand.addOption('gitSync', Command.Boolean('--git-sync'))
+UploadCommand.addOption('noGitSync', Command.Boolean('--no-git-sync'))
 UploadCommand.addOption('repositoryURL', Command.String('--repository-url'))
--- a/src/commands/git-metadata/upload.ts
+++ b/src/commands/git-metadata/upload.ts
@@ -42,6 +42,7 @@ export class UploadCommand extends Command {
   }
   private dryRun = false
   private verbose = false
+  private gitSync = false
   private noGitSync = false
   private directory = ''
   private logger: Logger = new Logger((s: string) => {
@@ -77,6 +78,10 @@ export class UploadCommand extends Command {
       )
 
       return 1
+    }
+
+    if (this.gitSync) {
+      this.logger.warn('option --git-sync is deprecated as it is now the default behavior')
     }
 
     const metricsLogger = getMetricsLogger({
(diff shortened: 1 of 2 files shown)


src/commands/git-metadata/upload.ts
```diff
@@ -213,5 +213,5 @@ UploadCommand.addPath('git-metadata', 'upload')
 UploadCommand.addOption('dryRun', Command.Boolean('--dry-run'))
 UploadCommand.addOption('verbose', Command.Boolean('--verbose'))
 UploadCommand.addOption('directory', Command.String('--directory'))
-UploadCommand.addOption('gitSync', Command.Boolean('--git-sync'))
```
Letiste, line 216: What happens for the users using this argument? Would it break?

src/commands/git-metadata/upload.ts
```diff
@@ -213,5 +213,5 @@ UploadCommand.addPath('git-metadata', 'upload')
...
+UploadCommand.addOption('noGitSync', Command.Boolean('--no-git-sync'))
```
sashacmc, line 216: Just a clarification, shouldn't we think about backwards compatibility? What if customers are already using this command line option, and after we remove it their pipeline breaks because datadog-ci failed on unknown option? Do we have some kind of policy on this?

sashacmc (approved):

src/commands/git-metadata/upload.ts
lraucy: very good point, I'll need to add the support for this arg back. Thanks!

src/commands/git-metadata/upload.ts
lraucy: Yes indeed, thanks! I'll fix it.

Letiste (approved):

rpelliard (approved):

sashacmc (approved):

src/commands/git-metadata/upload.ts
```diff
@@ -117,8 +125,13 @@ export class UploadCommand extends Command {
...
         this.logger.info(`${this.dryRun ? '[DRYRUN] ' : ''}Successfully synced git DB in ${elapsed} seconds.`)
       } catch (err) {
-        console.log('error writing to git db')
-        this.logger.warn(`Could not write to GitDB: ${err}`)
+        if (!this.isTargetingGov()) {
+          console.log('error writing to git db')
+          this.logger.warn(`Could not write to GitDB: ${err}`)
+        } else {
+          // Skip the warning for Gov DC since git sync is not available there yet.
+          this.logger.warn(`not writing to GitDB: not available for gov`)
```
Drarig29, line 133: nit: You are mixing sentences starting with uppercases and lowercases
```suggestion
          this.logger.warn(`Not writing to GitDB: not available for gov`)
```

src/commands/git-metadata/upload.ts
```diff
@@ -79,6 +83,10 @@ export class UploadCommand extends Command {
       return 1
     }
 
+    if (this.gitSync) {
+      this.logger.warn('option --git-sync is deprecated as it is now the default behavior')
```
Drarig29, line 87: Same uppercase/lowercase nit

src/commands/git-metadata/upload.ts
```diff
@@ -117,8 +125,13 @@ export class UploadCommand extends Command {
         metricsLogger.logger.increment('gitdb.success', 1)
         this.logger.info(`${this.dryRun ? '[DRYRUN] ' : ''}Successfully synced git DB in ${elapsed} seconds.`)
       } catch (err) {
-        console.log('error writing to git db')
-        this.logger.warn(`Could not write to GitDB: ${err}`)
+        if (!this.isTargetingGov()) {
+          console.log('error writing to git db')
+          this.logger.warn(`Could not write to GitDB: ${err}`)
```
Drarig29, line 130: Why both a `console.log` and a `this.logger.warn`?

Drarig29: (I understand it was there before, but maybe it was an oversight?)

lraucy: Good catch it looks like something I missed in a previous PR, thanks

Drarig29 (approved):

PR #857, opened 2023-05-03
s8s(mobile): Compute and forward application size to getMobileApplicationPresignedURL

KevinYeramian:
### What and why?

- Compute and forward application size to `getMobileApplicationPresignedURL`. The endpoint was changed to verify the app size and send back a pre-signed URL or an error message with context. 
- Add try-catch around `getMobileApplicationPresignedURL` and `uploadMobileApplication`.

### How?

Changing back-end code

### Review checklist

### DIFF
--- a/src/commands/synthetics/mobile.ts
+++ b/src/commands/synthetics/mobile.ts
@@ -4,7 +4,7 @@ import fs from 'fs'
 import {APIHelper, EndpointError, formatBackendErrors} from './api'
 import {PresignedUrlResponse, Test, TestPayload, UserConfigOverride} from './interfaces'
 
-export const getFileAndMD5HashFromFile = async (filePath: string): Promise<{appSize: number; md5: string}> => {
+export const getSizeAndMD5HashFromFile = async (filePath: string): Promise<{appSize: number; md5: string}> => {
   const hash = crypto.createHash('md5')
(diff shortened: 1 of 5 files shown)


src/commands/synthetics/mobile.ts
```diff
@@ -1,34 +1,41 @@
 import * as crypto from 'crypto'
 import fs from 'fs'
 
-import {APIHelper} from './api'
-import {Test, TestPayload, UserConfigOverride} from './interfaces'
+import {APIHelper, EndpointError, formatBackendErrors} from './api'
+import {PresignedUrlResponse, Test, TestPayload, UserConfigOverride} from './interfaces'
 
-export const getMD5HashFromFile = async (file: string): Promise<string> => {
+export const getFileAndMD5HashFromFile = async (filePath: string): Promise<{appSize: number; md5: string}> => {
```
ypetit, line 7: Shouldn't this function be called something like `getSizeAndMD5HashFromFile` as you get the size of the file and not the file itself ?

KevinYeramian: I completely miss it, completely agree 👍

ClaraJacintho (approved):
LGTM bar the nit!

PR #858, opened 2023-05-03
[lambda] Add python 3.10 instrumentation support

sfirrin:
### What and why?
Adds support for python 3.10

### How?
Updates constants and adds test

### Review checklist

### DIFF
--- a/src/commands/lambda/__tests__/functions/instrument.part2.test.ts
+++ b/src/commands/lambda/__tests__/functions/instrument.part2.test.ts
@@ -55,6 +55,44 @@ describe('instrument', () => {
       `)
     })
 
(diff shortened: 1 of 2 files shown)


duncanista (approved):
LGTM!

joeyzhao2018 (changes requested):
Thanks for the PR. I think we also need to add python3.10 [here](https://github.com/DataDog/datadog-ci/blob/master/src/commands/lambda/constants.ts#L41) and [here](https://github.com/DataDog/datadog-ci/blob/master/src/commands/lambda/constants.ts#L48)

joeyzhao2018 (approved):

PR #862, opened 2023-05-08
[lambda] Only change handler path when a layer version is specified

sfirrin:
### What and why?
We found a bug where the handler path would change incorrectlly when no layer is set

### How?
Checks that a layer version is set before changing the Handler of a function

### Review checklist


duncanista (approved):
LGTM – just update tests to pass!

PR #869, opened 2023-05-15
Upload JS sourcemaps with GZIP encoding

0xnm:
### What and why?

This small PR makes upload of JS source maps to use `gzip` encoding, as it was done previously for Flutter and React Native source maps uploads in https://github.com/DataDog/datadog-ci/pull/801.

Note: I didn't test it E2E (that upload is working and mapping file can be processed for the error), I will appreciate if @DataDog/rum-browser does this check.

### Review checklist


bcaudan (approved):
Tested on browser-sdk repo

rpelliard (approved):

PR #874, opened 2023-05-16
Show validation errors on SARIF command

drodriguezhdez:
### What and why?

This PR fixes the logic to show the validation errors when the `sarif upload` command is used.

### How?

Join strings composed by `instancePath` and `message` to build the final message.

### Review checklist


juan-fernandez (commented):
can we add a test for this?

PR #884, opened 2023-05-19
[lambda] deprecate `ruby2.5`

duncanista:
### What and why?

Removes support for `ruby2.5` instrumentation
This runtime has been deprecated by AWS (phase 2).

### How?

Removing it from the runtime lookups.

### Review checklist

### DIFF
--- a/src/commands/lambda/constants.ts
+++ b/src/commands/lambda/constants.ts
@@ -44,7 +44,6 @@ export const RUNTIME_LOOKUP = {
   'python3.8': RuntimeType.PYTHON,
   'python3.9': RuntimeType.PYTHON,
   'python3.10': RuntimeType.PYTHON,
-  'ruby2.5': RuntimeType.RUBY,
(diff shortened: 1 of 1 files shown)


zARODz11z (approved):
LGTM, seems like we're 1 day on time for this. Curious how you knew this was a needed change, were you just some how aware of the depreciation of these runtimes? I guess you probably check https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html more often than me.

duncanista:
> LGTM, seems like we're 1 day on time for this. Curious how you knew this was a needed change, were you just some how aware of the depreciation of these runtimes? I guess you probably check https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html more often than me.

We are 2 months late actually! But we sometimes leave some room after Phase 2.
And, it was just tech debt, we had this for a while, but having a lot of instrumentation tools, we sometimes miss this deprecations!

PR #886, opened 2023-05-24
fixed format of gitlab url

bgorriz:
### What and why?
With this change we do not reformat the GITLAB pipeline url.

Before we were removing a part from the url that should not be removed.

### How?

### Review checklist


juan-fernandez (approved):
LGTM

etnbrd (commented):
LGTM 👍 Let's release that with a bump of the minor version at least.
Out of curiosity, why removing the tests data?

PR #889, opened 2023-05-25
Adding check on required tags for gate evaluation

andrea-mosk:
### What and why?

Adding a check on required tags for gate evaluation. When the command is executed, we should be able to retrieve branch, pipeline name and repository URL. If we can't retrieve them, by default the command fails, and clients can override this behavior by specifying a flag.

### Review checklist

### DIFF
--- a/src/commands/gate/renderer.ts
+++ b/src/commands/gate/renderer.ts
@@ -95,6 +95,6 @@ export const renderMissingRequiredTag = (tag: string): string => {
 export const renderMissingTagsError = (): string => {
   // TODO include link to documentation
   return chalk.red(`Could not retrieve all the required information needed to evaluate the rules. 
-Use the "--allow-partial-evaluation" option in order to evaluate results only based on current information. 
+Use the "--allow-partial-evaluation" option to evaluate results based only on current information. 
 See the documentation for more details.`)
 }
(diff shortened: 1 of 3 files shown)


liashenko (approved):

src/commands/gate/renderer.ts
```diff
@@ -87,3 +87,14 @@ export const renderGateEvaluationError = (error: any): string => {
...
 }
+
+export const renderMissingRequiredTag = (tag: string): string => {
+  return chalk.red(`ERROR: the following information could not be retrieved: "${tag}".\n`)
+}
+
+export const renderMissingTagsError = (): string => {
+  // TODO include link to documentation
+  return chalk.red(`Could not retrieve all the required information needed to evaluate the rules. 
+Use the "--allow-partial-evaluation" option in order to evaluate results only based on current information.
```
juan-fernandez, line 98: nitpick
```suggestion
Use the "--allow-partial-evaluation" option to evaluate results based only on current information. 
```

juan-fernandez (approved):

andrea-mosk:
I'm closing this one as we decided to remove the pipeline name from the scope. The approach will now become:

1. `datadog-ci gate` will always send all the information
2. In the service we will check the information received. If commit, branch or repo is not there we will return 400 or something similar

PR #885, opened 2023-05-22
Changing gate command exit status based on gate evaluation result

andrea-mosk:
### What and why?

Changing the result exit status based on the gate evaluation result:

- Passed -> successful command
- Failed -> failed command
- Empty -> successful command unless `--fail-on-empty` is specified

### Review checklist

- [x] Feature or bugfix MUST have appropriate tests (unit, integration)

PR #879, opened 2023-05-18
DOCS-5436 Configuration Edits

alai97:
### What and why?

Add an Example Test File link.

### How?

DOCS-5436

### Preview

https://docs-staging.datadoghq.com/alai97/continuous-testing-anchor-link-cleanup/continuous_testing/cicd_integrations/configuration

PR #878, opened 2023-05-17
Logging useful information based on gate evaluation response

andrea-mosk:
### What and why?

Based on gate evaluation response returned, we log important information to customers. 

Example of logs:

Screenshot 2023-05-17 at 12 55 49

### Review checklist

- [x] Feature or bugfix MUST have appropriate tests (unit, integration)

PR #875, opened 2023-05-16
Initial implementation of the gate evaluation command

andrea-mosk:
### What and why?

Initial implementation of a new "gate evaluation" command.

### Review checklist

- [x] Feature or bugfix MUST have appropriate tests (unit, integration)

PR #890, opened 2023-05-25
v2.12.0

Drarig29:
Releasing:

- #888
- #800 
- #887
- #886
- #885
- #878
- #875
- #881
- #879
- #882

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@datadog/datadog-ci",
-  "version": "2.11.0",
+  "version": "2.12.0",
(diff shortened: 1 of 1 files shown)


juan-fernandez (approved):
LGTM.

For the release notes, could we add a note for this PR: https://github.com/DataDog/datadog-ci/pull/886 saying something like:
> Pipelines URLs for gitlab will change now that legacy URLs are removed (see https://docs.gitlab.com/ee/update/deprecations.html#legacy-urls-replaced-or-removed for more context)

PR #893, opened 2023-05-26
[test] Fix flaky standalone binary test on macOS

Drarig29:
### What and why?

The previous attempt (#838) to fixing this ended up working for a limited time. Even after "warming up" the binary with a `beforeAll`, the test is [still flaky](https://github.com/DataDog/datadog-ci/actions/runs/5071931748/jobs/9108968694).

### How?

Let's replace the "warming up" - which does an extra execution of the binary without a lot of benefits - by an increase of the timeout to 10 seconds (for macOS only).

### Review checklist


juan-fernandez (approved):
thanks for this 👍

PR #895, opened 2023-05-30
Add support for user defined scopes and tags

andrea-mosk:
### What and why?

Users will be able to provide:

- Scope, via the `--scope` option. They can provide multiple values for the same key (e.g. `--scope=team:backend --scope=team:frontend`). Scope will be used to further filter the rules that are being evaluated.
- Tags, via the `--tags` option. These will work exactly as the other commands.

### Review checklist

### DIFF
--- a/src/commands/gate/__tests__/scope.test.ts
+++ b/src/commands/gate/__tests__/scope.test.ts
@@ -18,4 +18,10 @@ describe('parseScope', () => {
       team: ['backend', 'frontend'],
     })
   })
+  test('should remove duplicated values', () => {
+    expect(parseScope(['key1:value1', 'team:backend', 'team:backend'])).toEqual({
+      key1: ['value1'],
+      team: ['backend'],
+    })
+  })
 })
--- a/src/commands/gate/__tests__/scope.test.ts
+++ b/src/commands/gate/__tests__/scope.test.ts
@@ -1,0 +1,21 @@
+import {parseScope} from '../utils'
+
+describe('parseScope', () => {
+  test('falls back to empty object if invalid format', () => {
+    expect(parseScope([''])).toEqual({})
+    expect(parseScope(['not.correct.format'])).toEqual({})
+    expect(parseScope(['not.correct.format,either'])).toEqual({})
+  })
+  test('returns an object with the scope variables with well formatted strings', () => {
+    expect(parseScope(['key1:value1', 'key2:value2'])).toEqual({key1: ['value1'], key2: ['value2']})
+  })
+  test('should not include invalid key:value pairs', () => {
+    expect(parseScope(['key1:value1', 'key2:value2', 'invalidkeyvalue'])).toEqual({key1: ['value1'], key2: ['value2']})
+  })
+  test('should merge values for the same key', () => {
+    expect(parseScope(['key1:value1', 'team:backend', 'team:frontend'])).toEqual({
+      key1: ['value1'],
+      team: ['backend', 'frontend'],
+    })
+  })
+})
--- a/src/commands/gate/evaluate.ts
+++ b/src/commands/gate/evaluate.ts
@@ -4,6 +4,7 @@ import {Command} from 'clipanion'
 import {getCISpanTags} from '../../helpers/ci'
 import {getGitMetadata} from '../../helpers/git/format-git-span-data'
 import {SpanTags} from '../../helpers/interfaces'
+import {parseTags} from '../../helpers/tags'
 import {getUserGitSpanTags} from '../../helpers/user-provided-git'
 
 import {apiConstructor} from './api'
@@ -14,7 +15,7 @@ import {
   renderGateEvaluationInput,
   renderGateEvaluationError,
 } from './renderer'
-import {getBaseIntakeUrl} from './utils'
+import {getBaseIntakeUrl, parseScope} from './utils'
 
 export class GateEvaluateCommand extends Command {
   // TODO add usage
@@ -22,16 +23,22 @@ export class GateEvaluateCommand extends Command {
   private config = {
     apiKey: process.env.DATADOG_API_KEY || process.env.DD_API_KEY,
     appKey: process.env.DATADOG_APP_KEY,
+    envVarTags: process.env.DD_TAGS,
   }
 
   private dryRun = false
   private failOnEmpty = false
+  private userScope?: string[]
+  private tags?: string[]
 
   public async execute() {
     const api = this.getApiHelper()
     const spanTags = await this.getSpanTags()
+    const userScope = this.userScope ? parseScope(this.userScope) : {}
+
     const payload = {
       spanTags,
+      userScope,
     }
 
     return this.evaluateRules(api, payload)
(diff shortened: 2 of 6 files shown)


src/commands/gate/utils.ts
```diff
@@ -5,3 +5,31 @@ export const getBaseIntakeUrl = () => {
...
+      }
+
+      const firstColon = keyValuePair.indexOf(':')
+      const key = keyValuePair.substring(0, firstColon)
+      const value = keyValuePair.substring(firstColon + 1)
+
+      if (acc.hasOwnProperty(key)) {
+        acc[key].push(value)
+      } else {
+        acc[key] = [value]
```
juan-fernandez, line 27: ❓ do you always want lists? Even for single values? Because there's the option to do: 

```javascript
if (acc.hasOwnProperty(key)) {
  if (acc[key].length) {
    acc[key].push(value)
  } else {
    acc[key] = [acc[key], value]
  }
}
```
it's not the prettiest but maybe it's easier to handle in the backend?

src/commands/gate/__tests__/scope.test.ts
```diff
@@ -0,0 +1,21 @@
...
+    expect(parseScope(['not.correct.format,either'])).toEqual({})
+  })
+  test('returns an object with the scope variables with well formatted strings', () => {
+    expect(parseScope(['key1:value1', 'key2:value2'])).toEqual({key1: ['value1'], key2: ['value2']})
+  })
+  test('should not include invalid key:value pairs', () => {
+    expect(parseScope(['key1:value1', 'key2:value2', 'invalidkeyvalue'])).toEqual({key1: ['value1'], key2: ['value2']})
+  })
+  test('should merge values for the same key', () => {
+    expect(parseScope(['key1:value1', 'team:backend', 'team:frontend'])).toEqual({
```
juan-fernandez, line 16: if the user specifies the same key:value pair twice, should the value be repeated in the resulting list? 
```javascript
expect(parseScope(['team:backend', 'team:backend']))
```

src/commands/gate/__tests__/scope.test.ts
```diff
@@ -0,0 +1,21 @@
...
+      key1: ['value1'],
+      team: ['backend', 'frontend'],
+    })
+  })
+})
```
juan-fernandez, line 21: we assume that every value will be treated as a string, even when specifying numbers, right?

juan-fernandez (approved):
I left some questions but LGTM

src/commands/gate/evaluate.ts
```diff
@@ -58,15 +65,20 @@ export class GateEvaluateCommand extends Command {
...
 
+    const envVarTags = this.config.envVarTags ? parseTags(this.config.envVarTags.split(',')) : {}
+    const cliTags = this.tags ? parseTags(this.tags) : {}
+
     return {
       ...gitSpanTags,
       ...ciSpanTags,
       ...userGitSpanTags,
+      ...cliTags,
+      ...envVarTags,
```
liashenko, line 76: What happens if the cli tags collide with the env tags? Let's say a user specifies `git.commit.sha` or `git.repository_url` in the cli. Which ones take precedence?

liashenko (approved):

src/commands/gate/utils.ts
andrea-mosk: I think we are fine with lists of single values. In the backend we will have a specific struct assuming a list there and we will iterate over the elements. If there'll be 1 element it will just do a single iteration.

src/commands/gate/__tests__/scope.test.ts
andrea-mosk: Good spot! The logic will still function properly but we will iterate over elements even if not needed. I'll change this to remove duplicates

src/commands/gate/__tests__/scope.test.ts
andrea-mosk: Correct

src/commands/gate/renderer.ts
```diff
@@ -67,11 +67,18 @@ export const renderDryRunEvaluation = (): string => {
...
+export const renderGateEvaluationInput = (evaluateRequest: Payload): string => {
   let fullStr = chalk.bold(`${ICONS.INFO} Evaluating rules matching the following information:\n`)
-  fullStr += `Repository: ${spanTags['git.repository_url']}\n`
-  fullStr += `Branch: ${spanTags['git.branch']}\n`
-  fullStr += `Pipeline Name: ${spanTags['ci.pipeline.name']}\n`
+  fullStr += `Repository: ${evaluateRequest.spanTags[GIT_REPOSITORY_URL]}\n`
+  fullStr += `Branch: ${evaluateRequest.spanTags[GIT_BRANCH]}\n`
+
+  for (const scopeKey in evaluateRequest.userScope) {
+    if (evaluateRequest.userScope.hasOwnProperty(scopeKey)) {
```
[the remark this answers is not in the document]
andrea-mosk: Yeah I know, but I was receiving the error from `yarn format`:

> error: The body of a for-in should be wrapped in an if statement to filter unwanted properties from the prototype

After a quick research I found that this is what is needed to put the error away. Do you know about a better way?

src/commands/gate/evaluate.ts
andrea-mosk: @juan-fernandez Correct me if I'm wrong, but the latest tags take precedence here. This gives the user the ability to overwrite the default tags that we retrieve if they want, similarly to other commands (junitxml, sarif)

src/commands/gate/renderer.ts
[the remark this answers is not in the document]
juan-fernandez: You could use `Object.keys(userScope).forEach` or `Object.entries` or even do it with a single `reduce`: 
```javascript
fullStr = Object.entries(evaluateRequest.userScope).reduce((acc, [scopeKey, scopeValue]) => 
  acc + `${scopeKey}: ${scopeValue.join(' OR ')}\n`
, fullStr)
```

src/commands/gate/__tests__/scope.test.ts
andrea-mosk: Done!

src/commands/gate/evaluate.ts
drodriguezhdez: 👋🏼 I think I've found a way to do it in the backend without changing all the ingestion pipeline. I'm trying this before adding the `ddTags` key in the client. But as Juan commented, it depends on the specific ingestion pipeline logic for the gate command.

src/commands/gate/renderer.ts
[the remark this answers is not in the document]
andrea-mosk: Thanks! Changed to using `Object.entries`

src/commands/gate/evaluate.ts
andrea-mosk: Mmm I'm not sure I'm getting what the issue was with the tags for the `sarif` command. I guess in the junitxml command they are treated in the same way and nothing was spotted so I'm trying to understand whether this can be an issue for us as well or not

andrea-mosk: We chatted in Slack. The issue was that the tags weren't deserialized correctly in the `sarif-processor` service because the corresponding struct only allowed specific fields. In the `quality-gates` service we are using a general `map[string]string` (similarly to `junitxml-processor`) so it shouldn't be a problem for us 👍

andrea-mosk: Chatted offline, it won't be an issue for us

PR #900, opened 2023-06-01
Add retries and handle status code based on response

andrea-mosk:
### What and why?

Enhancing the gate command:

- Add retries to the call
- Checking status code when there's an error. If it's 4xx, we also log the message to convey the validation failure to the customers. If it's 5xx, we just log the status. 
- By default, if the response is 5xx, the datadog-ci command still succeed. Clients can override this behavior by setting the `--fail-if-unavailable` flag

### Review checklist


src/commands/gate/evaluate.ts
```diff
@@ -85,33 +88,46 @@ export class GateEvaluateCommand extends Command {
...
-        // TODO Handle unavailability etc.
-        this.context.stderr.write(renderGateEvaluationError(error))
-
-        return 1
+        return this.handleEvaluationError(error)
       })
   }
 
-  private handleEvaluationResponse(evaluationResponse: EvaluationResponse): number {
+  private handleSuccessfulEvaluation(evaluationResponse: EvaluationResponse): number {
```
juan-fernandez, line 108: do we want to name this `handleEvaluationSuccess`? To follow the convention with `handleEvaluationError`

juan-fernandez (approved):
LGTM. Left some minor comments

src/commands/gate/__tests__/evaluate.test.ts
```diff
@@ -67,8 +67,63 @@ describe('evaluate', () => {
...
+      command['failIfUnavailable'] = true
+      command.context = {stderr: {write}} as any
+
+      const error = createError(500, 'internal issue')
+      expect(command['handleEvaluationError'].bind(command).call({}, error)).toEqual(1)
+
+      const stdErrLog = write.mock.calls[0][0]
+      expect(stdErrLog).toContain('ERROR: Could not evaluate the rules. Status code: 500')
+      expect(stdErrLog).toContain("Use the '--fail-if-unavailable' option to fail the command in this situation.")
+      expect(stdErrLog).not.toContain('internal issue')
```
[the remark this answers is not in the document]
andrea-mosk: Not if it's a 5xx, as the error will be something internal that we don't want the user to have visibility about, as they can't do much about it. If it's a 4xx (most likely a validation failure), we want them to have visibility as they might be able to fix it

src/commands/gate/evaluate.ts
andrea-mosk: Makes sense, renamed!

src/commands/gate/__tests__/evaluate.test.ts
```diff
@@ -67,8 +67,63 @@ describe('evaluate', () => {
...
+      const command = new GateEvaluateCommand()
+      command['failIfUnavailable'] = true
+      command.context = {stderr: {write}} as any
+
+      const error = createError(500, 'internal issue')
+      expect(command['handleEvaluationError'].bind(command).call({}, error)).toEqual(1)
+
+      const stdErrLog = write.mock.calls[0][0]
+      expect(stdErrLog).toContain('ERROR: Could not evaluate the rules. Status code: 500')
+      expect(stdErrLog).toContain("Use the '--fail-if-unavailable' option to fail the command in this situation.")
```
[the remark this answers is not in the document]
andrea-mosk: Changed it (see [comment](https://github.com/DataDog/datadog-ci/pull/900#discussion_r1213307080) below)

src/commands/gate/__tests__/evaluate.test.ts
```diff
@@ -22,7 +22,7 @@ describe('evaluate', () => {
       expect(write.mock.calls[0][0]).toContain('DATADOG_APP_KEY')
     })
   })
-  describe('handleEvaluationResponse', () => {
+  describe('handleSuccessfulEvaluation', () => {
```
juan-fernandez, line 25:
```suggestion
  describe('handleEvaluationSuccess', () => {
```

src/commands/gate/renderer.ts
```diff
@@ -82,13 +83,29 @@ export const renderGateEvaluationInput = (evaluateRequest: Payload): string => {
...
-    fullStr += chalk.red(` Error Status: ${error.response.status}`)
+export const renderGateEvaluationError = (error: any, failIfUnavailable: boolean): string => {
+  let errorStr = 'ERROR: Could not evaluate the rules.'
+  if (!error.response) {
+    return chalk.red(`${errorStr}\n`)
   }
 
-  fullStr += '\n'
+  errorStr += ` Status code: ${error.response.status}.\n`
+  if (isBadRequestError(error)) {
```
juan-fernandez, line 93: is this condition only valid for HTTP 400 errors? No other 4xx?

juan-fernandez (approved):

src/commands/gate/renderer.ts
andrea-mosk: Correct, we return 400 in case of validation failures and only in those cases we should convey the error message to the user

PR #903, opened 2023-06-02
Adding usage and README for gate command

andrea-mosk:
### What and why?

Adding usage and README for gate command

### Review checklist

### DIFF
--- a/src/commands/gate/README.md
+++ b/src/commands/gate/README.md
@@ -1,6 +1,6 @@
 # GATE command
 
-Evaluate Quality Gates rules in Datadog.
+Evaluate Quality Gates rules in Datadog. The exit status of the command will change depending on the result of the gate evaluation.
 
 ## Usage
 
--- a/src/commands/gate/README.md
+++ b/src/commands/gate/README.md
@@ -35,7 +35,7 @@ Additionally, you can configure the `gate` command with the following environmen
 - `DD_TAGS`: Sets global tags applied to all spans. The format must be `key1:value1,key2:value2`. The upload process merges the tags passed on the command line with the tags in the `--tags` parameter. If a key appears in both `--tags` and `DD_TAGS`, the value in `DD_TAGS` takes precedence.
 - `DATADOG_SITE`: Your Datadog site, for example, datadoghq.com or datadoghq.eu.
 
-### Optional dependencies
+### Dependencies
 
 - [`git`](https://git-scm.com/downloads) is used for extracting repository metadata.
 
--- a/src/commands/gate/README.md
+++ b/src/commands/gate/README.md
@@ -21,19 +21,19 @@ datadog-ci gate evaluate --scope team:backend --scope team:frontend --fail-on-em
 ```
 
 - `--scope` is an array of key value pairs of the form `key:value`. This parameter sets additional scope when retrieving matching rules. Only the rules matching the scope provided will be evaluated.
-- `--tags` is an array of key value pairs of the form `key:value`. This parameter sets global tags applied to all results. The upload process merges the tags passed on the command line with the tags in the `DD_TAGS` environment variable. If a key appears in both `--tags` and `DD_TAGS`, the value in `DD_TAGS` takes precedence.
-- `--dry-run` (default: `false`): runs the command without the final evaluation step. All other checks are performed.
-- `--fail-if-unavailable` (default: `false`): fails the command if there is unavailability on the Datadog side.
-- `--fail-on-empty` (default: `false`): fails the command if no matching rules were found in Datadog.
+- `--tags` is an array of key value pairs of the form `key:value`. This parameter sets global tags that are applied to all results. The upload process merges the tags passed on the command line with the tags in the `DD_TAGS` environment variable. If a key appears in both `--tags` and `DD_TAGS`, the value in `DD_TAGS` takes precedence.
+- `--dry-run` runs the command without the final evaluation step. All other checks are performed. The default value is `false`.
+- `--fail-if-unavailable` fails the command if Datadog is unavailable. The default value is `false`.
+- `--fail-on-empty` fails the command if no matching rules are found in Datadog. The default value is `false`.
 
 ### Environment variables
 
-Additionally, you may configure the `gate` command with environment variables:
+Additionally, you can configure the `gate` command with the following environment variables:
 
-- `DATADOG_API_KEY` or `DD_API_KEY` (**required**): API key used to authenticate the requests.
-- `DATADOG_APP_KEY` (**required**): APP key used to authenticate the requests.
-- `DD_TAGS`: Set global tags applied to all spans. The format must be `key1:value1,key2:value2`. The upload process merges the tags passed on the command line with the tags in the `--tags` parameter. If a key appears in both `--tags` and `DD_TAGS`, the value in `DD_TAGS` takes precedence.
-- `DATADOG_SITE`: choose your Datadog site, for example, datadoghq.com or datadoghq.eu.
+- `DATADOG_API_KEY` or `DD_API_KEY` (**required**): The API key used to authenticate the requests.
+- `DATADOG_APP_KEY` (**required**): The application key used to authenticate the requests.
+- `DD_TAGS`: Sets global tags applied to all spans. The format must be `key1:value1,key2:value2`. The upload process merges the tags passed on the command line with the tags in the `--tags` parameter. If a key appears in both `--tags` and `DD_TAGS`, the value in `DD_TAGS` takes precedence.
+- `DATADOG_SITE`: Your Datadog site, for example, datadoghq.com or datadoghq.eu.
 
 ### Optional dependencies
 
--- a/src/commands/gate/README.md
+++ b/src/commands/gate/README.md
@@ -1,1 +1,62 @@
-WORK IN PROGRESS
+# GATE command
+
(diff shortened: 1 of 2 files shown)


src/commands/gate/README.md
andrea-mosk, line 1: Both the README and the usage description have been done similarly to the `sarif` command:

- [Sarif README](https://github.com/DataDog/datadog-ci/blob/master/src/commands/sarif/README.md)
- [Sarif usage](https://github.com/DataDog/datadog-ci/blob/master/src/commands/sarif/upload.ts#L60-L81)

src/commands/gate/README.md
```diff
@@ -1 +1,62 @@
-WORK IN PROGRESS
\ No newline at end of file
+# GATE command
+
+Evaluate Quality Gates rules in Datadog.
```
liashenko, line 3: I feel like the description is too short.
Do you think we can add a short description of what exactly it's doing?
It might be useful to add that it returns different exit codes and can be used to gate (pass/fail) a pipeline based on the data in Datadog.

Also, maybe it's worth to add that it should be called in the pipeline after sending data to Datadog.

andrea-mosk: Just for context, this won't be the primary source for understanding how to setup quality gates, but we will have a dedicated documentation page, similarly to the `junit` command: https://docs.datadoghq.com/continuous_integration/tests/junit_upload/?tab=linux. 

The description here is short and aligned with Sarif command ("Upload your SARIF report files.") and junit command ("Upload your jUnit XML files."). I don't think we need much more than this as we'll have documentation, that for example will cover the fact that the events needs to be sent before the command is executed, etc.

src/commands/gate/README.md
```diff
@@ -1 +1,62 @@
...
+```
+
+For example:
+
+```bash
+datadog-ci gate evaluate --scope team:backend --scope team:frontend --fail-on-empty
+```
+
+- `--scope` is an array of key value pairs of the form `key:value`. This parameter sets additional scope when retrieving matching rules. Only the rules matching the scope provided will be evaluated.
+- `--tags` is an array of key value pairs of the form `key:value`. This parameter sets global tags applied to all results. The upload process merges the tags passed on the command line with the tags in the `DD_TAGS` environment variable. If a key appears in both `--tags` and `DD_TAGS`, the value in `DD_TAGS` takes precedence.
```
drichards-87, line 24:
```suggestion
- `--tags` is an array of key value pairs of the form `key:value`. This parameter sets global tags that are applied to all results. The upload process merges the tags passed on the command line with the tags in the `DD_TAGS` environment variable. If a key appears in both `--tags` and `DD_TAGS`, the value in `DD_TAGS` takes precedence.
```

src/commands/gate/README.md
```diff
@@ -1 +1,62 @@
...
+- `--dry-run` (default: `false`): runs the command without the final evaluation step. All other checks are performed.
+- `--fail-if-unavailable` (default: `false`): fails the command if there is unavailability on the Datadog side.
+- `--fail-on-empty` (default: `false`): fails the command if no matching rules were found in Datadog.
+
+### Environment variables
+
+Additionally, you may configure the `gate` command with environment variables:
```
drichards-87, line 31:
```suggestion
Additionally, you can configure the `gate` command with the following environment variables:
```

src/commands/gate/README.md
```diff
@@ -1 +1,62 @@
...
+
+- `DATADOG_API_KEY` or `DD_API_KEY` (**required**): API key used to authenticate the requests.
+- `DATADOG_APP_KEY` (**required**): APP key used to authenticate the requests.
+- `DD_TAGS`: Set global tags applied to all spans. The format must be `key1:value1,key2:value2`. The upload process merges the tags passed on the command line with the tags in the `--tags` parameter. If a key appears in both `--tags` and `DD_TAGS`, the value in `DD_TAGS` takes precedence.
+- `DATADOG_SITE`: choose your Datadog site, for example, datadoghq.com or datadoghq.eu.
```
drichards-87, line 36:
```suggestion
- `DATADOG_API_KEY` or `DD_API_KEY` (**required**): The API key used to authenticate the requests.
- `DATADOG_APP_KEY` (**required**): The application key used to authenticate the requests.
- `DD_TAGS`: Sets global tags applied to all spans. The format must be `key1:value1,key2:value2`. The upload process merges the tags passed on the command line with the tags in the `--tags` parameter. If a key appears in both `--tags` and `DD_TAGS`, the value in `DD_TAGS` takes precedence.
- `DATADOG_SITE`: Your Datadog site, for example, datadoghq.com or datadoghq.eu.
```

src/commands/gate/README.md
```diff
@@ -1 +1,62 @@
...
+
+```bash
+datadog-ci gate evaluate --scope team:backend --scope team:frontend --fail-on-empty
+```
+
+- `--scope` is an array of key value pairs of the form `key:value`. This parameter sets additional scope when retrieving matching rules. Only the rules matching the scope provided will be evaluated.
+- `--tags` is an array of key value pairs of the form `key:value`. This parameter sets global tags applied to all results. The upload process merges the tags passed on the command line with the tags in the `DD_TAGS` environment variable. If a key appears in both `--tags` and `DD_TAGS`, the value in `DD_TAGS` takes precedence.
+- `--dry-run` (default: `false`): runs the command without the final evaluation step. All other checks are performed.
+- `--fail-if-unavailable` (default: `false`): fails the command if there is unavailability on the Datadog side.
+- `--fail-on-empty` (default: `false`): fails the command if no matching rules were found in Datadog.
```
drichards-87, line 27:
```suggestion
- `--dry-run` runs the command without the final evaluation step. All other checks are performed. The default value is `false`.
- `--fail-if-unavailable` fails the command if Datadog is unavailable. The default value is `false`.
- `--fail-on-empty` fails the command if no matching rules are found in Datadog. The default value is `false`.
```

drichards-87 (changes requested):
Left you some minor feedback from Docs.

src/commands/gate/README.md
```diff
@@ -1 +1,62 @@
...
+- `--fail-if-unavailable` fails the command if Datadog is unavailable. The default value is `false`.
+- `--fail-on-empty` fails the command if no matching rules are found in Datadog. The default value is `false`.
+
+### Environment variables
+
+Additionally, you can configure the `gate` command with the following environment variables:
+
+- `DATADOG_API_KEY` or `DD_API_KEY` (**required**): The API key used to authenticate the requests.
+- `DATADOG_APP_KEY` (**required**): The application key used to authenticate the requests.
+- `DD_TAGS`: Sets global tags applied to all spans. The format must be `key1:value1,key2:value2`. The upload process merges the tags passed on the command line with the tags in the `--tags` parameter. If a key appears in both `--tags` and `DD_TAGS`, the value in `DD_TAGS` takes precedence.
```
juan-fernandez, line 35: unrelated to the changes, but would we want a `DD_SCOPE`?

src/commands/gate/README.md
```diff
@@ -1 +1,62 @@
...
+- `DATADOG_SITE`: Your Datadog site, for example, datadoghq.com or datadoghq.eu.
+
+### Optional dependencies
```
juan-fernandez, line 38: will this always be an optional dependency?

src/commands/gate/README.md
juan-fernandez: I agree we don't need to go into detail for this README, but the exit code bit might be small and useful. 

"The exit code of this command will change depending on the result of the evaluation rule"? or something like that

juan-fernandez (approved):
LGTM. Only minor comments

liashenko (approved):

drichards-87 (approved):

src/commands/gate/README.md
andrea-mosk: Mmm do you think it will be useful for clients? What's the main reason we have `DD_TAGS` and `--tags`?

src/commands/gate/README.md
andrea-mosk: I copied it from Junit and Sarif, which are basically doing the same thing (extracting git-related info from env var). Should I update this?

src/commands/gate/README.md
andrea-mosk: 👍 Added

src/commands/gate/README.md
juan-fernandez: ergonomics: sometimes it might be easier to set an env var than it is to modify the command. No big difference though

src/commands/gate/README.md
juan-fernandez: I was just curious whether gating would eventually _require_ git for getting the pertinent pipeline/job/anything. JUnit and Sarif only add this metadata as extra info, AFAIK

andrea-mosk: Oh, that's a good point. Should I mark this as a dependency then (without "optional")?

juan-fernandez: it seems like the safest option, as we can always go to "optional" later on

andrea-mosk: 👍 Looks good, changed

src/commands/gate/README.md
andrea-mosk: Makes sense, noted it down for future 👍

PR #906, opened 2023-06-02
[test] Improve standalone binary test timeout

Drarig29:
### What and why?

We managed to fix the flakiness on macOS (#893), but Windows can also [occasionally fail](https://github.com/DataDog/datadog-ci/actions/runs/5142676448/jobs/9256709607).

Let's fix this definitely.

### How?

Define timeouts for all platforms.

### Review checklist

### DIFF
--- a/standalone-e2e/standalone-binary.test.ts
+++ b/standalone-e2e/standalone-binary.test.ts
@@ -6,21 +6,25 @@ import {version} from '../package.json'
 const execPromise = promisify(exec)
 
 const isWin = process.platform === 'win32'
-
 const os = isWin ? 'win' : process.platform === 'darwin' ? 'darwin' : 'linux'
 
 const STANDALONE_BINARY = `datadog-ci_${os}-x64`
-
 const STANDALONE_BINARY_PATH = `${isWin ? '.\\' : './'}${STANDALONE_BINARY}${isWin ? '.exe' : ''}`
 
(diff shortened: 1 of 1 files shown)


standalone-e2e/standalone-binary.test.ts
```diff
@@ -6,21 +6,25 @@ import {version} from '../package.json'
...
 const sanitizeOutput = (output: string) => output.replace(/(\r\n|\n|\r)/gm, '')
 
+const timeoutPerPlatform: Record<typeof os, number> = {
+  // Some macOS agents sometimes run slower, making this test suite flaky on macOS only.
+  // The issue is tracked here: https://github.com/actions/runner-images/issues/3885
+  darwin: 10 * 1000,
+  // Keep the default timeout for Linux.
+  linux: 5 * 1000,
+  // Running the binary on Windows is also slower than on Linux, and sometimes times out by a very small margin.
+  win: 6 * 1000,
```
juan-fernandez, line 23: maybe at this point bumping this to 10 makes sense? The bottleneck is macos so it shouldn't be much of a difference

juan-fernandez (approved):

standalone-e2e/standalone-binary.test.ts
juan-fernandez: (IIRC these tests run in parallel so macos is the bottleneck of the pipeline)

Drarig29: True 😁

PR #912, opened 2023-06-06
https://datadoghq.atlassian.net/browse/SLS-3589

nhulston:
Fix ruby2.7 ARM layer typo

### What and why?

The ruby ARM layer had a typo

### How?

`ruby2.`  -> `ruby2.7`

### DIFF
--- a/src/commands/lambda/constants.ts
+++ b/src/commands/lambda/constants.ts
@@ -47,7 +47,7 @@ export const RUNTIME_LOOKUP = {
 
 export type Runtime = keyof typeof RUNTIME_LOOKUP
 export type LayerKey = keyof typeof LAYER_LOOKUP
-export const ARM_LAYERS = [EXTENSION_LAYER_KEY, 'dotnet6', 'python3.8', 'python3.9', 'python3.10', 'ruby2.']
+export const ARM_LAYERS = [EXTENSION_LAYER_KEY, 'dotnet6', 'python3.8', 'python3.9', 'python3.10', 'ruby2.7']
 export const ARM64_ARCHITECTURE = 'arm64'
 export const ARM_LAYER_SUFFIX = '-ARM'
 


src/commands/lambda/constants.ts
duncanista, line 1: Can we add a unit test instrumenting for Ruby 2.7 on ARM?

duncanista (changes requested):

src/commands/lambda/__tests__/instrument.test.ts
```diff
@@ -1073,46 +1073,71 @@ describe('lambda', () => {
...
       })
 
-      test('instruments Ruby application properly', async () => {
-        ;(fs.readFile as any).mockImplementation((a: any, b: any, callback: any) => callback({code: 'ENOENT'}))
-        mockLambdaConfigurations(lambdaClientMock, {
-          'arn:aws:lambda:us-east-1:123456789012:function:lambda-hello-world': {
-            config: {
-              FunctionArn: 'arn:aws:lambda:us-east-1:123456789012:function:lambda-hello-world',
-              Runtime: 'ruby2.7',
+      describe('instruments Ruby application properly', () => {
```
duncanista, line 1076: Although I agree there is a lot of boilerplate code and testing instrumentation could be simplified. I think making it a function and reusing it here would be a bit confusing if someone else jumps into the code, since it's only applied for this scenario. I strongly think this could be applied to every command, but a better testing suite could be implemented overall in our project.

That said, I think we are over engineering here, why don't we try an easier approach like:
https://github.com/DataDog/datadog-ci/blob/1e7a4ae91d25a52434af622ceda1556c06b1c397/src/commands/lambda/__tests__/instrument.test.ts#L723-L740

What do you think?

nhulston: Thanks for the feedback. I have rewritten the test, and I think it's a lot easier to understand now. Let me know your thoughts!

duncanista (approved):
LGTM! :shipit: 

Great job!

PR #917, opened 2023-06-09
[Synthetics] Create mobile application version upload command

ClaraJacintho:
### What and why?

Create a command to upload new mobile application versions to be used in Mobile Synthetics tests. The Mobile Application must have been set up in Datadog before.

### How?

The command will:
- Upload the application file to Datadog, as done previously for test overrides
- Create a new version for a given Mobile Application, that can then be used in Synthetics tests

### Review checklist


xgouchet (approved):
Nice work, can you also update the `src/commands/synthetics/README.md` doc to show how to use this new option?

src/commands/synthetics/interfaces.ts
```diff
@@ -380,6 +380,13 @@ export interface CommandConfig extends SyntheticsCIConfig {
   failOnTimeout: boolean
 }
 
+export interface UploadApplicationCommandConfig extends SyntheticsCIConfig {
+  mobileApplicationVersionFilePath?: string
+  mobileApplicationId?: string
+  versionName?: string
+  latest?: boolean
+}
+
```
etnbrd, line 389: Rather than extending `SyntheticsCIConfig` with a lot of unnecessary fields, let's just copy the fields that are required for this new command. And we can create a card to refactor these interfaces to avoid the duplication.

etnbrd (changes requested):
LGTM, with just one change to avoid all the unnecessary config field which are confusing.

etnbrd (approved):
LGTM 👍

src/commands/synthetics/mobile.ts
```diff
@@ -116,3 +124,50 @@ export const uploadApplicationAndOverrideConfig = async (
...
+}
+
+export const uploadMobileApplicationVersion = async (
+  config: UploadApplicationCommandConfig
+): Promise<MobileApplicationVersion> => {
+  const api = getApiHelper(config)
+
+  if (!config.mobileApplicationVersionFilePath) {
+    throw new CiError('MISSING_MOBILE_APPLICATION_PATH', 'Mobile application path is required.')
+  }
```
KevinYeramian, line 149: nit: Add new lines between  and after conditions for readability.

src/commands/synthetics/uploadApplicationCommand.ts
```diff
@@ -0,0 +1,85 @@
...
+
+    try {
+      const version = await uploadMobileApplicationVersion(this.config)
+      this.logger.info(`Created new version ${version.version_name}, with version ID: ${version.id}`)
+    } catch (error) {
+      if (error instanceof CiError || error instanceof EndpointError) {
+        this.logger.error(`Error: ${error.message}`)
+      }
+
+      return 1
```
KevinYeramian, line 69: Should we use `reportExitLogs` and `toExitCode`  here to correctly handle the log and exit code ?

src/commands/synthetics/__tests__/fixtures.ts
```diff
@@ -545,3 +547,25 @@ export const getTestPayload = (override?: Partial<TestPayload>) => ({
...
+  application_id: '789-dfg-987',
+  file_name: 'bla.',
+  original_file_name: 'test.apk',
+  is_latest: true,
+  version_name: 'test version',
+  created_at: '22-09-2022',
+  ...override,
+})
+
+export const uploadCommandCongif: UploadApplicationCommandConfig = {
```
Drarig29, line 562:
```suggestion
export const uploadCommandConfig: UploadApplicationCommandConfig = {
```

src/commands/synthetics/uploadApplicationCommand.ts
Drarig29, line 1: This file is camel case, but `run-tests-command.ts` is kebab case, could you rename your file to kebab case?

src/commands/synthetics/uploadApplicationCommand.ts
ClaraJacintho: `reportExitLogs` requires a `Reporter`, which we are avoiding in this command since it can cause problems with how it shows up in Datadog. 

`reportExitLogs` requires  a `Reporter` and a test result, which we won't have in this case. In other places in the `run-tests-command` a simple `return 1` is used to denote an error as well.

Drarig29: This ^

I was exactly writing that :grin:

src/commands/synthetics/uploadApplicationCommand.ts
ClaraJacintho: I am speed 🏃🏻‍♀️

KevinYeramian (approved):
LGTM

PR #918, opened 2023-06-09
[SYNTH] rename runTestsCommand

etnbrd:
### What and why?

In prevision to https://github.com/DataDog/datadog-ci/pull/917, and to avoid any confusion between the two commands, this PR renames the original `command.ts` file into `run-tests-command.ts`, and rename the original `RunTestCommand` class to `RunTestsCommand` (plural tests) to follow the `run-tests` command.

### How?

✨VSCode Rename Symbol✨ 

### Review checklist

- [X] Feature or bugfix MUST have appropriate tests (unit, integration) -> N/A


jakepruitt:
Not to be too picky, but could we rename the `run-test.ts` file as well ? Something like `run-tests-lib.ts` or even just `run-tests.ts` ?

etnbrd:
> Not to be too picky, but could we rename the `run-test.ts` file as well ? Something like `run-tests-lib.ts` or even just `run-tests.ts` ?

I pushed that to another PR to avoid rebasing everything again for this small change: https://github.com/DataDog/datadog-ci/pull/922

PR #919, opened 2023-06-09
Add waiting time to gate evaluate command

andrea-mosk:
### What and why?

Add waiting time to gate evaluate command

### Review checklist

### DIFF
--- a/src/commands/gate/README.md
+++ b/src/commands/gate/README.md
@@ -11,7 +11,7 @@ Evaluate Quality Gates rules in Datadog. The exit status of the command will cha
 The `evaluate` command evaluates the matching Quality Gates rules in Datadog.
 
 ```bash
-datadog-ci gate evaluate [--scope] [--tags] [--dry-run] [--fail-if-unavailable] [--fail-on-empty]
+datadog-ci gate evaluate [--scope] [--tags] [--dry-run] [--fail-if-unavailable] [--fail-on-empty] [--no-wait]
 ```
 
 For example:
@@ -25,6 +25,8 @@ datadog-ci gate evaluate --scope team:backend --scope team:frontend --fail-on-em
 - `--dry-run` runs the command without the final evaluation step. All other checks are performed. The default value is `false`.
 - `--fail-if-unavailable` fails the command if Datadog is unavailable. The default value is `false`.
 - `--fail-on-empty` fails the command if no matching rules are found in Datadog. The default value is `false`.
+- `--no-wait` can be used in order to remove the default waiting time (30s) that the command has to ensure that the events (e.g. tests) can be properly queried by the rules. This mechanism avoids that rules are incorrectly evaluated due to the absence of the events. Read more on the quality gates documentation before enabling this. The default value is `false`.
+
 
 ### Environment variables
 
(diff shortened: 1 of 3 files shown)


src/commands/gate/README.md
```diff
@@ -25,6 +25,8 @@ datadog-ci gate evaluate --scope team:backend --scope team:frontend --fail-on-em
 - `--dry-run` runs the command without the final evaluation step. All other checks are performed. The default value is `false`.
 - `--fail-if-unavailable` fails the command if Datadog is unavailable. The default value is `false`.
 - `--fail-on-empty` fails the command if no matching rules are found in Datadog. The default value is `false`.
+- `--no-wait` can be used in order to remove the default waiting time (30s) that the command has to ensure that the events (e.g. tests) can be properly queried by the rules. This mechanism avoids that rules are incorrectly evaluated due to the absence of the events. Read more on the quality gates documentation before enabling this. The default value is `false`.
```
juan-fernandez, line 28: I think this is a bit clearer:
```suggestion
- `--no-wait` removes the waiting time (30s) that is in place to ensure that events (e.g. tests) can be properly queried by the rules. This mechanism avoids that rules are incorrectly evaluated due to the absence of events. Read more on the quality gates documentation before passing this flag. The default value is `false`.
```

Also, why do we want to allow users to set this? Isn't it dangerous?

liashenko (approved):

src/commands/gate/README.md
andrea-mosk: Looks clearer, thanks! 

> Also, why do we want to allow users to set this? Isn't it dangerous?

It's not really dangerous, what can happen is that rules might be incorrectly evaluated, but we need to do a good job on the documentation side on making sure they understand the risks. There are many use cases in which they wouldn't want to wait for 30s, for example when there are intermediate jobs between when the events are sent and when `datadog-ci gate evaluate` is called.

juan-fernandez (approved):

src/commands/gate/README.md
```diff
@@ -25,6 +25,8 @@ datadog-ci gate evaluate --scope team:backend --scope team:frontend --fail-on-em
 - `--dry-run` runs the command without the final evaluation step. All other checks are performed. The default value is `false`.
 - `--fail-if-unavailable` fails the command if Datadog is unavailable. The default value is `false`.
 - `--fail-on-empty` fails the command if no matching rules are found in Datadog. The default value is `false`.
+- `--no-wait` removes the waiting time (30s) that is in place to ensure that events (e.g. tests) can be properly queried by the rules. This mechanism avoids that rules are incorrectly evaluated due to the absence of events. Read more on the quality gates documentation before passing this flag. The default value is `false`.
```
cswatt, line 28:
```suggestion
- `--no-wait` removes the waiting time (30s) that is in place to ensure that events (for example: tests) can be properly queried by the rules. This mechanism prevents rules from being incorrectly evaluated due to the absence of events. Read more on the quality gates documentation before passing this flag. The default value is `false`.
```

would be nice to add a link to the quality gates documentation

cswatt (approved):
leaving feedback on docs, approving so as not to block

src/commands/gate/README.md
andrea-mosk: Thanks, I have committed your suggestion.  We still need to create the documentation, so no link for now 😬

PR #924, opened 2023-06-09
[lambda] [flare] Create lambda flare subcommands

nhulston:
### What and why?

Create lambda flare subcommands.
The command will be used to easily gather configuration information from customers to help support staff quickly resolve issues.

### How?

Add the following options:
`--dry, --interactive, --allFunctions, --function, --region, --api-key, --case-id, --email`

Also, have some initial validation to ensure that required options are included.

### Review checklist

### DIFF
--- a/src/commands/lambda/__tests__/__snapshots__/flare.test.ts.snap
+++ b/src/commands/lambda/__tests__/__snapshots__/flare.test.ts.snap
@@ -3,11 +3,15 @@
 exports[`lambda flare prints dry-run header 1`] = `
 "
 [Dry Run] 🐶 Generating Lambda flare to send your configuration to Datadog
+
+[Error] No functions specified. [-f,--function] or [--allFunctions]
 "
 `;
 
 exports[`lambda flare prints non-dry-run header 1`] = `
 "
 🐶 Generating Lambda flare to send your configuration to Datadog
+
+[Error] No functions specified. [-f,--function] or [--allFunctions]
 "
 `;
--- a/src/commands/lambda/__tests__/flare.test.ts
+++ b/src/commands/lambda/__tests__/flare.test.ts
@@ -6,7 +6,7 @@ describe('lambda flare', () => {
     const context = createMockContext()
     const code = await cli.run(['lambda', 'flare'], context as any)
     const output = context.stdout.toString()
-    expect(code).toBe(0)
+    expect(code).toBe(1)
     expect(output).toMatchSnapshot()
   })
 
@@ -15,7 +15,62 @@ describe('lambda flare', () => {
     const context = createMockContext()
     const code = await cli.run(['lambda', 'flare', '-d'], context as any)
     const output = context.stdout.toString()
-    expect(code).toBe(0)
+    expect(code).toBe(1)
     expect(output).toMatchSnapshot()
+  })
+
+  it('prints error when no functions specified', async () => {
+    const cli = makeCli()
+    const context = createMockContext()
+    const code = await cli.run(
+      ['lambda', 'flare', '-r', 'us-west-2', '--api-key', '123', '-e', 'test@test.com'],
+      context as any
+    )
+    expect(code).toBe(1)
+    const output = context.stdout.toString()
+    expect(output).toContain('No functions specified. [-f,--function] or [--allFunctions]')
+  })
+
+  it('prints error when no region specified', async () => {
+    const cli = makeCli()
+    const context = createMockContext()
+    const code = await cli.run(
+      ['lambda', 'flare', '-f', 'func', '--api-key', '123', '-e', 'test@test.com'],
+      context as any
+    )
+    expect(code).toBe(1)
+    const output = context.stdout.toString()
+    expect(output).toContain('No region specified. [-r,--region]')
+  })
+
+  it('prints error when no apiKey specified', async () => {
+    const cli = makeCli()
+    const context = createMockContext()
+    const code = await cli.run(
+      ['lambda', 'flare', '-f', 'func', '-r', 'us-west-2', '-e', 'test@test.com'],
+      context as any
+    )
+    expect(code).toBe(1)
+    const output = context.stdout.toString()
+    expect(output).toContain('No API key specified. [--api-key]')
+  })
+
+  it('prints error when no email specified', async () => {
+    const cli = makeCli()
+    const context = createMockContext()
+    const code = await cli.run(['lambda', 'flare', '-f', 'func', '-r', 'us-west-2', '--api-key', '123'], context as any)
+    expect(code).toBe(1)
+    const output = context.stdout.toString()
+    expect(output).toContain('No email specified. [-e,--email]')
+  })
+
+  it('runs successfully with all required options specified', async () => {
+    const cli = makeCli()
+    const context = createMockContext()
+    const code = await cli.run(
+      ['lambda', 'flare', '-f', 'func', '-r', 'us-west-2', '--api-key', '123', '-e', 'test@test.com'],
+      context as any
+    )
+    expect(code).toBe(0)
   })
 })
--- a/src/commands/lambda/flare.ts
+++ b/src/commands/lambda/flare.ts
@@ -1,14 +1,57 @@
 import {Command} from 'clipanion'
 
-import {renderLambdaFlareHeader} from './renderers/flare-renderer'
+import {renderError, renderLambdaFlareHeader} from './renderers/flare-renderer'
 
 export class LambdaFlareCommand extends Command {
   private isDryRun = false
+  private isInteractive = false
+  private allFunctions = false
+  private functions: string[] = []
+  private region = ''
+  private apiKey = ''
+  private caseID = ''
+  private email = ''
 
+  /**
+   * @returns 0 if the command ran successfully, 1 otherwise.
+   */
   public async execute() {
     this.context.stdout.write(renderLambdaFlareHeader(this.isDryRun))
+
+    if (!this.validateCommand()) {
+      return 1
+    }
+
+    return 0
+  }
+
+  /**
+   * @returns true if the command is valid, false otherwise.
+   * Prints an error message if the command is invalid.
+   */
+  private validateCommand = () => {
+    if (this.functions.length === 0 && !this.allFunctions) {
+      this.context.stdout.write(renderError('No functions specified. [-f,--function] or [--allFunctions]'))
+    } else if (this.region === '') {
+      this.context.stdout.write(renderError('No region specified. [-r,--region]'))
+    } else if (this.apiKey === '') {
+      this.context.stdout.write(renderError('No API key specified. [--api-key]'))
+    } else if (this.email === '') {
+      this.context.stdout.write(renderError('No email specified. [-e,--email]'))
+    } else {
+      return true
+    }
+
+    return false
   }
 }
 
 LambdaFlareCommand.addPath('lambda', 'flare')
 LambdaFlareCommand.addOption('isDryRun', Command.Boolean('-d,--dry'))
+LambdaFlareCommand.addOption('isInteractive', Command.Boolean('-i,--interactive'))
+LambdaFlareCommand.addOption('allFunctions', Command.Boolean('--allFunctions'))
+LambdaFlareCommand.addOption('functions', Command.Array('-f,--function'))
+LambdaFlareCommand.addOption('region', Command.String('-r,--region'))
+LambdaFlareCommand.addOption('apiKey', Command.String('--api-key'))
+LambdaFlareCommand.addOption('caseID', Command.String('-c,--case-id'))
+LambdaFlareCommand.addOption('email', Command.String('-e,--email'))
--- a/src/commands/lambda/flare.ts
+++ b/src/commands/lambda/flare.ts
@@ -30,7 +30,9 @@ export class LambdaFlareCommand extends Command {
    * Prints an error message if the command is invalid.
    */
   private validateCommand = () => {
-    if (this.functions.length === 0 && !this.allFunctions) {
+    if (this.isInteractive) {
+      return true
+    } else if (this.functions.length === 0 && !this.allFunctions) {
       this.context.stdout.write(renderError('No functions specified. [-f,--function] or [--allFunctions]'))
     } else if (this.region === '') {
       this.context.stdout.write(renderError('No region specified. [-r,--region]'))
--- a/src/commands/lambda/renderers/flare-renderer.ts
+++ b/src/commands/lambda/renderers/flare-renderer.ts
@@ -1,8 +1,15 @@
-import {dryRunTag} from './common-renderer'
+import {
+  dryRunTag,
+  errorTag,
+  warningTag,
+  warningExclamationSignTag,
+  successCheckmarkTag,
+  failCrossTag,
+} from './common-renderer'
 
 /**
  * @returns a header indicating which `lambda` subcommand is running.
- * @param command current selected lambda subcommand.
+ * @param isDryRun whether or not the command is a dry run. Defaults to false.
  *
  * ```txt
  * [Dry Run] 🐶 Instrumenting Lambda function
@@ -13,3 +20,53 @@ export const renderLambdaFlareHeader = (isDryRun: boolean) => {
 
   return `\n${prefix}🐶 Generating Lambda flare to send your configuration to Datadog\n`
 }
+
+/**
+ * @param error an error message
+ * @returns the provided error prefixed by {@link errorTag}.
+ *
+ * ```txt
+ * [Error] The provided error goes here!
+ * ```
+ */
+export const renderError = (error: string) => `\n${errorTag} ${error}\n`
+
+/**
+ * @param warning the message to warn about.
+ * @returns the provided warning prefixed by {@link warningTag}.
+ *
+ * ```txt
+ * [Warning] The provided warning goes here!
+ * ```
+ */
+export const renderWarning = (warning: string) => `${warningTag} ${warning}\n`
+
+/**
+ * @param warning the message to warn about.
+ * @returns the provided warning prefixed by {@link warningExclamationSignTag}.
+ *
+ * ```txt
+ * [!] The provided warning goes here!
+ * ```
+ */
+export const renderSoftWarning = (warning: string) => `${warningExclamationSignTag} ${warning}\n`
+
+/**
+ * @param message the message to set with the success tag.
+ * @returns the provided message prefixed by {@link successCheckmarkTag}.
+ *
+ * ```txt
+ * [✔] The provided message goes here!
+ * ```
+ */
+export const renderSuccess = (message: string) => `${successCheckmarkTag} ${message}\n`
+
+/**
+ * @param message the message to set with the fail tag.
+ * @returns the provided message prefixed by {@link failCrossTag}.
+ *
+ * ```txt
+ * [✖] The provided message goes here!
+ * ```
+ */
+export const renderFail = (message: string) => `${failCrossTag} ${message}\n`


src/commands/lambda/flare.ts
```diff
@@ -1,14 +1,59 @@
...
 }
 
 LambdaFlareCommand.addPath('lambda', 'flare')
 LambdaFlareCommand.addOption('isDryRun', Command.Boolean('-d,--dry'))
+LambdaFlareCommand.addOption('isInteractive', Command.Boolean('-i,--interactive'))
+LambdaFlareCommand.addOption('allFunctions', Command.Boolean('--allFunctions'))
+LambdaFlareCommand.addOption('functions', Command.Array('-f,--function'))
+LambdaFlareCommand.addOption('region', Command.String('-r,--region'))
+LambdaFlareCommand.addOption('apiKey', Command.String('--api-key'))
+LambdaFlareCommand.addOption('caseID', Command.String('-c,--case-id'))
```
duncanista, line 58:
```suggestion
LambdaFlareCommand.addOption('caseId', Command.String('-c,--case-id'))
```

src/commands/lambda/flare.ts
```diff
@@ -1,14 +1,59 @@
...
+    }
+
+    return false
   }
 }
 
 LambdaFlareCommand.addPath('lambda', 'flare')
 LambdaFlareCommand.addOption('isDryRun', Command.Boolean('-d,--dry'))
+LambdaFlareCommand.addOption('isInteractive', Command.Boolean('-i,--interactive'))
+LambdaFlareCommand.addOption('allFunctions', Command.Boolean('--allFunctions'))
```
duncanista, line 54: Standard for CLI options is kebab case.
Also let's remove this option until an implementation to flare all functions is available.

src/commands/lambda/flare.ts
```diff
@@ -1,14 +1,59 @@
...
+    } else if (this.apiKey === '') {
+      this.context.stdout.write(renderError('No API key specified. [--api-key]'))
+    } else if (this.email === '') {
+      this.context.stdout.write(renderError('No email specified. [-e,--email]'))
+    } else {
+      return true
+    }
+
+    return false
   }
```
duncanista, line 48: Can we refactor this into a Chain of Responsibility pattern instead of having multiple `if-else` statements?

src/commands/lambda/flare.ts
```diff
@@ -1,14 +1,59 @@
...
 
 export class LambdaFlareCommand extends Command {
   private isDryRun = false
+  private isInteractive = false
+  private allFunctions = false
+  private functions: string[] = []
+  private region = ''
+  private apiKey = ''
+  private caseID = ''
+  private email = ''
```
duncanista, line 13: Let's not add an initial value for this. Make it a nullable and add typing.

duncanista (changes requested):
Couple comments

src/commands/lambda/flare.ts
```diff
@@ -1,14 +1,41 @@
...
+    }
+
+    return 0
   }
 }
 
 LambdaFlareCommand.addPath('lambda', 'flare')
 LambdaFlareCommand.addOption('isDryRun', Command.Boolean('-d,--dry'))
+LambdaFlareCommand.addOption('isInteractive', Command.Boolean('-i,--interactive'))
+// LambdaFlareCommand.addOption('allFunctions', Command.Boolean('--all-functions'))
```
duncanista, line 36: Remove it, not just comment it.

src/commands/lambda/flare-command-validator.ts
```diff
@@ -0,0 +1,89 @@
+// Uses the Chain of Responsibility design pattern to validate the command
+// eslint-disable-next-line max-classes-per-file
+abstract class RequestHandler {
```
duncanista, line 3: Can we change the naming to something related to validation? Like `Validator`?

Also, what is a request? And what is the typing `{[key: string]: any}`?

src/commands/lambda/flare-command-validator.ts
```diff
@@ -0,0 +1,89 @@
...
+  private nextHandler: RequestHandler | undefined
+
+  public setNext(handler: RequestHandler): RequestHandler {
+    this.nextHandler = handler
+
+    return handler
+  }
+
+  public handle(request: {[key: string]: any}): string | undefined {
```
duncanista, line 12: What does `handle` mean? I don't understand the name since what we are doing is checking if something is valid.

src/commands/lambda/flare-command-validator.ts
```diff
@@ -0,0 +1,89 @@
+// Uses the Chain of Responsibility design pattern to validate the command
```
duncanista, line 1: Document this on the class to extend, not on the file.

src/commands/lambda/flare-command-validator.ts
```diff
@@ -0,0 +1,89 @@
...
+  public handle(request: {[key: string]: any}): string | undefined {
+    if (this.nextHandler) {
+      return this.nextHandler.handle(request)
+    }
+
+    return undefined
+  }
+}
+
+// Concrete handlers
```
duncanista, line 21: Can we document these classes?

src/commands/lambda/flare-command-validator.ts
```diff
@@ -0,0 +1,89 @@
...
+const functionsCheckHandler = new FunctionsCheckHandler()
+const regionCheckHandler = new RegionCheckHandler()
+const apiKeyCheckHandler = new ApiKeyCheckHandler()
+const emailCheckHandler = new EmailCheckHandler()
+
+interactiveCheckHandler
+  .setNext(functionsCheckHandler)
+  .setNext(regionCheckHandler)
+  .setNext(apiKeyCheckHandler)
+  .setNext(emailCheckHandler)
```
duncanista, line 82: Wouldn't this be better in a Chain Builder?

src/commands/lambda/renderers/flare-renderer.ts
```diff
@@ -13,3 +20,53 @@ export const renderLambdaFlareHeader = (isDryRun: boolean) => {
...
+
+/**
+ * @param warning the message to warn about.
+ * @returns the provided warning prefixed by {@link warningTag}.
+ *
+ * ```txt
+ * [Warning] The provided warning goes here!
+ * ```
+ */
+export const renderWarning = (warning: string) => `${warningTag} ${warning}\n`
```
duncanista, line 42: Where is thing being used?
The same goes for the constants below.

src/commands/lambda/flare-command-validator.ts
```diff
@@ -0,0 +1,89 @@
+// Uses the Chain of Responsibility design pattern to validate the command
+// eslint-disable-next-line max-classes-per-file
```
duncanista, line 2: We need tests for this file too.

duncanista (changes requested):
Left some comments

src/commands/lambda/flare-command-validator.ts
nhulston: Renamed!
`request` is a bad name. It's given by passing in `this` from `flare.ts` -- it's an object with all of the variables `isDryRun`, `isInteractive`, `functions`, etc.

I have renamed `request` to `flags`

`flags = {
  isDryFun: false,
  isInteractive: false,
  ...
}`

src/commands/lambda/flare-command-validator.ts
nhulston: The method name of `handle` is based on the Chain of Responsibility pattern
https://refactoring.guru/design-patterns/chain-of-responsibility
Do you think I should rename it to something like `validate`?

src/commands/lambda/renderers/flare-renderer.ts
nhulston: These are methods I plan on using later in the project. Should they be removed right now?

duncanista: If we are not using them now, let's not add them until we get there.

src/commands/lambda/flare-command-validator.ts
duncanista: I understand. But this classes are meant to be used as a blueprint in which you then transform for the use case. In this case, `handle` should be something meaningful to what we are doing here.

nhulston: I see, that makes sense. Renamed

src/commands/lambda/__tests__/fixtures.ts
```diff
@@ -39,6 +39,12 @@ export const createMockContext = () => {
         data += input
       },
     },
+    stderr: {
+      toString: () => data,
+      write: (input: string) => {
+        data += input
+      },
+    },
```
nhulston, line 47: Note that this is needed to properly test outputs from `stderr`

package.json
```diff
@@ -76,6 +76,7 @@
     "inquirer": "^8.2.5",
     "inquirer-checkbox-plus-prompt": "^1.4.2",
     "js-yaml": "3.13.1",
+    "jszip": "^3.10.1",
```
nhulston, line 79: Package required for compatibility on Windows, Mac, and Linux

src/commands/lambda/flare.ts
```diff
@@ -1,14 +1,194 @@
...
+            fs.unlinkSync(currentPath)
+          }
+        })
+      }
+    }
+    deleteFolderContents(folderPath)
+    try {
+      fs.rmdirSync(folderPath)
+    } catch (e) {
+      this.context.stderr.write(renderSoftWarning('Failed to delete log files located at .lambda-flare-output'))
```
nhulston, line 179: `renderSoftWarning` or `renderError`?

src/commands/lambda/renderers/flare-renderer.ts
```diff
@@ -1,8 +1,8 @@
-import {dryRunTag} from './common-renderer'
+import {dryRunTag, errorTag, warningExclamationSignTag} from './common-renderer'
 
 /**
  * @returns a header indicating which `lambda` subcommand is running.
- * @param command current selected lambda subcommand.
+ * @param isDryRun whether or not the command is a dry run. Defaults to false.
```
duncanista, line 5: This is not true, since the function signature is not `= (isDryRun = false)`. So there is no default.

src/commands/lambda/renderers/flare-renderer.ts
```diff
@@ -11,5 +11,25 @@ import {dryRunTag} from './common-renderer'
...
+
+/**
+ * @param error an error message
+ * @returns the provided error prefixed by {@link errorTag}.
+ *
+ * ```txt
+ * [Error] The provided error goes here!
+ * ```
+ */
+export const renderError = (error: string) => `\n${errorTag} ${error}\n`
```
duncanista, line 25: This seems to be used in https://github.com/DataDog/datadog-ci/blob/df721c1b57bd2925bd008ab9f15d4cc50e4d7ec5/src/commands/lambda/renderers/instrument-uninstrument-renderer.ts#L144

Can we move them into `common-renderer` to reuse the code?

src/commands/lambda/flare.ts
```diff
@@ -1,14 +1,194 @@
...
+  private functionName?: string
+  private region?: string
+  private apiKey?: string
+  private caseId?: string
+  private email?: string
+  private credentials?: AwsCredentialIdentity
 
+  /**
+   * @returns 0 if the command ran successfully, 1 otherwise.
+   */
```
duncanista, line 28: This is not a very good explanation of what `execute` does 😅

src/commands/lambda/flare.ts
```diff
@@ -1,14 +1,194 @@
...
+    if (
+      this.region === undefined &&
+      this.functionName !== undefined &&
+      this.functionName.startsWith('arn:aws:lambda')
+    ) {
+      this.region = this.functionName.split(':')[3]
+    }
+    this.region = this.region ?? process.env[AWS_DEFAULT_REGION_ENV_VAR]
+    if (this.region === undefined) {
+      this.context.stderr.write(renderError('No region specified. [-r,--region]'))
```
duncanista, line 49: I think there is also a renderer that already does this, would it make sense to use it too?

src/commands/lambda/flare.ts
```diff
@@ -1,14 +1,194 @@
...
+
+    // Validate region
+    if (
+      this.region === undefined &&
+      this.functionName !== undefined &&
+      this.functionName.startsWith('arn:aws:lambda')
+    ) {
+      this.region = this.functionName.split(':')[3]
+    }
+    this.region = this.region ?? process.env[AWS_DEFAULT_REGION_ENV_VAR]
```
duncanista, line 47:
```suggestion
    if (
      this.region === undefined &&
      this.functionName !== undefined &&
      this.functionName.startsWith('arn:aws:lambda')
    ) {
      this.region = this.functionName.split(':')[3]
    }
    this.region = this.region ?? process.env[AWS_DEFAULT_REGION_ENV_VAR]
```

Also, there is already a function that gives you the region given an ARN.

src/commands/lambda/flare.ts
```diff
@@ -1,14 +1,194 @@
...
+    if (this.region === undefined) {
+      this.context.stderr.write(renderError('No region specified. [-r,--region]'))
+      errorFound = true
+    }
+
+    // Validate Datadog API key
+    this.apiKey = this.apiKey ?? process.env[CI_API_KEY_ENV_VAR] ?? process.env[API_KEY_ENV_VAR]
+    if (this.apiKey === undefined) {
+      this.context.stderr.write(renderError('No Datadog API key specified. [--api-key]'))
```
duncanista, line 56: Let's be more descriptive here, since user can directly pass their API key with this option, or the environment variable `DD_API_KEY`. I'm not sure if we should tell the user the existence of `DATADOG_API_KEY`, can you ask the `datadog-ci` team why they created the tool with that environment variable and not `DD_API_KEY`?

src/commands/lambda/flare.ts
```diff
@@ -1,14 +1,194 @@
...
+
+    // Get AWS credentials
+    this.context.stdout.write('\n🔑 Getting AWS credentials...\n')
+    const credentials = await getAWSCredentials()
+    if (credentials === undefined) {
+      this.context.stdout.write(renderSoftWarning("No AWS credentials found, let's set them up!"))
+      await requestAWSCredentials()
+    } else {
+      this.credentials = credentials
+    }
```
duncanista, line 84:
```suggestion
    this.credentials = await getAWSCredentials()
    if (this.credentials === undefined) {
      this.context.stdout.write(renderSoftWarning("No AWS credentials found, let's set them up!"))
      await requestAWSCredentials()
    }
```

Also, the renderer I think already exists for `instrument|uninstrument` can we reuse it too?

Also, there is no try-catch block in case `getAWSCredentials` or `requestAWSCredentials` fail.

src/commands/lambda/flare.ts
```diff
@@ -1,14 +1,194 @@
...
+    // Get Lambda function configuration
+    this.context.stdout.write('🔍 Getting Lambda function configuration...\n')
+    const lambdaClientConfig: LambdaClientConfig = {
+      region: this.region,
+      credentials: this.credentials,
+    }
+    const lambdaClient = new LambdaClient(lambdaClientConfig)
+    if (this.functionName === undefined) {
+      throw new Error('Function name is undefined')
+    }
```
duncanista, line 95:
```suggestion
    this.context.stdout.write('🔍 Getting Lambda function configuration...\n')
    const lambdaClientConfig: LambdaClientConfig = {
      region: this.region,
      credentials: this.credentials,
    }
    const lambdaClient = new LambdaClient(lambdaClientConfig)
    if (this.functionName === undefined) {
      throw new Error('Function name is undefined')
    }
```

Why don't we first check if `this.functionName` is undefined to render the error?

Also, don't throw errors. Instead, do a `stderr` write.

src/commands/lambda/flare.ts
```diff
@@ -1,14 +1,194 @@
...
+    const config = await getLambdaFunctionConfig(lambdaClient, this.functionName)
```
duncanista, line 96: Unhandled Promise Rejection here, no try-catch block is being used in case this function fails.

src/commands/lambda/flare.ts
```diff
@@ -1,14 +1,194 @@
...
+    const configStrColored = util.inspect(config, false, undefined, true)
```
duncanista, line 97: What does this line do?

src/commands/lambda/flare.ts
```diff
@@ -1,14 +1,194 @@
...
+      return 0
+    }
+
+    // Write config to file
+    const folderPath = path.join(process.cwd(), '.lambda-flare-output')
+    if (!fs.existsSync(folderPath)) {
+      fs.mkdirSync(folderPath)
+    }
+    const filePath = path.join(folderPath, 'function_config.json')
+    fs.writeFileSync(filePath, configStrUncolored)
```
duncanista, line 114:
```suggestion
    // Print config
    this.context.stdout.write(`\n${configStrColored}\n`)
    if (this.isDryRun) {
      this.context.stdout.write('\n🚫 Configuration not sent because the command was ran as a dry run.\n')

      return 0
    }

    // Write config to file
    const folderPath = path.join(process.cwd(), '.lambda-flare-output')
    if (!fs.existsSync(folderPath)) {
      fs.mkdirSync(folderPath)
    }
    const filePath = path.join(folderPath, 'function_config.json')
    const configStr = JSON.stringify(config, undefined, 2)
    fs.writeFileSync(filePath, configStr)
```

src/commands/lambda/flare.ts
```diff
@@ -1,14 +1,194 @@
...
+      throw new Error('Function name is undefined')
+    }
+    const config = await getLambdaFunctionConfig(lambdaClient, this.functionName)
+    const configStrColored = util.inspect(config, false, undefined, true)
+    const configStrUncolored = JSON.stringify(config, undefined, 2)
+
+    // Print config
+    this.context.stdout.write(`\n${configStrColored}\n`)
+    if (this.isDryRun) {
+      this.context.stdout.write('\n🚫 Configuration not sent because the command was ran as a dry run.\n')
```
duncanista, line 103:
```suggestion
      this.context.stdout.write('\n🚫 The configuration was not sent as it was executed in dry run mode.\n')
```

src/commands/lambda/flare.ts
```diff
@@ -1,14 +1,194 @@
...
+
+      return 0
+    }
+
+    // Write config to file
+    const folderPath = path.join(process.cwd(), '.lambda-flare-output')
```
duncanista, line 109: Change to `.datadog-ci` but make it a constant

```suggestion
    const folderPath = path.join(process.cwd(), FLARE_OUTPUT_DIRECTORY)
```

src/commands/lambda/flare.ts
```diff
@@ -1,14 +1,194 @@
...
+    if (!fs.existsSync(folderPath)) {
+      fs.mkdirSync(folderPath)
+    }
+    const filePath = path.join(folderPath, 'function_config.json')
```
duncanista, line 113: Make it a constant too. For example, feel free to change the name of the constant if needed.
```suggestion
    const filePath = path.join(folderPath, FUNCTION_CONFIG_FILE)
```

src/commands/lambda/flare.ts
duncanista:
```suggestion
    // Print config
    this.context.stdout.write(`\n${configStrColored}\n`)
    if (this.isDryRun) {
      this.context.stdout.write('\n🚫 Configuration not sent because the command was ran as a dry run.\n')

      return 0
    }

    // Write config to file
    const folderPath = path.join(process.cwd(), '.lambda-flare-output')
    if (!fs.existsSync(folderPath)) {
      fs.mkdirSync(folderPath)
    }
    const filePath = path.join(folderPath, 'function_config.json')
    const configStr = JSON.stringify(config, undefined, 2)
    fs.writeFileSync(filePath, configStr)
```

src/commands/lambda/flare.ts
```diff
@@ -1,14 +1,194 @@
...
+    const zipPath = path.join(folderPath, 'lambda-flare-output.zip')
+    const zipFolder = async (command: Command) => {
+      try {
+        const data = await fs.promises.readFile(filePath, 'utf8')
+        const zip = new JSZip()
+        zip.file('function_config.json', data)
+        const content = await zip.generateAsync({type: 'nodebuffer'})
+        await fs.promises.writeFile(zipPath, content)
+      } catch (error) {
+        command.context.stderr.write(renderError('Error reading file or writing zip file. Please try again.'))
```
duncanista, line 126: This would output `[Error] Error reading...`. Let's change the wording here so we don't repeat the word Error two times.

src/commands/lambda/flare.ts
```diff
@@ -1,14 +1,194 @@
...
+    fs.writeFileSync(filePath, configStrUncolored)
+
+    // Zip folder
+    const zipPath = path.join(folderPath, 'lambda-flare-output.zip')
+    const zipFolder = async (command: Command) => {
+      try {
+        const data = await fs.promises.readFile(filePath, 'utf8')
+        const zip = new JSZip()
+        zip.file('function_config.json', data)
+        const content = await zip.generateAsync({type: 'nodebuffer'})
```
duncanista, line 123: Why `nodebuffer`?

src/commands/lambda/flare.ts
```diff
@@ -1,14 +1,194 @@
...
+        command.context.stderr.write(renderError('Error reading file or writing zip file. Please try again.'))
+
+        return true
+      }
+
+      return false
+    }
+    if (await zipFolder(this)) {
+      return 1
+    }
```
duncanista, line 135: I don't like having a function which is going to be executed on the next line after definition and not being used again.
Let's not make this a function and just handle everything on the execution level.

If we are trying to hide relevant zipping logic so it looks cleaner, then let's get this function outside of the `execute` function and make it its own function.

Whichever you think is better, but let's not do this, since we are also sending the whole instance just to use the context to output.

src/commands/lambda/flare.ts
```diff
@@ -1,14 +1,194 @@
...
+
+    // Send to Datadog
+    this.context.stdout.write('\n🚀 Sending to Datadog Support...\n')
+    const form = new FormData()
+    form.append('case_id', this.caseId)
+    form.append('flare_file', fs.createReadStream(zipPath))
```
duncanista, line 141: Can this throw a rejection?

src/commands/lambda/flare.ts
```diff
@@ -1,14 +1,194 @@
...
+    form.append('operator_version', 7)
+    form.append('email', this.email)
+    const requestConfig = {
+      headers: {
+        ...form.getHeaders(),
```
duncanista, line 146: Why are we doing this?

src/commands/lambda/flare.ts
```diff
@@ -1,14 +1,194 @@
...
+        'DD-API-KEY': this.apiKey,
+      },
+    }
+    await axios
+      .post('https://datad0g.com/api/ui/support/serverless/flare', form, requestConfig)
```
duncanista, line 151: Make this a constant, also, do we need `https://`?

src/commands/lambda/flare.ts
```diff
@@ -1,14 +1,194 @@
...
+      .then(() => {
+        this.context.stdout.write('\n✅ Successfully sent function config to Datadog Support!\n')
+      })
+      .catch((err) => {
+        this.context.stderr.write(
+          renderError('Failed to send function config to Datadog Support. Is your email and case ID correct?')
+        )
+        this.context.stderr.write(renderError(err))
+      })
```
duncanista, line 160: Turn this into a try-catch block. Let's not use `then-catch`.

src/commands/lambda/flare.ts
```diff
@@ -1,14 +1,194 @@
...
+          const currentPath = path.join(dir, file)
+          if (fs.lstatSync(currentPath).isDirectory()) {
+            deleteFolderContents(currentPath)
+          } else {
+            fs.unlinkSync(currentPath)
+          }
+        })
+      }
+    }
+    deleteFolderContents(folderPath)
```
duncanista, line 175: Same observations as the function `zipFolder`.

src/commands/lambda/flare.ts
duncanista: Soft warning is perfect. Also what do you mean by `log files`? Wouldn't it be `Flare files?` Also, if you are going to give them the path, dont be arbritrary and specify the full path. The folder instance must have a `toString` that gives you the full path.

duncanista (changes requested):
Amazing work Nick! Left a couple comments. This is taking a really great shape.

src/commands/lambda/flare.ts
nhulston: `configStrColored` is used to print on the terminal.
`configStrUncolored` is used to write to the JSON file.
The reason is because if we try to write the colored version to the JSON file, there will be lots of weird color codes (and it won't be in valid JSON format)

src/commands/lambda/flare.ts
nhulston: Yes, they can return rejections. Good catch!

src/commands/lambda/flare.ts
nhulston: Good catch!

src/commands/lambda/flare.ts
nhulston: Yes good catch!

src/commands/lambda/flare.ts
nhulston: I believe it's the only way to add a custom header when using `axios.post`. However, I've renamed `requestConfig` to `headerConfig` to make it more claer

src/commands/lambda/flare.ts
nhulston: Yes looks like `https://` is required, post request fails if `https://` is not included

src/commands/lambda/uninstrument.ts
```diff
@@ -18,6 +18,11 @@ import {
 import {getUninstrumentedFunctionConfigs, getUninstrumentedFunctionConfigsFromRegEx} from './functions/uninstrument'
 import {FunctionConfiguration} from './interfaces'
 import {requestAWSCredentials, requestChangesConfirmation, requestFunctionSelection} from './prompt'
+import {
+  renderError,
+  renderNoAWSCredentialsFound,
+  renderNoDefaultRegionSpecifiedError,
+} from './renderers/common-renderer'
 import * as renderer from './renderers/instrument-uninstrument-renderer'
```
duncanista, line 26: Is there a way to have these imports as one?
```suggestion
import * as commonRenderer from './renderers/common-renderer'
import * as renderer from './renderers/instrument-uninstrument-renderer' // maybe change `renderer` to something else?
```

This is so we don't have tons of imports from renderers. 
We might need to change the name of whole imports to be descriptive.

src/commands/lambda/functions/commons.ts
```diff
@@ -39,6 +39,7 @@ import {
 import {FunctionConfiguration, InstrumentationSettings, InstrumentedConfigurationGroup} from '../interfaces'
 import {applyLogGroupConfig} from '../loggroup'
 import {awsProfileQuestion} from '../prompt'
+import {renderSoftWarning} from '../renderers/common-renderer'
```
duncanista, line 42: Same as `instrument.ts` and `uninstrument.ts`

src/commands/lambda/flare.ts
```diff
@@ -1,14 +1,229 @@
...
+import {API_KEY_ENV_VAR, AWS_DEFAULT_REGION_ENV_VAR, CI_API_KEY_ENV_VAR} from './constants'
+import {getAWSCredentials, getLambdaFunctionConfig, getRegion} from './functions/commons'
+import {requestAWSCredentials} from './prompt'
+import {
+  renderError,
+  renderNoAWSCredentialsFound,
+  renderNoDefaultRegionSpecifiedError,
+  renderSoftWarning,
+} from './renderers/common-renderer'
 import {renderLambdaFlareHeader} from './renderers/flare-renderer'
```
duncanista, line 21: Let's change these imports to `import * as ... from ...`

src/commands/lambda/flare.ts
```diff
@@ -1,14 +1,229 @@
...
+      errorFound = true
+    }
+
+    // Validate region
+    this.region = this.region ?? getRegion(this.functionName ?? '')
+    this.region = this.region ?? process.env[AWS_DEFAULT_REGION_ENV_VAR]
+    if (this.region === undefined) {
+      this.context.stderr.write(renderNoDefaultRegionSpecifiedError())
+      errorFound = true
+    }
```
duncanista, line 53: Let's refactor this code based on our conversation – let's make this a constant, keeping the provided customer value in `this.region`. 
Also, let's prioritize this way: 
1. Region from ARN.
2. Region from `this.region`.
3. AWS default region.

src/commands/lambda/flare.ts
```diff
@@ -1,14 +1,229 @@
...
+      try {
+        await requestAWSCredentials()
+      } catch (err) {
+        this.context.stderr.write(renderError(err))
+
+        return 1
+      }
+    } else {
+      this.credentials = credentials
+    }
```
duncanista, line 99:
```suggestion
    try {
      this.credentials = await getAWSCredentials()
    } catch (err) {
      this.context.stderr.write(renderError(err))

      return 1
    }
    if (this.credentials === undefined) {
      this.context.stdout.write(renderNoAWSCredentialsFound())
      try {
        await requestAWSCredentials()
      } catch (err) {
        this.context.stderr.write(renderError(err))

        return 1
      }
    }
```

src/commands/lambda/flare.ts
```diff
@@ -1,14 +1,229 @@
...
+
+    // Get Lambda function configuration
+    if (this.functionName === undefined) {
+      this.context.stderr.write(renderError('Function name is undefined.'))
+
+      return 1
+    }
```
duncanista, line 106: Already being checked. Removal needed.

src/commands/lambda/flare.ts
```diff
@@ -1,14 +1,229 @@
...
+    // Print config
+    this.context.stdout.write(`\n${configStrColored}\n`)
+    if (this.isDryRun) {
+      this.context.stdout.write('\n🚫 The configuration was not sent as it was executed in dry run mode.\n')
+
+      return 0
+    }
+
+    // Write config to file
+    const FLARE_OUTPUT_DIRECTORY = '.datadog-ci'
```
duncanista, line 133: Constants at the top of the file.

src/commands/lambda/flare.ts
```diff
@@ -1,14 +1,229 @@
...
+    if (!fs.existsSync(folderPath)) {
+      try {
+        fs.mkdirSync(folderPath)
+      } catch (err) {
+        this.context.stderr.write(renderError(`Unable to create flare folder: ${err}`))
+
+        return 1
+      }
+    }
+    const FUNCTION_CONFIG_FILE_NAME = 'function_config.json'
```
duncanista, line 144: Same here.

src/commands/lambda/flare.ts
```diff
@@ -1,14 +1,229 @@
...
+    }
+    form.append('operator_version', 7)
+    form.append('email', this.email)
+    const headerConfig = {
+      headers: {
+        ...form.getHeaders(),
+        'DD-API-KEY': this.apiKey,
+      },
+    }
+    const ENDPOINT_URL = 'https://datad0g.com/api/ui/support/serverless/flare'
```
duncanista, line 187: Same here.

src/commands/lambda/flare.ts
```diff
@@ -1,14 +1,229 @@
...
+    try {
+      await axios.post(ENDPOINT_URL, form, headerConfig).then(() => {
+        this.context.stdout.write('\n✅ Successfully sent function config to Datadog Support!\n')
+      })
+    } catch (err) {
+      this.context.stderr.write(renderError(`Failed to send function config to Datadog Support: ${err}`))
```
duncanista, line 193: Get error from response data.

src/commands/lambda/flare.ts
```diff
@@ -1,14 +1,229 @@
...
+      zip.file(FUNCTION_CONFIG_FILE_NAME, data)
+      const content = await zip.generateAsync({type: 'nodebuffer'})
+      await fs.promises.writeFile(zipPath, content)
+    } catch (err) {
+      this.context.stderr.write(renderError(`Unable to zip the flare file: ${err}`))
+
+      return 1
+    }
+
+    // Send to Datadog
```
duncanista, line 168: Can we move this into a function?

src/commands/lambda/flare.ts
```diff
@@ -1,14 +1,229 @@
...
+
+    // Print config
+    this.context.stdout.write(`\n${configStrColored}\n`)
+    if (this.isDryRun) {
+      this.context.stdout.write('\n🚫 The configuration was not sent as it was executed in dry run mode.\n')
+
+      return 0
+    }
+
+    // Write config to file
```
duncanista, line 132: Can we move all this code into a function too?
For readability.

duncanista (changes requested):
Amazing work!! Bit of refactoring.

src/commands/lambda/flare.ts
duncanista: We had a chat about this, so closing it.

src/commands/lambda/flare.ts
```diff
@@ -1,14 +1,239 @@
...
+    const folderPath = path.join(process.cwd(), FLARE_OUTPUT_DIRECTORY)
+    const filePath = path.join(folderPath, FUNCTION_CONFIG_FILE_NAME)
+    const zipPath = path.join(folderPath, ZIP_FILE_NAME)
+    const configStrUncolored = JSON.stringify(config, undefined, 2)
+    try {
+      await this.writeFile(folderPath, filePath, configStrUncolored)
+      await this.zipContents(filePath, zipPath)
+      this.context.stdout.write('\n🚀 Sending to Datadog Support...\n')
+      await this.sendToDatadog(zipPath)
+      this.deleteFolder(folderPath)
```
duncanista, line 139: This should be something like
```suggestion
      this.context.stderr.write(renderError(err.message))
```

src/commands/lambda/flare.ts
```diff
@@ -1,14 +1,239 @@
...
+   * Write the function config to a file
+   * @param folderPath
+   * @param filePath
+   * @param data
+   * @throws Error if the file cannot be written
+   */
+  private writeFile = async (folderPath: string, filePath: string, data: string) => {
+    if (!fs.existsSync(folderPath)) {
+      try {
+        fs.mkdirSync(folderPath)
```
duncanista, line 159: This should just throw the error as `throw Error('message here')`

src/commands/lambda/flare.ts
```diff
@@ -1,14 +1,239 @@
...
+      } catch (err) {
+        throw Error(commonRenderer.renderError(`Unable to create flare folder: ${err.message}`))
+      }
+    }
+    try {
+      fs.writeFileSync(filePath, data)
```
duncanista, line 165: Same here

src/commands/lambda/flare.ts
```diff
@@ -1,14 +1,239 @@
...
+   * @param zipPath
+   * @throws Error if the zip fails
+   */
+  private zipContents = async (filePath: string, zipPath: string) => {
+    try {
+      const data = await fs.promises.readFile(filePath, 'utf8')
+      const zip = new JSZip()
+      zip.file(FUNCTION_CONFIG_FILE_NAME, data)
+      const content = await zip.generateAsync({type: 'nodebuffer'})
+      await fs.promises.writeFile(zipPath, content)
```
duncanista, line 183: Same here

src/commands/lambda/flare.ts
```diff
@@ -1,14 +1,239 @@
...
+  private sendToDatadog = async (zipPath: string) => {
+    const form = new FormData()
+    form.append('case_id', this.caseId)
+    form.append('flare_file', fs.createReadStream(zipPath))
+    form.append('operator_version', 7)
+    form.append('email', this.email)
+    const headerConfig = {
+      headers: {
+        ...form.getHeaders(),
+        'DD-API-KEY': this.apiKey,
```
duncanista, line 203: Same here

src/commands/lambda/flare.ts
```diff
@@ -1,14 +1,239 @@
...
+    try {
+      fs.rmSync(folderPath, {recursive: true, force: true})
+    } catch (err) {
+      throw new Error(`Failed to delete files located at ${folderPath}: ${err.message}`)
+    }
   }
 }
 
 LambdaFlareCommand.addPath('lambda', 'flare')
 LambdaFlareCommand.addOption('isDryRun', Command.Boolean('-d,--dry'))
```
duncanista, line 235: Same here

src/commands/lambda/__tests__/flare.test.ts
```diff
@@ -1,12 +1,64 @@
...
+  },
+  FunctionArn: 'arn:aws:lambda:us-east-1:123456789012:function:some-function',
+  FunctionName: 'some-function',
+}
+
+jest.mock('../functions/commons', () => ({
+  getAWSCredentials: jest.fn(),
+  getLambdaFunctionConfig: jest.fn().mockImplementation(() => Promise.resolve(mockConfig)),
+  getRegion: jest.requireActual('../functions/commons').getRegion as () => string | undefined,
+}))
```
duncanista, line 29:
```suggestion
jest.mock('../functions/commons', () => ({
  ...jest.requireActual('../functions/commons')
  getAWSCredentials: jest.fn(),
  getLambdaFunctionConfig: jest.fn().mockImplementation(() => Promise.resolve(mockConfig))
}))
```

src/commands/lambda/__tests__/flare.test.ts
```diff
@@ -1,12 +1,64 @@
...
+
+jest.mock('../prompt', () => ({
+  requestAWSCredentials: jest.fn(),
+}))
```
duncanista, line 33: Wouldn't this do the same?
```suggestion
jest.mock('../prompt')
```

src/commands/lambda/__tests__/flare.test.ts
```diff
@@ -1,12 +1,64 @@
...
+    const mockReadStream = new stream.Readable({
+      read() {
+        this.push('mock file content')
+        this.push(undefined)
+      },
+    })
+    fs.promises.writeFile = jest.fn().mockResolvedValue(() => {})
+    fs.promises.readFile = jest.fn().mockResolvedValue(JSON.stringify(mockConfig))
+    fs.createReadStream = jest.fn().mockReturnValue(mockReadStream)
+  })
```
duncanista, line 50: Why are we mocking on `beforeEach`? Are there some tests that don't need this mocking?

src/commands/lambda/__tests__/__snapshots__/flare.test.ts.snap
```diff
@@ -1,13 +1,156 @@
...
+
+🔑 Getting AWS credentials...
+[!] No AWS credentials found, let's set them up! Or you can re-run the command and supply the AWS credentials in the same way when you invoke the AWS CLI.
+
+🔍 Getting Lambda function configuration...
+
+{
+  Environment: {
+    Variables: {
+      DD_API_KEY: [32m'some-api-key'[39m,
```
duncanista, line 40: Mock globally the package that does this, this snapshots are not human readable and can be confusing if someone is debugging them. An example is how we mock `chalk`.

src/commands/lambda/__tests__/flare.test.ts
```diff
@@ -15,7 +67,192 @@ describe('lambda flare', () => {
...
+        'arn:aws:lambda:us-west-2:123456789012:function:my-function',
+        '-c',
+        '123',
+        '-e',
+        'test@test.com',
+      ],
+      context as any
+    )
     expect(code).toBe(0)
+  })
```
duncanista, line 112: Missing snapshot.

src/commands/lambda/__tests__/flare.test.ts
```diff
@@ -15,7 +67,192 @@ describe('lambda flare', () => {
...
+    expect(code).toBe(0)
+
+    process.env[CI_API_KEY_ENV_VAR] = undefined
+    process.env[API_KEY_ENV_VAR] = 'test-api-key'
+    code = await cli.run(
+      ['lambda', 'flare', '-f', 'func', '-r', 'test-region', '-c', '123', '-e', 'test@test.com'],
+      context as any
+    )
+    expect(code).toBe(0)
+  })
```
duncanista, line 154: This could be its own describe for API KEY testing

src/commands/lambda/__tests__/flare.test.ts
```diff
@@ -15,7 +67,192 @@ describe('lambda flare', () => {
...
+    const code = await cli.run(
+      ['lambda', 'flare', '-f', 'func', '-r', 'us-west-2', '-c', '123', '-e', 'test@test.com'],
+      context as any
+    )
+    expect(code).toBe(1)
+    const output = context.stderr.toString()
+    expect(output).toMatchSnapshot()
+  })
+
+  it('uses API key ENV variable', async () => {
```
duncanista, line 135: This test is not descriptive plus it does two different things.

src/commands/lambda/__tests__/flare.test.ts
```diff
@@ -15,7 +67,192 @@ describe('lambda flare', () => {
...
+    const output = context.stderr.toString()
+    expect(output).toMatchSnapshot()
+  })
+
+  it('requests AWS credentials when none are found and logs output correctly', async () => {
+    const cli = makeCli()
+    const context = createMockContext()
+    await cli.run(
+      ['lambda', 'flare', '-f', 'func', '-r', 'us-west-2', '-c', '123', '-e', 'test@test.com'],
+      context as any
```
duncanista, line 182: expect `code` to be `0` is missing

src/commands/lambda/__tests__/flare.test.ts
```diff
@@ -15,7 +67,192 @@ describe('lambda flare', () => {
...
+      expect.objectContaining({
+        headers: expect.objectContaining({
+          'DD-API-KEY': 'test-api-key',
+        }) as Record<string, string>,
+      })
+    )
+    const output = context.stdout.toString()
+    expect(output).toMatchSnapshot()
+    postSpy.mockRestore()
+  })
```
duncanista, line 231: I like how we are separating what we test, but I found it a little bit confusing because the first and last test are interchangable in the sense that they both throw the same data? How could we improve this?

src/commands/lambda/__tests__/flare.test.ts
```diff
@@ -15,7 +67,192 @@ describe('lambda flare', () => {
...
+    const postSpy = (axios.post = jest.fn().mockRejectedValue({status: 500}))
+    const cli = makeCli()
+    const context = createMockContext()
+    await cli.run(
+      ['lambda', 'flare', '-f', 'func', '-r', 'us-west-2', '-c', '123', '-e', 'test@test.com', '-d'],
+      context as any
+    )
+    expect(postSpy).not.toHaveBeenCalled()
+    const output = context.stdout.toString()
+    expect(output).toMatchSnapshot()
```
duncanista, line 243: Missing `code` expect to be `0`

src/commands/lambda/__tests__/flare.test.ts
```diff
@@ -15,7 +67,192 @@ describe('lambda flare', () => {
...
+    axios.post = jest.fn().mockRejectedValue({status: 500})
+    const cli = makeCli()
+    const context = createMockContext()
+    await cli.run(
+      ['lambda', 'flare', '-f', 'func', '-r', 'us-west-2', '-c', '123', '-e', 'test@test.com'],
+      context as any
+    )
+    const output = context.stderr.toString()
     expect(output).toMatchSnapshot()
   })
```
duncanista, line 257: I've noticed we only test happy paths, but we never actually check for some errors to fail? What would the error message for this look like? Also `fails to send request to Datadog` isn't that descriptive. What about something like `throws an error when post request to Datadog fails`? Also missing the `code` expect to be `1`.

duncanista (changes requested):
We need to improve tests, some are not mocking what happens when a specific step fails. We could do a `describe` per step/section. I.e. when the tool gets data, zips, sends post request, deletes folders, etc.

src/commands/lambda/__tests__/flare.test.ts
duncanista: I don't understand, this was resolved, but no response was given in this thread, and the code is still the same. Why was this marked as resolved?

src/commands/lambda/__tests__/flare.test.ts
```diff
@@ -1,21 +1,444 @@
...
+  },
+  FunctionArn: 'arn:aws:lambda:us-east-1:123456789012:function:some-function',
+  FunctionName: 'some-function',
+}
+
+// Mock util.inspect to remove color codes
+const originalInspect = util.inspect
+jest.spyOn(util, 'inspect').mockImplementation((object: any, options?: InspectOptions) => {
+  return originalInspect(object, {...options, colors: false})
+})
```
duncanista, line 38: I suggested mocking globally but this is doing it in this file, what is the advantage of doing it here? and, why are we spying on it instead of mocking it? Wouldn't it make more sense to mock it like the module `chalk` was mocked globally?

src/commands/lambda/__tests__/flare.test.ts
```diff
@@ -1,21 +1,444 @@
...
+import {LambdaFlareCommand, writeFile, zipContents} from '../flare'
+import {requestAWSCredentials} from '../prompt'
+
+import {createMockContext, makeCli, mockAwsCredentials} from './fixtures'
+
+// Constants
+const MOCK_FOLDER_PATH = './mock_folder'
+const MOCK_FILE_PATH = 'function_config.json'
+const MOCK_ZIP_PATH = 'output.zip'
+const MOCK_API_KEY = 'test-api-key'
```
duncanista, line 20: There is already a mock in `fixtures.ts` of this.

src/commands/lambda/__tests__/flare.test.ts
```diff
@@ -1,21 +1,444 @@
...
+jest.mock('fs')
+fs.writeFileSync = jest.fn().mockImplementation(() => {})
+fs.readFileSync = jest.fn().mockReturnValue(JSON.stringify(MOCK_CONFIG, undefined, 2))
+const mockReadStream = new stream.Readable({
+  read() {
+    this.push(JSON.stringify(MOCK_CONFIG, undefined, 2))
+    this.push(undefined)
+  },
+})
+fs.createReadStream = jest.fn().mockReturnValue(mockReadStream)
```
duncanista, line 58: Can we have this mocks inside the `jest.mock`? Just like you did for mocking `commons`.

src/commands/lambda/__tests__/flare.test.ts
```diff
@@ -1,21 +1,444 @@
...
+      const code = await cli.run(VALID_INPUT, context as any)
+      expect(code).toBe(0)
+      const output = context.stdout.toString()
+      expect(output).toMatchSnapshot()
+    })
+  })
+
+  describe('writeFile', () => {
+    const MOCK_DATA = 'mock data'
+    const instance = new LambdaFlareCommand()
```
duncanista, line 214: An instance of the Flare command is being created just to use the context. But there is already a function that mocks the context.

You have previously used this context, why don't we follow the same approach here?

src/commands/lambda/__tests__/flare.test.ts
```diff
@@ -1,21 +1,444 @@
...
+      expect(fs.existsSync).toHaveBeenCalledWith(MOCK_FOLDER_PATH)
+      expect(fs.mkdirSync).toHaveBeenCalledWith(MOCK_FOLDER_PATH)
+      expect(fs.writeFileSync).toHaveBeenCalledWith(MOCK_FILE_PATH, MOCK_DATA)
+    })
+  })
+
+  describe('zipContents', () => {
+    beforeEach(() => {
+      fs.writeFileSync = jest.fn().mockImplementation(() => {})
+      fs.readFileSync = jest.fn().mockResolvedValue(JSON.stringify(MOCK_CONFIG, undefined, 2))
```
duncanista, line 253: Aren't these two already defined at the top of the file?

src/commands/lambda/__tests__/flare.test.ts
```diff
@@ -1,21 +1,444 @@
...
+    this.push(JSON.stringify(MOCK_CONFIG, undefined, 2))
+    this.push(undefined)
+  },
+})
+fs.createReadStream = jest.fn().mockReturnValue(mockReadStream)
+
+// Zip mocks
+jest.mock('jszip')
+const mockJSZip = new JSZip()
+require('jszip').mockImplementation(() => mockJSZip)
```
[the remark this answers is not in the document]
nhulston: I'm mocking the constructor so that `new JSZip()` returns `mockJSZip`. That way I have a const to the instance of `JSZip`, so I can I can have statements like:
`expect(mockJSZip.file).toHaveBeenCalled()
expect(mockJSZip.generateAsync).not.toHaveBeenCalled()`

src/commands/lambda/__tests__/flare.test.ts
```diff
@@ -1,21 +1,444 @@
...
+      expect(output).toMatchSnapshot()
+      postSpy.mockRestore()
+    })
+  })
+
+  describe('AWS Lambda configuration', () => {
+    it('stops and prints error when getLambdaFunctionConfig fails', async () => {
+      require('../functions/commons').getLambdaFunctionConfig.mockImplementation(() => {
+        throw new Error('MOCK ERROR: Some API error')
+      })
```
duncanista, line 375: Let's not do `require()` every time we want to mock something.

On the other hand, require it at the top, and then do:

```js
;(functionToMock as any).mockImplementation(() => Promise.reject('Unexpected error'))
```

Or whatever the outcome you are trying to achieve.

src/commands/lambda/flare.ts
```diff
@@ -1,14 +1,249 @@
...
+export const writeFile = async (folderPath: string, filePath: string, data: string, context: BaseContext) => {
+  if (fs.existsSync(folderPath)) {
+    deleteFolder(folderPath, context)
+  }
+
+  try {
+    fs.mkdirSync(folderPath)
+    fs.writeFileSync(filePath, data)
+  } catch (err) {
+    throw Error(`Unable to save function config: ${err.message}`)
```
duncanista, line 169: This is doing two things but we are printing an specific error. What if the error was creating the folder?

```suggestion
    throw Error(`Unable to create function configuration file: ${err.message}`)
```

duncanista: Or if we want to be consistent with the other errors `Unable to create flare file: ...`

src/commands/lambda/flare.ts
```diff
@@ -1,14 +1,249 @@
...
+      'DD-API-KEY': apiKey,
+    },
+  }
+
+  try {
+    await axios.post(ENDPOINT_URL, form, headerConfig)
+    context.stdout.write('\n✅ Successfully sent function config to Datadog Support!\n')
+  } catch (err) {
+    const errResponse: string = err.response?.data?.error
+    throw Error(`Failed to send function config to Datadog Support: ${err.message}. ${errResponse ?? ''}\n`)
```
duncanista, line 224:
```suggestion
    throw Error(`Failed to send flare file to Datadog Support: ${err.message}. ${errResponse ?? ''}\n`)
```

src/commands/lambda/__tests__/flare.test.ts
nhulston: Yes, but some tests in the describe block edit the implementation of these two. For example:
https://github.com/DataDog/datadog-ci/blob/4e84a9221001b2deb0c2bfe502c9b7b377e65fb2/src/commands/lambda/__tests__/flare.test.ts#L267-L269

src/commands/lambda/__tests__/flare.test.ts
nhulston: Resoled with `const context = createMockContext() as BaseContext`

src/commands/lambda/__tests__/flare.test.ts
nhulston: I can't because `jest.mock` is ran before all of the imports and consts are defined, even if the imports and consts appear above these lines of code in the file.

An option is to try something like this below, but I still get the error: `Cannot access 'fixtures_1' before initialization` (referring to `mockDatadogApiKey`)

```let MOCK_CONFIG: any
jest.mock('fs', () => {
  MOCK_CONFIG = {
    Environment: {
      Variables: {
        DD_API_KEY: mockDatadogApiKey, // <--- can't be used in jest.mock because it hasn't been initialized
        DD_SITE: 'datadoghq.com',
        DD_LOG_LEVEL: 'debug',
      },
    },
    FunctionArn: 'arn:aws:lambda:us-east-1:123456789012:function:some-function',
    FunctionName: 'some-function',
  }
  const mockReadStream = new stream.Readable({
    read() {
      this.push(JSON.stringify(MOCK_CONFIG, undefined, 2))
      this.push(undefined)
    },
  })

  return {
    writeFileSync: jest.fn().mockImplementation(() => {}),
    readFileSync: jest.fn().mockReturnValue(JSON.stringify(MOCK_CONFIG, undefined, 2)),
    createWriteStream: jest.fn().mockReturnValue(mockReadStream),
  }
})```

File: src/commands/lambda/__tests__/flare.test.ts (hunk 67 above)
Comment (nhulston) [reply src/commands/lambda/__tests__/flare.test.ts:38]: Converted it to a global mock in `__mocks__/chalk.js`

File: src/__mocks__/util.js
Comment (duncanista) [inline src/__mocks__/util.js:1]: Amazing! This will benefit everyone using this in the future!

File: src/commands/lambda/flare.ts [hunk 76]
```diff
@@ -1,14 +1,249 @@
...
+
+import {API_KEY_ENV_VAR, AWS_DEFAULT_REGION_ENV_VAR, CI_API_KEY_ENV_VAR} from './constants'
+import {getAWSCredentials, getLambdaFunctionConfig, getRegion} from './functions/commons'
+import {requestAWSCredentials} from './prompt'
+import * as commonRenderer from './renderers/common-renderer'
+import * as flareRenderer from './renderers/flare-renderer'
+
+const {version} = require('../../../package.json')
+
+const ENDPOINT_URL = 'https://datad0g.com/api/ui/support/serverless/flare'
```
juan-fernandez, line 20: do we want to hardcode a staging URL like this?

src/commands/lambda/flare.ts
```diff
@@ -1,14 +1,249 @@
...
+ * @param filePath
+ * @param zipPath
+ * @throws Error if the zip fails
+ */
+export const zipContents = async (filePath: string, zipPath: string) => {
+  try {
+    const data = fs.readFileSync(filePath, 'utf8')
+    const zip = new JSZip()
+    zip.file(FUNCTION_CONFIG_FILE_NAME, data)
+    const content = await zip.generateAsync({type: 'nodebuffer'})
```
juan-fernandez, line 184: couldn't we use `zlib` that is native? We wouldn't need a new dependency then.

juan-fernandez (commented):
I lack the context for reviewing the lambda bit but do we need `jszip`? Have we tried using `zlib`?

src/commands/lambda/flare.ts
nhulston: Thanks for the review, I will look into using `zlib` instead

nhulston: Hey @juan-fernandez 
Looks like `zlib` only supports zipping single files. While this would work for this PR, we do plan on zipping multiple files in the near future, so we would need a new dependency like `fszip`.

@duncanista What do you think? Should I use `zlib` for this PR or stick with `fszip`?

src/commands/lambda/flare.ts
duncanista: Hey @juan-fernandez, we will hard code it for staging in the meantime. We will not merge this to the `main` branch until we have the right endpoint merged in production. 

We will bring another PR to update this to the user specific site endpoint in the future, wdyt?

src/commands/lambda/flare.ts
duncanista: @nhulston do you mean `jszip`? 

@juan-fernandez I'd normally go for the `zlib` approach too, when I researched on this, it seemed that we needed more packages to compress folders (`tar` and `fstream`). By using `zlib` and its dependencies for our use case, previously mentioned, we would be using 550~600 kB less than if we used `jszip`.

It seems that change of implementation might be trivial. But of course, this could be refactored in the future if package size is a problem. This would also not be merged directly into main since our working branch `serverless-flare` is still a WIP before its first release.

What do you think @juan-fernandez?

src/commands/lambda/__tests__/flare.test.ts
duncanista: Let's leave this work for another PR. Let's log a ticket for this, since other people jumping on the project might have a hard time understanding all of what's going on with the mocks.

duncanista (approved):
Great first PR!

PR #925, opened 2023-06-12
Remove assumptions from command

andrea-mosk:
### What and why?

Remove assumptions from command

### Review checklist

### DIFF
--- a/src/commands/gate/interfaces.ts
+++ b/src/commands/gate/interfaces.ts
@@ -15,22 +15,17 @@ export interface EvaluationResponsePayload {
   }
 }
 
-type GateStatus = 'passed' | 'failed' | 'empty'
-
 export interface EvaluationResponse {
-  status: GateStatus
+  status: string
   rule_evaluations: RuleEvaluation[]
 }
-
-type RuleStatus = 'passed' | 'failed' | 'no_data'
 
 export interface RuleEvaluation {
   rule_id: string
   rule_name: string
-  status: RuleStatus
+  status: string
   is_blocking: boolean
   failure_reason: string
-  events_count: number
 }
 
 export interface APIHelper {
--- a/src/commands/gate/renderer.ts
+++ b/src/commands/gate/renderer.ts
@@ -45,7 +45,7 @@ export const renderStatus = (result: string): string => {
       return chalk.yellow(`No Data ${ICONS.WARNING} `)
   }
 
-  return 'Unknown Status: ' + result.toLowerCase()
+  return result.toLowerCase()
 }
 
 export const renderRuleEvaluation = (ruleEvaluation: RuleEvaluation): string => {


src/commands/gate/interfaces.ts
```diff
@@ -15,22 +15,17 @@ export interface EvaluationResponsePayload {
   }
 }
 
-type GateStatus = 'passed' | 'failed' | 'empty'
```
andrea-mosk, line 18: We can introduce new states in future on the service side, and I don't want it to break backward compatibility with the command

src/commands/gate/interfaces.ts
```diff
@@ -15,22 +15,17 @@ export interface EvaluationResponsePayload {
...
-type RuleStatus = 'passed' | 'failed' | 'no_data'
-
 export interface RuleEvaluation {
   rule_id: string
   rule_name: string
-  status: RuleStatus
+  status: string
   is_blocking: boolean
   failure_reason: string
-  events_count: number
```
andrea-mosk, line 33: Not used

juan-fernandez (commented):
I don't think these types are exposed to the user in any way, right? They are just for us, and we should always be developing on the latest version, so I fail to see an issue with having a fixed number of statuses: if we ever need more, we can just add them. 

This said, if you think they are not adding value, I'm fine with sticking with `string`

andrea-mosk:
> I don't think these types are exposed to the user in any way, right? They are just for us, and we should always be developing on the latest version, so I fail to see an issue with having a fixed number of statuses: if we ever need more, we can just add them.
> 
> This said, if you think they are not adding value, I'm fine with sticking with `string`

My concern about having those types is what happens if we introduce a new status on the service (let's say `blocked`). We can add it to datadog-ci as well, but what happens if clients are using older versions of the command? Will it fail to deserialize because it doesn't recognize the `blocked` status?

juan-fernandez:
> > I don't think these types are exposed to the user in any way, right? They are just for us, and we should always be developing on the latest version, so I fail to see an issue with having a fixed number of statuses: if we ever need more, we can just add them.
> > This said, if you think they are not adding value, I'm fine with sticking with `string`
> 
> My concern about having those types is what happens if we introduce a new status on the service (let's say `blocked`). We can add it to datadog-ci as well, but what happens if clients are using older versions of the command? Will it fail to deserialize because it doesn't recognize the `blocked` status?

no, types are not considered at runtime. They are just a way to prevent us developers from doing things that could result in a runtime error, such as a bad variable access.

What you did in https://github.com/DataDog/datadog-ci/pull/925/files#diff-2cf56c8a602d4d0ad9afeaeb1a0abf734cb385d1a75346f9c0ee4603b3dc73e4L48 makes sense though, because that does affect runtime.

juan-fernandez (approved):

PR #926, opened 2023-06-12
[junit] Replace unique filename with just filename

rodrigo-roca:
### What and why?

Replaces the old unique filename with just the filename since we don't need it to be unique anymore on our side.

### How?

Deleting that code and just using the filename

### Review checklist

### DIFF
--- a/src/helpers/file.ts
+++ b/src/helpers/file.ts
@@ -1,4 +1,4 @@
 // We need a unique file name so we use span tags like the pipeline URL,
 // which can contain dots and other unsafe characters for filenames.
 // We filter them out here.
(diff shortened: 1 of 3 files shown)


src/helpers/file.ts
```diff
@@ -1,4 +1,4 @@
 // We need a unique file name so we use span tags like the pipeline URL,
 // which can contain dots and other unsafe characters for filenames.
 // We filter them out here.
-export const getSafeFileName = (unsafeFileName: string) => unsafeFileName.replace(/[^a-z0-9]/gi, '_')
+export const getSafeFileName = (unsafeFileName: string) => unsafeFileName.replace('/', '\\')
```
juan-fernandez, line 4: why this change? wouldn't it still be useful to get rid of any character not in `a-z0-9`?

juan-fernandez (approved):

PR #928, opened 2023-06-14
[ci-visibility] Change default value of git upload

juan-fernandez:
### What and why?

We've had some issues with slowness doing this git upload operation. Until we have proper metrics reporting in place, it's safer to turn this into opt-in rather than opt-out.

### How?

* Change `--skip-git-metadata-upload` to `--enable-git-metadata-upload`. Now, only if you're passing it will `datadog-ci junit upload` attempt to upload your data. 

### Review checklist

### DIFF
--- a/src/commands/junit/README.md
+++ b/src/commands/junit/README.md
@@ -30,7 +30,7 @@ datadog-ci junit upload --service my-service --tags key1:value1 --tags key2:valu
 - `--max-concurrency` (default: `20`): number of concurrent uploads to the API.
 - `--dry-run` (default: `false`): it will run the command without the final upload step. All other checks are performed.
 - `--logs` (default: `false`): it will enable collecting logs from the content in the XML reports.
-- `--enable-git-metadata-upload` (default: `false`): it will upload git metadata.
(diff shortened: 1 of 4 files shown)


src/commands/junit/README.md
```diff
@@ -30,7 +30,7 @@ datadog-ci junit upload --service my-service --tags key1:value1 --tags key2:valu
 - `--max-concurrency` (default: `20`): number of concurrent uploads to the API.
 - `--dry-run` (default: `false`): it will run the command without the final upload step. All other checks are performed.
 - `--logs` (default: `false`): it will enable collecting logs from the content in the XML reports.
-- `--skip-git-metadata-upload` (default: `false`): it will run the command without uploading git metadata.
+- `--enable-git-metadata-upload` (default: `false`): it will upload git metadata.
```
edrevo, line 33: This is a breaking change (if someone was using `--skip-git-metadata-upload` then this will break them. I think we should just change the default value to "true"

juan-fernandez: of course 🤦, good catch

juan-fernandez: this forces us to use something like `--skip-git-metadata-upload=false` or `--skip-git-metadata-upload=0`, which isn't great, but it's better than a breaking change

edrevo (approved):

estherk15 (approved):

PR #930, opened 2023-06-20
Add details URL to rule logs

andrea-mosk:
### What and why?

Add details URL to rule logs

### Review checklist

### DIFF
--- a/src/commands/gate/renderer.ts
+++ b/src/commands/gate/renderer.ts
@@ -59,7 +59,7 @@ export const renderRuleEvaluation = (ruleEvaluation: RuleEvaluation): string => 
   }
 
   fullStr += `${chalk.yellow('Blocking')}: ${ruleEvaluation.is_blocking}\n`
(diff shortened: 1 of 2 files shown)


src/commands/gate/renderer.ts
```diff
@@ -59,6 +59,10 @@ export const renderRuleEvaluation = (ruleEvaluation: RuleEvaluation): string =>
   }
 
   fullStr += `${chalk.yellow('Blocking')}: ${ruleEvaluation.is_blocking}\n`
+  if (ruleEvaluation.details_url && ruleEvaluation.details_url.length > 0) {
```
juan-fernandez, line 62: is `details_url` always a `string`? If that's the case, this is enough. An empty string is falsy 
```suggestion
  if (ruleEvaluation.details_url) {
```

juan-fernandez (approved):

src/commands/gate/renderer.ts
andrea-mosk: Oh, thanks!

PR #931, opened 2023-06-21
[junit] Update to v3

rodrigo-roca:
### What and why?

- Changes the junit upload to v3
  - Adds support for "report tags" that are tags that are only applied to sessions
  - Changes the format that the payload is sent to simplify the backend side.
  - Replaces the unique filename with just the filename https://github.com/DataDog/datadog-ci/pull/926

### How?

Instead of spreading everything into the "spanTags" variable now it is separated into several variables.  Changed the junit payload interface to achieve this.

### Review checklist

### DIFF
--- a/src/commands/junit/api.ts
+++ b/src/commands/junit/api.ts
@@ -49,17 +49,8 @@ export const uploadJUnitXML = (request: (args: AxiosRequestConfig) => AxiosPromi
 
   form.append('event', JSON.stringify(metadata), {filename: 'event.json'})
 
-  let uniqueFileName = `${fileName}-${jUnitXML.service}-${metadata[GIT_SHA]}`
-
-  if (metadata[CI_PIPELINE_URL]) {
-    uniqueFileName = `${uniqueFileName}-${metadata[CI_PIPELINE_URL]}`
-  }
-  if (metadata[CI_JOB_URL]) {
-    uniqueFileName = `${uniqueFileName}-${metadata[CI_JOB_URL]}`
-  }
-
   form.append('junit_xml_report_file', fs.createReadStream(jUnitXML.xmlPath).pipe(createGzip()), {
-    filename: `${getSafeFileName(uniqueFileName)}.xml.gz`,
+    filename: `${getSafeFileName(fileName)}.xml.gz`,
   })
 
   return request({
--- a/src/commands/junit/api.ts
+++ b/src/commands/junit/api.ts
@@ -7,7 +7,7 @@ import type {AxiosPromise, AxiosRequestConfig, AxiosResponse} from 'axios'
 import FormData from 'form-data'
 
 import {getSafeFileName} from '../../helpers/file'
-import {CI_JOB_URL, CI_PIPELINE_URL, GIT_SHA} from '../../helpers/tags'
+import {CI_JOB_URL, CI_PIPELINE_URL, GIT_SHA, SERVICE} from '../../helpers/tags'
 import {getRequestBuilder} from '../../helpers/utils'
 
 import {Payload} from './interfaces'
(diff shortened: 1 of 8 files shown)


src/commands/junit/api.ts
```diff
@@ -49,17 +55,8 @@ export const uploadJUnitXML = (request: (args: AxiosRequestConfig) => AxiosPromi
...
-  if (metadata[CI_PIPELINE_URL]) {
-    uniqueFileName = `${uniqueFileName}-${metadata[CI_PIPELINE_URL]}`
-  }
-  if (metadata[CI_JOB_URL]) {
-    uniqueFileName = `${uniqueFileName}-${metadata[CI_JOB_URL]}`
-  }
-
   form.append('junit_xml_report_file', fs.createReadStream(jUnitXML.xmlPath).pipe(createGzip()), {
-    filename: `${getSafeFileName(uniqueFileName)}.xml.gz`,
+    filename: `${getSafeFileName(fileName)}.xml.gz`,
```
AdrianLC, line 59: let's do `.replace(/\//g, '\\')` here directly. It's 1 function and it's more obvious what it does as it's simpler that it was before.

If you want to keep it in a function rename it to replaceSlashes or similar

src/commands/junit/api.ts
```diff
@@ -32,10 +31,17 @@ export const uploadJUnitXML = (request: (args: AxiosRequestConfig) => AxiosPromi
     fileName = 'default_file_name'
   }
 
+  const reportTagsAndMetrics: Record<string, any> = {
+    tags: jUnitXML.reportTags,
+    metrics: jUnitXML.reportMetrics,
+  }
+
   const metadata: Record<string, any> = {
```
AdrianLC, line 39: maybe rename this now that it has an internal attribute 'metadata' it might be confusing

src/commands/junit/api.ts
rodrigo-roca: I would prefer to have it in a separate function even if it is simpler than before. The rename sounds good to me 👍

src/commands/junit/upload.ts
```diff
@@ -325,6 +353,8 @@ UploadJUnitXMLCommand.addOption('env', Command.String('--env'))
 UploadJUnitXMLCommand.addOption('dryRun', Command.Boolean('--dry-run'))
 UploadJUnitXMLCommand.addOption('tags', Command.Array('--tags'))
 UploadJUnitXMLCommand.addOption('metrics', Command.Array('--metrics'))
+UploadJUnitXMLCommand.addOption('reportTags', Command.Array('--report-tags'))
+UploadJUnitXMLCommand.addOption('reportMetrics', Command.Array('--report-metrics'))
```
Drarig29, line 357: Can you update the README too? https://github.com/DataDog/datadog-ci/blob/master/src/commands/junit/README.md#usage

Drarig29 (approved):
Looks good to me

estherk15 (approved):

src/commands/junit/README.md
```diff
@@ -22,10 +22,14 @@ datadog-ci junit upload --service my-service --tags key1:value1 --tags key2:valu
...
-- `--tags` is a array of key value pairs of the shape `key:value`. This will set global tags applied to all spans.
+- `--tags` is an array of key value pairs of the shape `key:value`. This will set global tags applied to all spans.
   - The resulting dictionary will be merged with whatever is in the `DD_TAGS` environment variable. If a `key` appears both in `--tags` and `DD_TAGS`, whatever value is in `DD_TAGS` will take precedence.
 - `--metrics` is an array of key numerical value pairs of the shape `key:123`. This will set global metrics applied to all spans.
   - The resulting dictionary will be merged with whatever is in the `DD_METRICS` environment variable. If a `key` appears both in `--metrics` and `DD_METRICS`, whatever value is in `DD_METRICS` will take precedence.
+- `--report-tags` is an array of key value pairs like the `--tags` argument, but the tags are only applied to the session instead of to every test.
+  - The resulting dictionary will NOT be merged with `DD_TAGS`.
+- `--report-metrics` is an array of key value pairs like the `--metrics` argument, but the tags are only applied to the session instead of to every test.
```
juan-fernandez, line 31:
```suggestion
- `--report-metrics` is an array of key numerical value pairs like the `--metrics` argument, but the metrics are only applied to the session instead of to every test.
```

juan-fernandez (approved):

AdrianLC (approved):

PR #932, opened 2023-06-21
[ci-visibility] Add trace id to git requests when using `junit upload`

juan-fernandez:
### What and why?

By adding a trace id in the git metadata upload requests we make sure that we can track the whole process. 

### How?

* Add an `id` module to generate span / trace ids. 
* Set trace id and parent id headers in the git requests. 

### Review checklist

### DIFF
--- a/src/commands/junit/id.ts
+++ b/src/commands/junit/id.ts
@@ -1,0 +1,64 @@
+/* eslint-disable */
+// Eslint disabled because we're doing bitwise logic, which isn't allowed by the linter
+// From https://github.com/DataDog/dd-trace-js/blob/e4b9a268f0429b6f1e92c384b61a0d104aeb1259/packages/dd-trace/src/id.js
+import {randomFillSync} from 'crypto'
+
+const UINT_MAX = 4294967296
+
+const data = new Uint8Array(8 * 8192)
+
+let batch = 0
+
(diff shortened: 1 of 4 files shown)


src/commands/junit/id.ts
```diff
@@ -0,0 +1,64 @@
+/* eslint-disable */
+// Eslint disabled because we're doing bitwise logic, which isn't allowed by the linter
+// From https://github.com/DataDog/dd-trace-js/blob/e4b9a268f0429b6f1e92c384b61a0d104aeb1259/packages/dd-trace/src/id.js
```
romainkomorndatadog, line 3: Would it be worth grabbing a test out of https://github.com/DataDog/dd-trace-js/blob/e4b9a268f0429b6f1e92c384b61a0d104aeb1259/packages/dd-trace/test/id.spec.js , too?

juan-fernandez: yeah that makes sense

juan-fernandez: done in https://github.com/DataDog/datadog-ci/pull/932/commits/bb8edbed0d9dfa42e5e322e39760af382b5a5977

Drarig29 (approved):
LGTM

gnufede (approved):

PR #937, opened 2023-06-23
[stepfunctions][test] Fix running tests locally

Drarig29:
### What and why?

The `creates step function log group, subscribes it to forwarder, and adds it to step function` test was timing out when running locally.

### How?

I mocked `createLogGroup` too, and updated the related assertions.

### Review checklist


kimi-p (approved):

kimi-p:
Thanks for fixing it!

PR #941, opened 2023-06-23
[lambda][flare] Collect CloudWatch Logs

nhulston:
### What and why?

This PR introduces the `--with-logs` flag to the datadog-ci lambda flare command, allowing the retrieval and forwarding of CloudWatch logs. This aims to provide the Datadog support team with better insights for assisting customers.

Logs are saved in `.datadog-ci/logs/<log name>` in a CSV format, which the Datadog support team will use to better assist customers in resolving issues.

Screenshot 2023-06-23 at 3 07 27 PM

### How?
An example command to collect CloudWatch logs would be `lambda flare -f function-name -r us-east-1 -c 1234567 -e nicholas.hulston@datadoghq.com --with-logs`

This PR does the following:
1. Adds the `--with-logs` flag and its corresponding boolean `withLogs`
2. Creates the `getLogStreamNames` function to retrieve the 3 latest log streams
3. Creates the `getLogEvents` function, which gets all logs in a specific log stream
4. Creates the `getAllLogs` function, which gets all log events for each stream name, and returns a Map of type `Map<string, OutputLogEvent[]>`
5. Rewrites the `zipContents` function to support zipping entire directories, instead of just a single file using the `jszip` package.
6. Creates the `convertToCSV` function, which turns `OutputLogEvent[]` into a valid CSV format consistent with files exported from AWS CloudWatch.
7. Creates the `createDirectories` function, which creates the `.datadog-ci` folder as well as `.datadog-ci/logs` when `--with-logs` is flagged.
8. Utilizes these helper functions to retrieve, save, zip, and send logs when `--with-logs` is included.
9. When running in dry mode with `-d` or `--dry`, the output files in `.datadog-ci` are no longer removed. This way, the user can see the output of the command without sending to support. Files are still automatically removed when `-d` is not included as a flag.

### Future Plans
- Customize the CSV files to include fields that the Support Team might find useful, such as changing `timestamp` to a human-readable time.
- Interactive mode
- Supporting multiple functions/regions
- Create support tickets via the command
- And more

### Review checklist


src/commands/lambda/flare.ts
```diff
@@ -40,105 +49,170 @@ export class LambdaFlareCommand extends Command {
...
     }
 
-    if (errorFound) {
+    // Exit if there are errors
+    if (errorMessages.length > 0) {
+      this.context.stderr.write(errorMessages.join('') + '\n')
+
       return 1
     }
+    assert(this.functionName !== undefined)
```
duncanista, line 89: What's the reason of changing the validation on this PR? This doesn't seem related to the PR description changes.

src/commands/lambda/flare.ts
```diff
@@ -40,105 +49,170 @@ export class LambdaFlareCommand extends Command {
...
 
     // Get AWS credentials
     this.context.stdout.write('\n🔑 Getting AWS credentials...\n')
     try {
       this.credentials = await getAWSCredentials()
     } catch (err) {
-      this.context.stderr.write(commonRenderer.renderError(err))
+      this.context.stderr.write(commonRenderer.renderError(err.message))
```
duncanista, line 96: What's the reason behind this change?

src/commands/lambda/flare.ts
```diff
@@ -40,105 +49,170 @@ export class LambdaFlareCommand extends Command {
...
 
       return 1
     }
     if (this.credentials === undefined) {
-      this.context.stdout.write(commonRenderer.renderNoAWSCredentialsFound())
+      this.context.stdout.write('\n' + commonRenderer.renderNoAWSCredentialsFound())
```
duncanista, line 101: Why is this needed now?

src/commands/lambda/flare.ts
duncanista: Also, why are we asserting? Isn't the first validaton check supposed to do this?

src/commands/lambda/flare.ts
```diff
@@ -40,105 +49,170 @@ export class LambdaFlareCommand extends Command {
...
+      try {
+        logStreamNames = await getLogStreamNames(cwlClient, logGroupName)
+      } catch (err) {
+        this.context.stderr.write(commonRenderer.renderError(`Unable to get log streams: ${err.message}`))
+
+        return 1
+      }
+      if (logStreamNames === undefined) {
+        this.context.stdout.write(
+          commonRenderer.renderSoftWarning('No CloudWatch streams were found. Logs will not be retrieved or sent.\n')
```
duncanista, line 148:
```suggestion
          commonRenderer.renderSoftWarning('No CloudWatch log streams were found. Logs will not be retrieved or sent.\n')
```

src/commands/lambda/flare.ts
```diff
@@ -40,105 +49,170 @@ export class LambdaFlareCommand extends Command {
...
+        )
+      } else {
```
duncanista, line 150: Avoid using `else`

src/commands/lambda/flare.ts
```diff
@@ -40,105 +49,170 @@ export class LambdaFlareCommand extends Command {
...
+    const configStr = util.inspect(config, false, undefined, true)
+    this.context.stdout.write(`\n${configStr}\n`)
 
-    // Print config
-    const configStrColored = util.inspect(config, false, undefined, true)
-    this.context.stdout.write(`\n${configStrColored}\n`)
-    if (this.isDryRun) {
-      this.context.stdout.write('\n🚫 The configuration was not sent as it was executed in dry run mode.\n')
+    // Get CloudWatch logs
+    const logs: [string, OutputLogEvent[]][] = []
```
duncanista, line 132: This typing is confusing. Also, why not use a map?

src/commands/lambda/flare.ts
```diff
@@ -150,41 +224,154 @@ export class LambdaFlareCommand extends Command {
...
+ * @throws Error if the log streams cannot be retrieved
+ */
+export const getLogStreamNames = async (cwlClient: CloudWatchLogsClient, logGroupName: string) => {
+  const getLogStreamsCommand = new DescribeLogStreamsCommand({
+    logGroupName,
+    limit: 3,
+    descending: true,
+    orderBy: 'LastEventTime',
+  })
+  const logStreams = (await cwlClient.send(getLogStreamsCommand)).logStreams
```
duncanista, line 271:
```suggestion
  const command = new DescribeLogStreamsCommand({
    logGroupName,
    limit: 3,
    descending: true,
    orderBy: 'LastEventTime',
  })
  const response = await cwlClient.send(command)
  const logStreams = response.logStreams
```

src/commands/lambda/flare.ts
```diff
@@ -150,41 +224,154 @@ export class LambdaFlareCommand extends Command {
...
+  if (logStreams === undefined) {
+    return undefined
+  }
+  const output: string[] = logStreams
+    .filter((logStream) => logStream !== undefined)
+    .map((logStream) => logStream.logStreamName)
+    .filter((logStreamName): logStreamName is string => logStreamName !== undefined && logStreamName.length > 0)
```
duncanista, line 278: What is this suppose to do?
This logic is way too complex.

src/commands/lambda/flare.ts
```diff
@@ -150,41 +224,154 @@ export class LambdaFlareCommand extends Command {
...
+    logStreamName,
+  })
+  try {
+    const logEvents = (await cwlClient.send(params)).events
+
+    return logEvents === undefined || logEvents.length === 0 ? undefined : logEvents
+  } catch (err) {
+    throw Error(err.message)
+  }
+}
```
duncanista, line 304:
```suggestion
export const getLogEvents = async (cwlClient: CloudWatchLogsClient, logGroupName: string, logStreamName: string) => {
  const command = new GetLogEventsCommand({
    logGroupName,
    logStreamName,
  })
  try {
    const response = await cwlClient.send(command)
    const logEvents = response.events

    return logEvents === undefined || logEvents.length === 0 ? undefined : logEvents
  } catch (err) {
    throw err
  }
}
```

src/commands/lambda/flare.ts
```diff
@@ -40,105 +49,170 @@ export class LambdaFlareCommand extends Command {
...
-    const filePath = path.join(folderPath, FUNCTION_CONFIG_FILE_NAME)
-    const zipPath = path.join(folderPath, ZIP_FILE_NAME)
-    const configStrUncolored = JSON.stringify(config, undefined, 2)
+    this.context.stdout.write('\n💾 Saving files...\n')
     try {
-      await writeFile(folderPath, filePath, configStrUncolored, this.context)
-      await zipContents(filePath, zipPath)
+      // Create folders
+      const rootFolderPath = path.join(process.cwd(), FLARE_OUTPUT_DIRECTORY)
+      const logsFolderPath = path.join(rootFolderPath, LOGS_DIRECTORY)
```
duncanista, line 178: So you create the logs folder even if user specified no logs?

src/commands/lambda/flare.ts
```diff
@@ -40,105 +49,170 @@ export class LambdaFlareCommand extends Command {
...
+      const configFilePath = path.join(rootFolderPath, FUNCTION_CONFIG_FILE_NAME)
+      writeFile(configFilePath, JSON.stringify(config, undefined, 2))
+      this.context.stdout.write(`${logs.length > 0 ? '• ' : ''}Saved function config to ${configFilePath}\n`)
+      for (const [logStreamName, logEvents] of logs) {
+        const logFilePath = path.join(logsFolderPath, `${logStreamName.split('/').join('-')}.csv`)
+        const data = convertToCSV(logEvents)
+        writeFile(logFilePath, data)
+        this.context.stdout.write(`• Saved logs to ${logFilePath}\n`)
+        // Sleep for 1 millisecond so OS can sort files by creation time
+        await new Promise((resolve) => setTimeout(resolve, 1))
```
duncanista, line 194: Where does the magic number come from? Why 1 ms?

src/commands/lambda/flare.ts
```diff
@@ -40,105 +49,170 @@ export class LambdaFlareCommand extends Command {
...
     try {
-      await writeFile(folderPath, filePath, configStrUncolored, this.context)
-      await zipContents(filePath, zipPath)
+      // Create folders
+      const rootFolderPath = path.join(process.cwd(), FLARE_OUTPUT_DIRECTORY)
+      const logsFolderPath = path.join(rootFolderPath, LOGS_DIRECTORY)
+      if (fs.existsSync(rootFolderPath)) {
+        deleteFolder(rootFolderPath)
+      }
+      createDirectories(rootFolderPath, logsFolderPath, logs.length > 0)
```
duncanista, line 182: I think this logic is too complex.

1. The logs folder can be calculated inside `createDirectores`, because all its dependencies are available inside the function.
2. Wouldn't it make sense to just send the logs instead of a boolean? If you are going to check either way if it's empty, you can do that inside the function. 

Wdyt?

src/commands/lambda/flare.ts
nhulston: This change is not necessary for this PR, but I think this solution would be more readable and take up less lines of code. Logically, it does the same thing as it did before.
This file, for example, uses the same approach: https://github.com/DataDog/datadog-ci/blob/3d41ad6db1fed0f0bcc6216b49d534961673fd8a/src/commands/synthetics/utils.ts#L567-L576

I am asserting because I am trying to force the type of `this.functionName` to be `string` instead of `string | undefined`. Do you think there is a better approach, such as doing `this.functionName ?? ''` instead?

src/commands/lambda/flare.ts
nhulston: This is just a nit change. I felt that there were too many log messages in a row without any spacing in between when this was not included.

The new line would create some logical divide between the 'getting credentials' part and the 'requesting credentials' part.

Before: 
Screenshot 2023-06-26 at 9 42 26 AM

After:
Screenshot 2023-06-26 at 9 43 33 AM

What do you think? Is this change necessary?

src/commands/lambda/flare.ts
nhulston: Change added

src/commands/lambda/flare.ts
nhulston: I felt that a map was overkill / slower since we are only getting 3 log streams. However, I agree that the typing would be a lot more readable. Do you think I should change it to a map?

src/commands/lambda/flare.ts
nhulston: Done

src/commands/lambda/flare.ts
nhulston: I am trying to force the output to be of type `string[]`.

I am now realizing that the first filter on line 276 is unnecessary since `logStreams` is of type `LogStream[]` and therefore can't have any undefined items in the array. Therefore, the new code looks like this after the last commit:
```  const output = logStreams
    .map((logStream) => logStream.logStreamName)
    .filter((logStreamName): logStreamName is string => logStreamName !== undefined && logStreamName.length > 0)
```

I agree that the second `.filter` is really confusing. Since `logStream.logStreamName` can be of type `string | undefined`, I want to ensure that we only include log stream names in the array that were not undefined or empty.

Do you propose I break the block of code into multiple separate lines to make the logic more readable?

src/commands/lambda/flare.ts
nhulston: Why should I avoid using `else` in this scenario?
I'm also not sure how to achieve the same logic without using `else` and without adding additional complexity.

src/commands/lambda/flare.ts
nhulston: Thanks, I have made this change.
However, I want to let you know that I still throw using `throw Error(err.message)` to remove the `Error: ` prefix. This is because I catch any errors that `getLogEvents` throws and add an error prefix of `[Error] ` by using `commonRenderer.renderError()`

src/commands/lambda/flare.ts
nhulston: No, the `createDirectories()` accepts a parameter `createLogsFolder: boolean`, and I pass in `logs.length > 0` as the parameter. However, I agree that this is really confusing.
Here is my change to make it more readable:

```
      const rootFolderPath = path.join(process.cwd(), FLARE_OUTPUT_DIRECTORY)
      const logsFolderPath = path.join(rootFolderPath, LOGS_DIRECTORY)
      if (fs.existsSync(rootFolderPath)) {
        deleteFolder(rootFolderPath)
      }
      const shouldCreateLogsFolder = this.withLogs && logs.length > 0
      createDirectories(rootFolderPath, logsFolderPath, shouldCreateLogsFolder)
```

What do you think of this change?

src/commands/lambda/flare.ts
nhulston: I noticed from my own manual testing that if we did not sleep, all the files were created at the same instant (as far as the operating system was concerned). Therefore, I could not sort the files by creating time.
If we instead wait any amount of time, even only 1 millisecond, the files could now be sorted.

I picked 1 millisecond because that was the minimum amount of time I could sleep using `await new Promise`. Do you think I should move the magic number to a const? It seems unnecessary since the magic number is 1 and not some complex number.

src/commands/lambda/flare.ts
duncanista: There are better ways of forcing a type in TypeScript.
Either you type cast, type convert, or simply use the non-null assertion operator. Why not do that instead of assert?

Nice, sounds good, let's do it in another PR that focuses only in refactoring this!

Normally, reviewers expect to find changes in the logic in which the description is referring to. Those changes add value to the whole project, but are not directly tied to the PR description. So yeah, generally, if a change is not related, we would expect a separate PR.

src/commands/lambda/flare.ts
duncanista: Sounds good. If you come up with a PR to refactor validation, it would be a good opportunity to type guard the errors, I noticed you didn't do it, which caused this errors to not show.

src/commands/lambda/flare.ts
duncanista: sg

src/commands/lambda/flare.ts
duncanista: Why would a map be an overkill? And if we are only getting three log streams, how would it be slower?

I think whatever is more readable is the way to go.

src/commands/lambda/flare.ts
nhulston: Understood. Logged a ticket for improving flag validation!

src/commands/lambda/flare.ts
duncanista: Yeah, the general idea is to make more readable code. Breaking it down, makes it easier for anyone jumping on the project to get the picture!

src/commands/lambda/flare.ts
nhulston: Just because of the overhead of hashing and other calculations. Since we're not deleting entries or retrieving by key, I thought it would be unnecessary. But yes, using a map would have an unnoticeable difference in speed and be more readable.

src/commands/lambda/flare.ts
duncanista: This makes sense, sounds good. Make a `sleep` function for readability!

src/commands/lambda/flare.ts
duncanista: I see, I didn't catch that it was being used again. Thanks for clarifying!

src/commands/lambda/flare.ts
nhulston: Because I'm throwing an error, it will reappend the `Error: `, so there will be to `Error: ` prefixes.
But I'm realizing that the whole try/catch in `getLogEvents` is unnecessary.

I think it should just be changed to:
```
export const getLogEvents = async (cwlClient: CloudWatchLogsClient, logGroupName: string, logStreamName: string) => {
  const command = new GetLogEventsCommand({
    logGroupName,
    logStreamName,
  })

  try {
    const response = await cwlClient.send(command)
    const logEvents = response.events

    return logEvents === undefined || logEvents.length === 0 ? undefined : logEvents
  } catch (err) {
    throw Error(err.message)
  }
}
```

This way, if `await cliClient.send` throws an error, it will be caught by the caller:
```
          try {
            logEvents = await getLogEvents(cwlClient, logGroupName, logStreamName)
          } catch (err) {
            this.context.stderr.write(
              commonRenderer.renderError(`Unable to get log events for stream ${logStreamName}: ${err.message}`)
            )

            return 1
          }
```

What do you think?

src/commands/lambda/flare.ts
nhulston: Thanks. I have created a `sleep` function that accepts a parameter `ms`!
```
/**
 * @param ms number of milliseconds to sleep
 */
export const sleep = async (ms: number) => {
  await new Promise((resolve) => setTimeout(resolve, ms))
}
```

src/commands/lambda/flare.ts
nhulston: As we talked about, I have broken this down into its own function `getAllLogs`.
Usage: `logs = await getAllLogs(region!, this.functionName)`

What do you think of these changes?

src/commands/lambda/flare.ts
```diff
@@ -86,61 +94,132 @@ export class LambdaFlareCommand extends Command {
...
+          this.context.stdout.write('\n✅ Found log streams:\n')
+          for (const [logStreamName, logEvents] of logs) {
+            const warningMessage =
+              logEvents && logEvents.length > 0
+                ? '\n'
+                : ' - ' + commonRenderer.renderSoftWarning('No log events found in this stream')
+            this.context.stdout.write(`• ${logStreamName}${warningMessage}`)
+          }
+        }
+      }
```
nhulston, line 172: This part is not very readable, but I think it might be necessary. Do you have any suggestions on how to improve it?

duncanista: Following this suggestion, could you find a way to remove the if-else block?
```suggestion
      if (this.withLogs) {
        if (logs.size === 0) {
          this.context.stdout.write(
            commonRenderer.renderSoftWarning(
              'No CloudWatch log streams were found. Logs will not be retrieved or sent.'
            )
          )
        } else {
          this.context.stdout.write('\n✅ Found log streams:\n')
          for (const [logStreamName, logEvents] of logs) {
            let warningMessage = '\n'
            if (logEvents.length === 0) {
                warningMessage = ' - ' + commonRenderer.renderSoftWarning('No log events found in this stream')
            }
            this.context.stdout.write(`• ${logStreamName}${warningMessage}`)
          }
        }
      }
```

nhulston: I have moved the for loop outside of the if/else block. WDYT?
```
if (this.withLogs) {
        if (logs.size === 0) {
          this.context.stdout.write(
            commonRenderer.renderSoftWarning(
              'No CloudWatch log streams were found. Logs will not be retrieved or sent.'
            )
          )
        } else {
          this.context.stdout.write('\n✅ Found log streams:\n')
        }

        for (const [logStreamName, logEvents] of logs) {
          let warningMessage = '\n'
          if (logEvents.length === 0) {
            warningMessage = ' - ' + commonRenderer.renderSoftWarning('No log events found in this stream')
          }
          this.context.stdout.write(`• ${logStreamName}${warningMessage}`)
        }
      }
```

src/commands/lambda/flare.ts
```diff
@@ -218,31 +470,18 @@ export const sendToDatadog = async (
 
   try {
     await axios.post(ENDPOINT_URL, form, headerConfig)
-    context.stdout.write('\n✅ Successfully sent flare file to Datadog Support!\n')
   } catch (err) {
+    // Ensure the root folder is deleted if the request fails
+    deleteFolder(rootFolderPath)
+
     const errResponse: string = err.response?.data?.error
     throw Error(`Failed to send flare file to Datadog Support: ${err.message}. ${errResponse ?? ''}\n`)
```
nhulston, line 478: Note that `err.response` does not exist when `err instanceof Error`, so I did not type guard here

src/commands/lambda/flare.ts
```diff
@@ -150,41 +228,215 @@ export class LambdaFlareCommand extends Command {
...
+ * @param logGroupName name of the log group
+ * @returns an array of the last 3 log stream names or an empty array if no log streams are found
+ * @throws Error if the log streams cannot be retrieved
+ */
+export const getLogStreamNames = async (cwlClient: CloudWatchLogsClient, logGroupName: string) => {
+  const command = new DescribeLogStreamsCommand({
+    logGroupName,
+    limit: 3,
+    descending: true,
+    orderBy: 'LastEventTime',
```
duncanista, line 281: Is there an AWS SDK constant for this?

src/commands/lambda/flare.ts
```diff
@@ -150,41 +228,215 @@ export class LambdaFlareCommand extends Command {
...
+ * Gets the 3 latest log stream names, sorted by last event time
+ * @param cwlClient CloudWatch Logs client
+ * @param logGroupName name of the log group
+ * @returns an array of the last 3 log stream names or an empty array if no log streams are found
+ * @throws Error if the log streams cannot be retrieved
+ */
+export const getLogStreamNames = async (cwlClient: CloudWatchLogsClient, logGroupName: string) => {
+  const command = new DescribeLogStreamsCommand({
+    logGroupName,
+    limit: 3,
```
duncanista, line 279: Maybe turn into a constant?

src/commands/lambda/flare.ts
nhulston: Yes! Changed to `OrderBy.LastEventTime`

src/commands/lambda/flare.ts
nhulston: Done

src/commands/lambda/__tests__/flare.test.ts
```diff
@@ -204,111 +228,405 @@ describe('lambda flare', () => {
...
-
-      expect(fs.existsSync).toHaveBeenCalledWith(MOCK_FOLDER_PATH)
+      expect(() => createDirectories(MOCK_FOLDER_PATH, MOCK_LOG_PATH, new Map())).toThrowErrorMatchingSnapshot()
       expect(fs.mkdirSync).toHaveBeenCalledWith(MOCK_FOLDER_PATH)
-      ;(fs.mkdirSync as jest.Mock).mockRestore()
+      fs.mkdirSync = jest.fn().mockImplementation(() => {})
+    })
+  })
+
+  describe('getLogStreamNames', () => {
```
duncanista, line 275: This tests are not using the AWS SDK mocking library, there should be examples about how client commands are mocked in the repo.

Let's change this tests to use the same mocking strategy for the AWS SDK.

duncanista (changes requested):
Looks good, only change needed is to use the right strategy to mock the AWS SDK. 

Nice refactoring, way easier to review this time!

src/commands/lambda/__tests__/flare.test.ts
nhulston: Changed my `getLogStreamNames` and `getLogEvents` tests, can you take another look?

src/commands/lambda/__tests__/flare.test.ts
```diff
@@ -204,111 +238,405 @@ describe('lambda flare', () => {
...
+    it('throws error when log streams cannot be retrieved', async () => {
+      mockCloudWatchStreams(CLOUDWATCH_CLIENT_MOCK, [])
+      const mockCwlClient = {
+        send: jest.fn().mockRejectedValue(new Error('Cannot retrieve log streams')),
+      }
+
+      await expect(
+        getLogStreamNames((mockCwlClient as unknown) as CloudWatchLogsClient, MOCK_LOG_GROUP)
+      ).rejects.toThrow('Cannot retrieve log streams')
+      expect(mockCwlClient.send).toHaveBeenCalled()
```
duncanista, line 322: This is not the way of mocking an error for an AWS mock client.

src/commands/lambda/__tests__/flare.test.ts
```diff
@@ -204,111 +238,405 @@ describe('lambda flare', () => {
...
+      const logEvents = await getLogEvents(
+        (CLOUDWATCH_CLIENT_MOCK as unknown) as CloudWatchLogsClient,
+        MOCK_LOG_GROUP,
+        MOCK_LOG_STREAM
+      )
+
+      expect(logEvents).toEqual(expectedEvents)
+    })
+
+    it('returns undefined when no log events are found', async () => {
```
duncanista, line 353:
```suggestion
    it('returns empty array when no log events are found', async () => {
```

src/commands/lambda/__tests__/flare.test.ts
```diff
@@ -204,111 +238,405 @@ describe('lambda flare', () => {
...
+      const logEvents = await getLogEvents(
+        (CLOUDWATCH_CLIENT_MOCK as unknown) as CloudWatchLogsClient,
+        MOCK_LOG_GROUP,
+        MOCK_LOG_STREAM
+      )
+
+      expect(logEvents).toEqual([])
+    })
+
+    it('throws error when log events cannot be retrieved', async () => {
```
duncanista, line 365: Same here for mocking error

duncanista (changes requested):
Looks good, couple changes needed and we are good to go.

src/commands/lambda/__tests__/flare.test.ts
```diff
@@ -30,6 +53,10 @@ const MOCK_CONFIG = {
   FunctionArn: 'arn:aws:lambda:us-east-1:123456789012:function:some-function',
   FunctionName: 'some-function',
 }
+const MOCK_LOG_GROUP = 'mockLogGroup'
+const MOCK_OUTPUT_EVENT: OutputLogEvent[] = [{timestamp: 123, message: 'Log 1'}]
+const MOCK_LOGS = new Map().set('log1', MOCK_OUTPUT_EVENT)
+const CLOUDWATCH_CLIENT_MOCK = mockClient(CloudWatchLogsClient)
```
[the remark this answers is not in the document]
nhulston: Good catch!

src/commands/lambda/__tests__/flare.test.ts
nhulston: Good catch, changed to `cloudWatchLogsClientMock.on(DescribeLogStreamsCommand).rejects('Cannot retrieve log streams')`

duncanista (approved):
:shipit:

Issue #823, opened 2023-04-07
Inconsistent usage of DATADOG_API_KEY and DD_API_KEY


albertvaka:
Most Datadog products (Agent, Tracers, etc.) use `DD_API_KEY` as the environment variable for the API key.

In `datadog-ci`, we mostly use `DATADOG_API_KEY` while some commands accept both `DATADOG_API_KEY` and `DD_API_KEY`.

For consistency, I think we should accept `DD_API_KEY` in all commands. We probably want to still accept `DATADOG_API_KEY` for backwards compatibility, but I think we should prefer `DD_API_KEY` instead  (eg: in documentation).

The same applies to `DATADOG_SITE` vs `DD_SITE`.

Drarig29:
Fixed in https://github.com/DataDog/datadog-ci/pull/1723

PR #944, opened 2023-06-27
[ci-visibility] Update docs to recommend `DD_*` env vars everywhere

juan-fernandez:
### What and why?

We want to be as consistent as possible. `DD_*` env vars are the standard. We still want to support `DATADOG_*` but at least the documentation should stick with `DD_*`.

### How?

Update docs to recommend using `DD_*`.


rtrieu (approved):
Looks good!

Drarig29:
Just commenting to link this PR to the relevant issue: https://github.com/DataDog/datadog-ci/issues/823

PR #946, opened 2023-06-29
[ci-visibility] Do not use extension for checking if a path is a file

juan-fernandez:
### What and why?

When passing folders with this shape: `./reports/junit.xml/`, `datadog-ci` considered that to be a valid file (because it used `path.extname`), so when attempting to read its contents it failed. 

This PR fixes that.

### How?

Using `isFile` util function to check if the input is actually a file or not.

### Review checklist

### DIFF
--- a/src/commands/junit/upload.ts
+++ b/src/commands/junit/upload.ts
@@ -37,7 +37,7 @@ import {
   renderSuccessfulUpload,
   renderUpload,
 } from './renderer'
-import {isFalse} from './utils'
+import {isFalse, isFile} from './utils'
 
 const TRACE_ID_HTTP_HEADER = 'x-datadog-trace-id'
 const PARENT_ID_HTTP_HEADER = 'x-datadog-parent-id'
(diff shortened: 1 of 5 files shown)


src/commands/junit/upload.ts
```diff
@@ -251,14 +251,16 @@ export class UploadJUnitXMLCommand extends Command {
...
   private async getMatchingJUnitXMLFiles(spanTags: SpanTags): Promise<Payload[]> {
-    const jUnitXMLFiles = (this.basePaths || []).reduce((acc: string[], basePath: string) => {
-      const isFile = !!path.extname(basePath)
-      if (isFile) {
-        return acc.concat(fs.existsSync(basePath) ? [basePath] : [])
-      }
+    const jUnitXMLFiles = (this.basePaths || [])
+      .reduce((acc: string[], basePath: string) => {
+        const isBasePathFile = isFile(basePath)
+        if (isBasePathFile) {
```
romainkomorndatadog, line 257: Is there a reason ```if (isFile(basePath))``` isn't applicable here?

romainkomorndatadog (approved):

src/commands/junit/upload.ts
juan-fernandez: did that change in https://github.com/DataDog/datadog-ci/pull/946/commits/37d1d26afdfff06550c417a0de21ea0a886414f7

PR #947, opened 2023-06-29
[lambda][flare] Obfuscate secrets

nhulston:
### What and why?

We want to obfuscate secrets (passwords, API keys, and other unknown environment variables) in Lambda function configuration so customers don't send private information to Datadog support. Any non-Datadog environment variables should also be obfuscated.

### How?

1. Create a whitelist of 'allowed' environment variables to not be obfuscated. These are stored in a set in the constant `NON_OBFUSCATED_ENV_VARS` for O(1) lookup.
2. If an environment variable is not in this set, we will pass it to the `getObfuscation` function which returns an obfuscated version of the string.
3. Short strings (<12 characters) will be entirely obfuscated and replaced with `*`'s. Longer strings will preserve the first 2 and last 4 characters of the string, but obfuscate everything in between.
4. Numbers and boolean values are unlikely to be secrets, so they can be skipped and returned as-is
5. Obfuscate the function config before logging to console, saving the file, and sending to Datadog support

### Review checklist

### DIFF
--- a/src/commands/lambda/constants.ts
+++ b/src/commands/lambda/constants.ts
@@ -118,7 +118,8 @@ export const AWS_SECRET_ARN_REG_EXP = /arn:aws:secretsmanager:[\w-]+:\d{12}:secr
 export const DATADOG_API_KEY_REG_EXP = /(?<![a-f0-9])[a-f0-9]{32}(?![a-f0-9])/g
 export const DATADOG_APP_KEY_REG_EXP = /(?<![a-f0-9])[a-f0-9]{40}(?![a-f0-9])/g
 
-// Lambda Flare environment variables
+// Environment Variables whose values don't need to be
+// masked in a Flare
 export const SKIP_MASKING_ENV_VARS = new Set([
   AWS_LAMBDA_EXEC_WRAPPER_VAR,
   SITE_ENV_VAR,
@@ -128,8 +129,7 @@ export const SKIP_MASKING_ENV_VARS = new Set([
   VERSION_ENV_VAR,
   ENVIRONMENT_ENV_VAR,
   EXTRA_TAGS_ENV_VAR,
+  PROFILER_ENV_VAR,
   PROFILER_PATH_ENV_VAR,
   DOTNET_TRACER_HOME_ENV_VAR,
-  CI_SITE_ENV_VAR,
-  AWS_DEFAULT_REGION_ENV_VAR,
 ])
--- a/src/commands/lambda/constants.ts
+++ b/src/commands/lambda/constants.ts
@@ -117,3 +117,19 @@ export const AWS_SECRET_ACCESS_KEY_REG_EXP = /(?<![A-Za-z0-9/+=])[A-Za-z0-9/+=]{
 export const AWS_SECRET_ARN_REG_EXP = /arn:aws:secretsmanager:[\w-]+:\d{12}:secret:.+/
 export const DATADOG_API_KEY_REG_EXP = /(?<![a-f0-9])[a-f0-9]{32}(?![a-f0-9])/g
 export const DATADOG_APP_KEY_REG_EXP = /(?<![a-f0-9])[a-f0-9]{40}(?![a-f0-9])/g
+
+// Lambda Flare environment variables
+export const SKIP_MASKING_ENV_VARS = new Set([
+  AWS_LAMBDA_EXEC_WRAPPER_VAR,
+  SITE_ENV_VAR,
+  LOG_LEVEL_ENV_VAR,
+  LAMBDA_HANDLER_ENV_VAR,
+  SERVICE_ENV_VAR,
+  VERSION_ENV_VAR,
+  ENVIRONMENT_ENV_VAR,
+  EXTRA_TAGS_ENV_VAR,
+  PROFILER_PATH_ENV_VAR,
+  DOTNET_TRACER_HOME_ENV_VAR,
+  CI_SITE_ENV_VAR,
+  AWS_DEFAULT_REGION_ENV_VAR,
+])
--- a/src/commands/lambda/flare.ts
+++ b/src/commands/lambda/flare.ts
@@ -30,6 +30,8 @@ const LOGS_DIRECTORY = 'logs'
 const FUNCTION_CONFIG_FILE_NAME = 'function_config.json'
 const ZIP_FILE_NAME = 'lambda-flare-output.zip'
 const LOG_STREAM_COUNT = 3
+const FULL_OBFUSCATION = '****************'
+const MIDDLE_OBFUSCATION = '**********'
 const NON_OBFUSCATED_ENV_VARS = new Set([
   constants.SITE_ENV_VAR,
   constants.LOG_LEVEL_ENV_VAR,
@@ -286,16 +288,15 @@ export const getObfuscation = (original: string) => {
   }
 
   // Obfuscate entire string if it's short
-  if (original.length < 20) {
-    return '*'.repeat(original.length)
+  if (original.length < 12) {
+    return FULL_OBFUSCATION
   }
 
   // Keep first two and last four characters if it's long
   const front = original.substring(0, 2)
-  const middle = '*'.repeat(original.length - 6)
   const end = original.substring(original.length - 4)
 
-  return front + middle + end
+  return front + MIDDLE_OBFUSCATION + end
 }
 
 /**
--- a/src/commands/lambda/flare.ts
+++ b/src/commands/lambda/flare.ts
@@ -9,14 +9,14 @@ import {
   OrderBy,
   OutputLogEvent,
 } from '@aws-sdk/client-cloudwatch-logs'
-import {LambdaClient, LambdaClientConfig} from '@aws-sdk/client-lambda'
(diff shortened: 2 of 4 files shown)


src/commands/lambda/flare.ts
```diff
@@ -30,6 +30,18 @@ const LOGS_DIRECTORY = 'logs'
...
+const NON_OBFUSCATED_ENV_VARS = new Set([
+  constants.SITE_ENV_VAR,
+  constants.LOG_LEVEL_ENV_VAR,
+  constants.LAMBDA_HANDLER_ENV_VAR,
+  constants.SERVICE_ENV_VAR,
+  constants.VERSION_ENV_VAR,
+  constants.ENVIRONMENT_ENV_VAR,
+  constants.EXTRA_TAGS_ENV_VAR,
+  constants.DOTNET_TRACER_HOME_ENV_VAR,
+])
```
nhulston, line 44: Note that boolean/number environment variables are not included since we don't obfuscate booleans and numbers. An example is `DD_TRACE_ENABLED`

src/commands/lambda/flare.ts
```diff
@@ -229,6 +242,62 @@ export class LambdaFlareCommand extends Command {
...
+export const getObfuscation = (original: string) => {
+  // Don't obfuscate booleans
+  if (original.toLowerCase() === 'true' || original.toLowerCase() === 'false') {
+    return original
+  }
+
+  // Dont obfuscate numbers
+  if (!isNaN(Number(original))) {
+    return original
+  }
```
nhulston, line 287: I used `Number` instead of `parseInt` because `Number('123abc')` will return `NaN` while `parseInt('123abc')` will return `123`.

src/commands/lambda/flare.ts
```diff
@@ -30,6 +30,18 @@ const LOGS_DIRECTORY = 'logs'
 const FUNCTION_CONFIG_FILE_NAME = 'function_config.json'
 const ZIP_FILE_NAME = 'lambda-flare-output.zip'
 const LOG_STREAM_COUNT = 3
+const FULL_OBFUSCATION = '****************'
+const MIDDLE_OBFUSCATION = '**********'
```
duncanista, line 34: This constants since one is the subset of the other.
And this can be easily calculated within the method that masks data.

Why don't we go for a different approach?

src/commands/lambda/flare.ts
```diff
@@ -229,6 +242,62 @@ export class LambdaFlareCommand extends Command {
   }
 }
 
+/**
+ * Obfuscate the environment variables in a Lambda function configuration
+ * @param config
+ */
+export const obfuscateConfig = (config: FunctionConfiguration) => {
```
duncanista, line 249: I think the correct name for this would be `maskConfig`, obfuscation is the process, but what you are doing here is a tactic of it, you are masking the data.

src/commands/lambda/flare.ts
```diff
@@ -229,6 +242,62 @@ export class LambdaFlareCommand extends Command {
...
+  }
+}
+
+/**
+ * Obfuscate a string but keep the first two and last four characters
+ * Obfuscate the entire string if it's short
+ * @param original the string to obfuscate
+ * @returns the obfuscated string
+ */
+export const getObfuscation = (original: string) => {
```
duncanista, line 278: Same here, we are masking data.

src/commands/lambda/flare.ts
```diff
@@ -229,6 +242,62 @@ export class LambdaFlareCommand extends Command {
...
+ * Obfuscate the environment variables in a Lambda function configuration
+ * @param config
+ */
+export const obfuscateConfig = (config: FunctionConfiguration) => {
+  const environmentVariables = config.Environment?.Variables
+  if (!environmentVariables) {
+    return config
+  }
+
+  const obfuscatedEnvironmentVariables: {[key: string]: string} = {}
```
duncanista, line 255: Same here.

src/commands/lambda/flare.ts
```diff
@@ -229,6 +242,62 @@ export class LambdaFlareCommand extends Command {
...
+  for (const [key, value] of Object.entries(environmentVariables)) {
+    obfuscatedEnvironmentVariables[key] = value
+    if (!NON_OBFUSCATED_ENV_VARS.has(key)) {
```
duncanista, line 258: This reads as `if not not_array has key` could we improve the naming to make it simpler?

Something that could be read as `if key in array skip`, maybe even just improving the name can lead to a change.

Something like `SKIP_MASKING_ENV_VARS`, whatever you think its best
```suggestion
    if (SKIP_MASKING_ENV_VARS.has(key)) {
        continue
    }
    // mask here
```

src/commands/lambda/flare.ts
duncanista: Let's move this array into `constants.ts`

duncanista: This is also missing some environmet variables.

nhulston: I added `CI_SITE_ENV_VAR` and `AWS_DEFAULT_REGION_ENV_VAR`. Is there any other missing environment variables? Everything else seems to be a boolean/number or a secret.

src/commands/lambda/constants.ts
```diff
@@ -117,3 +117,19 @@ export const AWS_SECRET_ACCESS_KEY_REG_EXP = /(?<![A-Za-z0-9/+=])[A-Za-z0-9/+=]{
 export const AWS_SECRET_ARN_REG_EXP = /arn:aws:secretsmanager:[\w-]+:\d{12}:secret:.+/
 export const DATADOG_API_KEY_REG_EXP = /(?<![a-f0-9])[a-f0-9]{32}(?![a-f0-9])/g
 export const DATADOG_APP_KEY_REG_EXP = /(?<![a-f0-9])[a-f0-9]{40}(?![a-f0-9])/g
+
+// Lambda Flare environment variables
```
duncanista, line 121:
```suggestion
// Environment Variables whose values don't need to be
// masked in a Flare
```

src/commands/lambda/constants.ts
```diff
@@ -117,3 +117,19 @@ export const AWS_SECRET_ACCESS_KEY_REG_EXP = /(?<![A-Za-z0-9/+=])[A-Za-z0-9/+=]{
...
+  LOG_LEVEL_ENV_VAR,
+  LAMBDA_HANDLER_ENV_VAR,
+  SERVICE_ENV_VAR,
+  VERSION_ENV_VAR,
+  ENVIRONMENT_ENV_VAR,
+  EXTRA_TAGS_ENV_VAR,
+  PROFILER_PATH_ENV_VAR,
+  DOTNET_TRACER_HOME_ENV_VAR,
+  CI_SITE_ENV_VAR,
+  AWS_DEFAULT_REGION_ENV_VAR,
```
duncanista, line 134: This is not going to be in an AWS Lambda env vars.

src/commands/lambda/constants.ts
```diff
@@ -117,3 +117,19 @@ export const AWS_SECRET_ACCESS_KEY_REG_EXP = /(?<![A-Za-z0-9/+=])[A-Za-z0-9/+=]{
...
+  SITE_ENV_VAR,
+  LOG_LEVEL_ENV_VAR,
+  LAMBDA_HANDLER_ENV_VAR,
+  SERVICE_ENV_VAR,
+  VERSION_ENV_VAR,
+  ENVIRONMENT_ENV_VAR,
+  EXTRA_TAGS_ENV_VAR,
+  PROFILER_PATH_ENV_VAR,
+  DOTNET_TRACER_HOME_ENV_VAR,
+  CI_SITE_ENV_VAR,
```
duncanista, line 133: Same for this.

src/commands/lambda/constants.ts
```diff
@@ -117,3 +117,19 @@ export const AWS_SECRET_ACCESS_KEY_REG_EXP = /(?<![A-Za-z0-9/+=])[A-Za-z0-9/+=]{
 export const AWS_SECRET_ARN_REG_EXP = /arn:aws:secretsmanager:[\w-]+:\d{12}:secret:.+/
 export const DATADOG_API_KEY_REG_EXP = /(?<![a-f0-9])[a-f0-9]{32}(?![a-f0-9])/g
 export const DATADOG_APP_KEY_REG_EXP = /(?<![a-f0-9])[a-f0-9]{40}(?![a-f0-9])/g
+
+// Lambda Flare environment variables
+export const SKIP_MASKING_ENV_VARS = new Set([
```
duncanista, line 122: Missing `PROFILER_ENV_VAR`

src/commands/lambda/flare.ts
nhulston: Resolved as we talked about it in Slack

src/commands/lambda/flare.ts
nhulston: Good suggestion! Changed

src/commands/lambda/constants.ts
nhulston: Added

duncanista (approved):

PR #953, opened 2023-07-03
[lambda][flare] Ask for user confirmation before sending files to support

nhulston:
### What and why?

We want to ask users for confirmation before sending their files to support.
If they answer `no` or anything other than `y` or `Y`, the flare files are not sent and are deleted
They must answer `y` or `Y` before the files are sent to support.
This is intended to:
1. Give the user an opportunity to check the files and see if there is anything they don't want to send to support
2. Prevent users from accidentally spamming their support ticket with files

Screenshot 2023-07-03 at 2 38 49 PM

Screenshot 2023-07-03 at 2 39 09 PM

### How?

Use `inquirer` to prompt the user for confirmation.
Exit the program early if they answer `no`, continue otherwise.

### Review checklist

### DIFF
--- a/src/commands/lambda/__tests__/flare.test.ts
+++ b/src/commands/lambda/__tests__/flare.test.ts
@@ -809,7 +809,7 @@ describe('lambda flare', () => {
     })
   })
 
-  describe('prompt to send works as expected', () => {
+  describe('prompts for confirmation before sending', () => {
     beforeEach(() => {
       ;(getAWSCredentials as any).mockResolvedValue(mockAwsCredentials)
     })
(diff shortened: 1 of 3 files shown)


src/commands/lambda/__tests__/flare.test.ts
```diff
@@ -803,4 +808,30 @@ describe('lambda flare', () => {
       expect(output).toMatchSnapshot()
     })
   })
+
+  describe('prompt to send works as expected', () => {
```
nhulston, line 812: Not really a good name

sfirrin (approved):

PR #954, opened 2023-07-03
[lambda][flare] Add human readable time to CloudWatch CSV files

nhulston:
### What and why?

Currently, only the timestamp in milliseconds since UNIX epoch is recorded in the CSV files for CloudWatch logs.
This change adds human readable time as a column in the CSV file in the format `YYYY-MM-DD HH:mm:ss.<milliseconds>`.
This is intended to improve the readability of these CSV files to help the support team.

### How?

1. Convert UNIX milliseconds to Date object using `new Date(timestamp)`
2. Convert Date object to string in the format described above using 
```
date.toISOString().replace('T', ' ').replace('Z', '')
```

Screenshot 2023-07-03 at 2 51 27 PM

### Review checklist

### DIFF
--- a/src/commands/lambda/flare.ts
+++ b/src/commands/lambda/flare.ts
@@ -441,11 +441,15 @@ export const writeFile = (filePath: string, data: string) => {
  * @returns the CSV string
  */
 export const convertToCSV = (logEvents: OutputLogEvent[]) => {
-  const rows = [['timestamp', 'message']]
+  const rows = [['timestamp', 'datetime', 'message']]
   for (const logEvent of logEvents) {
     const timestamp = `"${logEvent.timestamp ?? ''}"`
+    let datetime = ''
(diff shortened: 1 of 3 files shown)


src/commands/lambda/flare.ts
```diff
@@ -441,11 +441,16 @@ export const writeFile = (filePath: string, data: string) => {
  * @returns the CSV string
  */
 export const convertToCSV = (logEvents: OutputLogEvent[]) => {
-  const rows = [['timestamp', 'message']]
+  const rows = [['timestamp', 'datetime', 'message']]
```
nhulston, line 444: Do we want to call the new column `datetime` or is there a better name?

sfirrin: That's good

sfirrin (approved):

PR #955, opened 2023-07-03
[lambda][flare] Get resource tags

nhulston:
### What and why?

- Get resource tags for an AWS lambda function
- Save them in `tags.json`
- This is because resource tags can be useful to the Datadog support team when resolving customer issues

Screenshot 2023-07-03 at 3 44 15 PM

Screenshot 2023-07-03 at 3 53 02 PM

I also refactored tests a little bit by standardizing `region` in `flare.test.ts` to use `const MOCK_REGION = 'us-east-1'`

### How?

1. Create a `getTags` function to gather resource tags
2. Use `ListTagsCommand` in the AWS SDK to get resource tags for a particular function ARN
3. The function ARN is retrieved from the function config, which is fetched previously in the command
4. Use try/catch statements to ensure human-readable error logging
5. Write to `tags.json` when tags are found using the `writeFile` method. When no tags are found, no file is written.

### Review checklist

### DIFF
--- a/src/commands/lambda/__tests__/fixtures.ts
+++ b/src/commands/lambda/__tests__/fixtures.ts
@@ -199,10 +199,10 @@ export const mockCloudWatchLogEvents = (
 }
 
 export const mockResourceTags = (
-  cloudWatchLogsClientMock: AwsStub<ServiceInputTypes, ServiceOutputTypes>,
+  lambdaClientMock: AwsStub<ServiceInputTypes, ServiceOutputTypes>,
   output: ListTagsCommandOutput
 ) => {
-  cloudWatchLogsClientMock.on(ListTagsCommand).resolves(output)
+  lambdaClientMock.on(ListTagsCommand).resolves(output)
 }
 
 export const mockAwsAccount = '123456789012'
--- a/src/commands/lambda/__tests__/fixtures.ts
+++ b/src/commands/lambda/__tests__/fixtures.ts
@@ -24,7 +24,7 @@ import {
   TagResourceCommand,
   ListTagsCommand,
   ListTagsResponse,
-  GetLayerVersionCommandInput,
+  GetLayerVersionCommandInput, ServiceInputTypes, ServiceOutputTypes, ListTagsCommandOutput,
 } from '@aws-sdk/client-lambda'
 import {AwsStub} from 'aws-sdk-client-mock'
 import {Cli, Command} from 'clipanion/lib/advanced'
@@ -193,6 +193,13 @@ export const mockCloudWatchLogEvents = (
   events: OutputLogEvent[]
 ) => {
   cloudWatchLogsClientMock.on(GetLogEventsCommand).resolves({events})
+}
+
+export const mockResourceTags = (
+  cloudWatchLogsClientMock: AwsStub<ServiceInputTypes, ServiceOutputTypes>,
+  output: ListTagsCommandOutput
+) => {
+  cloudWatchLogsClientMock.on(ListTagsCommand).resolves(output)
 }
 
 export const mockAwsAccount = '123456789012'
--- a/src/commands/lambda/__tests__/fixtures.ts
+++ b/src/commands/lambda/__tests__/fixtures.ts
@@ -24,7 +24,10 @@ import {
   TagResourceCommand,
   ListTagsCommand,
   ListTagsResponse,
-  GetLayerVersionCommandInput, ServiceInputTypes, ServiceOutputTypes, ListTagsCommandOutput,
+  GetLayerVersionCommandInput,
+  ServiceInputTypes,
+  ServiceOutputTypes,
+  ListTagsCommandOutput,
 } from '@aws-sdk/client-lambda'
 import {AwsStub} from 'aws-sdk-client-mock'
 import {Cli, Command} from 'clipanion/lib/advanced'
--- a/src/commands/lambda/__tests__/flare.test.ts
+++ b/src/commands/lambda/__tests__/flare.test.ts
@@ -538,6 +538,11 @@ describe('lambda flare', () => {
 
   describe('getTags', () => {
     const MOCK_ARN = 'arn:aws:lambda:us-east-1:123456789012:function:my-function'
+
+    afterAll(() => {
+      mockResourceTags(lambdaClientMock, MOCK_TAGS)
+    })
+
     it('should return the tags when they exist', async () => {
       const mockTags: any = {Tags: {Key1: 'Value1', Key2: 'Value2'}}
 
@@ -560,7 +565,6 @@ describe('lambda flare', () => {
       lambdaClientMock.on(ListTagsCommand).rejects(new Error('Test Error'))
 
       await expect(getTags(lambdaClientMock, MOCK_REGION, MOCK_ARN)).rejects.toThrow(errorMessage)
-      mockResourceTags(lambdaClientMock, MOCK_TAGS)
     })
   })
 
--- a/src/commands/lambda/__tests__/flare.test.ts
+++ b/src/commands/lambda/__tests__/flare.test.ts
@@ -53,7 +53,8 @@ const MOCK_FOLDER_PATH = path.join(MOCK_CWD, MOCK_FOLDER_NAME)
 const MOCK_FILE_NAME = 'function_config.json'
 const MOCK_FILES = new Set([MOCK_FILE_NAME, 'file1.csv', 'file2.csv', 'file3.csv'])
 const MOCK_ZIP_PATH = 'output.zip'
-const MOCK_REQUIRED_FLAGS = ['lambda', 'flare', '-f', 'func', '-r', 'us-east-1', '-c', '123', '-e', 'test@test.com']
+const MOCK_REGION = 'us-east-1'
+const MOCK_REQUIRED_FLAGS = ['lambda', 'flare', '-f', 'func', '-r', MOCK_REGION, '-c', '123', '-e', 'test@test.com']
 const MOCK_CONFIG = {
   Environment: {
     Variables: {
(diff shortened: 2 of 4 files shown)


src/commands/lambda/flare.ts
```diff
@@ -29,6 +29,7 @@ const ENDPOINT_URL = 'https://datad0g.com/api/ui/support/serverless/flare'
 const FLARE_OUTPUT_DIRECTORY = '.datadog-ci'
 const LOGS_DIRECTORY = 'logs'
 const FUNCTION_CONFIG_FILE_NAME = 'function_config.json'
+const TAGS_FILE_NAME = 'tags.json'
```
nhulston, line 32: Is this a good name? Or should it be more descriptive like `resource_tags.json`?
Looks like AWS just calls them "Tags"

src/commands/lambda/flare.ts
```diff
@@ -141,6 +142,30 @@ export class LambdaFlareCommand extends Command {
     const configStr = util.inspect(config, false, undefined, true)
     this.context.stdout.write(`\n${configStr}\n`)
 
+    // Get tags
+    this.context.stdout.write('\n🏷 Getting Resource Tags...\n')
+    const arn = config.FunctionArn ?? this.functionName
```
nhulston, line 147: Do we want to add this `?? this.functionName` to handle the case where `config.FunctionArn` is undefined?
Or would it be better practice to immediately write an error and exit when `config.FunctionArn` is undefined?

src/commands/lambda/flare.ts
duncanista: Mmm, let's stick with what AWS calls them. These are indeed the resource tags, but the resource is Lambda, so, I think it makes sense to have them like this.

src/commands/lambda/flare.ts
duncanista: FunctionArn will never be undefined, so we could do `config.FunctionArn!`.

I think it's a safe assumption. As per the things you said.

src/commands/lambda/__tests__/flare.test.ts
```diff
@@ -66,6 +70,7 @@ const MOCK_LOG_GROUP = 'mockLogGroup'
 const MOCK_OUTPUT_EVENT: OutputLogEvent[] = [{timestamp: 123, message: 'Log 1'}]
 const MOCK_LOGS = new Map().set('log1', MOCK_OUTPUT_EVENT)
 const cloudWatchLogsClientMock = mockClient(CloudWatchLogsClient)
+const lambdaClientMock: any = mockClient(LambdaClient)
```
duncanista, line 73: Why are we typing as `any` here, but not in the CWL?

src/commands/lambda/__tests__/flare.test.ts
```diff
@@ -104,6 +109,10 @@ const mockJSZip = {
 }
 ;(JSZip as any).mockImplementation(() => mockJSZip)
 
+// Resource tags mock
+const mockTags: any = {Tags: {}}
+mockResourceTags(lambdaClientMock, mockTags)
```
duncanista, line 114: Why is the purpose of having this all the way to the top?

Is this because we want to ensure we mock the tags for every test? If so, why are we mocking on the top level and not on a `beforeAll`? Are there any advantages in comparison to that?

src/commands/lambda/__tests__/fixtures.ts
```diff
@@ -195,6 +198,13 @@ export const mockCloudWatchLogEvents = (
   cloudWatchLogsClientMock.on(GetLogEventsCommand).resolves({events})
 }
 
+export const mockResourceTags = (
+  cloudWatchLogsClientMock: AwsStub<ServiceInputTypes, ServiceOutputTypes>,
```
duncanista, line 202: Is this only for a CWL client mock? I see you are passing a Lambda client instead.

src/commands/lambda/flare.ts
```diff
@@ -141,6 +142,30 @@ export class LambdaFlareCommand extends Command {
...
+        this.context.stderr.write(commonRenderer.renderError(err.message))
+      }
+
+      return 1
+    }
+    const tagLength = Object.keys(tags).length
+    if (tagLength === 0) {
+      this.context.stdout.write(
+        commonRenderer.renderSoftWarning(
+          `No resource tags were found, so the \`${TAGS_FILE_NAME}\` file will not be created or sent.`
```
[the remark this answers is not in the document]
nhulston: I agree, I don't think we need to explicitly say this. Changed to your suggestion

src/commands/lambda/__tests__/flare.test.ts
nhulston: If I remove the typing as `any`, then `await getTags(lambdaClientMock, MOCK_REGION, MOCK_ARN)` has a type error:
```
TS2345: Argument of type 'AwsStub<ServiceInputTypes, ServiceOutputTypes>' is not assignable to parameter of type 'LambdaClient'.
```

Typing it as `LambdaClient` instead of `any` gives the error:
```
TS2739: Type 'AwsStub ' is missing the following properties from type 'LambdaClient': config, destroy, middlewareStack
```

src/commands/lambda/__tests__/flare.test.ts
nhulston: This mock is required to be a global mock, or else all other tests will fail.
`[Error] Unable to get resource tags: Cannot read properties of undefined (reading 'Tags')`

This could be done in a `beforeAll` but it seems unnecessary to re-mock this before every test when the mocked tags aren't being changed.

src/commands/lambda/__tests__/fixtures.ts
nhulston: Looks like I misnamed the parameter. I have changed `cloudWatchLogsClientMock` to `lambdaClientMock`

src/commands/lambda/__tests__/flare.test.ts
duncanista: I see, sounds about right, normally we pass it as `lambdaClientMock as any`, could we do that on the function in which you are passing it? Just to keep the typing consistent in cases where the `any` typing is not needed?

src/commands/lambda/__tests__/flare.test.ts
duncanista: But a `beforeAll` isn't re-mocking this if set on the top level describe. Wouldn't that be preferred? It's the same behavior for testing.

src/commands/lambda/__tests__/flare.test.ts
```diff
@@ -527,6 +535,34 @@ describe('lambda flare', () => {
     })
   })
 
+  describe('getTags', () => {
```
duncanista, line 538: For example, here, you are never mocking back to the original value. That means that you are overwriting the previous value. What happens if you get the tags after the last test? i.e. `afterEach` get the tags again. Would the last call throw an error again? 

That's why I mentioned about the mocking on the top level and never restoring that value.

src/commands/lambda/__tests__/flare.test.ts
nhulston: Yes I'll pass it into functions as `lambdaClientMock as any` instead!

src/commands/lambda/__tests__/flare.test.ts
nhulston: Oh I was confusing `beforeAll` with `beforeEach`.

I can't have `const mockTags` in a `beforeAll` because the const needs to be in global scope. This way, the tags can be re-mocked in later tests.

For example, this test has the tags throw an error, so it must be overidden to ensure later tests succeed.

```
    it('should throw an error when the command fails', async () => {
      const errorMessage = 'Unable to get resource tags: Test Error'
      lambdaClientMock.on(ListTagsCommand).rejects(new Error('Test Error'))

      await expect(getTags(lambdaClientMock, MOCK_REGION, MOCK_ARN)).rejects.toThrow(errorMessage)
      mockResourceTags(lambdaClientMock, mockTags)
    })
```

src/commands/lambda/__tests__/flare.test.ts
nhulston: Good point, I moved the original mock to be in an `afterAll`

src/commands/lambda/__tests__/flare.test.ts
duncanista: I don't see the change 😯

nhulston: Oops, it's changed now

src/commands/lambda/__tests__/flare.test.ts
```diff
@@ -105,6 +111,10 @@ const mockJSZip = {
 ;(JSZip as any).mockImplementation(() => mockJSZip)
 
 describe('lambda flare', () => {
+  beforeAll(() => {
+    mockResourceTags(lambdaClientMock, MOCK_TAGS)
+  })
```
duncanista, line 116: Nice

src/commands/lambda/__tests__/flare.test.ts
```diff
@@ -50,7 +53,8 @@ const MOCK_FOLDER_PATH = path.join(MOCK_CWD, MOCK_FOLDER_NAME)
 const MOCK_FILE_NAME = 'function_config.json'
 const MOCK_FILES = new Set([MOCK_FILE_NAME, 'file1.csv', 'file2.csv', 'file3.csv'])
 const MOCK_ZIP_PATH = 'output.zip'
-const MOCK_REQUIRED_FLAGS = ['lambda', 'flare', '-f', 'func', '-r', 'us-west-2', '-c', '123', '-e', 'test@test.com']
+const MOCK_REGION = 'us-east-1'
```
duncanista, line 56: Looks good overall, the only thing I didn't quite get was the updating in every command as `MOCK_REGION`, I understand in the future we would allow the user to flare multiple functions, which is going to make testing harder if we keep this approach.

duncanista (approved):

src/commands/lambda/__tests__/flare.test.ts
nhulston: Ah good point. In the future I won't use `MOCK_REGION` if the function specified is a function ARN.
This will only be used in the cases where `--region` is required as a flag

PR #956, opened 2023-07-04
Improve unshallowing and git metadata upload

edrevo:
### What and why?

Improve git metadata upload:
- We now work with git repositories whose default remote is not "origin"
- We do not change the git config when doing git metadata upload
- Improved git metadata upload on error conditions
- Reduced code duplication

### Review checklist

### DIFF
--- a/src/commands/git-metadata/__tests__/gitdb.test.ts
+++ b/src/commands/git-metadata/__tests__/gitdb.test.ts
@@ -269,7 +269,7 @@ describe('gitdb', () => {
           input: [
             '--shallow-since="1 month ago"',
             '--update-shallow',
-            '--filter="blob:none"',
+            '--filter=blob:none',
             '--recurse-submodules=no',
             'origin',
             'COMMIT',
@@ -330,7 +330,7 @@ describe('gitdb', () => {
           input: [
             '--shallow-since="1 month ago"',
             '--update-shallow',
-            '--filter="blob:none"',
+            '--filter=blob:none',
             '--recurse-submodules=no',
             'myorigin',
             'COMMIT',
@@ -435,7 +435,7 @@ describe('gitdb', () => {
         },
       ],
       revparse: [{input: '--is-shallow-repository', output: 'false'}],
-      version: [{input: undefined, output: newGitVersion}],
+      version: [],
       execSync: [
         {
           input: `git pack-objects --compression=9 --max-pack-size=3m ${tmpdir}/1000`,
@@ -520,7 +520,7 @@ describe('gitdb', () => {
         },
       ],
       revparse: [{input: '--is-shallow-repository', output: 'false'}],
-      version: [{input: undefined, output: newGitVersion}],
+      version: [],
       execSync: [
         {
           input: `git pack-objects --compression=9 --max-pack-size=3m ${tmpdir}/1000`,
@@ -621,7 +621,7 @@ describe('gitdb', () => {
         },
       ],
       revparse: [{input: '--is-shallow-repository', output: 'false'}],
-      version: [{input: undefined, output: newGitVersion}],
+      version: [],
       execSync: [
         {
           input: `git pack-objects --compression=9 --max-pack-size=3m ${tmpdir}/1000`,
@@ -718,7 +718,7 @@ describe('gitdb', () => {
         },
       ],
       revparse: [{input: '--is-shallow-repository', output: 'false'}],
-      version: [{input: undefined, output: newGitVersion}],
+      version: [],
       execSync: [
         {
           input: `git pack-objects --compression=9 --max-pack-size=3m ${tmpdir}/1000`,
@@ -823,7 +823,7 @@ describe('gitdb', () => {
       ],
       raw: [],
       revparse: [{input: '--is-shallow-repository', output: 'false'}],
-      version: [{input: undefined, output: newGitVersion}],
+      version: [],
       execSync: [],
       axios: [
         {
@@ -928,7 +928,7 @@ describe('gitdb', () => {
       ],
       raw: [],
       revparse: [{input: '--is-shallow-repository', output: 'false'}],
-      version: [{input: undefined, output: newGitVersion}],
+      version: [],
       execSync: [],
       axios: [
         {
@@ -1015,7 +1015,7 @@ describe('gitdb', () => {
         },
       ],
       revparse: [{input: '--is-shallow-repository', output: 'false'}],
-      version: [{input: undefined, output: newGitVersion}],
+      version: [],
       execSync: [],
       axios: [
         {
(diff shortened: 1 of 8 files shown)


src/helpers/git/get-git-data.ts
```diff
@@ -9,10 +9,11 @@ export const gitRemote = async (git: simpleGit.SimpleGit): Promise<string> => {
   if (remotes.length === 0) {
     throw new Error('No git remotes available')
   }
+  const defaultRemote = await getDefaultRemoteName(git)
 
   for (const remote of remotes) {
     // We're trying to pick the remote called with the default git name 'origin'.
```
lraucy, line 15: the comment is now outdated

lraucy (approved):
lgtm, one small nit for keeping comments updated!

src/commands/git-metadata/__tests__/gitdb.test.ts
```diff
@@ -47,17 +48,66 @@ describe('gitdb', () => {
   })
   const testError = new Error('call failed')
 
+  // This is a bit hacky but the documentation of simpleGit asks you to depend on
+  // a behavior of VersionResult which is not debined in its typings: it has an
```
juan-fernandez, line 52:
```suggestion
  // a behavior of VersionResult which is not defined in its typings: it has an
```

src/commands/git-metadata/__tests__/gitdb.test.ts
```diff
@@ -47,17 +48,66 @@ describe('gitdb', () => {
...
+  // override on toString. Here I am replicating this toString behavior so that
+  // the tests pass
```
juan-fernandez, line 54: any chance we can add a permalink to where they're adding this `toString` behavior?

src/commands/git-metadata/gitdb.ts
```diff
@@ -116,13 +118,25 @@ const getLatestLocalCommits = async (git: simpleGit.SimpleGit) => {
...
 
 const unshallowRepositoryWhenNeeded = async (log: Logger, git: simpleGit.SimpleGit) => {
-  const isShallow = (await git.revparse('--is-shallow-repository')) === 'true'
-  if (isShallow) {
+  const shallowCmdPromise = git.revparse('--is-shallow-repository')
+  const versionCmdPromise: simpleGit.Response<simpleGit.VersionResult> = git.version()
+  const isShallow = (await shallowCmdPromise) === 'true'
+  const gitversion = String(await versionCmdPromise)
+  if (isShallow && gte(gitversion, '2.27.0')) {
     log.info('[unshallow] Git repository is a shallow clone, unshallowing it...')
```
juan-fernandez, line 126: nitpick/idea: how would this look?
```suggestion
  const isShallow = (await git.revparse('--is-shallow-repository')) === 'true'
  if (!isShallow) {
    return
  }
  const gitversion = String(await git.version())
  if (lte(gitversion, '2.27.0')) {
    return
  }
  log.info('[unshallow] Git repository is a shallow clone, unshallowing it...')
```

the early returns are a matter of taste, so feel free to keep the logic as is. We could, however: 
* get `isShallow` and `gitversion` directly, without passing through `shallowCmdPromise` and `versionCmdPromise`
* avoid calling `git.version()` if the repository isn't shallow

src/commands/git-metadata/gitdb.ts
```diff
@@ -116,13 +118,25 @@ const getLatestLocalCommits = async (git: simpleGit.SimpleGit) => {
...
-    await git.fetch([`--shallow-since="${MAX_HISTORY.oldestCommits}"`, '--update-shallow', '--refetch'])
+    const headCmdPromise = git.revparse('HEAD')
+    const remoteNameCmdPromise = getDefaultRemoteName(git)
+    log.info(
+      `[unshallow] Running git fetch --shallow-since="${MAX_HISTORY.oldestCommits}" --update-shallow --filter="blob:none" --recurse-submodules=no`
+    )
+    await git.fetch([
+      `--shallow-since="${MAX_HISTORY.oldestCommits}"`,
+      '--update-shallow',
+      '--filter="blob:none"',
```
juan-fernandez, line 135: this has bitten me in the past 😅 
```suggestion
      '--filter=blob:none',
```

src/commands/git-metadata/gitdb.ts
```diff
@@ -116,13 +118,25 @@ const getLatestLocalCommits = async (git: simpleGit.SimpleGit) => {
...
+  if (isShallow && gte(gitversion, '2.27.0')) {
     log.info('[unshallow] Git repository is a shallow clone, unshallowing it...')
-    log.info('[unshallow] Setting remote.origin.partialclonefilter to "blob:none" to avoid fetching file content')
-    await git.addConfig('remote.origin.partialclonefilter', 'blob:none')
-    log.info(`[unshallow] Running git fetch --shallow-since="${MAX_HISTORY.oldestCommits}" --update-shallow --refetch`)
-    await git.fetch([`--shallow-since="${MAX_HISTORY.oldestCommits}"`, '--update-shallow', '--refetch'])
+    const headCmdPromise = git.revparse('HEAD')
+    const remoteNameCmdPromise = getDefaultRemoteName(git)
+    log.info(
+      `[unshallow] Running git fetch --shallow-since="${MAX_HISTORY.oldestCommits}" --update-shallow --filter="blob:none" --recurse-submodules=no`
```
juan-fernandez, line 130:
```suggestion
      `[unshallow] Running git fetch --shallow-since="${MAX_HISTORY.oldestCommits}" --update-shallow --filter=blob:none --recurse-submodules=no`
```

src/helpers/git/get-git-data.ts
```diff
@@ -9,18 +9,28 @@ export const gitRemote = async (git: simpleGit.SimpleGit): Promise<string> => {
...
-  // Falling back to picking the first remote in the list if 'origin' is not found.
+  // Falling back to picking the first remote in the list if the default remote is not found.
   return stripCredentials(remotes[0].refs.push)
 }
 
+export const getDefaultRemoteName = async (git: simpleGit.SimpleGit): Promise<string> => {
+  try {
+    const defaultRemoteConfig = (await git.getConfig('clone.defaultRemoteName')) ?? {}
+
+    return defaultRemoteConfig.value ?? 'origin'
```
juan-fernandez, line 28: nitpick: this is up to preference too, but you can do this in a single line: 
```suggestion
    return (await git.getConfig('clone.defaultRemoteName'))?.value ?? 'origin'
```

juan-fernandez (approved):
only nitpicks, otherwise LGTM :-)

src/helpers/git/get-git-data.ts
edrevo: Nice, I didn't know about that syntax!

src/commands/git-metadata/__tests__/gitdb.test.ts
edrevo:
```suggestion
  // the tests pass. See https://github.com/steveukx/git-js/blob/d184c13273abca4b6572c260f9625c19f944d4f7/simple-git/src/lib/tasks/version.ts#L15-L39
```

Issue #658, opened 2022-10-06
Vulnerability regarding vm2 version 3.9.9


antonpqv:
Hi,

We are using your package `datadog-ci` version `1.15.0`. Dependabot has detected the vulnerability regarding `vm2` version `3.9.9` pasted below. 

I tracked the dependancies and can confirm that `vm2` is a descendant dependency of `datadog-ci`.

Would it be possible for you to bump the version of `vm2` that you are using to version `3.9.11` in. order to get rid of this vulnerability?

Best,
Anton

---

vm2 vulnerable to Sandbox Escape resulting in Remote Code Execution on host #98

Impact
A threat actor can bypass the sandbox protections to gain remote code execution rights on the host running the sandbox.

Patches
This vulnerability was patched in the release of version 3.9.11 of vm2

Workarounds
None.

References
Github Issue - https://github.com/patriksimek/vm2/issues/467
The file that was patched - https://github.com/patriksimek/vm2/blob/master/lib/setup-sandbox.js#L71
The commit with the patch - https://github.com/patriksimek/vm2/commit/d9a7f3cc995d3d861e1380eafb886cb3c5e2b873#diff-b1a515a627d820118e76d0e323fe2f0589ed50a1eacb490f6c3278fe3698f164

For more information
If you have any questions or comments about this advisory:

Open an issue in [VM2](https://github.com/patriksimek/vm2)

Severity:
Critical 10.0/10

CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:C/C:H/I:H/A:H

Weaknesses:
CWE-913

CVE ID:
CVE-2022-36067

GHSA ID:
GHSA-mrgp-mrhc-5jrq

---

m-rousse:
Hi @antonpqv,

The dependency has been bumped in PR https://github.com/DataDog/datadog-ci/pull/651.
It is also what I get when running `yarn why vm2`, which properly shows version 3.9.11:
```
yarn why v1.22.19
[1/4] Why do we have the module "vm2"...?
[2/4] Initialising dependency graph...
[3/4] Finding dependency...
[4/4] Calculating file sizes...
info 
=> Found "vm2@3.9.11"
info Reasons this module exists
   - "proxy-agent#pac-proxy-agent#pac-resolver#degenerator" depends on it
   - Hoisted from "proxy-agent#pac-proxy-agent#pac-resolver#degenerator#vm2"
info Disk size without dependencies: "268KB"
[...]

PR #957, opened 2023-07-04
[dep] Migrate to `proxy-agent@6.2.1`

Drarig29:
### What and why?

Closes https://github.com/DataDog/datadog-ci/issues/658
Closes https://github.com/DataDog/synthetics-ci-github-action/issues/119

This PR bumps `proxy-agent` to version 6, which removes our `ftp` dependency, and bumps `vm2`.

### How?

The main difference in the new version is that the constructor of `ProxyAgent` doesn't take a proxy URL as argument anymore. So we now have to use the `getProxyForUrl` callback.

#### Environment variables

However, setting this callback overrides the default `getProxyForUrl`, which [uses the `proxy-from-env` package](https://github.com/TooTallNate/proxy-agents/blob/08c4625d8b0c015265e27b260a17a23f55f486d1/packages/proxy-agent/src/index.ts#L99). So we have to update `utils.getProxyAgent()` to take this into account.

#### WebSocket

Also, `proxy-agent@5.0.0` [didn't support websocket](https://github.com/TooTallNate/node-proxy-agent/blob/3f4571097d22c9b9ac399d5b060c36b6e2caa993/index.js#L163-L174) at the time, but [it now does](https://github.com/TooTallNate/proxy-agents/blob/08c4625d8b0c015265e27b260a17a23f55f486d1/packages/proxy-agent/src/index.ts#L106-L117) and we don't need it in the codebase.

Moreover, part of the [Continuous Testing Tunnel](https://docs.datadoghq.com/continuous_testing/testing_tunnel/) uses websocket and that traffic should not be proxied. So we have to prevent any `wss?:` connections from being proxied in `utils.getProxyAgent()`.

I tested the tunnel feature (with `yarn launch synthetics run-tests --tunnel ...`) and it worked as expected.

### Review checklist

### DIFF
--- a/src/helpers/utils.ts
+++ b/src/helpers/utils.ts
@@ -209,18 +209,13 @@ export const getRequestBuilder = (options: RequestOptions) => {
 }
 
 export const getProxyAgent = (proxyOpts?: ProxyConfiguration): ProxyAgent => {
-  if (!proxyOpts) {
+  const proxyUrlFromConfiguration = getProxyUrl(proxyOpts)
+  if (!proxyOpts || proxyUrlFromConfiguration === '') {
+    // Let the default proxy agent discover environment variables.
     return new ProxyAgent()
   }
 
-  const {auth, host, port, protocol} = proxyOpts
-
-  return new ProxyAgent({
-    auth: auth ? `${auth.username}:${auth.password}` : undefined,
-    host,
-    port,
-    protocol: `${protocol}:`,
-  })
+  return new ProxyAgent({getProxyForUrl: () => proxyUrlFromConfiguration})
 }
 
 export const getApiHostForSite = (site: string) => {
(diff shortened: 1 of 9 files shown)


src/helpers/utils.ts
```diff
@@ -208,10 +208,23 @@ export const getRequestBuilder = (options: RequestOptions) => {
...
+export const getProxyAgent = (proxyOpts?: ProxyConfiguration): ProxyAgent => {
   const proxyUrlFromConfiguration = getProxyUrl(proxyOpts)
+  if (!proxyOpts || proxyUrlFromConfiguration === '') {
+    // Let the default proxy agent discover environment variables.
+    return new ProxyAgent()
+  }
+
+  return new ProxyAgent({
+    getProxyForUrl: (url) => {
+      // Do not proxy the WebSocket connections.
```
jakepruitt, line 220: Hmmmm why do we need this?

jakepruitt (approved):
Looks good!

src/helpers/utils.ts
jakepruitt: Nevermind, read the PR description

lefebvree (approved):
LGTM, indeed the tunnel did not go through the user defined proxy if that wasn't supported by proxy-agent. Better to disable it altogether until we can ensure it works properly :+1:

PR #960, opened 2023-07-05
[lambda][flare] Add optional start and end flags for getting CloudWatch logs

nhulston:
### What and why?

This PR adds optional `--start` and `--end` flags to specify a time range for CloudWatch logs.
Only log streams and log events that lie within the range will be saved.
Even though both are optional flags, if one is specified, the other is required.
Each timestamp is in milliseconds since UNIX Epoch.

### How?

1. The `--start` and `--end` validation is complicated, so I created a helper function `validateStartEndFlags()` that handles the validation.
2. The `getLogEvents()` function has been updated to set the `startTime` and `endTime` options in `GetLogEventsCommand`. The logic for this is abstracted and handled by the AWS SDK. If `--start` and `--end` were not specified, then these options won't be added to the `GetLogEventsCommand` config.
3. `DescribeLogStreamsCommand` does not have start or end time config options. Therefore, I had to manually filter the streams. The `getLogStreamNames()` function has been modified to accept `startMillis` and `endMillis` as parameters. We can get the first and last event times for a log stream using `logStream.lastEventTimestamp` and `logStream.firstEventTimestamp`. Then, we can filter the log streams to only include log streams whose range falls inside `[startMillis, endMillis]`

### Review checklist

### DIFF
--- a/src/commands/lambda/__tests__/flare.test.ts
+++ b/src/commands/lambda/__tests__/flare.test.ts
@@ -307,53 +307,38 @@ describe('lambda flare', () => {
   })
 
   describe('validateStartEndFlags', () => {
-    it('returns undefined when start and end flags are not specified', () => {
+    it('returns [0, 0] when start and end flags are not specified', () => {
       const errorMessages: string[] = []
-      const res = validateStartEndFlags(undefined, undefined, errorMessages)
-      expect(res).toBeUndefined()
+      const res = validateStartEndFlags(undefined, undefined)
+      expect(res).toEqual([0, 0])
       expect(errorMessages).toEqual([])
     })
 
-    it('returns undefined when start is specified but end is not specified', () => {
-      const errorMessages: string[] = []
-      const res = validateStartEndFlags('123', undefined, errorMessages)
-      expect(res).toBeUndefined()
-      expect(errorMessages).toMatchSnapshot()
+    it('throws error when start is specified but end is not specified', () => {
+      expect(() => validateStartEndFlags('123', undefined)).toThrowErrorMatchingSnapshot()
     })
 
-    it('returns undefined when end is specified but start is not specified', () => {
-      const errorMessages: string[] = []
-      const res = validateStartEndFlags(undefined, '123', errorMessages)
-      expect(res).toBeUndefined()
-      expect(errorMessages).toMatchSnapshot()
+    it('throws error when end is specified but start is not specified', () => {
+      expect(() => validateStartEndFlags(undefined, '123')).toThrowErrorMatchingSnapshot()
     })
 
-    it('returns undefined when start is invalid', () => {
-      const errorMessages: string[] = []
-      const res = validateStartEndFlags('123abc', '200', errorMessages)
-      expect(res).toBeUndefined()
-      expect(errorMessages).toMatchSnapshot()
+    it('throws error when start is invalid', () => {
+      expect(() => validateStartEndFlags('123abc', '200')).toThrowErrorMatchingSnapshot()
     })
 
-    it('returns undefined when end is invalid', () => {
-      const errorMessages: string[] = []
-      const res = validateStartEndFlags('100', '234abc', errorMessages)
-      expect(res).toBeUndefined()
-      expect(errorMessages).toMatchSnapshot()
+    it('throws error when end is invalid', () => {
+      expect(() => validateStartEndFlags('100', '234abc')).toThrowErrorMatchingSnapshot()
     })
 
-    it('returns undefined when start is not before the end time', () => {
-      const errorMessages: string[] = []
-      const res = validateStartEndFlags('200', '100', errorMessages)
-      expect(res).toBeUndefined()
-      expect(errorMessages).toMatchSnapshot()
+    it('throws error when start is not before the end time', () => {
+      expect(() => validateStartEndFlags('200', '100')).toThrowErrorMatchingSnapshot()
     })
 
     it('sets end time to current time if end time is too large', () => {
       const now = Date.now()
-      const res = validateStartEndFlags('0', '9999999999999', [])
+      const res = validateStartEndFlags('0', '9999999999999')
       expect(res).not.toBeUndefined()
-      const [start, end] = res!
+      const [start, end] = res
       expect(start).toBe(0)
       expect(end).toBeGreaterThanOrEqual(now - 1000)
       expect(end).toBeLessThanOrEqual(now + 1000)
--- a/src/commands/lambda/__tests__/flare.test.ts
+++ b/src/commands/lambda/__tests__/flare.test.ts
@@ -333,10 +333,10 @@ describe('lambda flare', () => {
   })
 
   describe('validateStartEndFlags', () => {
-    it('returns [0, 0] when start and end flags are not specified', () => {
+    it('returns [undefined, undefined] when start and end flags are not specified', () => {
       const errorMessages: string[] = []
       const res = validateStartEndFlags(undefined, undefined)
-      expect(res).toEqual([0, 0])
+      expect(res).toEqual([undefined, undefined])
       expect(errorMessages).toEqual([])
     })
 
@@ -468,7 +468,7 @@ describe('lambda flare', () => {
       mockCloudWatchLogStreams(cloudWatchLogsClientMock, mockStreams)
 
       const expectedLogStreams = ['Stream1', 'Stream2', 'Stream3']
-      const logStreams = await getLogStreamNames(new CloudWatchLogsClient({}), MOCK_LOG_GROUP, 0, 0)
+      const logStreams = await getLogStreamNames(new CloudWatchLogsClient({}), MOCK_LOG_GROUP, undefined, undefined)
 
       expect(logStreams).toEqual(expectedLogStreams)
     })
@@ -476,7 +476,7 @@ describe('lambda flare', () => {
     it('returns empty array when no log streams are found', async () => {
       mockCloudWatchLogStreams(cloudWatchLogsClientMock, [])
 
-      const logStreams = await getLogStreamNames(new CloudWatchLogsClient({}), MOCK_LOG_GROUP, 0, 0)
+      const logStreams = await getLogStreamNames(new CloudWatchLogsClient({}), MOCK_LOG_GROUP, undefined, undefined)
 
       expect(logStreams).toEqual([])
     })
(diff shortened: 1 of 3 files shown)


src/commands/lambda/flare.ts
```diff
@@ -30,7 +30,9 @@ const FLARE_OUTPUT_DIRECTORY = '.datadog-ci'
 const LOGS_DIRECTORY = 'logs'
 const FUNCTION_CONFIG_FILE_NAME = 'function_config.json'
 const ZIP_FILE_NAME = 'lambda-flare-output.zip'
-const LOG_STREAM_COUNT = 3
+const MAX_LOG_STREAMS = 50
+const DEFAULT_LOG_STREAMS = 3
+const MAX_LOG_EVENTS_PER_STREAM = 1000
```
nhulston, line 35: I chose the values for `MAX_LOG_STREAMS` and `MAX_LOG_EVENTS_PER_STREAM` somewhat arbitrarily.

I don't really see a scenario where support needs to see past the last 50 log streams, but let me know if there would be a better value for this const.

Assuming each log event is about 200 characters long, a log stream with 1,000 log events would be about 200KB.

Meaning there could a maximum file size of about 200KB per stream * 50 streams = 10MB, but the real size would likely be far far less.

If 10MB is too large, we can lower either `MAX_LOG_STREAMS` or `MAX_LOG_EVENTS_PER_STREAM`.

src/commands/lambda/flare.ts
```diff
@@ -365,9 +431,20 @@ export const getLogStreamNames = async (cwlClient: CloudWatchLogsClient, logGrou
...
+      continue
+    }
+    if (rangeSpecified) {
+      const lastEventTime = logStream.lastEventTimestamp
+      const firstEventTime = logStream.firstEventTimestamp
+      if (!lastEventTime || !firstEventTime) {
+        continue
+      }
+      if (lastEventTime < startMillis || firstEventTime > endMillis) {
+        continue
```
nhulston, line 444: `DescribeLogStreamsCommand` doesn't accept a start or end time, so i had to filter log streams manually

src/commands/lambda/flare.ts
```diff
@@ -379,14 +456,28 @@ export const getLogStreamNames = async (cwlClient: CloudWatchLogsClient, logGrou
...
+  const config: any = {
     logGroupName,
     logStreamName,
-  })
+    limit: MAX_LOG_EVENTS_PER_STREAM,
+  }
+  if (startMillis !== 0 || endMillis !== 0) {
+    config.startTime = startMillis
+    config.endTime = endMillis
+  }
```
nhulston, line 479: Unlike `DescribeLogStreamsCommand`, `GetLogEventsCommand` does accept optional `startTime` and `endTime` config values

src/commands/lambda/flare.ts
```diff
@@ -86,6 +90,17 @@ export class LambdaFlareCommand extends Command {
       errorMessages.push(commonRenderer.renderError('No email specified. [-e,--email]'))
     }
 
+    // Validate start/end flags if both are specified
+    let startMillis = 0
+    let endMillis = 0
+    try {
+      ;[startMillis, endMillis] = validateStartEndFlags(this.start, this.end)
```
nhulston, line 97: If no start/end flags are specified, I set them to 0.
Would it be more readable to leave `startMillis` and `endMillis` as undefined if the start/end flags were not specified?

src/commands/lambda/flare.ts
duncanista: Thanks for the in-depth explanation and calculations. Sounds about right.

src/commands/lambda/flare.ts
```diff
@@ -271,6 +286,42 @@ export class LambdaFlareCommand extends Command {
...
+export const validateStartEndFlags = (start: string | undefined, end: string | undefined) => {
+  if (!start && !end) {
+    return [0, 0]
+  }
+
+  if (!start) {
+    throw Error('Start time is required when end time is specified. [--start]')
+  } else if (!end) {
+    throw Error('End time is required when start time is specified. [--end]')
+  }
```
duncanista, line 305:
```suggestion
  if (!start) {
    throw Error('Start time is required when end time is specified. [--start]')
  }
  if (!end) {
    throw Error('End time is required when start time is specified. [--end]')
  }
```

src/commands/lambda/flare.ts
duncanista: Why `;` at the beginning of the array definition? Is it because you want to assign the value to a previously defined variable? Maybe a comment about this could help to understand why this is necessary.

src/commands/lambda/flare.ts
```diff
@@ -371,16 +422,28 @@ export const createDirectories = (
...
+) => {
+  const config = {
     logGroupName,
-    limit: LOG_STREAM_COUNT,
     descending: true,
     orderBy: OrderBy.LastEventTime,
-  })
+    limit: DEFAULT_LOG_STREAMS,
+  }
+  const rangeSpecified = startMillis !== 0 || endMillis !== 0
```
duncanista, line 442: Why `||`? Isn't it guaranteed that we only run this when both of them are set?

src/commands/lambda/flare.ts
nhulston: It was required by ESLint, no other reason

src/commands/lambda/flare.ts
```diff
@@ -390,9 +453,20 @@ export const getLogStreamNames = async (cwlClient: CloudWatchLogsClient, logGrou
...
-      output.push(logStreamName)
+    if (!logStreamName) {
+      continue
     }
+    if (rangeSpecified) {
+      const lastEventTime = logStream.lastEventTimestamp
+      const firstEventTime = logStream.firstEventTimestamp
+      if (!lastEventTime || !firstEventTime) {
+        continue
+      }
```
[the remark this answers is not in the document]
nhulston: Good point. According to the documentation, the timestamps can take up to an hour to update. So a very recently executed log could return undefined for these values. 
I think it makes more sense to add these streams when the values are undefined instead of continuing.
Updated the code to:
```
      if (lastEventTime && firstEventTime && (lastEventTime < startMillis || firstEventTime > endMillis)) {
        continue
      }
```

src/commands/lambda/flare.ts
nhulston: It's possible that the user intentionally set `startMillis` to 0.
For example, if they wanted to get all log streams, they might do: `--start 0 --end 9999999999999`.
In this case, `||` is required to detect that a range was specified.

src/commands/lambda/flare.ts
duncanista: I think we should change it to `undefined`, since it is more readable, now seeing how I came up with questions below around `rangeSpecified`, it makes sense to leave it to undefined. Although if user specifies 0, we'd have to also change that.

What happens if it's a negative number?

src/commands/lambda/flare.ts
```diff
@@ -271,6 +286,44 @@ export class LambdaFlareCommand extends Command {
...
 }
 
+/**
+ * Validate the start and end flags and adds error messages if found
+ * @param start start time as a string
+ * @param end end time as a string
+ * @throws error if start or end are not valid numbers
+ * @returns [startMillis, endMillis] as numbers
+ */
+export const validateStartEndFlags = (start: string | undefined, end: string | undefined) => {
```
duncanista, line 296: Do we cover edge caes for negative numbers?

src/commands/lambda/flare.ts
duncanista: Makes sense, this would've helped when this values are `undefined` as mentioned on a thread above.

duncanista (changes requested):
One last comment around redability.

src/commands/lambda/flare.ts
```diff
@@ -390,9 +455,17 @@ export const getLogStreamNames = async (cwlClient: CloudWatchLogsClient, logGrou
...
     const logStreamName = logStream.logStreamName
-    if (logStreamName) {
-      output.push(logStreamName)
+    if (!logStreamName) {
+      continue
+    }
+    if (rangeSpecified) {
+      const lastEventTime = logStream.lastEventTimestamp
+      const firstEventTime = logStream.firstEventTimestamp
+      if (lastEventTime && firstEventTime && (lastEventTime < startMillis || firstEventTime > endMillis)) {
```
duncanista, line 464: Is there a way to improve the readability of this condition?

src/commands/lambda/flare.ts
nhulston: The command doesn't run when trying to enter a negative number for start/end flags:

Screenshot 2023-07-11 at 7 58 52 PM

src/commands/lambda/flare.ts
nhulston: Changed to:
```
if (rangeSpecified) {
  const firstEventTime = logStream.firstEventTimestamp
  const lastEventTime = logStream.lastEventTimestamp
  if (lastEventTime && lastEventTime < startMillis) {
    continue
  }
  if (firstEventTime && firstEventTime > endMillis) {
    continue
  }
}
```

What do you think?

src/commands/lambda/flare.ts
```diff
@@ -383,16 +436,28 @@ export const createDirectories = (
...
+) => {
+  const config = {
     logGroupName,
-    limit: LOG_STREAM_COUNT,
     descending: true,
     orderBy: OrderBy.LastEventTime,
-  })
+    limit: DEFAULT_LOG_STREAMS,
+  }
+  const rangeSpecified = startMillis !== undefined && endMillis !== undefined
```
nhulston, line 456: This cannot be simplified to `!startMillis && !endMillis` because if `startMillis = 0`, then `rangeSpecified` will incorrectly have the value of `false`.

src/commands/lambda/flare.ts
```diff
@@ -99,6 +103,17 @@ export class LambdaFlareCommand extends Command {
       }
     }
 
+    // Validate start/end flags if both are specified
+    let startMillis
+    let endMillis
```
duncanista, line 108:
```suggestion
    let startMillis, endMillis
```

nhulston: ESLint doesn't let us do that in this repo

src/commands/lambda/__tests__/flare.test.ts
```diff
@@ -35,6 +35,7 @@ import {
   getMasking,
   getTags,
   maskConfig,
+  validateStartEndFlags,
```
duncanista, line 38: Amazing tests :)

duncanista (approved):
LGTM!

PR #970, opened 2023-07-11
[synthetics] Add support for `--pollingTimeout`

Drarig29:
- https://github.com/DataDog/datadog-ci/pull/970 ←
- https://github.com/DataDog/datadog-ci/pull/971
---

### What and why?

This PR adds a `--pollingTimeout` CLI argument.

This will allow us to add a `polling_timeout` input in our [CircleCI orb](https://github.com/DataDog/synthetics-test-automation-circleci-orb).

### How?

- Add a new argument
- Add a utility function to parse an optional integer
  - We can later reuse it in the github action ([here](https://github.com/DataDog/synthetics-ci-github-action/blob/84ab83ecd7f52c6daec005ced8883a461cef51c3/src/resolve-config.ts#L106-L111)) and azure devops extension ([here](https://github.com/DataDog/datadog-ci-azure-devops/blob/main/SyntheticsRunTestsTask/utils.ts#L12-L17))
- Throw an `INVALID_CONFIG` error if the validation fails
- Satisfy this order of precedence:
  - `--pollingTimeout` >> `config.global.pollingTimeout` >> `this.config.pollingTimeout` ([deprecated](https://github.com/DataDog/datadog-ci/pull/971))

### Review checklist

- [x] Feature or bugfix MUST have appropriate tests (unit, integration)

PR #971, opened 2023-07-11
[synthetics] Deprecate `pollingTimeout` at root of config file

Drarig29:
- https://github.com/DataDog/datadog-ci/pull/970 
- https://github.com/DataDog/datadog-ci/pull/971 ←
---

### What and why?

The `pollingTimeout` can currently be set in two places inside a `datadog-ci.json` file:

```json
{
  "apiKey": "api-key",
  "appKey": "app-key",
  "pollingTimeout": 300000
  "global": {
    "pollingTimeout": 180000
  },
}
```

This is the result of making it possible to override the `pollingTimeout` in test files (we take the maximum in all the `pollingTimeout` values to poll for the batch).

Because the overrides in `global` and in test files are the same, we ended up having `pollingTimeout` both at the root of the config file and in `global`. However, only the latter makes sense.

### How?

- Undocument the option in the README
- Add a deprecation log

### Review checklist

### DIFF
--- a/src/commands/synthetics/README.md
+++ b/src/commands/synthetics/README.md
@@ -77,17 +77,13 @@ See below for the list of advanced options in the global configuration file. For
 : A boolean flag that fails the CI job if at least one test exceeds the default test timeout. The default is set to `true`.
 
 `files`
-: Glob patterns to detect Synthetic test [configuration files](#test-files).
+: Glob patterns to detect Synthetic [test files](#test-files).
 
 `global`
-: Overrides for Synthetic tests applied to all tests.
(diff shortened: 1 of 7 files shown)


jakepruitt:
> This is the result of making it possible to override the pollingTimeout in test files (we take the maximum in all the pollingTimeout values to poll for the batch).

The fact that it's still a batch-wide setting, and not a test-specific setting, leads me to want to deprecate the `global` setting and the test-based override, which could lead to confusion.

Drarig29:
Closed the pull request because we have a better idea for the future of the `pollingTimeout` option.

PR #972, opened 2023-07-11
Add `mergeStepFunctionAndLambdaTraces` option to `stepfunctions` command

kimi-p:
### What and why?
We are asking customers to manually change their step functions definition to inject Step Functions Context Object into Lambda payload. To help customers do this automatically, we add `mergeStepFunctionAndLambdaTraces` flag for customers to opt (default to false) in merging their step functions trace with lambda traces.

Also fixed the issue where the same step functions arn are printed while looping all steps.

### How?
For each step functions' steps, we find all default lambda api steps and check if their lambda step payload has special logic. If not, we change their `Payload.$` to `'States.JsonMerge($$, $, false)'` to inject step functions context (which contains the info that is to be hashed by lambda Python/NodeJS layer) into payload. The scenario that we are updating their payload are:
- without `Payload.$` field
- with `Payload.$` but the value is `$`, which means the original lambda payload.

### Test
Tested in sandbox account. Traces showed up merging as expected.
```bash
aws-vault exec serverless-sandbox-account-admin -- node ./dist/cli.js stepfunctions instrument \
--forwarder arn:aws:lambda:sa-east-1:425362996713:function:step-functions-tracing-forwarder-self-monitoring-us1-prod \
--step-function arn:aws:states:sa-east-1:425362996713:stateMachine:kimi-testing-state-machine \
--env kimi-dev \
--service kimi-dev-service \
--merge-lambda-traces

```

image

image

### Review checklist

### DIFF
--- a/src/commands/stepfunctions/helpers.ts
+++ b/src/commands/stepfunctions/helpers.ts
@@ -79,3 +79,65 @@ export const parseArn = (
 export const buildLogAccessPolicyName = (stepFunction: DescribeStateMachineCommandOutput): string => {
   return `LogsDeliveryAccessPolicy-${stepFunction.name}`
 }
+
+export const updateStateMachineDefinition = (
+  describeStateMachineCommandOutput: DescribeStateMachineCommandOutput,
+  context: BaseContext
+): void => {
+  if (typeof describeStateMachineCommandOutput.definition !== "string") {
+    return
+  }
+  const definitionObj = JSON.parse(describeStateMachineCommandOutput.definition) as StateMachineDefinitionType
+  for (const stepName in definitionObj.States) {
+    if (definitionObj.States.hasOwnProperty(stepName)) {
+      const step = definitionObj.States[stepName]
+      if (shouldUpdateStepForTracesMerging(step)) {
+        updatePayloadInStateMachineDefinition()
+      }
+    }
+  }
+
+
+}
+
+export const updatePayloadInStateMachineDefinition() => {
+  return
+}
+
+export const shouldUpdateStepForTracesMerging = (step: StepType): boolean => {
+  // is default lambda api
+  if (step.Resource === 'arn:aws:states:::lambda:invoke') {
+    if (step.Parameters === undefined) {
+      return false
+    }
+    // payload field not set
+    if (!step.Parameters.hasOwnProperty('Payload.$')) {
+      return true
+    }
+    // default payload
+    if (step.Parameters['Payload.$'] === '$') {
+      return true
+    }
+  }
+
+  return false
+}
+
+type StateMachineDefinitionType = {
+  Comment?: string
+  StartAt?: string
+  States?: StatesType
+}
+
+type StatesType = Record<string, StepType>
+type StepType = {
+  Type: string
+  Parameters?: ParametersType
+  Resource: string
+  Next?: string
+  End?: string
+}
+
+type ParametersType = {
+  'Payload.$'?: string
+}
--- a/src/commands/stepfunctions/helpers.ts
+++ b/src/commands/stepfunctions/helpers.ts
@@ -1,7 +1,9 @@
 import {DescribeStateMachineCommandOutput} from '@aws-sdk/client-sfn'
+import {SFNClient} from '@aws-sdk/client-sfn/dist-types/SFNClient'
 import {BaseContext} from 'clipanion'
 import {diff} from 'deep-object-diff'
 
+import {updateStateMachineDefinition} from './awsCommands'
 import {DD_CI_IDENTIFYING_STRING} from './constants'
 
 export const displayChanges = (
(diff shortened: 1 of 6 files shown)


src/commands/stepfunctions/instrument.ts
```diff
@@ -88,6 +90,9 @@ export class InstrumentStepFunctionsCommand extends Command {
 
     // loop over step functions passed as parameters and generate a list of requests to make to AWS for each step function
     for (const stepFunctionArn of stepFunctionArns) {
+      this.context.stdout.write(
```
duncanista, line 93: Why are we printing this now?

src/commands/stepfunctions/instrument.ts
```diff
@@ -328,6 +333,17 @@ export class InstrumentStepFunctionsCommand extends Command {
...
+
+      if (this.mergeStepFunctionAndLambdaTraces) {
+        // Not putting the update operation into the business logic of logs subscription. This will
+        // add additional API call, but it would also allow easier testing and cleaner code.
+        await injectContextIntoLambdaPayload(
+          describeStateMachineCommandOutput,
+          stepFunctionsClient,
+          this.context,
+          this.dryRun
+        )
```
duncanista, line 345: This needs to be in a try-catch block in case of failure.

src/commands/stepfunctions/instrument.ts
```diff
@@ -342,5 +358,9 @@ InstrumentStepFunctionsCommand.addPath('stepfunctions', 'instrument')
 InstrumentStepFunctionsCommand.addOption('dryRun', Command.Boolean('-d,--dry-run'))
 InstrumentStepFunctionsCommand.addOption('environment', Command.String('-e,--env'))
 InstrumentStepFunctionsCommand.addOption('forwarderArn', Command.String('--forwarder'))
+InstrumentStepFunctionsCommand.addOption(
+  'mergeStepFunctionAndLambdaTraces',
+  Command.Boolean('-mlt,--merge-lambda-traces,--merge-step-function-and-lambda-traces')
```
duncanista, line 363: nice

src/commands/stepfunctions/helpers.ts
```diff
@@ -12,9 +14,7 @@ export const displayChanges = (
   params: any,
   previousParams?: any
 ): void => {
-  context.stdout.write(`${'='.repeat(50)}`)
-  context.stdout.write(`\n${dryRun ? '\n[Dry Run] Planning for' : 'Will apply'} the following change:\n`)
-  context.stdout.write(`\nChanges for ${stepFunctionArn}\n`)
```
duncanista, line 17: Is this why we are changing the print?

src/commands/stepfunctions/instrument.ts
duncanista: Resolving since `awsCommands.ts` has the error handling and printing.

duncanista (approved):
Left a question, but so far, LGTM!

src/commands/stepfunctions/helpers.ts
kimi-p: This section is too verbose and updating it would resolve the issue you raised in the original PR where step function ARN are showing too many times.

src/commands/stepfunctions/instrument.ts
kimi-p: For the same reason as https://github.com/DataDog/datadog-ci/pull/972/files#r1265488424

src/commands/stepfunctions/helpers.ts
duncanista: Oh, perfect!

kimi-p:
> Left a question, but so far, LGTM!

@duncanista  I cannot find the question. Can you check if they are sent? Thanks!

src/commands/stepfunctions/helpers.ts
```diff
@@ -79,3 +79,80 @@ export const parseArn = (
...
+export const updateStepObject = ({Parameters}: StepType): void => {
+  if (Parameters) {
+    Parameters[`Payload.$`] = 'States.JsonMerge($$, $, false)'
+  }
+}
+
+export const shouldUpdateStepForTracesMerging = (step: StepType): boolean => {
+  // is default lambda api
+  if (step.Resource === 'arn:aws:states:::lambda:invoke') {
+    if (step.Parameters === undefined) {
```
nine5two7, line 123: nit,  if (step.Parameters) is expected to be better. Check [this](https://stackoverflow.com/questions/28975896/is-there-a-way-to-check-for-both-null-and-undefined#:~:text=450-,if(%20value%20)%20%7B%0A%7D,-will%20evaluate%20to)

kimi-p: Thanks for the comment. Updated.

PR #977, opened 2023-07-12
[lambda][flare] Generate insights file

nhulston:
### What and why?

This PR creates a file called `INSIGHTS.md` that includes additional information that can be helpful to support staff.
- Directory that the command was ran in
- CLI version
- Timestamp
- Framework (Serverless Framework, AWS CDK, CloudFormation, or unknown)
- Runtime
- Layers
- Package size

Screenshot 2023-07-26 at 11 28 02 AM

### How?

1. Create the `getFramework()` function which scans files in the current directory to determine what framework is being used. For example, if `serverless.yml` exists, then we can conclude that the Serverless Framework is being used.
2. Create the `generateInsightsFile()` that calculates any unknown values and writes it to a file using the existing `writeFile()` method. The funciton uses markdown to generate an easy-to-read file.

### Review checklist


src/commands/lambda/flare.ts
```diff
@@ -771,6 +784,75 @@ export const sleep = async (ms: number) => {
...
+ * @returns the framework used or undefined if no framework is found
+ */
+export const getFramework = () => {
+  const files = new Set(fs.readdirSync(process.cwd()))
+
+  // Check for serverless framework
+  const serverlessFiles = ['serverless.yml', 'serverless.yaml']
+  for (const file of serverlessFiles) {
+    if (files.has(file)) {
+      return 'Serverless Framework'
```
duncanista, line 798: Turn into `Enum`

src/commands/lambda/flare.ts
```diff
@@ -771,6 +784,75 @@ export const sleep = async (ms: number) => {
...
 
+/**
+ * Get the framework used based on the files in the directory
+ * @returns the framework used or undefined if no framework is found
+ */
+export const getFramework = () => {
+  const files = new Set(fs.readdirSync(process.cwd()))
+
+  // Check for serverless framework
+  const serverlessFiles = ['serverless.yml', 'serverless.yaml']
```
duncanista, line 795: nit: repeating constants, we need to find a way to get a subset of these from the `PROJECT_FILES`

src/commands/lambda/flare.ts
```diff
@@ -771,6 +784,75 @@ export const sleep = async (ms: number) => {
...
+  return 'Unknown'
+}
+
+/**
+ * Generate the insights file
+ * @param insightsFilePath path to the insights file
+ * @param isDryRun whether or not this is a dry run
+ * @param config Lambda function configuration
+ */
+export const generateInsightsFile = (insightsFilePath: string, isDryRun: boolean, config: FunctionConfiguration) => {
```
duncanista, line 824: Missing:
1. FunctionName,
2. ARN,
3. Handler – or equivalent if Image
4. Environment Variables,
5. Timeout,
6. Memory Size,
7. Architecture.

This file does not contain any separation between what's CLI run data and what's customer's AWS Lambda Data.

I'd suggest creating a markdown template, the current approach is just a list in a markdown file.

src/commands/lambda/flare.ts
```diff
@@ -771,6 +784,75 @@ export const sleep = async (ms: number) => {
...
+  const lines: string[] = []
+  lines.push('_Autogenerated file from `lambda flare`_')
+  if (isDryRun) {
+    lines.push('</br>_This command was run in dry mode._')
```
duncanista, line 828: What's the difference between `</br>` and `\n`?

src/helpers/utils.ts
```diff
@@ -352,3 +352,26 @@ export const timedExecAsync = async <I, O>(f: (input: I) => Promise<O>, input: I
...
+  if (bytes < 0) {
+    throw Error("'bytes' can't be negative.")
+  }
+
+  const k = 1024
+  const dm = decimals < 0 ? 0 : decimals
+  const sizes = ['Bytes', 'KB', 'MB', 'GB', 'TB', 'PB', 'EB', 'ZB', 'YB']
+  const i = Math.floor(Math.log(bytes) / Math.log(k))
+
+  return `${parseFloat((bytes / Math.pow(k, i)).toFixed(dm))} ${sizes[i]}`
```
duncanista, line 376: Not sure it's very readable, maybe we could improve this method?

For example
```suggestion
  const k = 1024
  const dm = decimals < 0 ? 0 : decimals
  const units = ['Bytes', 'KB', 'MB', 'GB', 'TB', 'PB', 'EB', 'ZB', 'YB']
  const i = Math.floor(Math.log(bytes) / Math.log(k))
  const formattedBytes = parseFloat((bytes / Math.pow(k, i)).toFixed(dm))

  return `${formattedBytes} ${units[i]}`
```

src/commands/lambda/flare.ts
nhulston: Good question! In markdown, `</br>` creates a small line break. Using `\n` creates a bigger gap between the two lines.

With `</br>`:
Screenshot 2023-07-25 at 2 55 11 PM

With new lines:
Screenshot 2023-07-25 at 2 55 29 PM

duncanista (commented):
Left some comments!

src/commands/lambda/flare.ts
nhulston: Thanks, I added the missing fields and separated the CLI data vs Lambda config data

src/commands/lambda/flare.ts
nhulston: Is this what you were looking for?
```
enum Frameworks {
  Serverless_Framework = 'Serverless Framework',
  AWS_CDK = 'AWS CDK',
  AWS_CloudFormation = 'AWS CloudFormation',
  Unknown = 'Unknown',
}
```

And then: `return Frameworks.Serverless_Framework` instead of returning a string literal.

src/helpers/utils.ts
nhulston: I've adjusted the function a little bit for more readability, LMK what you think

src/commands/lambda/flare.ts
duncanista: Camel case for Enums: https://www.typescriptlang.org/docs/handbook/enums.html

src/commands/lambda/flare.ts
nhulston: Fixed, accomplished using the array spread operator:
```
export const FRAMEWORK_MAPPINGS = new Map([
  ['serverless.yaml', Frameworks.ServerlessFramework],
  ['serverless.yml', Frameworks.ServerlessFramework],
  ...
])

// Project files to search for in Flare
export const PROJECT_FILES = [
  ...FRAMEWORK_MAPPINGS.keys(),
  // Datadog CloudFormation Template
  'datadog-cloudfomation-macro.yaml',
  ...
]
```

src/commands/lambda/constants.ts
```diff
@@ -135,18 +135,30 @@ export const SKIP_MASKING_ENV_VARS = new Set([
...
+export const FRAMEWORK_MAPPINGS = new Map([
+  ['serverless.yaml', Frameworks.ServerlessFramework],
+  ['serverless.yml', Frameworks.ServerlessFramework],
+  ['serverless.js', Frameworks.ServerlessFramework],
+  ['cdk.json', Frameworks.AwsCdk],
+  ['.cdk.json', Frameworks.AwsCdk],
+  ['template.yaml', Frameworks.AwsCloudFormation],
+  ['template.yml', Frameworks.AwsCloudFormation],
+  ['template.json', Frameworks.AwsCloudFormation],
+])
```
duncanista, line 156: Why not reverse it?

export const FRAMEWORK_FILES_LOOKUP = {
  [DeploymentFrameworks.ServerlessFramework]: ['serverless.yaml', 'serverless.js', 'serverless.yml']
} as const

src/commands/lambda/constants.ts
```diff
@@ -135,18 +135,30 @@ export const SKIP_MASKING_ENV_VARS = new Set([
   VERSION_ENV_VAR,
 ])
 
+export enum Frameworks {
```
duncanista, line 138:
```suggestion
export enum DeploymentFrameworks {
```

I think `Framework` is not very specific

src/commands/lambda/flare.ts
duncanista: Left a comment

src/commands/lambda/constants.ts
nhulston: For `O(1)` lookup, otherwise we have to loop through all the values to find out what the framework is.

duncanista: Sounds good, let's change the constant name to `FRAMEWORK_FILES_MAPPING`!

src/commands/lambda/constants.ts
```diff
@@ -135,18 +135,30 @@ export const SKIP_MASKING_ENV_VARS = new Set([
...
+export enum DeploymentFrameworks {
+  ServerlessFramework = 'Serverless Framework',
+  AwsCdk = 'AWS CDK',
+  AwsCloudFormation = 'AWS CloudFormation',
+  Unknown = 'Unknown',
+}
+
+// Mappings of files to frameworks.
+// For example, if `serverless.yml` exists, we know it's the Serverless Framework
+export const FRAMEWORK_FILES_MAPPINGS = new Map([
```
duncanista, line 147: nit
```suggestion
export const FRAMEWORK_FILES_MAPPING = new Map([
```

duncanista:
Could you post an updated screenshot?

duncanista (approved):

PR #978, opened 2023-07-12
[lambda][flare] Change config obfuscation to use functions from `commons.ts`

nhulston:
### What and why?

Config obfuscation was using a function in `flare.ts`. However, a function that does the same thing already exists in `commons.ts`.
This PR modifies the `maskConfig()` function to use helper functions from `commons.ts`. It also removes any unused functions and tests.

### How?

Uses `maskStringifiedEnvVar()` in `commons.ts` to obfuscate the config in the form of a string. We then convert the obfuscated string back into type `FunctionConfig` using `JSON.parse()`.

### Review checklist

Existing tests for `maskConfig()` still pass as expected. Unused tests for the now deleted `getMasking()` function have been removed.

### DIFF
--- a/src/commands/lambda/__tests__/flare.test.ts
+++ b/src/commands/lambda/__tests__/flare.test.ts
@@ -32,7 +32,6 @@ import {
   getEndpointUrl,
   getLogEvents,
   getLogStreamNames,
-  getMasking,
   getTags,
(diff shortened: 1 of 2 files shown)


duncanista (approved):
LGTM!

PR #987, opened 2023-07-14
[junit] Use uuidV4 for filename

rodrigo-roca:
### What and why?

If we send the normal filename it can contain some characters that would cause 400 errors at the intake. Now we create a [uuidv4](https://en.wikipedia.org/wiki/Universally_unique_identifier#Version_4_(random)) for the filename and send the original filename somewhere else.

### How?

Use uuid package and send filename somewhere else as `_dd.report_name`

### Review checklist

### DIFF
--- a/src/commands/junit/api.ts
+++ b/src/commands/junit/api.ts
@@ -6,7 +6,7 @@ import type {AxiosPromise, AxiosRequestConfig, AxiosResponse} from 'axios'
 
 import FormData from 'form-data'
 
-import {replaceForwardSlashes} from '../../helpers/file'
+import {getSafeFilename} from '../../helpers/file'
 import {getRequestBuilder} from '../../helpers/utils'
 
 import {Payload} from './interfaces'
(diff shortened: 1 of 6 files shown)


src/commands/junit/api.ts
```diff
@@ -56,7 +57,7 @@ export const uploadJUnitXML = (request: (args: AxiosRequestConfig) => AxiosPromi
   form.append('event', JSON.stringify(custom), {filename: 'event.json'})
 
   form.append('junit_xml_report_file', fs.createReadStream(jUnitXML.xmlPath).pipe(createGzip()), {
-    filename: `${replaceForwardSlashes(fileName)}.xml.gz`,
+    filename: `${uuidv4()}.xml.gz`,
```
AdrianLC, line 60: We want to apply this also to the filename in sarif, don't we?

rodrigo-roca: Is it the same as junit that EvP handles the uniqueness of the upload? I am not very knowledgeable about that part

AdrianLC: yes the upload to s3 is common

AdrianLC (approved):
LGTM but let's wait until the backend change is deployed

juan-fernandez (approved):
LGTM, but we need to add the new dependencies to https://github.com/DataDog/datadog-ci/blob/master/LICENSE-3rdparty.csv.

Also, let's wait for the approval from someone else from datadog-ci team, since this is adding a new dependency and it affects other commands

duncanista (approved):
LGTM – on the dependencies side!

PR #991, opened 2023-07-17
Set static analysis team as code owners of `sarif`

juan-fernandez:
### What and why?

Change owners of `sarif` command from `ci-app-libraries` to `static-analysis-core`. The reason being, they're the actual owners of the command and should review the PRs affecting it.

### How?

Change row in `CODEOWNERS` file.


duncanista (approved):
LGTM!

PR #997, opened 2023-07-19
[quality-gates] Implement dry run

calvarjorge:
### What and why?

We add a dry run feature. When executing a dry run evaluation, this will be sent in the request to the quality-gates service.

The dry run request should return the rules that would have been evaluated if the request was not dry run.

These rules will be printed in the console.

JIRA: https://datadoghq.atlassian.net/browse/CIAPP-6582

### How?

### Review checklist

### DIFF
--- a/src/commands/gate/renderer.ts
+++ b/src/commands/gate/renderer.ts
@@ -18,7 +18,11 @@ export const renderEvaluationResponse = (evaluationResponse: EvaluationResponse)
   }
 
   let fullStr = ''
-  fullStr += chalk.green('Successfully evaluated all matching rules.\n')
+  fullStr += chalk.green(
+    evaluationResponse.status.toLowerCase() === 'dry_run'
+      ? 'Successfully completed a dry run request\n'
+      : 'Successfully evaluated all matching rules.\n'
+  )
   fullStr += `Overall result: ${renderStatus(evaluationResponse.status)}\n`
   fullStr += `Number of rules evaluated: ${chalk.bold(evaluationResponse.rule_evaluations.length)}\n`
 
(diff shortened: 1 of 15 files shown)


src/commands/gate/renderer.ts
```diff
@@ -18,7 +18,11 @@ export const renderEvaluationResponse = (evaluationResponse: EvaluationResponse)
   }
 
   let fullStr = ''
-  fullStr += chalk.green('Successfully evaluated all matching rules.\n')
+  fullStr += chalk.green(
```
andrea-mosk, line 21: Rather than doing this, I would do for dry-run the same thing we are doing for empty, i.e. catching it in a if statement above and render something specific in that case. That allows us to modify thing in "normal" states without thinking on whether the dry-run case supports is or not. So something like:

```
if (evaluationResponse.status.toLowerCase() === 'empty') {
    return renderDryRunEvaluation(evaluationResponse)
  }
```

With the `renderDryRunEvaluation` method printing some specific dry-run information

calvarjorge: I have created a separate rendering function for dry run responses. There is some repeated code, but it is now easier to make changes in rendering differentiating  by dry run / no dry run.

juan-fernandez (approved):
LGTM. Two minor comments:

1. Just for your knowledge: other CI Vis commands do not fire a request at all when doing dry runs, so this'd be the first command doing it. I wonder if customers will be confused, but I realise the dry run wouldn't be very valuable without the request.
2. Are there tests we want to add for this?

calvarjorge:
> LGTM. Two minor comments:
> 
> 1. Just for your knowledge: other CI Vis commands do not fire a request at all when doing dry runs, so this'd be the first command doing it. I wonder if customers will be confused, but I realise the dry run wouldn't be very valuable without the request.
> 2. Are there tests we want to add for this?

Thanks for the review.

1. In this case, I think that, when making a dry run request, we would like to show the user the rules that would be evaluated in case the request was not dry run.
2. I have added a unit test just in case.

PR #998, opened 2023-07-19
Move base URL computation to common package and fix CI Visibility URLs

alxch-:
### What and why?

We have two issues with the URLs that are displayed once a JunitXML upload is finished:
- the App URL is not correct for all datacenters except US1
- the App URL does not take into account subdomains.

### How?

This PR:
- adds a `getCommonAppBaseURL` helper, copied from Synthetics code, that could be common to all commands.
- this fixes the URLs for all datacenters for the JunitXML upload as well as supporting subdomains.

### Review checklist

### DIFF
--- a/src/commands/junit/utils.ts
+++ b/src/commands/junit/utils.ts
@@ -2,20 +2,12 @@ import {lstatSync} from 'fs'
 
 import {SpanTags} from '../../helpers/interfaces'
 import {CI_JOB_URL, CI_PIPELINE_URL, GIT_BRANCH, GIT_REPOSITORY_URL, GIT_SHA} from '../../helpers/tags'
+import {getCommonAppBaseURL} from "../../helpers/app";
 
 export const getBaseUrl = () => {
-  if (process.env.DATADOG_SITE || process.env.DD_SITE) {
-    let site = process.env.DATADOG_SITE || process.env.DD_SITE
-    if (site === 'datadoghq.com') {
-      site = 'app.datadoghq.com'
-    } else if (site === 'datad0g.com') {
-      site = 'dd.datad0g.com'
-    }
-
-    return `https://${site}`
-  }
-
-  return 'https://app.datadoghq.com'
+  const site = process.env.DATADOG_SITE || process.env.DD_SITE || 'datadoghq.com';
+  const subdomain = process.env.DATADOG_SUBDOMAIN || '';
+  return getCommonAppBaseURL(site, subdomain);
 }
 
 export const getTestRunsUrl = (spanTags: SpanTags): string => {
--- a/src/helpers/__tests__/app.test.ts
+++ b/src/helpers/__tests__/app.test.ts
@@ -1,0 +1,33 @@
+import {getCommonAppBaseURL} from "../app";
+
+describe('getCommonAppBaseUrl', () => {
+  test('the base URL that is correct', () => {
+    // Usual datadog site.
+    expect(getCommonAppBaseURL('datadoghq.com', '')).toBe('https://app.datadoghq.com/')
+    expect(getCommonAppBaseURL('datadoghq.com', 'app')).toBe('https://app.datadoghq.com/')
+    expect(getCommonAppBaseURL('datadoghq.com', 'myorg')).toBe('https://myorg.datadoghq.com/')
+
+    // Different top-level domain.
+    expect(getCommonAppBaseURL('datadoghq.eu', '')).toBe('https://app.datadoghq.eu/')
+    expect(getCommonAppBaseURL('datadoghq.eu', 'app')).toBe('https://app.datadoghq.eu/')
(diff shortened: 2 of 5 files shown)


src/commands/junit/utils.ts
```diff
@@ -1,21 +1,14 @@
...
 import {CI_JOB_URL, CI_PIPELINE_URL, GIT_BRANCH, GIT_REPOSITORY_URL, GIT_SHA} from '../../helpers/tags'
 
 export const getBaseUrl = () => {
-  if (process.env.DATADOG_SITE || process.env.DD_SITE) {
-    let site = process.env.DATADOG_SITE || process.env.DD_SITE
-    if (site === 'datadoghq.com') {
-      site = 'app.datadoghq.com'
-    } else if (site === 'datad0g.com') {
-      site = 'dd.datad0g.com'
-    }
```
juan-fernandez, line 13: don't we need this staging logic?

src/helpers/__tests__/app.test.ts
```diff
@@ -0,0 +1,27 @@
...
+    // The correct Main DC (US3 in this case) is resolved automatically.
+    expect(getCommonAppBaseURL('ap1.datadoghq.com', '')).toBe('https://ap1.datadoghq.com/')
+    expect(getCommonAppBaseURL('ap1.datadoghq.com', 'app')).toBe('https://ap1.datadoghq.com/')
+    expect(getCommonAppBaseURL('ap1.datadoghq.com', 'myorg')).toBe('https://myorg.datadoghq.com/')
+    expect(getCommonAppBaseURL('us3.datadoghq.com', '')).toBe('https://us3.datadoghq.com/')
+    expect(getCommonAppBaseURL('us3.datadoghq.com', 'app')).toBe('https://us3.datadoghq.com/')
+    expect(getCommonAppBaseURL('us3.datadoghq.com', 'myorg')).toBe('https://myorg.datadoghq.com/')
+    expect(getCommonAppBaseURL('us5.datadoghq.com', '')).toBe('https://us5.datadoghq.com/')
+    expect(getCommonAppBaseURL('us5.datadoghq.com', 'app')).toBe('https://us5.datadoghq.com/')
+    expect(getCommonAppBaseURL('us5.datadoghq.com', 'myorg')).toBe('https://myorg.datadoghq.com/')
```
juan-fernandez, line 25: would we want to add tests for staging?

juan-fernandez (approved):
LGTM. Only a minor comment about datad0g.com

src/commands/junit/utils.ts
alxch-: no, it should be covered by the common function already I think

src/commands/junit/README.md
```diff
@@ -51,6 +51,7 @@ Additionally you might configure the `junit` command with environment variables:
   - The resulting dictionary will be merged with whatever is in the `--metrics` parameter. If a `key` appears both in `--metrics` and `DD_METRICS`, whatever value is in `DD_METRICS` will take precedence.
 - `DD_SITE`: choose your Datadog site, e.g. datadoghq.com or datadoghq.eu.
 - `DD_CIVISIBILITY_LOGS_ENABLED`: it will enable collecting logs from the content in the XML reports.
+- `DD_SUBDOMAIN`: if you have a [custom sub-domain enabled](https://docs.datadoghq.com/account_management/multi_organization/#custom-sub-domains) for your organization, this value should be set with the subdomain so that the link  to the Datadog Application that the library logs once the upload finishes is accurate.
```
alai97, line 54: I hope this makes sense, and isn't too wordy. Feel free to rephrase!

```suggestion
- `DD_SUBDOMAIN`: if you have a [custom subdomain enabled](https://docs.datadoghq.com/account_management/multi_organization/#custom-sub-domains) for your organization, set this value with the subdomain so that the link to the Datadog site that the library logs once the upload finishes is accurate.
```

alai97 (approved):
Looks good, one suggestion!

src/commands/junit/README.md
alxch-: I'm a bit concerned about using "site" given that this is the name of another variable here that means something else, that's why I used "application". Should I use "website"?

src/helpers/__tests__/app.test.ts
juan-fernandez: I see a change for that has been pushed but my comment was not considered stale.

src/helpers/app.ts
```diff
@@ -0,0 +1,16 @@
+const DEFAULT_SUBDOMAIN = 'app'
```
Drarig29, line 1: Can you export this constant too, and maybe rename it to `DEFAULT_DATADOG_SUBDOMAIN`?

Drarig29 (approved):
LGTM, except one nit

PR #1000, opened 2023-07-21
Add DD_TRACE_ENABLED for stepfunctions instrument and warning for legacy lambda integration

kimi-p:
https://datadoghq.atlassian.net/browse/SLS-3822

- To help customers enable step functions tracing when they run instrument command, we add `DD_TRACE_ENABLED=true` tag automatically. 
- If customer enables `mergeStepFunctionAndLambdaTraces` flag but they are using legacy lambda integration, we show warning with instructions to them so that they can know their legacy lambda steps' spans cannot merge with step functions span.

Tested in sandbox
```

======= For arn:aws:states:sa-east-1:425362996713:stateMachine:Kimi-Testing-State-machine =========

Will apply the following change:

TagResource:
{
  "resourceArn": "arn:aws:states:sa-east-1:425362996713:stateMachine:Kimi-Testing-State-machine",
  "tags": [
    {
      "key": "DD_TRACE_ENABLED",
      "value": "true"
    }
  ]
}
TagResource finished successfully!


Will apply the following change:

PutSubscriptionFilter:
{
  "destinationArn": "arn:aws:lambda:sa-east-1:425362996713:function:step-functions-tracing-forwarder-self-monitoring-us1-prod",
  "filterName": "Kimi-Testing-State-machine-DdCiLogGroupSubscription",
  "filterPattern": "",
  "logGroupName": "/aws/vendedlogs/states/Kimi-Testing-State-machine-Logs-dev"
}
Subscription filter Kimi-Testing-State-machine-DdCiLogGroupSubscription is created or the original filter Kimi-Testing-State-machine-DdCiLogGroupSubscription is overwritten.

[Warn] Step InvokeLambda2 may be using the basic legacy integration, which does not support merging lambda trace(s) with Step Functions trace. 
          To merge lambda trace(s) with Step Functions trace, please consider using the latest integration. 
          More details can be found on https://docs.aws.amazon.com/step-functions/latest/dg/connect-lambda.html 

```
image

### Review checklist

### DIFF
--- a/src/commands/stepfunctions/helpers.ts
+++ b/src/commands/stepfunctions/helpers.ts
@@ -97,9 +97,9 @@ export const injectContextIntoLambdaPayload = async (
       if (shouldUpdateStepForTracesMerging(step)) {
         updateStepObject(step)
         definitionHasBeenUpdated = true
-      } else if (step.Resource.startsWith('arn:aws:lambda')) {
+      } else if (step.Resource?.startsWith('arn:aws:lambda')) {
         context.stdout.write(
-          `\n[Warn] Step ${stepName} may be using the basic legacy integration, which does not support merging lambda trace(s) with Step Functions trace. 
+          `[Warn] Step ${stepName} may be using the basic legacy integration, which does not support merging lambda trace(s) with Step Functions trace. 
           To merge lambda trace(s) with Step Functions trace, please consider using the latest integration. 
           More details can be found on https://docs.aws.amazon.com/step-functions/latest/dg/connect-lambda.html \n`
         )
@@ -152,7 +152,7 @@ export type StatesType = Record<string, StepType>
 export type StepType = {
   Type: string
   Parameters?: ParametersType
-  Resource: string
+  Resource?: string
   Next?: string
   End?: boolean
 }
--- a/src/commands/stepfunctions/helpers.ts
+++ b/src/commands/stepfunctions/helpers.ts
@@ -97,6 +97,12 @@ export const injectContextIntoLambdaPayload = async (
       if (shouldUpdateStepForTracesMerging(step)) {
         updateStepObject(step)
         definitionHasBeenUpdated = true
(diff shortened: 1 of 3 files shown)


src/commands/stepfunctions/instrument.ts
```diff
@@ -155,6 +155,10 @@ export class InstrumentStepFunctionsCommand extends Command {
         stepFunctionTagsToAdd.push({key: TAG_VERSION_NAME, value: `v${cliVersion}`})
       }
 
+      if (!listStepFunctionTagsResponse?.tags?.some((tag) => tag.key === DD_TRACE_ENABLED)) {
```
nine5two7, line 158: So, here, it can cover all the below three cases?
1) DD_TRACE_ENABLED not appear, then add tag true
2) DD_TRACE_ENABLED is set to true by customer, then add tag true
3) DD_TRACE_ENABLED is set to false by customer

src/commands/stepfunctions/helpers.ts
```diff
@@ -146,7 +152,7 @@ export type StatesType = Record<string, StepType>
 export type StepType = {
   Type: string
   Parameters?: ParametersType
-  Resource: string
+  Resource?: string
```
kimi-p, line 155: Need to update this since I just learned that `Choice` doesn't have `Resource` field. Good that we catch this first.
```
    "ChoiceStep": {
      "Type": "Choice",
      "Comment": "some choice to make",
      "Choices": [
        {
          "Variable": "$.Comment",
          "StringEquals": "Insert your JSON here",
          "Next": "InvokeLambda"
        }
      ]
    },
```

src/commands/stepfunctions/instrument.ts
kimi-p: Thanks for catching this. I've update the PR so that flag set to true when:
- there are no such flag
- flag exist but value is not `true`

If the flag is already true and there are no there tags to update, ci won't make an additional call.

nine5two7 (approved):
lgtm

PR #1003, opened 2023-07-24
[ci] Make `bump-ci-integrations` depend on `build`

Drarig29:
### What and why?

In #942, we tried to invalidate the cache by passing a specific version to `yarn add`, but the downstream pipelines still occasionally fail ([example](https://github.com/DataDog/synthetics-ci-github-action/actions/runs/5587506326/job/15132248097)).

### How?

> ~~If this doesn't work, then we can move the `bump-ci-integrations` job matrix at the very bottom of `.github/workflows/release.yml`, which should add enough delay. **Source: #942**~~

Add a `needs:` dependency on `build`.

### Review checklist

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -15,34 +15,6 @@ jobs:
         env:
           YARN_NPM_AUTH_TOKEN: ${{ secrets.YARN_NPM_AUTH_TOKEN }}
 
-  bump-ci-integrations:
-    name: Bump datadog-ci in integration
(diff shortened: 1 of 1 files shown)


Drarig29:
@juan-fernandez oh yes, good point! When looking at the UI, it felt like the jobs were run sequentially...

When enabling timestamps though, we can see that the PRs are created before the end of `yarn npm publish`:

image

image

I'll add a dependency as you said 🙇

juan-fernandez (approved):

PR #1007, opened 2023-07-25
[synthetics] Fix max parallelization message

Drarig29:
### What and why?

Currently, the wrong property is used for the max parallelization message.

### How?

Replace `orgMaxConcurrencyCap` with `onDemandConcurrencyCap`

### Review checklist


etnbrd (approved):
LGTM 👍 
I don't see any validation that `onDemandeConcurrencyCap` is returned with the `orgSettings`, but I guess that's expected from the endpoint.

PR #1008, opened 2023-07-26
[flare] UX Improvements

nhulston:
### What and why?

- Explain post request errors (400 or 403=wrong API key, 500 most likely means invalid case ID or email) to avoid confusion. These two status codes are the most common errors that customers might run into.
- Require `caseId` and `email` in dry mode so that the dry mode is the same experience as running in non-dry mode
- Summarize configs (for both lambda and cloud-run) when printing as to not flood the user's terminal.
- Some small wording changes for better clarity

Screenshot 2023-08-09 at 10 52 29 AM
Screenshot 2023-08-09 at 11 29 03 AM

### How?

- Create the `summarizeConfig()` methods in `cloud-run/flare.ts` and `lambda/flare.ts` that only keep the most important information. We then stringify the new object and print that to terminal, instead of the original config
- Update `sendToDatadog()` function to print additional messages in the case of 500 or 403 status codes.
```
switch (code) {
  case 500:
    message += 'Are your case ID and email correct?\n'
    break
  case 400:
  case 403:
    message += 'Is your Datadog API key correct?\n'
    break
}
```

### Review checklist


src/commands/lambda/__tests__/flare.test.ts
```diff
@@ -986,4 +978,8 @@ describe('lambda flare', () => {
       expect(result).toEqual(expectedFiles)
     })
   })
+
+  test('summarizeConfig', () => {
+    expect(summarizeConfig(MOCK_LAMBDA_CONFIG)).toMatchSnapshot()
+  })
```
nhulston, line 984: Is this test necessary? This is already being tested by snapshots in other tests, but I still think it doesn't hurt to have this test

sfirrin: Yeah I think it doesn't hurt

src/helpers/flare.ts
```diff
@@ -61,7 +61,16 @@ export const sendToDatadog = async (
       const errResponse: string = (err.response?.data.error as string) ?? ''
       const errorMessage = err.message ?? ''
 
-      throw Error(`Failed to send flare file to Datadog Support: ${errorMessage}. ${errResponse}\n`)
+      let message = `Failed to send flare file to Datadog Support: ${errorMessage}. ${errResponse}\n`
+      const code = err.response?.status
+      // The error message doesn't say why there was an error, so it's important to tell the user why the request failed.
```
duncanista, line 66: What do you mean by this?

nhulston: Good question. Before this PR, the message just says:
`[Error] Failed to send flare file to Datadog Support: Request failed with status code 500.`
There is not any explanation as to why the status code is 500 though.
I will make sure to update the comment to clarify this

src/helpers/flare.ts
```diff
@@ -61,7 +61,16 @@ export const sendToDatadog = async (
...
+      if (code === 500) {
+        message += 'Are your case ID and email correct?\n'
+      } else if (code === 403) {
+        message += 'Is your Datadog API key correct?\n'
+      }
```
duncanista, line 71: Prefer a switch. 403 is indeed bad auth. 400 is there too, and 500 as well. Sent you the endpoint rejects privately.

nhulston: Thanks, added 400 and converted to switch statements.

src/helpers/flare.ts
```diff
@@ -61,7 +61,22 @@ export const sendToDatadog = async (
       const errResponse: string = (err.response?.data.error as string) ?? ''
       const errorMessage = err.message ?? ''
 
-      throw Error(`Failed to send flare file to Datadog Support: ${errorMessage}. ${errResponse}\n`)
+      let message = `Failed to send flare file to Datadog Support: ${errorMessage}. ${errResponse}\n`
+      const code = err.response?.status
+      // The error message doesn't say why there was an error. All it says is:
+      // "[Error] Failed to send flare file to Datadog Support: Request failed with status code 500."
+      // Therefore, we need to add an explanation to clarify when the code is 500 or 403.
+      switch (code) {
```
duncanista, line 69: Also, the Flare endpoint provides an `errors` field, why don't we use it?

Drarig29: We have a `formatBackendErrors()` function which uses this `errors` field:

https://github.com/DataDog/datadog-ci/blob/7ce750567e0f28506b88b3193ac176e30656a40a/src/commands/synthetics/api.ts#L40-L56

Feel free to move it and reuse if it suits your needs 😉

nhulston: > We have a `formatBackendErrors()` function which uses this `errors` field:
> 
> https://github.com/DataDog/datadog-ci/blob/7ce750567e0f28506b88b3193ac176e30656a40a/src/commands/synthetics/api.ts#L40-L56
> 
> Feel free to move it and reuse if it suits your needs 😉

Thanks for the suggestion! I don't think we'll need this here because we want some custom messages for handling errors

duncanista (approved):

Drarig29 (approved):
LGTM

PR #1010, opened 2023-07-27
[Static analysis] Use uuid for sarif's report filename

juan-fernandez:
### What and why?

We're not using the filename for anything. By using a `uuid` we're sure we're not going to find weird characters or collisions. 

### How?

Use `uuidv4` for filenames when uploading `sarif` reports. 

### Review checklist

~- [ ] Feature or bugfix MUST have appropriate tests (unit, integration)~

We don't have e2e tests right now, so this change can't be tested other than manually (I tested it and the report is uploaded correctly). We have another task for creating e2e tests


Drarig29 (approved):
Looks good to me

juli1 (approved):
Thank you for doing this for us 🙏

tonyredondo (approved):

PR #1011, opened 2023-07-27
[cloud-run][flare] Get recent logs

nhulston:
### What and why?

`cloud-run flare` is a command that gets important information about a Cloud Run service and sends the info to Datadog support for troubleshooting.

This PR adds the `--with-logs` argument, which gets recent logs for a cloud-run service.

Screenshot 2023-08-04 at 12 57 20 PM
Screenshot 2023-08-02 at 5 16 31 PM
Screenshot 2023-08-01 at 3 44 10 PM

### How?

1. Add the `@google-cloud/logging` package. We also explicitly add the `agent-base` and `http-proxy-agent` packages to resolve conflicting dependencies. If these two are not explicitly defined in `package.json`, the CLI will fail to build.
2. Add the `getLogs()` function, which gets recent logs using the cloud-run logging SDK.
  We add the `AND (textPayload:* OR httpRequest:*)` filter to only get logs that are an HTTP request or have a text payload (a standard output log). Other logs are just cloud-run audit logs, which aren't important.
3. After retrieving logs, we save them to a file using the `saveLogsFile()` method, which takes an array of logs and the file path to save the logs to.
4. If no logs are found, we don't write that file.
5. Currently, there are 4 different types of logs: `total` meaning the most recent logs of any kind, `warning` meaning only logs with a severity >= to WARNING, `error` meaning only logs with a severity >= to ERROR, and `debug` meaning only logs with a severity == to DEBUG. More types of logs can easily be added by adjusting the `logsConfig` array.

### Review checklist


src/commands/cloud-run/flare.ts
```diff
@@ -141,19 +200,10 @@ export class CloudRunFlareCommand extends Command {
...
-        if (err instanceof Error) {
-          this.context.stderr.write(helpersRenderer.renderError(err.message))
-        }
-
-        return 1
-      }
+      const confirmSendFiles = await requestConfirmation(
+        'Are you sure you want to send the flare file to Datadog Support?',
+        false
+      )
```
nhulston, line 206: Since this is all wrapped by an outer try/catch, this try/catch here is redundant.

src/commands/cloud-run/flare.ts
```diff
@@ -4,27 +4,43 @@ import path from 'path'
...
+const TEXT_LOGS_FILE_NAME = 'text_logs.csv'
+const WARNING_LOGS_FILE_NAME = 'warning_logs.csv'
+const ERRORS_LOGS_FILE_NAME = 'error_logs.csv'
+const DEBUG_LOGS_FILE_NAME = 'DEBUG_logs.csv'
+
+// Must be in range 0 - 1000
+export const MAX_LOGS_PER_PAGE = 1000
+// There will be a maximum of (MAX_LOGS_PER_PAGE * MAX_PAGES) logs for each log file
+// The more pages there are, the longer the program will take to run
+export const MAX_PAGES = 3
```
nhulston, line 37: Do we need more than 1000 logs per file? Or is the pagination overkill?

src/commands/lambda/flare.ts
```diff
@@ -362,19 +362,11 @@ export class LambdaFlareCommand extends Command {
...
-        if (err instanceof Error) {
-          this.context.stderr.write(helpersRenderer.renderError(err.message))
-        }
+      const confirmSendFiles = await requestConfirmation(
+        'Are you sure you want to send the flare file to Datadog Support?',
+        false
+      )
 
-        return 1
-      }
```
nhulston, line 377: Same here, the inner try/catch is redundant.

src/commands/cloud-run/flare.ts
```diff
@@ -228,25 +281,134 @@ export const getCloudRunServiceConfig = async (
...
+
+  // Only get recent logs
+  const date = new Date()
+  date.setMinutes(date.getMinutes() - MAX_LOG_AGE_MINUTES)
+  const formattedDate = date.toISOString()
+
+  // Query options
+  let filter = `resource.labels.service_name="${serviceId}" AND resource.labels.location="${location}" AND timestamp>="${formattedDate}"`
+  filter += severityFilter ? ` AND ${severityFilter}` : ''
+  filter += isOnlyTextLogs ? ' AND textPayload:*' : ' AND (textPayload:* OR httpRequest:*)'
```
nhulston, line 332: We only care about logs that are an HTTP request or that have a text payload. Other logs are just audit logs which cause a lot of unnecessary spam.

src/commands/cloud-run/flare.ts
```diff
@@ -114,21 +130,60 @@ export class CloudRunFlareCommand extends Command {
...
+          if (logs.length === 0) {
+            this.context.stdout.write(`• No ${logConfig.type} logs were found\n`)
+          } else {
+            this.context.stdout.write(`• Found ${logs.length} ${logConfig.type} logs\n`)
+            logFileMappings.set(logs, logConfig.fileName)
+          }
+        } catch (err) {
+          const msg = err instanceof Error ? err.message : ''
+          this.context.stderr.write(`• Unable to get ${logConfig.type} logs: ${msg}\n`)
+        }
```
nhulston, line 158: If there is an error getting logs, do we want to end the entire program? As the code is now, we don't exit the program when an error is hit.

src/commands/cloud-run/flare.ts
```diff
@@ -114,21 +130,60 @@ export class CloudRunFlareCommand extends Command {
...
+      const rootFolderPath = path.join(process.cwd(), FLARE_OUTPUT_DIRECTORY)
+      const logsFolderPath = path.join(rootFolderPath, LOGS_DIRECTORY)
+      this.context.stdout.write(chalk.bold(`\n💾 Saving files to ${rootFolderPath}...\n`))
       if (fs.existsSync(rootFolderPath)) {
         deleteFolder(rootFolderPath)
       }
+      const subFolders = []
+      if (logFileMappings.size > 0) {
+        subFolders.push(logsFolderPath)
+      }
```
nhulston, line 173: I use this boolean to determine whether or not to create a logs folder. There is an edge case where `--with-logs` is included, but for whatever reason no logs are found (maybe the service hasn't been ran in 24 hours). In this case, it's unnecessary to create a logs folder.

src/commands/cloud-run/flare.ts
```diff
@@ -174,11 +220,11 @@ export class CloudRunFlareCommand extends Command {
       deleteFolder(rootFolderPath)
     } catch (err) {
       if (err instanceof Error) {
-        this.context.stderr.write(helpersRenderer.renderError(`Unable to save configuration: ${err.message}`))
+        this.context.stderr.write(helpersRenderer.renderError(err.message))
       }
-    }
 
-    return 0
```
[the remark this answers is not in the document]
nhulston: I agree, thanks for the comment. Re-added this line

src/commands/cloud-run/flare.ts
```diff
@@ -4,27 +4,43 @@ import path from 'path'
...
 import {SKIP_MASKING_CLOUDRUN_ENV_VARS} from './constants'
+import {CloudRunLog, LogConfig} from './interfaces'
 import {renderAuthenticationInstructions} from './renderer'
 
 const SERVICE_CONFIG_FILE_NAME = 'service_config.json'
 const FLARE_ZIP_FILE_NAME = 'cloudrun-flare-output.zip'
+const ALL_LOGS_FILE_NAME = 'all_logs.csv'
+const WARNING_LOGS_FILE_NAME = 'warning_logs.csv'
+const ERRORS_LOGS_FILE_NAME = 'error_logs.csv'
+const DEBUG_LOGS_FILE_NAME = 'DEBUG_logs.csv'
```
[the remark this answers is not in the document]
nhulston: Good catch, that was a mistake on my end. Fixed!

src/commands/cloud-run/__tests__/flare.test.ts
```diff
@@ -86,6 +98,8 @@ jest.mock('@google-cloud/run', () => {
 jest.spyOn(helpersPromptModule, 'requestConfirmation').mockResolvedValue(true)
 jest.mock('util')
 jest.mock('jszip')
+jest.mock('@google-cloud/logging-min')
+jest.useFakeTimers({advanceTimers: true, now: new Date(Date.UTC(2023, 0))})
```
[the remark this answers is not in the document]
nhulston: Makes sense, thanks for the comment. I have removed `advanceTimers`

src/commands/lambda/__tests__/flare.test.ts
```diff
@@ -100,7 +100,7 @@ const mockJSZip = {
 ;(JSZip as any).mockImplementation(() => mockJSZip)
 
 // Date
-jest.useFakeTimers({advanceTimers: true, now: new Date(Date.UTC(2023, 0))})
```
nhulston, line 103: This `advanceTimers` was unnecessary. All tests still pass without this

package.json
```diff
@@ -66,8 +66,10 @@
     "@aws-sdk/client-sfn": "^3.358.0",
     "@aws-sdk/credential-providers": "^3.358.0",
     "@aws-sdk/property-provider": "^3.357.0",
+    "@google-cloud/logging-min": "^10.4.0",
```
duncanista, line 69: This is an experimental library, likely to be modified.

> Note: logging-min is experimental, and its feature surface is subject to change.

Do we really want to rely on an experimental package?

LICENSE-3rdparty.csv
```diff
@@ -79,3 +79,6 @@ ajv-formats,import,MIT,Copyright (c) 2020 Evgeny Poberezkin
 jszip,import,MIT,Copyright (c) 2009-2016 Stuart Knightley and other contributors
 @google-cloud/run,import,Apache-2.0,Copyright (c) 2023 Google LLC and other contributors
 google-auth-library,import,Apache-2.0,Copyright (c) 2023 Google LLC and other contributors
+@google-cloud/logging-min,import,Apache-2.0,Copyright (c) 2023 Google LLC and other contributors
+agent-base,import,MIT,Copyright (c) 2013 Nathan Rajlich
```
duncanista, line 83: Still unsure about adding new libraries which are only dependents of the GCP logging one.

src/commands/cloud-run/flare.ts
duncanista: Maybe ask to Support?

src/commands/cloud-run/flare.ts
duncanista: Good question to ask Support, do we rather get some logs? What if the re-run allows us to get the whole file? What about a retry mechanism?

src/commands/cloud-run/flare.ts
duncanista: What about telling the user that since no logs were found, we are not creating it? I guess this will only apply if they have `--with-logs` set

src/commands/cloud-run/interfaces.ts
```diff
@@ -0,0 +1,14 @@
+// Info that will be used to build the CSV file
```
duncanista, line 1: You can do better JSDocs for these interfaces.

src/commands/cloud-run/flare.ts
```diff
@@ -114,21 +130,60 @@ export class CloudRunFlareCommand extends Command {
...
+    const logFileMappings = new Map<CloudRunLog[], string>()
+    if (this.withLogs) {
+      this.context.stdout.write(chalk.bold('\n📖 Getting logs...\n'))
+
+      const logsConfig: LogConfig[] = [
+        {type: 'total', fileName: ALL_LOGS_FILE_NAME},
+        {type: 'warning', severityFilter: ' AND severity>="WARNING"', fileName: WARNING_LOGS_FILE_NAME},
+        {type: 'error', severityFilter: ' AND severity>="ERROR"', fileName: ERRORS_LOGS_FILE_NAME},
+        {type: 'debug', severityFilter: ' AND severity="DEBUG"', fileName: DEBUG_LOGS_FILE_NAME},
+      ]
```
duncanista, line 143: This seems to be a constant?

src/commands/cloud-run/flare.ts
duncanista: NVM, this is handled in "no logs were found".

src/commands/cloud-run/flare.ts
```diff
@@ -114,21 +130,60 @@ export class CloudRunFlareCommand extends Command {
     const configStr = util.inspect(config, false, 10, true)
     this.context.stdout.write(`\n${configStr}\n`)
 
-    // Save and zip service configuration
-    this.context.stdout.write(chalk.bold('\n💾 Saving configuration...\n'))
-    const rootFolderPath = path.join(process.cwd(), FLARE_OUTPUT_DIRECTORY)
+    // Get logs
+    const logFileMappings = new Map<CloudRunLog[], string>()
```
duncanista, line 134: Map definitions are usually `key,value`.

src/commands/cloud-run/flare.ts
```diff
@@ -174,8 +220,10 @@ export class CloudRunFlareCommand extends Command {
       deleteFolder(rootFolderPath)
     } catch (err) {
       if (err instanceof Error) {
-        this.context.stderr.write(helpersRenderer.renderError(`Unable to save configuration: ${err.message}`))
+        this.context.stderr.write(helpersRenderer.renderError(err.message))
```
duncanista, line 223: Why are we removing the message here now?

src/commands/cloud-run/flare.ts
```diff
@@ -228,25 +276,129 @@ export const getCloudRunServiceConfig = async (
...
+  // We only want to get logs from the last `MAX_LOG_AGE_MINUTES` to make sure they are relevant.
+  // We also only want to include logs with a textPayload or logs that were an HTTP request.
+  // Any other logs are just audit logs which are spammy and don't have any relevant information.
+  filter += severityFilter ?? ''
+
+  const options = {
+    filter,
+    orderBy: FILTER_ORDER,
+    pageSize: MAX_LOGS_PER_PAGE,
+    page: '',
```
duncanista, line 329: Is page needed even if it's empty?

duncanista (changes requested):
Left some comments, we must not use an external experimental library.

package.json
nhulston: Good point, we can use `@google-cloud/logging` instead

src/commands/cloud-run/flare.ts
nhulston: True, this can be moved to the top of the file.

src/commands/cloud-run/flare.ts
nhulston: Before, the only thing we were doing was saving the configuration file. Since we are now saving log files, zipping, and sending in this try/catch block, this message is outdated.

This is okay because each of the helper methods used here has proper message handling: `deleteFolder()`, `createDirectories()`, `writeFile()`, `zipContents()`, etc. all have descriptive error handling build in.

duncanista: Thanks for the explanation!

src/commands/cloud-run/flare.ts
nhulston: We only need 1000 logs per file, so I removed pagination and we get a max of 1000 logs now

src/commands/cloud-run/interfaces.ts
```diff
@@ -0,0 +1,28 @@
+/**
+ * Used to store important info about a CloudRun log.
+ * This info will be used to build the CSV log files.
```
duncanista, line 3:
```suggestion
 * Summarize relevant information about a Cloud Run log.
 * Also used to build CSV log files.
```

duncanista (approved):
Looks good from Serverless perspective!

package.json
```diff
@@ -66,8 +66,10 @@
     "@aws-sdk/client-sfn": "^3.358.0",
     "@aws-sdk/credential-providers": "^3.358.0",
     "@aws-sdk/property-provider": "^3.357.0",
+    "@google-cloud/logging": "^10.5.0",
     "@google-cloud/run": "^0.6.0",
     "@types/datadog-metrics": "0.6.1",
+    "agent-base": "^7.1.0",
```
juan-fernandez, line 72: is this really needed? From my tests, just adding `http-proxy-agent` solves the issues

juan-fernandez (changes requested):
blocking until we figure out if `agent-base` is needed

package.json
nhulston: I have removed `agent-base` and looks like it is not needed. Thanks for the reivew

juan-fernandez (approved):

PR #1013, opened 2023-07-31
[CIAPP-7210] datadog-ci gate wait for the time returned by service

liashenko:
### What and why?

Update `datadog-ci gate` wait logic to wait for the time returned by the service, not the predefined time, to increase flexibility on handling the data latency, and not depend on datadog-ci releases. 

### How?
Handle in datadog-ci `200 OK`, `status:wait` and `wait_time_ms` returned from the service, in case the gate is not ready to be evaluated yet. 

### Review checklist


src/commands/gate/evaluate.ts
```diff
@@ -143,7 +149,42 @@ export class GateEvaluateCommand extends Command {
...
+        .catch((err) => {
+          setTimeout(() => {
+            reject(err)
+          }, this.getDelay(attempt ?? 1))
+        })
+    })
+  }
+
+  private getDelay(attempt: number): number {
+    return attempt * this.initialRetryMs
```
juan-fernandez, line 184: this is not exponential but linear with the attempt. I'd say it's ok, but let's not say exponential in the comment. If you actually want exponential you may do `2 ** attempt * this.initialRetryMs`

src/commands/gate/evaluate.ts
```diff
@@ -119,22 +126,21 @@ export class GateEvaluateCommand extends Command {
...
-      await this.delay(this.waitingTime)
-    }
-
     this.context.stdout.write(renderGateEvaluationInput(evaluateRequest))
 
-    return retryRequest(
-      () => api.evaluateGateRules(evaluateRequest, this.context.stdout.write.bind(this.context.stdout)),
-      {
-        onRetry: (e, attempt) => {
+    // retry is handled in evaluateRulesWithWait, so we set maxTimeout and minTimeout as 0 for retryRequest
```
juan-fernandez, line 131: this is going to be confusing so let's use the chance to be clearer in the comment
```suggestion
    /** 
    * `retryRequest` does not allow setting a wait time dependent on a backend response,
    * so we handle the wait time in `evaluateRulesWithWait`: we'll wait whatever is necessary 
    * for the returned promise to resolve or be rejected. The retry will start immediately after 
    * and will be handled by `retryRequest`.
    */
```

src/commands/gate/evaluate.ts
```diff
@@ -143,7 +149,42 @@ export class GateEvaluateCommand extends Command {
       })
   }
 
-  private handleEvaluationSuccess(evaluationResponse: EvaluationResponse): number {
+  /**
+   * Evaluate rules with retry
+   * then if service returns status:wait, retry for wait_time_ms
+   * if service returns 5xx, exponential backoff retry
+   */
```
juan-fernandez, line 156: it's not quite a "retry". The retry will still be handled by `retryRequest`, but we "fake" a wait time here. It's not very natural so it's worth investing in a longer comment 

```suggestion
  /**
   * Evaluate gate rules and return a promise that will behave as follows:
   * - If the request is successful, the promise will be resolved with the response
   * - If the request is successful but the status is 'wait', the promise will be rejected after the received wait time (wait_time_ms)
   * - If the request is not successful, the promise will be rejected after `initialRetryMs`, with an exponential factor that depends on the attempt (exponential backoff).
   * If the promise is rejected, `retryRequest` will handle the retry immediately.
   */
```

src/commands/gate/evaluate.ts
```diff
@@ -119,22 +126,21 @@ export class GateEvaluateCommand extends Command {
...
+        // render retry message if error is not wait
+        if (e.message !== 'wait') {
           this.context.stderr.write(renderEvaluationRetry(attempt, e))
-        },
-        retries: 5,
-      }
-    )
+        }
+      },
+      maxRetryTime: this.maxRetryTimeMs,
```
juan-fernandez, line 139: I think we need not to set this, as this seems to be the _maximum_ time that we'll allow all the retries to last (https://github.com/tim-kos/node-retry/blob/master/lib/retry_operation.js#L10)
```suggestion
```

juan-fernandez (commented):
left some comments to improve comments, because this way we implemented the wait time is not the most intuitive

src/commands/gate/evaluate.ts
liashenko: That was my idea, but now I'm thinking about it, we should do it on the backend, to make the timeout more flexible as well.

src/commands/gate/evaluate.ts
liashenko: I'll make it exponential, but change `initialRetryMs` to 1s, because it's gonna be waiting a lot as is.

src/commands/gate/evaluate.ts
```diff
@@ -119,22 +125,25 @@ export class GateEvaluateCommand extends Command {
...
     this.context.stdout.write(renderGateEvaluationInput(evaluateRequest))
 
-    return retryRequest(
-      () => api.evaluateGateRules(evaluateRequest, this.context.stdout.write.bind(this.context.stdout)),
-      {
-        onRetry: (e, attempt) => {
+    /**
+     * `retryRequest` does not allow setting a wait time dependent on a backend response,
+     * so we handle the wait time in `evaluateRulesWithWait`: we'll wait whatever is necessary 
+     * for the returned promise to resolve or be rejected. The retry will start immediately after
```
juan-fernandez, line 133: my bad: there's only a retry if the promise is rejected
```suggestion
     * for the returned promise to be rejected. The retry will start immediately after
```

juan-fernandez (approved):

juan-fernandez:
it seems the lint job is failing. Probably because my suggestions in the PR do not follow standards 😅

src/commands/gate/__tests__/evaluate.test.ts
```diff
@@ -40,6 +43,9 @@ describe('evaluate', () => {
       const response: EvaluationResponse = {
         status: 'passed',
         rule_evaluations: [],
+        metadata: {
+          wait_time_ms: 0,
+        },
```
andrea-mosk, line 48: Is this needed? If so, why? We should be able to handle responses without it no (which should happen in the normal case)?

andrea-mosk (approved):
Overall it's a bit weird that we are conflating normal retries with retries in which we just needed to wait for X seconds. I think this should be done so that a single retry consists of N requests, based on how much we need to wait.

Also, is there a limit on the number of seconds we need to wait? Can the service return 10 minutes? Are we ok with waiting 10 minutes in that case?

This said, overall LGTM so I won't block it

src/commands/gate/__tests__/evaluate.test.ts
liashenko: It's because of ts interface, it fails when running tests without it, but when running the command it works :D 
I'll try to fix it.

liashenko:
> Overall it's a bit weird that we are conflating normal retries with retries in which we just needed to wait for X seconds. I think this should be done so that a single retry consists of N requests, based on how much we need to wait.

To do this we need to implement our own retry mechanism, because with the current framework we cannot wait inside retry.

> Also, is there a limit on the number of seconds we need to wait? Can the service return 10 minutes? Are we ok with waiting 10 minutes in that case?

We can fail the request on the service side if time passed is more than 5 minutes, this way we can control timeout on the service side as well.

src/commands/gate/evaluate.ts
```diff
@@ -145,7 +162,48 @@ export class GateEvaluateCommand extends Command {
...
+   * - If the request is successful but the status is 'wait', the promise will be rejected after the received wait time (wait_time_ms)
+   * - If the request is not successful, the promise will be rejected after `initialRetryMs`, with an exponential factor that depends on the attempt (exponential backoff).
+   * If the promise is rejected, `retryRequest` will handle the retry immediately.
+   */
+  private async evaluateRulesWithWait(
+    api: APIHelper,
+    evaluateRequest: Payload,
+    attempt?: number
+  ): Promise<AxiosResponse<EvaluationResponsePayload>> {
+    const timePassed = new Date().getTime() - evaluateRequest.startTimeMs
```
juan-fernandez, line 177: this measures the time between the beginning of `execute` and this line? This is always going to be _very_ little time, right? Or am I missing something? It seems unnecessary complexity if it's going to be so little time.

src/commands/gate/evaluate.ts
```diff
@@ -53,14 +54,23 @@ export class GateEvaluateCommand extends Command {
...
 
+  private initialRetryMs = 1000
+  private maxRetries = 5
+  private defaultTimeout = '1800' // 30 min
+
   private dryRun = Option.Boolean('--dry-run', false)
   private failOnEmpty = Option.Boolean('--fail-on-empty', false)
   private failIfUnavailable = Option.Boolean('--fail-if-unavailable', false)
   private noWait = Option.Boolean('--no-wait', false)
+  private timeoutS = Option.String('--timeout', this.defaultTimeout, {validator: validation.isInteger()})
```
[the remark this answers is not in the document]
juan-fernandez: is it seconds? If so, let's say seconds

src/commands/gate/evaluate.ts
```diff
@@ -53,14 +54,23 @@ export class GateEvaluateCommand extends Command {
...
+        'Evaluate matching quality gate rules in Datadog with a timeout of 120 seconds',
+        'datadog-ci gate evaluate --timeout 120',
+      ],
       ['Evaluate matching quality gate rules in Datadog without waiting', 'datadog-ci gate evaluate --no-wait'],
     ],
   })
 
+  private initialRetryMs = 1000
+  private maxRetries = 5
+  private defaultTimeout = '1800' // 30 min
```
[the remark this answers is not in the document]
liashenko: it's used as a default value for `Option.String`

src/commands/gate/evaluate.ts
liashenko: For the first request it's going to be small, but for retries it might be minutes, it depends on the wait time returned from the service (the processing delay).
The main thing here is `remainingWait`, so we timeout the command if a lot of time has passed since the beginning.

src/commands/gate/evaluate.ts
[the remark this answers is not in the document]
liashenko: I guess it's better to convert it there

src/commands/gate/__tests__/evaluate.test.ts
```diff
@@ -116,6 +116,22 @@ describe('evaluate', () => {
       expect(stdErrLog).not.toContain('internal issue')
     })
   })
+  describe('wait', () => {
+    test('should pass the command if the response status is wait', () => {
+      const write = jest.fn()
+      const command = new GateEvaluateCommand()
+      command.context = {stdout: {write}} as any
```
Drarig29, line 123: You can now refactor this. It will take care of resolving the default values for your command options.
```suggestion
      const command = createCommand(GateEvaluateCommand, {stderr: {write}} as any)
```

Drarig29 (approved):
LGTM, just one nit on unit tests

src/commands/gate/evaluate.ts
juan-fernandez: I see. The way it's implemented, if `remainingWait` is 0 it seems there'll be a single retry more (because we'll reject with `'wait'`), but maybe that'd be handled by `maxRetryTime: this.timeoutInSeconds * 1000,` ? 

I think at some point we'll need to add proper e2e tests because the logic has become quite custom

juan-fernandez (approved):
LGTM but at some point we want to add proper end to end tests to this, or at least some kind of integration tests with a mocked backend: the retry logic has become convoluted and it is bound to have corner cases we haven't covered. We can do that at a different PR so I'm not blocking

src/commands/gate/evaluate.ts
liashenko: You're right, apparently `maxRetryTime` doesn't do anything useful in this case. 

I also missed a couple of edge cases: like failing the command if `--fail-if-unavailble` is enabled, or if the timeout was 0, it just retried a lot.

liashenko: Added integration tests.

buraizu (approved):

src/commands/gate/evaluate.ts
```diff
@@ -177,27 +177,31 @@ export class GateEvaluateCommand extends Command {
...
 
     return new Promise((resolve, reject) => {
-      api
-        .evaluateGateRules(evaluateRequest, this.context.stdout.write.bind(this.context.stdout))
-        .then((response) => {
-          if (response.data.data.attributes.status === 'wait') {
-            this.context.stdout.write(renderWaiting())
-            const waitTime = response.data.data.attributes.metadata?.wait_time_ms ?? 0
+      if (remainingWait <= 0) {
+        reject(new Error('wait'))
```
juan-fernandez, line 184: if `remainingWait <= 0` we want to stop trying, right? Does it make sense to reject with `'wait'` here?

juan-fernandez (approved):

src/commands/gate/evaluate.ts
liashenko: you're right, I'm going to do it in the next PR!

PR #962, opened 2023-07-07
[lambda][flare] Autodiscover project files and prompt for additional files

nhulston:
### What and why?

For more context on what Lambda Flare is, please read the description in this PR: https://github.com/DataDog/datadog-ci/pull/924

1. Gather important project files such as `package.json` to assist Datadog support in resolving customer issues.
2. Interactively prompt the user to specify any additional files they would like to send to Datadog support.
3. All files to be sent are clearly listed to ensure users can review them before confirming and sending the mto support.
4. Update command line messages using `chalk.bold`

Screenshot 2023-07-11 at 10 34 56 AM

Screenshot 2023-07-11 at 10 36 11 AM

### How?

1. Loop through files in `PROJECT_FILES` located in `constants.ts` and determine if any of those files exist in the current directory. If so, automatically copy them to the flare project files directory.
2. Continue to prompt the user for additional file paths they would like to manually send until they enter an empty path.
3. All files are copied using the `fs.copyFileSync()` method into the flare additional files directory.
4. Check for duplicates before adding custom files so the user doesn't copy the same file multiple times.

### Review checklist

- [x] Feature or bugfix MUST have appropriate tests (unit, integration)

PR #1019, opened 2023-08-03
[cloud-run][flare] Autodiscover project files and allow additional files

nhulston:
### What and why?

This PR adds support for autodiscovering project files in cloud run flare. Also, it allows the user to specify additional files they would like to upload and send to Datadgo support.

The logic for these two features have already been implemented for lambda flare in #962. Therefore, this PR mostly involves refactoring and reuse of code.

Screenshot 2023-08-07 at 2 39 07 PM

### How?

Added support for auto-discovering project files & specifying additional files in `cloud-run/flare.ts`. Lots of helper functions that were used by `lambda/flare.ts` have been refactored and reused by cloud-run. I also updated the `subFolders` array to include folders for project files & additional files, and we copy these files to those folders.

As for refactoring, here are some specifics:
1. Created `renderProjectFiles()` method in `helpers` to be used by both cloud run and lambda.
2. Moved `validateFilePath()` and `getProjectFiles()` from `lambda/flare.ts` to `helpers/flare.ts`. 
3. Moved `requestFilePath` from `lambda/prompt.ts` to `helpers/prompt.ts`.
4. Moved `PROJECT_FILES_DIRECTORY` and `ADDITIONAL_FILES_DIRECTORY` consts from `lambda/flare.ts` to `src/constants.ts`

### Review checklist


src/commands/lambda/__tests__/flare.test.ts
```diff
@@ -91,20 +88,13 @@ fs.readdirSync = jest.fn().mockReturnValue([])
...
 }))
 
-// Zip mocks
-jest.mock('jszip')
-const mockJSZip = {
-  file: jest.fn(),
-  generateAsync: jest.fn().mockResolvedValue('zip content'),
-}
-;(JSZip as any).mockImplementation(() => mockJSZip)
-
```
nhulston, line 101: This mock is not needed anymore since we moved the zip logic and zip tests to helpers

src/helpers/__tests__/flare.test.ts
```diff
@@ -4,13 +4,20 @@ import process from 'process'
...
-import {getEndpointUrl, sendToDatadog} from '../flare'
+import {getEndpointUrl, getProjectFiles, sendToDatadog, validateFilePath} from '../flare'
+import * as flareModule from '../flare'
 
-jest.mock('jszip')
+import {MOCK_CWD} from './fixtures'
+
+// Mocks
 jest.mock('fs')
+process.cwd = jest.fn().mockReturnValue(MOCK_CWD)
```
Drarig29, line 17: Wasn't it possible to do `jest.spyOn(process, 'cwd').mockReturnValue()`?

src/helpers/flare.ts
```diff
@@ -3,10 +3,12 @@
  */
 
 import fs from 'fs'
+import path from 'path'
 
 import axios from 'axios'
 import FormData from 'form-data'
 
+import {PROJECT_FILES} from '../commands/lambda/constants'
```
duncanista, line 11: A helper function which is dependent on an internal command file?

src/helpers/flare.ts
```diff
@@ -88,3 +91,43 @@ export const getEndpointUrl = () => {
 
   return 'https://' + endpointUrl + FLARE_ENDPOINT_PATH
 }
+
+/**
+ * Searches current directory for project files
+ * @returns a set of file paths of project files
+ */
+export const getProjectFiles = async () => {
```
duncanista, line 99: Not sure if the AWS Lambda project files constant should be used for Cloud Run too, at least not the framework ones.

src/helpers/__tests__/flare.test.ts
```diff
@@ -4,13 +4,20 @@ import process from 'process'
 import axios from 'axios'
 import FormData from 'form-data'
 
+import {PROJECT_FILES} from '../../commands/lambda/constants'
```
duncanista, line 7: Same here, we are testing a `lambda` constant for a helper method.

src/commands/cloud-run/__tests__/renderer.test.ts
```diff
@@ -1,5 +1,36 @@
...
 
-test('renderAuthenticationInstructions', () => {
-  expect(renderAuthenticationInstructions()).toMatchSnapshot()
+process.cwd = jest.fn().mockReturnValue(MOCK_CWD)
+
+describe('renderer', () => {
+  test('renderAuthenticationInstructions', () => {
+    expect(renderAuthenticationInstructions()).toMatchSnapshot()
+  })
+
```
duncanista, line 14: I don't think we should be testing this inside of the `cloud-run` folder.

src/commands/cloud-run/__tests__/renderer.test.ts
```diff
@@ -1,5 +1,36 @@
+import process from 'process'
```
duncanista, line 1: I don't think we test renderers at all, is it worth it?

duncanista (changes requested):
Left some comments.

src/helpers/__tests__/flare.test.ts
nhulston: Yes, that is possible. I have changed all appearances of `process.cwd = ...`. Thanks for the comment!

src/helpers/flare.ts
nhulston: Good catch, I have refactored to avoid this

src/helpers/flare.ts
nhulston: I agree, cloud run now uses `FLARE_PROJECT_FILES`, and lambda uses `LAMBDA_PROJECT_FILES` which contains everything in `FLARE_PROJECT_FILES` and all the keys from `FRAMEWORK_FILES_MAPPING`.

src/commands/cloud-run/__tests__/renderer.test.ts
nhulston: Why not? `renderAuthenticationInstructions()` lives in `cloud-run/renderer.ts`

duncanista: Oh, not that, the ones below it, idk why it shows that one.

We normally don't test renderers because your snapshots already show them.

nhulston: Makes sense, I have removed the renderer tests

src/commands/cloud-run/flare.ts
```diff
@@ -11,19 +11,30 @@ import chalk from 'chalk'
...
 import {GoogleAuth} from 'google-auth-library'
 
-import {API_KEY_ENV_VAR, CI_API_KEY_ENV_VAR, FLARE_OUTPUT_DIRECTORY, LOGS_DIRECTORY} from '../../constants'
-import {sendToDatadog} from '../../helpers/flare'
+import {
+  ADDITIONAL_FILES_DIRECTORY,
+  API_KEY_ENV_VAR,
+  CI_API_KEY_ENV_VAR,
+  FLARE_OUTPUT_DIRECTORY,
+  FLARE_PROJECT_FILES,
```
duncanista, line 19: Cool, I wonder if in the future, the flare constant should be inside the `helpers/flare.ts` file? nit, not a blocker

duncanista (approved):
LGTM from the Serverless side

src/commands/cloud-run/flare.ts
nhulston: Good question. I could actually move all the flare constants to `helpers/flare.ts`, or we could keep them here in `src/constants.ts`

Currently, these constants are only used by the `cloud-run` and `lambda` commands, so they are kind of flooding the constants file. But eventually, they will also be used by the Azure `app-service` command.

@Drarig29 do you think they should be moved?

Drarig29: I think for now, the `src/constants.ts` is a good place. As long as it's well organized, I don't think this file being long is an issue. And IIUC, those flare constants will soon be shared by 3 commands.

Maybe in the future we can split `src/constants.ts` into multiple files, but I don't think it's needed right now. Let alone in this PR.

Drarig29: @nhulston I would rather keep constants in `src/constants.ts`, even if they are specific to sending a flare, which can be achieved my multiple commands.

PR #1020, opened 2023-08-04
[cloud-run][flare] Generate insights file

nhulston:
### What and why?

This PR creates a file called `INSIGHTS.md` that includes additional information that can be helpful to support staff. It is basically a TLDR of the service config, as well as some additional information including:
- Directory that the command was ran in
- CLI version
- Timestamp
- Environment variables
- List of recent revisions
- and more...

Screenshot 2023-08-11 at 1 56 07 PM

### How?

The `getRecentRevisions()` function uses the cloud-run SDK to fetch recent revisions. We achieve this by creating a `RevisionsClient` object, and calling the cloud-run `listRevisions()` function. For each revision, we create a string which includes the revision name and deployment time. Finally, we save these to the insights file.

As for the other information in the insights file, we pull out the most important information from the service config and write it to the `INSIGHTS.md` file using the `writeFile()` method. There are a few other pieces of information that we add to the file, including the current directory, CLI version, and timestamp.

If either `getRecentRevisions()` or `generateInsightsFile()` fail, we don't exit the program but instead print a warning and continue. This is because the insights file is a "nice to have", but does not warrant blocking the program from running when it fails.

### Review checklist


src/commands/lambda/flare.ts
```diff
@@ -349,9 +349,16 @@ export class LambdaFlareCommand extends Command {
...
+      try {
+        const insightsFilePath = path.join(rootFolderPath, INSIGHTS_FILE_NAME)
+        generateInsightsFile(insightsFilePath, this.isDryRun, config)
+        this.context.stdout.write(`• Saved insights file to ./${INSIGHTS_FILE_NAME}\n`)
+      } catch (err) {
+        const errorDetails = err instanceof Error ? err.message : ''
+        this.context.stdout.write(
+          helpersRenderer.renderSoftWarning(`Unable to create INSIGHTS.md file. ${errorDetails}`)
+        )
+      }
```
nhulston, line 361: Adding an inner try/catch because we don't want to exit the program just because there was an error in the INSIGHTS.md generation.
Insights is a nice to have, but shouldn't block the flare program from running

src/commands/cloud-run/flare.ts
```diff
@@ -34,13 +35,18 @@ import {SKIP_MASKING_CLOUDRUN_ENV_VARS} from './constants'
 import {CloudRunLog, LogConfig} from './interfaces'
 import {renderAuthenticationInstructions} from './renderer'
 
+const version = require('../../../package.json').version
```
duncanista, line 38: Use version constant from helpers.

src/commands/cloud-run/flare.ts
```diff
@@ -204,6 +210,17 @@ export class CloudRunFlareCommand extends Command {
       }
     }
 
+    // Get recent revisions, which will be used to generate insights file
+    this.context.stdout.write(chalk.bold('\n🌧 Fetching recent revisions...\n'))
```
duncanista, line 214: What are revisions?

src/commands/cloud-run/flare.ts
```diff
@@ -501,3 +538,123 @@ export const saveLogsFile = (logs: CloudRunLog[], filePath: string) => {
...
+ * @param location
+ * @param project
+ * @returns a string array of recent revisions
+ */
+export const getRecentRevisions = async (service: string, location: string, project: string) => {
+  const client = new RevisionsClient()
+  const request = {
+    parent: client.servicePath(project, location, service),
+  }
+  const iterable = (await client.listRevisions(request)) as any
```
duncanista, line 554: Why typing as any?

duncanista (commented):
Left some comments

src/commands/cloud-run/flare.ts
nhulston: Screenshot 2023-08-11 at 1 27 03 PM

Revisions basically just keep track of every "version" of the cloud-run service. Whenever it is modified and redeployed, a new revision is created.
I talked to @IvanTopolcic and he mentioned that this could be useful to support. It also appears in the cloud-run UI (see image above).

I will also include the deployment time since deployment time is included on the cloud-run dashboard

Screenshot 2023-08-11 at 1 45 35 PM

src/commands/cloud-run/flare.ts
nhulston: Good catch, I removed the `as any`

src/commands/cloud-run/flare.ts
```diff
@@ -501,3 +538,123 @@ export const saveLogsFile = (logs: CloudRunLog[], filePath: string) => {
...
+
+    for (const entry of response) {
+      const fullName: string = entry.name
+      const nameSplit = fullName.split('/')
+      const revisionName = nameSplit[nameSplit.length - 1]
+      revisions.push(revisionName)
+    }
+  }
+
+  return revisions.slice(0, MAX_REVISIONS)
```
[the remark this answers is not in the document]
nhulston: Agreed, that will be more efficient

src/commands/cloud-run/flare.ts
duncanista: Nice, thanks!

src/commands/cloud-run/flare.ts
```diff
@@ -501,3 +537,133 @@ export const saveLogsFile = (logs: CloudRunLog[], filePath: string) => {
...
+ * @param location
+ * @param project
+ * @returns a string array of recent revisions and their deployment timestamp
+ */
+export const getRecentRevisions = async (service: string, location: string, project: string) => {
+  const client = new RevisionsClient()
+  const request = {
+    parent: client.servicePath(project, location, service),
+  }
+  const revisions = (await client.listRevisions(request))[0]
```
duncanista, line 553: Previously you would do 

```
(await client.listRevisions(request)) as any
```
But now it's the first element? I'm confused what's happening here.

duncanista: Is it guaranteed that the first element will always exist?

nhulston: Yes, the `listRevisions()` method returns a tuple:
```
Promise<[
        protos.google.cloud.run.v2.IRevision[],
        protos.google.cloud.run.v2.IListRevisionsRequest | null,
        protos.google.cloud.run.v2.IListRevisionsResponse
    ]
```

I am using `[0]` to get the first element, the array of revisions

nhulston: I will add a comment explaining this ^^

duncanista (approved):

PR #1024, opened 2023-08-07
[yarn] Fix dev env

yoannmoinet:
### What and why?

There was an issue when setting up my dev env for datadog-ci:

yarn `3.4.1` isn't downloadable
image

### How?

Instead, we should download latest version (`1.22.19`) which will use the correct release (`3.4.1`) that we embed in the repo at `.yarn/releases/yarn-3.4.1.cjs`.

### QA

- [x] `yarn -v` still outputs the correct version without being stopped by volta.

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,9 +10,9 @@
   },
   "volta": {
     "node": "14.20.1",
-    "yarn": "3.4.1"
+    "yarn": "1.22.19"
   },
-  "packageManager": "yarn@3.4.1",
+  "packageManager": "yarn@1.22.19",
   "files": [
     "dist/**/*",
     "README",


yoannmoinet:
@Drarig29 `1.0.8`

🤦 I see that `1.1.1` has:

> Better detection of executables provided by yarn (#1388, #1393)

Still, it doesn't really make a difference in the end, as we embed the release in the repo.
But this way we also support older versions of `volta`.

I'm fine closing this if you think it doesn't matter.

Drarig29:
Now that we are familiar with this possible issue, I think we can just answer anybody with the same issue to use the version `>=1.1.1` of `volta`. The tooling was broken, now it's fixed 😁 

Thanks for the PR and sorry for the disruption, but I think
https://github.com/DataDog/datadog-ci/blob/a54b1011f61bc52d7abd4c387a254909ba25bc69/package.json#L11-L15
doesn't make much sense.

The version that we're actually using is `3.4.1` right now, so having this in `package.json` would be kind of a lie.

PR #1030, opened 2023-08-09
Provide `version` as a shared and exported constant

Drarig29:
### What and why?

This PR replaces all `require('../path/to/package.json')` to get the package's version by a shared `version` constant.

It also exports that constant for lib users:
```ts
import {version} from '@datadog/datadog-ci`
```

This also fixes a bunch of warnings, and makes all the commands more consistent.

### How?

- Create a new `src/helpers/version.ts` module, which only exports a `version` constant.
- I chose not to change the namings in the commands. They sometimes redeclare the variable globally or as a class field.
- I chose to name it `version` **and not `cliVersion`** because for lib users, what we export is not a CLI, but a "datadog-ci" library.

### Review checklist

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -3,6 +3,8 @@ import fs from 'fs'
 import {Cli} from 'clipanion'
 import {CommandClass} from 'clipanion/lib/advanced/Command'
 
+import {version} from './helpers/version'
+
 const BETA_COMMANDS = ['gate']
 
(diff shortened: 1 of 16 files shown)


duncanista (approved):
LGTM from the Serverless side!

xgouchet (approved):
Good for RUM Mobile

Letiste (approved):

PR #1033, opened 2023-08-11
[cloud-run][flare] Add optional `--start` and `--end` flags when getting logs

nhulston:
### What and why?

Add optional `--start` and `--end` flags to specify a time range when getting logs.
When included, we only query cloud-run logs that fall inside of that time range.
The flags must be numbers in milliseconds since Unix epoch.

### How?

1. Add the `--start` and `--end` flags
2. Move `validateStartEndFlags()` from `lambda/flare.ts` to `helpers/flare.ts`
3. Validate start/end flags using the `validateStartEndFlags()` method.
4. Update `getLogs()` to accept `startMillis` and `endMillis` parameters. We then update the query filter to only search for logs in that time range. 

### Review checklist


src/commands/cloud-run/flare.ts
```diff
@@ -462,21 +482,38 @@ export const summarizeConfig = (config: IService) => {
...
  * @returns array of logs as CloudRunLog interfaces
  */
-export const getLogs = async (logClient: Logging, serviceId: string, location: string, severityFilter?: string) => {
+export const getLogs = async (
+  logClient: Logging,
+  serviceId: string,
+  location: string,
+  startMillis: number | undefined,
+  endMillis: number | undefined,
+  severityFilter?: string
```
duncanista, line 496: Isn't this the same?
```suggestion
  startMillis?: number,
  endMillis?: number,
  severityFilter?: string
```

src/commands/cloud-run/flare.ts
```diff
@@ -462,21 +482,38 @@ export const summarizeConfig = (config: IService) => {
...
+  let endDate: string
+  if (startMillis && endMillis) {
+    startDate = new Date(startMillis).toISOString()
+    endDate = new Date(endMillis).toISOString()
+  } else {
+    const date = new Date()
+    date.setMinutes(date.getMinutes() - MAX_LOG_AGE_MINUTES)
+    startDate = date.toISOString()
+    endDate = new Date().toISOString() // Current time
+  }
```
duncanista, line 512: Can't we revert this so we don't use `else`?

duncanista (commented):
Left some comments

src/commands/cloud-run/flare.ts
nhulston: The only difference is the second option `startMillis?` doesn't require us the pass in any argument. What we currently have requires us to explicitly pass something in, even if what we pass in is `undefined`.
Since we always pass in `startMillis` and `endMillis` (and just pass in undefined when the flags aren't specified), I figured the current method header is better. 
WDYT?

duncanista (approved):
LGTM!

PR #1035, opened 2023-08-15
[lambda] fix handler not being set when running interactive mode

duncanista:
### What and why?

Handler for NodeJS and Python AWS Lambdas was not being set properly on interactive mode.
Due to not checking if we were running on such mode when deciding to update the handler.

### How?

Check either if layer version is set or we are in interactive mode.

### Review checklist

- [x] Unit tests
- [x] Tested instrumenting an AWS Lambda

### DIFF
--- a/src/commands/lambda/functions/instrument.ts
+++ b/src/commands/lambda/functions/instrument.ts
@@ -184,7 +184,7 @@ export const calculateUpdateRequest = async (
   }
 
   // Update Python Handler
-  if (runtimeType === RuntimeType.PYTHON && settings.layerVersion !== undefined) {
+  if (runtimeType === RuntimeType.PYTHON && (settings.layerVersion !== undefined || settings.interactive)) {
     const expectedHandler = PYTHON_HANDLER_LOCATION
     if (config.Handler !== expectedHandler) {
       needsUpdate = true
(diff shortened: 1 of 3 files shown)


src/commands/lambda/functions/instrument.ts
```diff
@@ -184,7 +184,7 @@ export const calculateUpdateRequest = async (
   }
 
   // Update Python Handler
-  if (runtimeType === RuntimeType.PYTHON && settings.layerVersion !== undefined) {
+  if (runtimeType === RuntimeType.PYTHON && (settings.layerVersion !== undefined || settings.interactive)) {
```
nhulston, line 187: Why were we checking that `settings.layerVersion !== undefined` before?

nhulston (approved):
I manually tested these changes and this does indeed fix the handler issue.

I had a question, but other than that LGTM!

src/commands/lambda/functions/instrument.ts
duncanista: Because sometimes, user might only want to instrument their AWS Lambda w/o changing their current layer version, or they simply don't want to add it.

When using the Layer, we need to change the handler, because that's how we are able to provide monitoring.

PR #1039, opened 2023-08-16
Fix error when importing `chalk` with named imports

Drarig29:
### What and why?

The TS upgrade changed something in the emitted JS, which removed some logic to save us when using named imports with `chalk`, which shouldn't be done. (remark from Sindre himself: https://github.com/sindresorhus/eslint-plugin-unicorn/issues/211#issuecomment-457173568)

```
Type Error: Cannot read properties of undefined (reading '_styler')
    at ./datadog-ci/node_modules/chalk/source/index.js:85:123
    at Object.fetchingFunctionsConfigSpinner (./datadog-ci/dist/commands/lambda/renderers/instrument-uninstrument-renderer.js:473:75)
    at InstrumentCommand.<anonymous> (./datadog-ci/dist/commands/lambda/instrument.js:245:56)
    at Generator.throw (<anonymous>)
    at rejected (./datadog-ci/dist/commands/lambda/instrument.js:29:65)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
```

### How?

- Add an eslint rule to forbid the named import of every chalk member
  - It's possible to forbid the `default` import, but not to _only allow_ `default` imports with that rule
- Fix the linting errors (use `import chalk from 'chalk'`, and update every usage with `chalk.<member>`)

### Review checklist

- [x] Feature or bugfix MUST have appropriate tests (~~unit, integration~~, **ESLint rule**)


nhulston (approved):
LGTM!

duncanista (approved):
Great catch @Drarig29 !
Thank you so much! 🙇🏽

PR #1040, opened 2023-08-16
[junit] Add support for junit reports without tests

rodrigo-roca:
### What and why?

It allows users to send junit xml reports where there are no testcases and instead of showing the misleading error `Neither <testsuites> nor <testsuite> are the root tag`. For example:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<testsuites tests="0" failures="0" disabled="0" errors="0" time="0.001" name="AllTests">
</testsuites>
```

### How?

By modifying the if statement to detect invalid xmls.

### Review checklist


anmarchenko (approved):

romainkomorndatadog (approved):
LGTM

PR #1042, opened 2023-08-18
[gate] If the command execution time is greater than the command timeout, bail to avoid retrying

liashenko:
### What and why?

If the command execution time is greater than the command timeout, bail to avoid retrying.

### How?

### Review checklist

### DIFF
--- a/src/commands/gate/evaluate.ts
+++ b/src/commands/gate/evaluate.ts
@@ -145,7 +145,7 @@ export class GateEvaluateCommand extends Command {
      * for the returned promise to be rejected. The retry will start immediately after
      * and will be handled by `retryRequest`.
      */
-    return retryRequest((attempt) => this.evaluateRulesWithWait(api, evaluateRequest, attempt), {
+    return retryRequest((bail, attempt) => this.evaluateRulesWithWait(api, evaluateRequest, attempt, bail), {
       onRetry: (e, attempt) => {
         // render retry message if error is not wait
         if (e.message !== 'wait') {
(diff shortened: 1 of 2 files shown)


Drarig29 (approved):
LGTM

src/commands/gate/evaluate.ts
```diff
@@ -169,20 +169,21 @@ export class GateEvaluateCommand extends Command {
...
+    attempt?: number,
+    bail?: (e: Error) => void
   ): Promise<AxiosResponse<EvaluationResponsePayload>> {
     const timePassed = new Date().getTime() - evaluateRequest.startTimeMs
     const remainingWait = this.timeoutInSeconds * 1000 - timePassed
 
     return new Promise((resolve, reject) => {
       if (remainingWait <= 0) {
-        reject(new Error('wait'))
+        bail?.(new Error('wait'))
```
juan-fernandez, line 186: do we want to bail out with `'wait'` here? There's nothing to wait for, right?

liashenko: It was more convenient, because we already process `wait` errors. I can change the bail to throw `new Error('timeout')`, and process `timeout` errors as well, wdyt?

juan-fernandez: but we don't want to process wait errors the same way, right?

liashenko: In`retryRequest` catch it's the same I think, It can happen if we exhaust retries and the response from the service is  still `wait`.

juan-fernandez (approved):

PR #1043, opened 2023-08-19
Add e2e tests and update protobuf definition

juli1:
### What

1. Update the protobuf definition
2. Add e2e tests

### Why?

1. We need to add the `CI_PIPELINE` as a source, which has been specified in the server protobuf recently
2. We want to make sure that at each commit, SBOM are being correctly processed.

### How?

1. Updating the protobuf definition and their TS definition
2. Adding a new e2e test in the GitHub actions

### DIFF
--- a/src/commands/sbom/upload.ts
+++ b/src/commands/sbom/upload.ts
@@ -25,7 +25,7 @@ const generatePayload = (payloadData: SbomPayloadData, service: string, tags: Sp
     entities: [
       SBOMEntity.create({
         id: service,
-        type: SBOMSourceType.UNSPECIFIED,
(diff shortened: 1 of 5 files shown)


Drarig29 (approved):

src/commands/sbom/upload.ts
```diff
@@ -70,6 +71,14 @@ export class UploadSbomCommand extends Command {
       return 1
     }
 
+    this.config.env = this.config.env || this.env
```
dastrong, line 74:
```suggestion
    this.config.env = this.env || this.config.env
```

I'd rather a passed in value override an env var. This behavior is consistent with the service on line 66 above

dastrong (approved):

PR #1049, opened 2023-08-22
[test] Fix flaky test for `validateStartEndFlags()`

Drarig29:
### What and why?

This unit test sometimes fail because `Math.min('9999999999999', Date.now())` always returns `Date.now()`, which may be different than a previous `Date.now()`.

### How?

- Use fake timers to get rid of the flakiness

### Review checklist

### DIFF
--- a/src/helpers/__tests__/flare.test.ts
+++ b/src/helpers/__tests__/flare.test.ts
@@ -236,7 +236,8 @@ describe('flare', () => {
       expect(res).not.toBeUndefined()
       const [start, end] = res
       expect(start).toBe(0)
-      expect(end).toEqual(now)
+      expect(end).toBeLessThan(9999999999999)
+      expect(end).toStrictEqual(now)
     })
   })
 })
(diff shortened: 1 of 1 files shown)


src/helpers/__tests__/flare.test.ts
```diff
@@ -228,7 +236,8 @@ describe('flare', () => {
       expect(res).not.toBeUndefined()
       const [start, end] = res
       expect(start).toBe(0)
-      expect(end).toEqual(now)
+      expect(end).toBeLessThan(9999999999999)
+      expect(end).toStrictEqual(now)
```
Drarig29, line 240: We could also remove line 234 (`const now = Date.now()`) and only assert that `end` is less than `9999999999999`. But returning `8888888888888` would make the assertion pass.

So I kept `now` and used fake timers instead. In real world, we don't rely on two calls of `Date.now()`.

Drarig29: cc. @nhulston

juan-fernandez (approved):

PR #1050, opened 2023-08-23
fix(deps): add missing typanion dependency

tusbar:
### What and why?

`serverless-plugin-datadog` fails with pnpm because of `@datadog/datadog-ci` missing this dependency.

### How?

Just add the missing dependency.

### Review checklist


Drarig29 (approved):
Thanks for your contribution!

PR #1051, opened 2023-08-24
[chore] Add `no-extraneous-dependencies` eslint rule

Drarig29:
### What and why?

Typanion was not listed in our dependencies (fixed by #1050). This PR adds an eslint rule so that it doesn't happen again.

Using extraneous (i.e. unlisted) dependencies makes some package managers (e.g. `pnpm`) fail to install the dependencies.

### How?

- Enable an eslint rule provided by `eslint-plugin-import`

### Review checklist

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -165,6 +165,7 @@ module.exports = {
     'guard-for-in': 'error',
     'id-denylist': 'error',
     'id-match': 'error',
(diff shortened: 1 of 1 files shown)


juan-fernandez (approved):

tusbar:
I was going to suggest this. :)

PR #1055, opened 2023-08-30
[CIAPP-7523] Mark last retry from datadog-ci gate

liashenko:
### What and why?
Mark last retry from datadog-ci gate to be able to send failed gate executions to EvP from the service.

We want to provide better visibility in Datadog in the following cases:
* one or more rule evaluations failed during the last request from datadog-ci
* datadog-ci timed out before finishing the evaluation

Currently Quality Gates service is not aware of internal datadog-ci retries, so we can’t send CIGATE events to EvP on the last request for the cases mentioned above.

More context in the [RFC](https://docs.google.com/document/d/1Cj5i5rs_S85AueP7m60gg4iWYVZ-Opr-SB2VpHzvEvE/edit)

### How?
When it's a last retry attempt or the remaining wait time is 0, add `isLastRetry:true` to the API request.

### Review checklist

### DIFF
--- a/src/commands/gate/evaluate.ts
+++ b/src/commands/gate/evaluate.ts
@@ -182,28 +182,39 @@ export class GateEvaluateCommand extends Command {
     const remainingWait = this.timeoutInSeconds * 1000 - timePassed
 
     return new Promise((resolve, reject) => {
-      if (remainingWait <= 0) {
-        bail?.(new Error('wait'))
-      } else {
-        api
-          .evaluateGateRules(evaluateRequest, this.context.stdout.write.bind(this.context.stdout))
-          .then((response) => {
-            if (response.data.data.attributes.status === 'wait') {
-              this.context.stdout.write(renderWaiting())
-              const waitTime = response.data.data.attributes.metadata?.wait_time_ms ?? 0
-              setTimeout(() => {
-                reject(new Error('wait'))
-              }, Math.min(remainingWait, waitTime))
-            } else {
-              resolve(response)
-            }
-          })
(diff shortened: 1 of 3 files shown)


src/commands/gate/evaluate.ts
```diff
@@ -182,28 +182,39 @@ export class GateEvaluateCommand extends Command {
...
-              resolve(response)
-            }
-          })
-          .catch((err) => {
-            setTimeout(() => {
-              reject(err)
-            }, Math.min(remainingWait, this.getDelay(attempt ?? 1)))
-          })
+      const request = {...evaluateRequest, options: {...evaluateRequest.options}}
+      if (remainingWait <= 0 || attempt === this.maxRetries + 1) {
```
romainkomorndatadog, line 186: Not that we should ever get to `attempt === this.maxRetries + 2`, but would this be more appropriate?
```
attempt > this.maxRetries
```

romainkomorndatadog (approved):
I think this looks good but I'm feeling a little bit shy about approving since I'm not all that familiar with this repo.

Maybe get a second opinion. :)

andrea-mosk (approved):
LGTM if you have tested it

src/commands/gate/evaluate.ts
liashenko: it shouldn't happen ever, if it happens datadog-ci retry is broken.

but if it does, we expect to get a single `lastRetry` in the service (because we'll send the event to EvP, and we want to have 1 event per gate execution), so I'd keep it as is if you don't mind

PR #1058, opened 2023-09-06
[utils] Format all git domains, not only github.com

AndrewChubatiuk:
### What and why?

Enabled formatting of all git remotes, not only github.com to fix datadog serverless plugin tags, while using in projects with repos that are not hosted on github

### How?

updated regular expression to replace fix remote urls with any domain name

### DIFF
--- a/src/helpers/__tests__/utils.test.ts
+++ b/src/helpers/__tests__/utils.test.ts
@@ -371,22 +371,22 @@ describe('utils', () => {
     })
   })
 
-  describe('filterAndFormatGithubRemote', () => {
+  describe('filterAndFormatGitRemote', () => {
     test('git remotes get formatted correctly', async () => {
-      expect(ciUtils.filterAndFormatGithubRemote('https://github.com/datadog/test.git')).toEqual(
+      expect(ciUtils.filterAndFormatGitRemote('https://github.com/datadog/test.git')).toEqual(
         'github.com/datadog/test.git'
       )
-      expect(ciUtils.filterAndFormatGithubRemote('git@github.com:datadog/test.git')).toEqual(
+      expect(ciUtils.filterAndFormatGitRemote('git@github.com:datadog/test.git')).toEqual(
         'github.com/datadog/test.git'
       )
-      expect(ciUtils.filterAndFormatGithubRemote('github.com/datadog/test.git')).toEqual('github.com/datadog/test.git')
-      expect(ciUtils.filterAndFormatGithubRemote('https://git.some.domain.com:8080/datadog/test.git')).toEqual(
+      expect(ciUtils.filterAndFormatGitRemote('github.com/datadog/test.git')).toEqual('github.com/datadog/test.git')
+      expect(ciUtils.filterAndFormatGitRemote('https://git.some.domain.com:8080/datadog/test.git')).toEqual(
         'git.some.domain.com/datadog/test.git'
       )
-      expect(ciUtils.filterAndFormatGithubRemote('git@test.domain.com:datadog/test.git')).toEqual(
+      expect(ciUtils.filterAndFormatGitRemote('git@test.domain.com:datadog/test.git')).toEqual(
         'test.domain.com/datadog/test.git'
       )
-      expect(ciUtils.filterAndFormatGithubRemote('example.domain.com/datadog/test.git')).toEqual('example.domain.com/datadog/test.git')
+      expect(ciUtils.filterAndFormatGitRemote('example.domain.com/datadog/test.git')).toEqual('example.domain.com/datadog/test.git')
     })
   })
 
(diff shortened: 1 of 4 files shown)


Drarig29 (changes requested):
Hi, thanks for your contribution!

Can you test case(s) for what you changed, and ensure the unit tests are passing?

You can find the existing tests here:

https://github.com/DataDog/datadog-ci/blob/19a9e0c7ef5302b905bd9ca41e19538b601ceeea/src/helpers/__tests__/utils.test.ts#L374-L384

src/helpers/__tests__/utils.test.ts
```diff
@@ -380,6 +380,13 @@ describe('utils', () => {
         'github.com/datadog/test.git'
       )
       expect(ciUtils.filterAndFormatGithubRemote('github.com/datadog/test.git')).toEqual('github.com/datadog/test.git')
+      expect(ciUtils.filterAndFormatGithubRemote('https://git.some.domain.com:8080/datadog/test.git')).toEqual(
+        'git.some.domain.com/datadog/test.git'
+      )
+      expect(ciUtils.filterAndFormatGithubRemote('git@test.domain.com:datadog/test.git')).toEqual(
+        'test.domain.com/datadog/test.git'
+      )
+      expect(ciUtils.filterAndFormatGithubRemote('todo.domain.com/datadog/test.git')).toEqual('todo.domain.com/datadog/test.git')
```
Drarig29, line 389:
```suggestion
      expect(ciUtils.filterAndFormatGithubRemote('example.domain.com/datadog/test.git')).toEqual('example.domain.com/datadog/test.git')
```

src/helpers/__tests__/utils.test.ts
```diff
@@ -380,6 +380,13 @@ describe('utils', () => {
         'github.com/datadog/test.git'
       )
       expect(ciUtils.filterAndFormatGithubRemote('github.com/datadog/test.git')).toEqual('github.com/datadog/test.git')
+      expect(ciUtils.filterAndFormatGithubRemote('https://git.some.domain.com:8080/datadog/test.git')).toEqual(
```
Drarig29, line 383: Why would you remove the port? Do you have an example of scenario?

juan-fernandez (approved):

src/helpers/__tests__/utils.test.ts
```diff
@@ -380,6 +380,13 @@ describe('utils', () => {
...
+        'git.some.domain.com/datadog/test.git'
+      )
+      expect(ciUtils.filterAndFormatGithubRemote('git@test.domain.com:datadog/test.git')).toEqual(
+        'test.domain.com/datadog/test.git'
+      )
+      expect(ciUtils.filterAndFormatGithubRemote('example.domain.com/datadog/test.git')).toEqual('example.domain.com/datadog/test.git')
```
duncanista, line 389: If we are formatting all git domains, shouldn't we also change the function name (and jsdocs) to be something different than `filterAndFormatGithubRemote`? Maybe just `filterAndFormatGitRemote`? Just a nit, but would be better to keep consistency.

duncanista (approved):
LGTM – left a comment!

src/helpers/__tests__/utils.test.ts
Drarig29: @AndrewChubatiuk good point!

Can you add the original `filterAndFormatGithubRemote()` function back (with its old implementation and unit tests) and mark it as deprecated like [that function](https://github.com/andrewchubatiuk/datadog-ci/blob/0e2e9fac8e0c20502245b1cb2898e6f763d7bae1/src/helpers/utils.ts#L100)?

Thanks in advance, and sorry for the delay!

Drarig29:
@AndrewChubatiuk I triggered the CI for your latest commit, and the lint phase fails. Running `yarn format` should fix it!

src/helpers/__tests__/utils.test.ts
Drarig29: Nice, indeed the unit test did not change, so the new implementation did not change in behavior for the old supported cases.

Drarig29 (approved):
LGTM, thanks for the contribution! 🎉

rpelliard:
Hello @AndrewChubatiuk  !

> Enabled formatting of all git remotes, not only github.com to fix datadog serverless plugin tags, while using in projects with repos that are not hosted on github

Out of curiosity, are you encountering an issue with the current version ? For context, this normalization for GitHub URLs was introduced to circumvent some limitations around some old tracer versions that don't support the URL scheme in tags, but it shouldn't be necessary in most cases now.

For GitHub repositories, the `git@` and `https://` repository URLs are predictable / interchangable, so we don't lose any information by stripping the scheme. However, this is not the case for all Git repo hosting providers, so I'd like to make sure we fully understand your issue before merging this change - Thanks !

rpelliard:
@AndrewChubatiuk What version of [serverless-plugin-datadog](https://github.com/DataDog/serverless-plugin-datadog), and which language are you using ? The latest one should only bundle versions of the tracing libraries that support the double columns in the tags

PR #1062, opened 2023-09-11
[lambda] add asm/app-sec option to lambda commands

thedavl:
### What and why?

Add the `--appsec` flag to the lambda CI instrument command, setting the `DD_SERVERLESS_APPSEC_ENABLED` to enable ASM

Update the uninstrument command to remove this env var.

Manually tested

### How?

### Review checklist


src/commands/lambda/instrument.ts
```diff
@@ -52,6 +52,7 @@ export class InstrumentCommand extends Command {
   })
 
   private apmFlushDeadline = Option.String('--apm-flush-deadline')
+  private appSecEnabled = Option.Boolean('--app-sec,--appSec', false)
```
duncanista, line 55: do we really need `--appSec`? I think `--app-sec` should be enough

duncanista: remember to add this option for `uninstrument` too

duncanista (commented):
left some comments!

src/commands/lambda/instrument.ts
thedavl: seems like most of the other commands offer a camelCase option (aside from dryRun, which we should probably add for consistency sake). 

Yup! I've modified uninstrument to remove the env var too.

duncanista: That camel case are deprecated flags which we cannot remove in order to not cause a breaking change with customers. We moved from camel case to slug case a while ago. That's why you will see som flags not having camel case.

duncanista: For `uninstrument`, I see that it removes the environment variable, that is correct. Nonetheless, you are not adding the `Option.Boolean`, which is just required in order for customers to be able to use `instrument/uninstrument` interchangeably. If I did `lambda uninstrument -f ... -r ... --app-sec` it would throw an error, because the flag is not set as an option.

Julio-Guerra (changes requested):
Enabling ASM requires two env vars ([doc](https://docs.datadoghq.com/security/application_security/enabling/serverless/?tab=serverlessframework)):
1. `AWS_LAMBDA_EXEC_WRAPPER=/opt/datadog_wrapper`
2. `DD_SERVERLESS_APPSEC_ENABLED=true`

The first was forgotten here. This was released last month in https://github.com/DataDog/datadog-lambda-extension/releases/tag/v47

src/commands/lambda/instrument.ts
Julio-Guerra: Our keyword really is just `appsec`, so why separate both?  ^^
I find --appsec way cleaner and aligned with the way we call ASM in all the Datadog libraries today.

thedavl: ah makes sense, thanks for the context. I've switched the flag over to be `--appsec`, and added it to the uninstrument command.

src/commands/lambda/functions/instrument.ts
```diff
@@ -283,6 +285,11 @@ export const calculateUpdateRequest = async (
     }
   }
 
+  // Enable ASM
+  if (settings['appsecEnabled'] === true) {
+    newEnvVars[AWS_LAMBDA_EXEC_WRAPPER_VAR] = AWS_LAMBDA_EXEC_WRAPPER
```
duncanista, line 290: I think this is onlu needed for .NET, Go, and Java. We should update this.

duncanista (commented):
left a question

src/commands/lambda/functions/instrument.ts
Julio-Guerra: No, the appsec setup for AWS lambda is no longer conditional and you need both env vars no matter the language.

duncanista (approved):

duncanista (approved):

src/commands/lambda/functions/instrument.ts
Julio-Guerra: Yes this is why indeed.

src/commands/lambda/functions/uninstrument.ts
```diff
@@ -149,8 +150,12 @@ export const calculateUpdateRequest = (config: LFunctionConfiguration, runtime:
     }
   }
 
-  // Remove AWS_LAMBDA_EXEC_WRAPPER for .NET and Java
-  if (runtimeType === RuntimeType.DOTNET || runtimeType === RuntimeType.JAVA) {
+  // Remove AWS_LAMBDA_EXEC_WRAPPER for .NET and Java or if ASM is enabled
+  if (
+    runtimeType === RuntimeType.DOTNET ||
+    runtimeType === RuntimeType.JAVA ||
```
Julio-Guerra, line 156: IIUC the overall implementation, this is no longer needed, because the ASM setup has no conditions anymore:
```suggestion
```

thedavl: yep, Jordan is correct. The only 3 cases the wrapper will be added is if the runtime is dotnet, java, or ASM is enabled

Julio-Guerra: Ok, my bad, sorry for the noise.

Julio-Guerra (approved):

src/commands/lambda/functions/uninstrument.ts
duncanista: don't be sorry! good question for clarification!

PR #1076, opened 2023-10-02
[git-upload] Improve unshallow process

juan-fernandez:
### What and why?

There are situations where the unshallow command fails: 
* If the local HEAD is a commit that has not been pushed to the remote
* If the CI is working on a detached HEAD or branch tracking hasn’t been set up

We want to try to work in those situations.

### How?

Adding fallbacks if either of those cases happen.

### Review checklist

### DIFF
--- a/src/commands/git-metadata/gitdb.ts
+++ b/src/commands/git-metadata/gitdb.ts
@@ -127,20 +127,21 @@ const unshallowRepositoryWhenNeeded = async (log: Logger, git: simpleGit.SimpleG
     return
   }
   log.info('[unshallow] Git repository is a shallow clone, unshallowing it...')
-  const headCmdPromise = git.revparse('HEAD')
-  const remoteNameCmdPromise = getDefaultRemoteName(git)
+  const headCommit = await git.revparse('HEAD')
+  const remoteName = (await getDefaultRemoteName(git)) ?? 'origin'
 
   const baseCommandLogLine = `[unshallow] Running git fetch --shallow-since="${MAX_HISTORY.oldestCommits}" --update-shallow --filter=blob:none --recurse-submodules=no`
 
   log.info(`${baseCommandLogLine} $(git config --default origin --get clone.defaultRemoteName) $(git rev-parse HEAD)`)
+
   try {
     await git.fetch([
       `--shallow-since="${MAX_HISTORY.oldestCommits}"`,
       '--update-shallow',
       '--filter=blob:none',
       '--recurse-submodules=no',
-      (await remoteNameCmdPromise) ?? 'origin',
-      await headCmdPromise,
+      remoteName,
+      headCommit,
     ])
   } catch (err) {
     // If the local HEAD is a commit that has not been pushed to the remote, the above command will fail.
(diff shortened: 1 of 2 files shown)


src/commands/git-metadata/gitdb.ts
```diff
@@ -127,19 +127,51 @@ const unshallowRepositoryWhenNeeded = async (log: Logger, git: simpleGit.SimpleG
...
-  await git.fetch([
-    `--shallow-since="${MAX_HISTORY.oldestCommits}"`,
-    '--update-shallow',
-    '--filter=blob:none',
-    '--recurse-submodules=no',
-    (await remoteNameCmdPromise) ?? 'origin',
-    await headCmdPromise,
-  ])
+  const headCommit = await git.revparse('HEAD')
+  const remoteName = (await getDefaultRemoteName(git)) ?? 'origin'
```
edrevo, line 131: Isn't this change making the code slower?

edrevo (approved):

src/commands/git-metadata/gitdb.ts
Drarig29: Why would it? It used to be [below before](https://github.com/DataDog/datadog-ci/pull/1076/files/3e6f0dc005f20f83464893a760af1fc86d81526a#diff-65f586ed98beaf8953587b0bfc797f3e538d3c61f0079d7cec4b25edae3c41f4L140). The only difference I would see is that if it fails, it won't be caught in the below try/catch as before, maybe it was intended?

I don't know why we had `remoteNameCmdPromise` before though 🤔

edrevo: I'm not sure how `await` works in Typescript, but my current reading of the code if that we don't get the default remote name until we are gone getting the HEAD's revparse, whereas before they were run in parallel.

Drarig29: Oh, well I think it behaves (almost) exactly the same as before here:

- Before:
  - start both promises, so yes the 2nd is started without awaiting the 1st one
  - but then, you were still awaiting [the 1st, **then the 2nd**](https://github.com/DataDog/datadog-ci/pull/1076/files/3e6f0dc005f20f83464893a760af1fc86d81526a#diff-65f586ed98beaf8953587b0bfc797f3e538d3c61f0079d7cec4b25edae3c41f4L140-L141)
- After:
  - we directly await the 1st, **then the 2nd**

If we really wanted to make them parallel, we would have to do something like this:
https://github.com/DataDog/datadog-ci/blob/3313ed0be792b0c35e3a1af070d9ca6b38acfcf7/src/helpers/git/format-git-span-data.ts#L22-L35

@juan-fernandez do you want to do that?

juan-fernandez: >  but my current reading of the code if that we don't get the default remote name until we are gone getting the HEAD's revparse, whereas before they were run in parallel.

this was not the case. Check with this simple example: 
```javascript
const log = (msg) => {
  console.log(`${new Date().toLocaleTimeString()}: ${msg}`)
}

const getValue = (val) => 
  new Promise(resolve => {
    setTimeout(() => {
      log(val)
      resolve(val)
    }, 3000)
  })


async function run () {
  log('start')
  const resolvedValues = [await getValue(1), await getValue(2)]
  log('resolved values', resolvedValues)
}

run()
```
which outputs something like 
```
4:39:52 PM: start
4:39:55 PM: 1
4:39:58 PM: 2
4:39:58 PM: resolved values
```

the first promise is awaited before the second one, so there is no parallelism. 

We could do some `Promise.all` logic but I don't think it's worth the extra complexity for the few ms we'll save, considering the git fetch will take significantly more

juan-fernandez: I was wrong. The two child processes were spawned when calling 
```
  const headCmdPromise = git.revparse('HEAD')
  const remoteNameCmdPromise = getDefaultRemoteName(git)
```
so it was indeed running in parallel, which we lost in this PR. My example is not valid. I still think it's not worth the complexity, but I can do a follow up PR

Drarig29: Oh interesting, yes since actual processes are created, then it's real parallelization (not concurrency like we are used to with Node.js).

PR #1077, opened 2023-10-05
Add e2e check for junit xml test data

juan-fernandez:
### What and why?

Add e2e test for JUnit XML report upload: we now not only upload data and check the return status, but query the datadog API to check that test data has actually arrived to the platform. 

### How?

Add a `bin/check-junit-upload.js` script to query the datadog API.


Drarig29 (approved):
Nice!

PR #1081, opened 2023-10-06
Update CI spec tests

albertvaka:
### What and why?

* Update the fixtures using the latest spec.
* Updated the code that reads the bitbucket env vars to make the tests pass with the new fixtures.


Drarig29 (approved):

src/helpers/__tests__/ci-env/appveyor.json
```diff
@@ -6,7 +6,7 @@
       "APPVEYOR_BUILD_ID": "appveyor-build-id",
       "APPVEYOR_BUILD_NUMBER": "appveyor-pipeline-number",
       "APPVEYOR_REPO_BRANCH": "master",
-      "APPVEYOR_REPO_COMMIT": "appveyor-repo-commit",
+      "APPVEYOR_REPO_COMMIT": "b9f0fb3fdbb94c9d24b2c75b49663122a529e123",
```
romainkomorndatadog, line 9: Are we actually doing anything with hashes to make sure they're valid? Otherwise, isn't it just string comparison?

romainkomorndatadog (approved):

src/helpers/__tests__/ci-env/appveyor.json
albertvaka: It's just string comparison. Eric changed them to actually look like hashes in the files that we use to generate these fixtures, but it's mostly an aesthetic change at this point.

PR #1084, opened 2023-10-11
Revert "[SYNTH-10966] Support blocking IP addresses in the tunnel"

jakepruitt:
Reverts DataDog/datadog-ci#1068, which introduced an issue with the tunnel that is incorrectly blocking URLs that should be allowed


lefebvree (approved):

Drarig29 (approved):
LGTM

PR #1090, opened 2023-10-21
adapt SBOM payload

juli1:
### What and why?

We are adapting the `sbom` command to our new API that handles SBOM.

The API keeps parsing SBOM. However, instead of sending the file to an intake, it parses the file, extract relevant information for each dependency (name, version, language and licenses) and send the data to the API. The API then validates the data, sending a `200` if the result is successful.

### How?

We are changing the implementation from sending the SBOM file to an intake to parsing the SBOM file, extract relevant information and sends only the important information.

Note that this is a very preliminary support, we will support more SBOM in the future.

### Review checklist


dastrong (approved):

src/commands/sbom/__tests__/payload.test.ts
```diff
@@ -1,68 +1,62 @@
...
+      env: undefined,
+      envVarTags: undefined,
+    }
+    const tags = await getSpanTags(config, [])
+
+    const payload = generatePayload(sbomContent, tags)
+    expect(payload).not.toBeNull()
+    expect(payload?.id).not.toBeNull()
+
+    expect(payload?.commit.sha).not.toBeNull()
```
Drarig29, line 51: `not.toBeNull()` doesn't say much about what those properties **should** contain. You could use `expect.any(String)`.

And instead of having multiple of those assertions, I think you could have something like:

```js
expect(payload).toStrictEqual({
  id: expect.any(String),
  commit: {
    author_name: expect.any(String),
    ...
  },
  ...
})
```

src/commands/sbom/language.ts
```diff
@@ -0,0 +1,23 @@
+import {DependencyLanguage} from './types'
+
+// Attempt to find the language from a SBOM component. For now, we get the source either from
+// the bom-ref or the purl property of the SBOM.
+export const getLanguageFromComponent = (component: any): DependencyLanguage | undefined => {
+  const componentName = component['name']
+
+  if (component['bom-ref']) {
+    if (component['bom-ref'].indexOf('pkg:npm') !== -1) {
```
Drarig29, line 9:
```suggestion
    if (component['bom-ref'].includes('pkg:npm')) {
```
or even:
```suggestion
    if (component['bom-ref'].startsWith('pkg:npm')) {
```

src/commands/sbom/license.ts
```diff
@@ -0,0 +1,93 @@
...
+// such as "MIT OR Apache-2.0". In this case, we return all the licenses in this condition.
+export const getLicensesFromString = (s: string): DependencyLicense[] => {
+  if (!s) {
+    return []
+  }
+  const licenses: DependencyLicense[] = []
+
+  if (s.indexOf('OR') !== -1) {
+    for (const lic of s.toLowerCase().split(' or ')) {
+      const l = getLicenseFromString(lic.replace(' ', ''))
```
Drarig29, line 43: Why are you removing spaces? I don't see any test cases about that.

If it's to remove possible duplicate spaces like `aaa OR    bbb`, then you can use `.trim()`.

But if think there may be something like `Apache 2.0`, then removing spaces won't work with your switch case in `getLicenseFromString()`

src/commands/sbom/license.ts
```diff
@@ -0,0 +1,93 @@
...
+    const lic = getLicenseFromString(s)
+    if (lic) {
+      licenses.push(lic)
+    }
+  }
+
+  return licenses
+}
+
+// Get all the licenses of this component We extract the "licenses" element from the SBOM component.
```
Drarig29, line 58:
```suggestion
// Get all the licenses of this component. We extract the "licenses" element from the SBOM component.
```

src/commands/sbom/license.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  const componentName = component['name']
+  const licenses: DependencyLicense[] = []
+
+  // Get the "licenses" attribute of the SBOM component.
+  if (component['licenses']) {
+    for (const license of component['licenses']) {
+      for (const el of elementsForLicense) {
+        // Handle "license": [ {"license": {"id": <license>}}]
+        if (license['license'] && license['license'][el]) {
```
Drarig29, line 72:
```suggestion
        if (license['license']?.[el]) {
```

src/commands/sbom/license.ts
```diff
@@ -0,0 +1,93 @@
...
+  const elementsForLicense = ['id', 'name']
+
+  const componentName = component['name']
+  const licenses: DependencyLicense[] = []
+
+  // Get the "licenses" attribute of the SBOM component.
+  if (component['licenses']) {
+    for (const license of component['licenses']) {
+      for (const el of elementsForLicense) {
+        // Handle "license": [ {"license": {"id": <license>}}]
```
Drarig29, line 71:
```suggestion
        // Handle "license": [ {"license": {"id": <license>}} ]
```

src/commands/sbom/license.ts
```diff
@@ -0,0 +1,93 @@
...
+        if (license['license'] && license['license'][el]) {
+          for (const l of getLicensesFromString(license['license'][el])) {
+            licenses.push(l)
+          }
+        }
+
+        // Handle "license": [ {"expression": "MIT"}]
```
Drarig29, line 78:
```suggestion
        // Handle "license": [ {"expression": "MIT"} ]
```

src/commands/sbom/__tests__/payload.test.ts
```diff
@@ -1,68 +1,62 @@
...
+    expect(payload?.commit.author_name).not.toBeNull()
+    expect(payload?.commit.author_email).not.toBeNull()
+    expect(payload?.commit.branch).not.toBeNull()
+    expect(payload?.repository.url).toContain('github.com')
+    expect(payload?.repository.url).toContain('DataDog/datadog-ci')
```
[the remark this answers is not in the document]
juli1: Because it's not equal and will depends on how you clone the `datadog-ci` repository. If you clone via `git` ot `https`, the result will be different.

src/commands/sbom/__tests__/payload.test.ts
juli1: I did not know this, thank you! Updated the tests!

src/commands/sbom/license.ts
Drarig29: > There is never a space in a license

Ok 🙂 And since `.split(' or ')` includes the spaces in the delimiter, your values are `['MIT', 'Apache-2.0']` so I think you should be able to remove `.replace(' ', '')`

src/commands/sbom/license.ts
```diff
@@ -0,0 +1,96 @@
...
+
+// Get all the licenses from a string. Sometimes, there are two licenses in one string
+// such as "MIT OR Apache-2.0". In this case, we return all the licenses in this condition.
+export const getLicensesFromString = (s: string): DependencyLicense[] => {
+  if (!s) {
+    return []
+  }
+  const licenses: DependencyLicense[] = []
+
+  if (s.indexOf('OR') !== -1) {
```
Drarig29, line 44: You can use `includes()` here as well:
```suggestion
  if (s.includes('OR')) {
```

src/commands/sbom/language.ts
```diff
@@ -0,0 +1,23 @@
...
+  if (component['bom-ref']) {
+    if (component['bom-ref'].includes('pkg:npm')) {
+      return DependencyLanguage.NPM
+    }
+    if (component['purl'].indexOf('pkg:composer') !== -1) {
+      return DependencyLanguage.PHP
+    }
+    if (component['purl'].indexOf('pkg:cargo') !== -1) {
+      return DependencyLanguage.RUST
+    }
```
Drarig29, line 17: Sorry for not saying it, but I also meant to change those other `indexOf()`
```suggestion
    if (component['purl'].includes('pkg:composer')) {
      return DependencyLanguage.PHP
    }
    if (component['purl'].includes('pkg:cargo')) {
      return DependencyLanguage.RUST
    }
```

src/commands/sbom/license.ts
juli1: The problem with `split` is that if I have `MIT OR  Apache-2.0`, it will not work. Some files have this, which is also why I remove the spaces

src/commands/sbom/language.ts
juli1: done!

src/commands/sbom/README.md
```diff
@@ -25,24 +24,11 @@ The following environment variables must be defined:
...
-brew install protobuf
-```
-
-Then, install `ts-proto` as a development dependency.
-You can find information about this binary [in the ts-proto repo](https://github.com/stephenh/ts-proto).
 
+## Development
 
-Finally, generate the TypeScript definition of the protobuf format:
+When developing software, you can try with the following command
```
Drarig29, line 30:
```suggestion
When developing software, you can try with the following command:
```

hestonhoffman (approved):

PR #1091, opened 2023-10-23
[synthetics] Improve "Finding files" message

Drarig29:
### What and why?

Before this PR, when using `datadog-ci synthetics run-tests -f ~/tests.synthetics.json`, the terminal used to interpret `~` and pass `$HOME/tests.synthetics.json` to datadog-ci, which then printed:

```
Finding files in /Users/john.doe/path/to/current/directory/Users/john.doe/tests.synthetics.json
```

if the command was run from `~/path/to/current/directory`.

### How?

"Finding files in" makes it sound like the passed glob is a directory (i.e. something which **contains** files), whereas what we're actually doing is to find files which match a specific glob, which may also be the real path of a single file.

So I changed "in" to "matching". And if the glob is an absolute path, we just show that absolute path.

### Review checklist

### DIFF
--- a/src/commands/synthetics/utils.ts
+++ b/src/commands/synthetics/utils.ts
@@ -203,7 +203,7 @@ export const getResultOutcome = (result: Result): ResultOutcome => {
 }
 
 export const getSuites = async (GLOB: string, reporter: MainReporter): Promise<Suite[]> => {
-  reporter.log(`Finding files matching ${getGlobForDisplay(GLOB)}\n`)
+  reporter.log(`Finding files matching ${path.resolve(process.cwd(), GLOB)}\n`)
 
   const files: string[] = await promisify(glob)(GLOB)
   if (files.length) {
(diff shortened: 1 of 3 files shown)


src/commands/synthetics/utils.ts
```diff
@@ -203,7 +203,8 @@ export const getResultOutcome = (result: Result): ResultOutcome => {
 }
 
 export const getSuites = async (GLOB: string, reporter: MainReporter): Promise<Suite[]> => {
-  reporter.log(`Finding files in ${path.join(process.cwd(), GLOB)}\n`)
+  reporter.log(`Finding files matching ${getGlobForDisplay(GLOB)}\n`)
```
etnbrd, line 206: For simplicity, what do you think about using `resolve`?
https://nodejs.org/api/path.html#pathresolvepaths

```suggestion
  reporter.log(`Finding files matching ${path.resolve(process.cwd(), GLOB)}\n`)
```

Drarig29: Thank you, I was so convinced that a glob like `{,!(node_modules)/**/}*.synthetics.json` would not be supported that I didn't try it.

Indeed, the following works:
```
> require('path').resolve('.', '{,!(node_modules)/**/}*.synthetics.json')
'/Users/corentin.girard/go/src/github.com/DataDog/synthetics-worker.git/prod/{,!(node_modules)/**/}*.synthetics.json'
> require('path').resolve('.', '/example/*.synthetics.json')
'/example/*.synthetics.json'
```

PR #1093, opened 2023-10-24
SBOM: support ruby, go and maven/java

juli1:
### What and why?

We want to make sure we can parse SBOM for Ruby, Go and Maven/Java

### How?

We are adapting the language and license detection. We are testing by having an SBOM generated for each of those languages.

### Review checklist

### DIFF
--- a/src/commands/sbom/language.ts
+++ b/src/commands/sbom/language.ts
@@ -19,7 +19,7 @@ export const getLanguageFromComponent = (component: any): DependencyLanguage | u
       return DependencyLanguage.RUBY
     }
     if (component['purl'].includes('pkg:maven')) {
-      return DependencyLanguage.RUBY
+      return DependencyLanguage.JVM
     }
     if (component['purl'].includes('pkg:golang')) {
       return DependencyLanguage.GO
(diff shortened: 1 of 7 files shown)


src/commands/sbom/language.ts
```diff
@@ -15,6 +15,15 @@ export const getLanguageFromComponent = (component: any): DependencyLanguage | u
     if (component['purl'].includes('pkg:cargo')) {
       return DependencyLanguage.RUST
     }
+    if (component['purl'].includes('pkg:gem')) {
+      return DependencyLanguage.RUBY
+    }
+    if (component['purl'].includes('pkg:maven')) {
+      return DependencyLanguage.RUBY
+    }
```
dastrong, line 23: Is `pkg:maven` strictly equal to Ruby in our tool? Isn't Maven used for different language like Java or Scala also?

dastrong (approved):
Having Java in the PR title/description seems misleading here, but maybe once the comment I left is answered that will clear it up. It's not blocking, so approved though.

src/commands/sbom/language.ts
juli1: Thanks for catching this. This should be JVM.

PR #1097, opened 2023-10-28
[synthetics] Show test runs as they arrive

Drarig29:
### What and why?

This PR changes how often the `run-tests` command displays test runs. Before this PR, we were only showing them once, at the very end.

With this PR, datadog-ci will show each test run as soon as we detect it's failed or passed.

### How?

Before:
- `getBatch()` every 5 seconds
- Use `emittedResultIndexes` to keep track of the in-batch results we emitted with `Reporter.resultReceived()` (minimal information about the result)
- Once the batch is not `in_progress` anymore, use `pollResults()` to get the full information of all results, and display their information with `renderResults()`, which calls `Reporter.resultEnd()` multiple times

After:
- `getBatch()` and `pollResults()` every 5 seconds
- Use `emittedResultIndexes` to keep track of the in-batch results we emitted with `Reporter.resultReceived()` (minimal information about the result) **and** `Reporter.resultEnd()` (full information about the result)
- Once the batch is not `in_progress` anymore, `pollResults()` isn't used an extra time, and `renderResults()` solely display the summary

The output looks exactly the same as before when all results end roughly at the same time:
- [Before](https://github.com/DataDog/datadog-ci/actions/runs/6672701863/job/18137226335#step:6:17) (github action)
- [After](https://github.com/DataDog/datadog-ci/actions/runs/6677903400/job/18148423474#step:6:17) (github action)

Here is how it looks when results arrive later, in a CI environment or not.

When run locally:

https://github.com/DataDog/datadog-ci/assets/9317502/a8aa437a-ea22-43f6-9749-2af14c52db88

When run in a CI environment:

https://github.com/DataDog/datadog-ci/assets/9317502/b3c01cb8-12a6-422e-83eb-53b1c4598eee

### Review checklist

### DIFF
--- a/src/commands/synthetics/__tests__/utils.test.ts
+++ b/src/commands/synthetics/__tests__/utils.test.ts
@@ -764,11 +764,11 @@ describe('utils', () => {
     } = {}) => {
       const getBatchMock = jest
         .spyOn(api, 'getBatch')
-        .mockImplementation(getBatchImplementation || (async () => JSON.parse(JSON.stringify(batch))))
+        .mockImplementation(getBatchImplementation || (async () => deepExtend({}, batch)))
 
       const pollResultsMock = jest
         .spyOn(api, 'pollResults')
-        .mockImplementation(pollResultsImplementation || (async () => JSON.parse(JSON.stringify([pollResult]))))
+        .mockImplementation(pollResultsImplementation || (async () => [deepExtend({}, pollResult)]))
 
       return {getBatchMock, pollResultsMock}
     }
@@ -969,10 +969,13 @@ describe('utils', () => {
     })
 
     test('correct number of pass and timeout results', async () => {
-      const pollTimeoutResult: PollResult = {...pollResult, resultID: 'another-id'}
+      const pollTimeoutResult: PollResult = {...deepExtend({}, pollResult), resultID: 'another-id'}
       const batchWithTimeoutResult: Batch = {
         ...batch,
-        results: [batch.results[0], {...batch.results[0], timed_out: true, result_id: pollTimeoutResult.resultID}],
+        results: [
+          batch.results[0],
+          {...batch.results[0], timed_out: true, result_id: pollTimeoutResult.resultID}, // backend is the source of truth for timeout
+        ],
       }
 
       mockApi({
@@ -988,12 +991,33 @@ describe('utils', () => {
           {
             datadogSite: DEFAULT_COMMAND_CONFIG.datadogSite,
             failOnCriticalErrors: false,
+            failOnTimeout: false,
             maxPollingTimeout: 2000,
             subdomain: DEFAULT_COMMAND_CONFIG.subdomain,
           },
           mockReporter
         )
-      ).toEqual([result, {...result, resultId: pollTimeoutResult.resultID, timedOut: true}])
+      ).toEqual([
+        {
+          ...result,
+          passed: true,
+          timedOut: false,
+        },
+        {
+          ...result,
+          passed: true, // because `failOnTimeout` is false
+          timedOut: true,
+          resultId: pollTimeoutResult.resultID,
+          result: {
+            ...result.result,
+            failure: {
+              code: 'TIMEOUT',
+              message: 'Result timed out',
+            },
+            passed: false,
+          },
+        },
+      ])
     })
 
     test('tunnel failure', async () => {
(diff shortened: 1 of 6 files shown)


src/commands/synthetics/utils.ts
```diff
@@ -844,8 +881,6 @@ export const renderResults = ({
     } else {
       summary.failed++
     }
-
-    reporter.resultEnd(result, getAppBaseURL(config))
```
jakepruitt, line 848: Since we no longer print the results as part of this function, could we rename the function from `renderResults` to something like `renderResultSummary`?

src/commands/synthetics/__tests__/utils.test.ts
```diff
@@ -969,10 +1114,13 @@ describe('utils', () => {
     })
 
     test('correct number of pass and timeout results', async () => {
-      const pollTimeoutResult: PollResult = {...pollResult, resultID: 'another-id'}
+      const pollTimeoutResult: PollResult = {...deepExtend({}, pollResult), resultID: 'another-id'}
```
jakepruitt, line 1117: What's the intention of the changes to this test? It seems unrelated to the in-progress logging.

jakepruitt (changes requested):
This looks good to me, though I'm interested in the case where the batch times out and there are still `in_progress` test runs. Would the timed out result show up in the logs? Could you add that as a test and confirm that `reporter.resultEnd()` is still called for those timed out tests?

src/commands/synthetics/__tests__/utils.test.ts
Drarig29: I fixed this test as a separate commit first, before doing any code change: https://github.com/DataDog/datadog-ci/pull/1097/commits/f9aa266004b380e9da73317ec6a56ff58b1dc5ba

This test was buggy because this line of code makes side effects:
https://github.com/DataDog/datadog-ci/blob/d045ad3aada023ff6c7d334cb2fc1fe2a1427583/src/commands/synthetics/utils.ts#L331-L332

This side effect was setting the `failure` property in **both the passed `pollResult`** and failed `pollTimeoutResult` here: https://github.com/DataDog/datadog-ci/blob/f9aa266004b380e9da73317ec6a56ff58b1dc5ba/src/commands/synthetics/__tests__/utils.test.ts#L983

The rest of the edits made to this test is to make it more readable.

Drarig29: I had to fix this test because it started failing **_because of the side effect which makes it buggy_** once I made my code changes.

src/commands/synthetics/utils.ts
Drarig29: This is a breaking change and I wanted to avoid that. But I can deprecate the old one with:

```ts
/**
 * @deprecated Use filterAndFormatGitRemote instead
 */
export const renderResults = renderResultSummary
```

jakepruitt: 👍 let's do that.

src/commands/synthetics/__tests__/utils.test.ts
jakepruitt: 👍 sounds good, thanks for the explanation

Drarig29:
> I'm interested in the case where the batch times out and there are still `in_progress` test runs. Would the timed out result show up in the logs?

@jakepruitt in case you missed it, this is not possible in theory.

If our backend works correctly, then when it detects the batch has timed out, it should update the batch to `status: failed`, and all its `in_progress` results to `status: failed` and `timed_out: true`. This behavior is tested here:

https://github.com/DataDog/datadog-ci/blob/8c39207fe69ed9e42e51a13abed4d5b8cf5ed133/src/commands/synthetics/__tests__/utils.test.ts#L1119-L1172

But in case something goes wrong with our backend and we have to rely on our `hasBatchExceededMaxPollingDate` fallback, I added a `emitResidualResultsIfAny` function. Its behavior is tested in the last assertions here:

https://github.com/DataDog/datadog-ci/blob/c9985c7d8dd11ed4628db9658e5770b999af582b/src/commands/synthetics/__tests__/utils.test.ts#L971-L1016

src/commands/synthetics/utils.ts
jakepruitt: Per conversation, we'll address this in a follow-up

jakepruitt (approved):
👍 Looks good to me, let's follow up later for the refactor into a separate `batch.ts` file.

PR #1102, opened 2023-11-07
[dep] Bump `@aws-sdk/*` to fix undeclared dependencies

Drarig29:
### What and why?

This PR will fix the following errors:
- https://github.com/aws/aws-sdk-js-v3/issues/5362
- https://github.com/aws/aws-sdk-js-v3/issues/5437

Those errors make the CLI crash when uploading JUnit reports [when using Yarn P'n'P](https://yarnpkg.com/features/pnp#semantic-erroring).

Note that this PR will introduce the following warning for Node.js 14.x:
```
(node:2367) Warning: NodeDeprecationWarning: The AWS SDK for JavaScript (v3) will
no longer support Node.js 14.x on May 1, 2024.

To continue receiving updates to AWS services, bug fixes, and security
updates please upgrade to an active Node.js LTS version.

More information can be found at: https://a.co/dzr2AJd
```

You can see it [in the unit tests that are run on Node 14](https://github.com/DataDog/datadog-ci/actions/runs/6785564730/job/18444172933?pr=1102#step:8:61).

### How?

Bump all the `@aws-sdk/*` packages.

### Review checklist


src/commands/lambda/constants.ts
```diff
@@ -29,29 +31,44 @@ export enum RuntimeType {
   NODE,
   PYTHON,
   RUBY,
+  GO,
 }
 
-export const RUNTIME_LOOKUP = {
+export const RUNTIME_LOOKUP: Record<Runtime, RuntimeType> = {
```
duncanista, line 37: We shouldn't be adding previous runtimes, since we don't have support for those who were not previously listed here.

We constantly keep removing runtimes which are out of support for AWS Lambda, so we need to revert this change.

duncanista (commented):
Everything looks good – but the runtime lookup map, thanks for fixing this! Once that is changed, I'll aprove :shipit:

src/commands/lambda/constants.ts
Drarig29: Oh ok, thanks for the context 😄

duncanista (approved):

duncanista (approved):

PR #1104, opened 2023-11-07
[SVLS-3931] Adds retry strategy with exponential backoff to AWS requests in Lambda Command

duncanpharvey:
### What and why?

Adds retries to AWS requests made by the lambda cli due to `Throttling Exception: Rate Exceeded` errors when instrumenting a large number of Lambda functions.

https://datadoghq.atlassian.net/browse/SVLS-3931

### How?

* Uses [@smithy/util-retry](https://github.com/awslabs/smithy-typescript) package to retry failed requests to AWS
* Retries 4 times after initial attempt with exponential backoff starting at 1 second

### Review checklist

### Additional Notes

* `getLambdaFunctionConfigsFromRegex` previously had dedicated retry logic but this was removed in favor of the retry strategy being used for all AWS requests
* From manual testing, Throttling Exceptions occur on `ListTagsCommand` requests. However the retry logic is applied to all AWS requests to handle potential Throttling Exceptions in environments different than the one used to test.

Validated the retry behavior by adding debug logs after the `ListTagsCommand` is sent:

```
  if (response.$metadata.attempts !== undefined && response.$metadata.attempts > 1) {
    console.log('ListTagsCommand')
    console.log(`statusCode: ${response.$metadata.httpStatusCode}`)
    console.log(`attempts: ${response.$metadata.attempts}`)
    console.log(`totalRetryDelay: ${response.$metadata.totalRetryDelay}`)
  }
```

```
ListTagsCommand
statusCode: 200
attempts: 2
totalRetryDelay: 1000
ListTagsCommand
statusCode: 200
attempts: 2
totalRetryDelay: 1000
ListTagsCommand
statusCode: 200
attempts: 3
totalRetryDelay: 3000
ListTagsCommand
statusCode: 200
attempts: 2
totalRetryDelay: 1000
ListTagsCommand
statusCode: 200
attempts: 3
totalRetryDelay: 3000
```

### DIFF
--- a/src/commands/lambda/__tests__/uninstrument.test.ts
+++ b/src/commands/lambda/__tests__/uninstrument.test.ts
@@ -303,7 +303,7 @@ describe('lambda', () => {
         expect(output).toMatchInlineSnapshot(`
           "
           🐶 Uninstrumenting Lambda function
-          [Error] Couldn't fetch Lambda functions. Error: Max retry count exceeded. Error: ListFunctionsError
+          [Error] Couldn't fetch Lambda functions. Error: ListFunctionsError
           "
         `)
       })
@@ -567,7 +567,7 @@ describe('lambda', () => {
         expect(output).toMatchInlineSnapshot(`
           "
           🐶 Uninstrumenting Lambda function
-          [Error] Couldn't fetch Lambda functions. Error: Max retry count exceeded. Error: ListFunctionsError
+          [Error] Couldn't fetch Lambda functions. Error: ListFunctionsError
           "
         `)
       })
--- a/src/commands/lambda/constants.ts
+++ b/src/commands/lambda/constants.ts
@@ -167,4 +167,7 @@ export const LAMBDA_PROJECT_FILES = [...FLARE_PROJECT_FILES, ...FRAMEWORK_FILES_
 
 // Configures max number of attempts and exponential backoff function for AWS requests
 // First retry is attempt 1
-export const EXPONENTIAL_BACKOFF_RETRY_STRATEGY = new ConfiguredRetryStrategy(4, (attempt: number) => 1000 * 2 ** (attempt - 1))
+export const EXPONENTIAL_BACKOFF_RETRY_STRATEGY = new ConfiguredRetryStrategy(
+  4,
+  (attempt: number) => 1000 * 2 ** (attempt - 1)
+)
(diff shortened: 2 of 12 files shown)


src/commands/lambda/constants.ts
```diff
@@ -165,3 +164,7 @@ export const FRAMEWORK_FILES_MAPPING = new Map([
 ])
 
 export const LAMBDA_PROJECT_FILES = [...FLARE_PROJECT_FILES, ...FRAMEWORK_FILES_MAPPING.keys()]
+
+// Configures max number of attempts and exponential backoff function for AWS requests
+// First retry is attempt 1
+export const RETRY_STRATEGY = new ConfiguredRetryStrategy(4, (attempt: number) => 1000 * 2 ** (attempt - 1))
```
IvanTopolcic, line 170: I'd probably rename `RETRY_STRATEGY` to something more descriptive. If we have two different retry strategies, the name `RETRY_STRATEGY` wouldn't make much sense.

src/commands/lambda/__tests__/uninstrument.test.ts
```diff
@@ -567,7 +567,7 @@ describe('lambda', () => {
         expect(output).toMatchInlineSnapshot(`
           "
           🐶 Uninstrumenting Lambda function
-          [Error] Couldn't fetch Lambda functions. Error: Max retry count exceeded. Error: ListFunctionsError
+          [Error] Couldn't fetch Lambda functions. Error: ListFunctionsError
```
IvanTopolcic, line 570: Would it be possible to add the reason why it couldn't fetch the lambda functions? Like a 403, 500, etc?

IvanTopolcic (commented):
Looks good! Just two small things.

src/commands/lambda/__tests__/uninstrument.test.ts
IvanTopolcic: I guess if we have multiple retries and they fail for different reasons it would be a bit difficult, so I'm not sure if this is worth the effort, but it would be nice to have.

duncanpharvey: The error should bubble up from the last request that caused a failure. Here's what was shared from a customer that prompted this change:

Image 2023-11-16 at 4 39 20 PM

`ListFunctionsError` is just a generic error for the purposes of unit testing.

Let me know if you think additional changes are needed though!

src/commands/lambda/constants.ts
duncanpharvey: I renamed it to `EXPONENTIAL_BACKOFF_RETRY_STRATEGY`. Let me know if that's too verbose

IvanTopolcic (approved):

Drarig29 (approved):

Issue #1096, opened 2023-10-26
security vulnerability in axios SNYK-JS-AXIOS-6032459


mishabruml:
### Bug description

https://security.snyk.io/vuln/SNYK-JS-AXIOS-6032459

Looks like it's not fixed at the moment in axios but I suppose it will be soon.

As an aside, the current version 0.21.4 used in this package is around 2 years old, is there any reason it couldn't be updated to the latest version 1.5.1? It would mean a major version change, there doesn't seem to be official breaking changes guide, there are some community ones though - see https://github.com/axios/axios/discussions/4996

### Command

None

mishabruml:
looks like axios are on it https://github.com/axios/axios/issues/6022

Drarig29:
Hi @mishabruml! We've had issues bumping axios in the past, because its move to ESM was not working properly with our standalone binary. Also, the latest fix that was published by axios seems to be a breaking change: https://github.com/axios/axios/pull/6028#issuecomment-1786789863

From the maintainer of axios:
> I felt that this was not an actual CVE in all honesty as the implementor has control over this and if they decide to set the flag to true they should have understood what that meant.

This is only a vulnerability when `withCredentials` is set to `true`, which isn't our case.

tobiasweibel:
I can see the ESM issues and don't really like to fix such issues in projects of my company.
BUT: As my company runs security scanners it will assert the axios issue and we basically have to address it somehow.

Drarig29:
Hi, just a heads up: we are working on it, a PR is being reviewed.

mishabruml:
@Drarig29 thanks for this, could you re-open the issue until this is released? 😄

Drarig29:
@mishabruml you can now use https://github.com/DataDog/datadog-ci/releases/tag/v2.23.1

PR #1108, opened 2023-11-09
[dep] Upgrade axios from 0.21.4 to 1.6.0

Drarig29:
### What and why?

Closes https://github.com/DataDog/datadog-ci/issues/1096

Bumping `axios` to `1.6.0` came with 2 issues:
- https://github.com/axios/axios/discussions/5866, fixed in a5c801aba9fa8131aeefc4ffa75002df67156c8e
- Sending requests with `FormData` (which sends `multipart/form-data` payloads) was failing silently in our backend with only a warning.
  - For reference, the warning was `Decoder has partially decoded field`, which was probably caused by an incorrect boundary. Bumping `form-data` from `3.0.1` to `4.0.0` ([compare changes](https://github.com/form-data/form-data/compare/v3.0.1...v4.0.0)) solved the issue: 41ce71b54e19deacf6953865dd14ee3031cf61c4
  - This bug was caught in [one of our e2e tests](https://github.com/DataDog/datadog-ci/actions/runs/6704428841/job/18217012787#step:9:43), for the `junit upload` command. I ensured this command also works with the standalone binary. I also tested with `synthetics upload-application`, which is another command using the  `form-data` package.

### How?

Bump `axios` and `form-data`, and add `node_modules/axios/dist/node/axios.cjs` to the `pkg.assets` to prevent the following error when running the standalone binary:

```
pkg/prelude/bootstrap.js:1872
      throw error;
      ^

Error: Cannot find module '/snapshot/my_app/node_modules/axios/dist/node/axios.cjs'
1) If you want to compile the package/file into executable, please pay attention to compilation warnings and specify a literal in 'require' call. 2) If you don't want to compile the package/file into executable and want to 'require' it from filesystem (likely plugin), specify an absolute path in 'require' call using process.cwd() or process.execPath.
```

### Review checklist

### DIFF
--- a/jest.config-standalone.js
+++ b/jest.config-standalone.js
@@ -1,9 +1,12 @@
 module.exports = {
   preset: 'ts-jest',
-  globals: {
-    'ts-jest': {
-      tsconfig: 'tsconfig.json',
-    },
+  transform: {
+    '^.+\\.tsx?$': [
+      'ts-jest',
+      {
+        tsconfig: 'tsconfig.json',
+      },
+    ],
   },
   roots: ['standalone-e2e'],
 }
--- a/src/commands/synthetics/api.ts
+++ b/src/commands/synthetics/api.ts
@@ -39,12 +39,8 @@ export class EndpointError extends Error {
 }
 
 export const formatBackendErrors = (requestError: AxiosError<BackendError>) => {
-  if (!requestError.config) {
-    return `config unavailable\n${requestError.message}`
-  }
-
   if (requestError.response?.data?.errors) {
-    const serverHead = `query on ${requestError.config.baseURL}${requestError.config.url} returned:`
+    const serverHead = `query on ${requestError.config?.baseURL}${requestError.config?.url} returned:`
     const errors = requestError.response.data.errors
     if (errors.length > 1) {
       const formattedErrors = errors.map((message: string) => `  - ${message}`)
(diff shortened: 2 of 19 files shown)


src/commands/synthetics/api.ts
```diff
@@ -38,6 +39,10 @@ export class EndpointError extends Error {
 }
 
 export const formatBackendErrors = (requestError: AxiosError<BackendError>) => {
+  if (!requestError.config) {
+    return `config unavailable\n${requestError.message}`
+  }
```
jakepruitt, line 44: Why this change? And why only for the syntehtics API client?

Drarig29: It was to make TypeScript happy, as the `config` can now be `undefined`. But I agree, this adds a new possible error, so let's use `requestError.config?.baseURL` instead.

And only on Synthetics because it's the only place where we read properties from `AxiosError.config`

Drarig29: Addressed in a1fb70da5b09aa69d7ac6b435a5421223cb338f7

juli1 (approved):

Letiste (approved):

romainkomorndatadog (approved):

jest.config-standalone.js
```diff
@@ -1,9 +1,12 @@
...
-    'ts-jest': {
-      tsconfig: 'tsconfig.json',
-    },
+  transform: {
+    '^.+\\.tsx?$': [
+      'ts-jest',
+      {
+        tsconfig: 'tsconfig.json',
+      },
+    ],
```
etnbrd, line 9: Why this change?

etnbrd (approved):
LGTM 👍 
Still curious about the standalone jest config.

jest.config-standalone.js
Drarig29: For a moment I thought it was related to the first issue I had with axios: 
image

Turns out it was _really_ just a warning, but it's still a good thing to fix 😄

PR #1114, opened 2023-11-14
[synthetics] Refactor `waitForBatchToFinish`

Drarig29:
### What and why?

Refactor `waitForBatchToFinish` because it could be easier to read.

### How?

- Change `do {} while ()` to `while () {}`
- Generally in `while (true)` loops, it's clearer to have a condition for `break`ing or `return`ing, rather than `continue`ing.

### Review checklist


etnbrd (approved):
LGTM 👍

Issue #1117, opened 2023-11-16
flutter-symbols not uploading android mapping.txt with --android-mapping-location


bounth:
### Bug description

The command `--android-mapping-location ./build/app/outputs/mapping/devRelease/mapping.txt` doesn't upload the mapping file as intended from the datadog-ci log output

```
Uploading Android Mapping File ./build/app/outputs/mapping/devRelease/mapping.txt
Mapping upload finished: 0
```

The command `ls` shows that the `mapping.txt` location is correct.
```
ls ./build/app/outputs/mapping/devRelease
configuration.txt
mapping.txt
resources.txt
seeds.txt
usage.txt
```

### Describe what you expected

Expect to see something similar to successful upload of dSYM and symbols file
```
✅ Uploaded 58 dSYMs in 58.895 seconds.
✅ Uploaded symbol 6 files in 61.192 seconds.
```

### Steps to reproduce the issue

Full log
```
==> Start
==> Switching to working directory: /Users/vagrant/git
+ curl -L --fail https://github.com/DataDog/datadog-ci/releases/latest/download/datadog-ci_darwin-x64 --output /usr/local/bin/datadog-ci
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0

 37 73.6M   37 27.6M    0     0  65.4M      0  0:00:01 --:--:--  0:00:01 65.4M
100 73.6M  100 73.6M    0     0   106M      0 --:--:-- --:--:-- --:--:--  169M
+ chmod +x /usr/local/bin/datadog-ci
+ datadog-ci version
v2.23.1
+ '[' -d ./build/app/outputs/symbols ']'
+ datadog-ci flutter-symbols upload --service-name myi-mobile --flavor dev --version 3.10.0-19 --dart-symbols-location ./build/app/outputs/symbols --android-mapping-location ./build/app/outputs/mapping/devRelease/mapping.txt --ios-dsyms
Starting upload. 
Uploading ios dSYMs at location ./build/ios/archive/Runner.xcarchive/dSYMs
[...]

PR #1118, opened 2023-11-17
[flutter-symbols] Improve message for mapping upload

Drarig29:
### What and why?

Reported in https://github.com/DataDog/datadog-ci/issues/1117

### How?

No need for more information: this PR just removes the reporting of `0` which could be mistaken for a number of uploaded files.

### Review checklist

- ~[ ] Feature or bugfix MUST have appropriate tests (unit, integration)~ (the text output is not unit tested in this command)

### DIFF
--- a/src/commands/flutter-symbols/upload.ts
+++ b/src/commands/flutter-symbols/upload.ts
@@ -329,7 +329,7 @@ export class UploadCommand extends Command {
       payload.content.set('repository', this.getGitDataPayload(this.gitData))
     }
 
-    const result = await uploadMultipartHelper(requestBuilder, payload, {
+    const status = await uploadMultipartHelper(requestBuilder, payload, {
       apiKeyValidator,
       onError: (e) => {
         this.context.stdout.write(renderFailedUpload(this.androidMappingLocation!, e.message))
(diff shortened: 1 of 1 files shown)


xgouchet (approved):

src/commands/flutter-symbols/upload.ts
```diff
@@ -345,9 +345,14 @@ export class UploadCommand extends Command {
       retries: 5,
       useGzip: true,
     })
-    this.context.stdout.write(`Mapping upload finished: ${result}\n`)
 
-    return result
+    if (status === UploadStatus.Success) {
+      this.context.stdout.write('Mapping upload finished\n')
+    } else {
+      this.context.stdout.write(`Mapping upload failed\n`)
```
fuzzybinary, line 352: Does the error code get output in `verbose`? If not, can we output the error code here? Just thinking that if someone has a problem the extra information would be nice.

Drarig29: The error code can only be `0`, `1` or `2`:
https://github.com/DataDog/datadog-ci/blob/50a2a70b23bad6e36ddc6d9f54bb886f44a4058e/src/helpers/upload.ts#L49-L53

And in this case ([`uploadMultipartHelper()`](https://github.com/DataDog/datadog-ci/blob/50a2a70b23bad6e36ddc6d9f54bb886f44a4058e/src/commands/flutter-symbols/helpers.ts#L20-L24) directly returns [`upload()`](https://github.com/DataDog/datadog-ci/blob/50a2a70b23bad6e36ddc6d9f54bb886f44a4058e/src/helpers/upload.ts#L59-L86)'s return value), we never possibly return a `UploadStatus.Skipped`, so I figured "Mapping upload finished" or "Mapping upload failed" was enough.

WDYT @fuzzybinary?

fuzzybinary: Yeah sounds good. Thanks for taking care of this 👍

fuzzybinary (approved):

PR #1121, opened 2023-11-21
[lambda] Add support for Node 20

duncanista:
### What and why?

Adds support for Node 20 as it's [available](https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html) for AWS Lambda.

### How?

Added `nodejs20.x` key to layer lookup and some unit tests.

### Review checklist


duncanista:
Failure is expected, blocked while waiting on #1120 to be merged so the type for the runtime exist.

PR #1125, opened 2023-11-25
[synthetics] Only send `executionRule` when overridden

Drarig29:
### What and why?

Currently, when a user changes the **execution rule in the UI**, the test is considered as a new test (`reason: new` for the selective re-run) if the execution rule is not overridden with `datadog-ci` (through a global config file or a test file).

That's because:
- The user changed the test configuration in Datadog
- Since there is no override, [`getExecutionRule()`](https://github.com/DataDog/datadog-ci/blob/18127f3126235433fa1cc46fb127ab1bf7b0cfdb/src/commands/synthetics/utils.ts#L130) falls back to the test configuration's execution rule
- Since the resulting [`TestPayload`](https://github.com/DataDog/datadog-ci/blob/18127f3126235433fa1cc46fb127ab1bf7b0cfdb/src/commands/synthetics/interfaces.ts#L329-L332) is different than the last time the tests were triggered, then the selective re-run will think it's a different user intent, i.e. `reason: new`

### How?

To fix this bad behavior, we have to send the `executionRule` only when it's overridden with configuration files.

On the backend, we already use the test configuration's execution rule when there is no `executionRule` override. So this will be a noop in terms of behavior.

This will only make sure that a test is considered as `reason: edited` instead of `reason: new` when the execution rule is changed in Datadog and not overriden.

### Review checklist


etnbrd (approved):
LGTM 👍 

Said otherwise: `executionRule` used to be handled differently than the rest of the config override, and this PR removes this inconsistency.

Also, bear in mind that `getExecutionRule` is part of the public interface. So this PR might require a major version bump.

PR #1128, opened 2023-12-03
[synthetics] Add `executeWithDetails()` method

jakub-g:
### What and why?

We want to `execute` tests programmatically and benefit from everything that `execute` provides (setting up reports, rendering results etc.), but also gather the results and summary, and process them on our side (log a custom summary in console etc.)

Currently datadog-ci just logs things itself, and returns only a `0 | 1` exit code at the end of execution.

### How?

Instead of returning `0 | 1`, return an object with detailed info.

### Review checklist

### DIFF
--- a/src/commands/synthetics/__tests__/run-tests-lib.test.ts
+++ b/src/commands/synthetics/__tests__/run-tests-lib.test.ts
@@ -429,7 +429,7 @@ describe('run-test', () => {
     })
   })
 
-  describe('execute', () => {
+  describe('executeWithDetails', () => {
     beforeEach(() => {
       jest.restoreAllMocks()
       jest.spyOn(api, 'getApiHelper').mockImplementation(
@@ -445,14 +445,14 @@ describe('run-test', () => {
     })
 
     test('should call executeTests and renderResults', async () => {
-      await runTests.execute({}, {})
+      await runTests.executeWithDetails({}, {})
       expect(runTests.executeTests).toHaveBeenCalled()
       expect(utils.renderResults).toHaveBeenCalled()
     })
 
     test('should extend config', async () => {
       const runConfig = {apiKey: 'apiKey', appKey: 'appKey'}
-      await runTests.execute(runConfig, {})
+      await runTests.executeWithDetails(runConfig, {})
       expect(runTests.executeTests).toHaveBeenCalledWith(
         expect.anything(),
         expect.objectContaining(runConfig),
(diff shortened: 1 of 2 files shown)


jakub-g:
@arcanis this method (`execute` from `run-test-lib` is not used by clipanion). The clipanion command is in a sibling file (`run-test-command`).

The `run-test-lib#execute` is only reexported from `index` for convenience, it's not used anywhere internally.

But aside from that, yes, I think a separate method without breaking public API contract would be better. CC @etnbrd @Drarig29

etnbrd:
Why don't we use the existing `executeTests` function which is meant to return the `result`?

yoannmoinet (dismissed):
LGTM

src/commands/synthetics/run-tests-lib.ts
```diff
@@ -300,5 +304,33 @@ export const execute = async (
...
+    jUnitReport,
+    reporters,
+    runId,
+    suites,
+  }: {
+    jUnitReport?: string
+    reporters?: (SupportedReporter | Reporter)[]
+    runId?: string
+    suites?: Suite[]
+  }
```
jakub-g, line 328: Should I extract the object type into a something like `type ExecuteOptions = `, or we're fine leaving it as-is?

jakub-g:
Updated the code as discussed, adding new method

src/commands/synthetics/run-tests-lib.ts
```diff
@@ -300,5 +304,33 @@ export const execute = async (
...
+): Promise<0 | 1> => {
+  return executeWithDetails(runConfig, {
+    jUnitReport,
+    reporters,
+    runId,
+    suites,
+  }).then((value) => value.exitCode)
```
Drarig29, line 335: Can you `await` instead of using `.then()`?

yoannmoinet (approved):
LGTM

src/commands/synthetics/__tests__/run-tests-lib.test.ts
```diff
@@ -476,20 +476,39 @@ describe('run-test', () => {
       })
```
jakub-g, line 476: Do we also want a test on return values, or is TS signature enough?

Could add something like 
```
    test('should return result values returned by executeTests', async () => {
      const returnValue = await runTests.executeWithDetails({}, {})
      expect(returnValue.results).toBeDefined()
      expect(returnValue.summary).toBeDefined()
    })
]
```

src/commands/synthetics/run-tests-lib.ts
Drarig29: Yes, and please define that type in this file without exporting it to avoid making the API surface even bigger. It's only to factor the type

jakub-g: @Drarig29 done

Drarig29 (approved):
LGTM

PR #1133, opened 2023-12-06
[synthetics] Synthetics mobile upload-application - use multipart upload

kevin-pietsch:
### What and why?

For synthetics-mobile commands that upload an application (`upload-application` and `run-test` with a temporary app), this PR  unifies behavior with `web-ui` by using multipart presigned upload instead of the previous presigned form method.

Using multi-part is potentially more efficient and will be supported by EU1 datacenter.

### Review checklist

### DIFF
--- a/src/commands/synthetics/api.ts
+++ b/src/commands/synthetics/api.ts
@@ -199,7 +199,7 @@ const uploadMobileApplicationPart = (request: (args: AxiosRequestConfig) => Axio
           'Content-MD5': parts[Number(partNumber) - 1].md5,
           // Presigned URL *requires* unset content-type since it's used for signature
           // We can clear axios default by setting to null
-          // https://github.com/axios/axios/issues/382
+          // https://github.com/axios/axios/pull/1845
           // eslint-disable-next-line no-null/no-null
           'Content-Type': null,
         },
--- a/src/commands/synthetics/api.ts
+++ b/src/commands/synthetics/api.ts
@@ -2,8 +2,6 @@ import {stringify} from 'querystring'
 
 import type {AxiosError, AxiosPromise, AxiosRequestConfig} from 'axios'
 
-import FormData from 'form-data'
-
 import {getRequestBuilder} from '../../helpers/utils'
 
 import {CriticalError} from './errors'
@@ -11,10 +9,12 @@ import {
   APIConfiguration,
   APIHelperConfig,
   Batch,
+  MobileApplicationUploadPart,
+  MobileApplicationUploadPartResponse,
   MobileApplicationVersion,
   Payload,
   PollResult,
-  PresignedUrlResponse,
+  MultipartPresignedUrlsResponse,
   ServerBatch,
   ServerTest,
   SyntheticsOrgSettings,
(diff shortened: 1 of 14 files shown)


nmizoguchi (approved):
LGTM overall, left a nit about casting :)

src/commands/synthetics/mobile.ts
```diff
@@ -4,46 +4,80 @@ import fs from 'fs'
...
   Test,
   TestPayload,
   UploadApplicationCommandConfig,
   UserConfigOverride,
 } from './interfaces'
 
-export const getSizeAndMD5HashFromFile = async (filePath: string): Promise<{appSize: number; md5: string}> => {
-  const hash = crypto.createHash('md5')
-  const fileStream = fs.createReadStream(filePath)
+const UPLOAD_FILE_MAX_PART_SIZE = 10 * 1024 * 1024
```
Drarig29, line 17:
```suggestion
const UPLOAD_FILE_MAX_PART_SIZE = 10 * 1024 * 1024 // MiB
```

src/commands/synthetics/api.ts
```diff
@@ -160,39 +160,85 @@ const getTunnelPresignedURL = (request: (args: AxiosRequestConfig) => AxiosPromi
...
+): Promise<MobileApplicationUploadPartResponse[]> => {
+  const promises = Object.entries(multipartPresignedUrlsParams.urls).map(async ([partNumber, presignedUrl]) => {
+    const resp = await retryRequest(
+      {
+        data: parts[Number(partNumber) - 1].blob,
+        headers: {
+          'Content-MD5': parts[Number(partNumber) - 1].md5,
+          // Presigned URL *requires* unset content-type since it's used for signature
+          // We can clear axios default by setting to null
+          // https://github.com/axios/axios/issues/382
```
Drarig29, line 202: **nit**: this link is the PR that closes the issue, and it explicitly talks about `null` (for the curious who would try to use `undefined` instead)
```suggestion
          // https://github.com/axios/axios/pull/1845
```

src/commands/synthetics/mobile.ts
```diff
@@ -4,46 +4,80 @@ import fs from 'fs'
...
+
+export const getSizeAndPartsFromFile = async (
+  filePath: string
+): Promise<{appSize: number; parts: MobileApplicationUploadPart[]}> => {
+  const readStreamOptions = {
+    // this oddly-named parameter will cause chunks to be no more than the given size
```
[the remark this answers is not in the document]
kevin-pietsch: Lol -- I'll take the suggestion to remove the commentary. IMO `bufferMaxSize` would have been a lot clearer for this very public API. Even if you know what high-water-mark typically means (which is not a good assumption), it's not completely obvious what it means for this specific case without looking up docs.

src/commands/synthetics/api.ts
kevin-pietsch: Oh ya -- good suggestion 👍

src/commands/synthetics/mobile.ts
```diff
@@ -4,46 +4,80 @@ import fs from 'fs'
...
+    highWaterMark: UPLOAD_FILE_MAX_PART_SIZE,
+  }
+  const fileStream = fs.createReadStream(filePath, readStreamOptions)
+  const parts: MobileApplicationUploadPart[] = []
   for await (const chunk of fileStream) {
-    hash.update(chunk)
+    parts.push({
+      md5: crypto
+        .createHash('md5')
+        .update(chunk as Buffer)
```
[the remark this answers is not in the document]
kevin-pietsch: Yes that's a good idea. I think we can `if (!(chunk instanceof Buffer)) throw`

PR #1134, opened 2023-12-06
[STAL-898] Add static analysis rules

juli1:
### What and why?

This PR adds support for static analysis rules for the `datadog-ci` repository. It will bring static analysis checks for the repository and works in the IDE and later in the CI/CD pipeline.

### How?

Adding a file in the repo


juli1:
> Is it complaining for the existing code or not at all?

In your IDE it should but not in the CI (yet). We will enable GitHub comments for it later to complain about new code.

PR #1156, opened 2024-01-08
[synthetics] Fix enabling selective re-run with config file

Drarig29:
### What and why?

This PR fixes a bug which makes it impossible to turn on `selectiveRerun` via config files. This is because the default value of the CLI parameter is `false` and not `undefined`, thus the `selectiveRerun` value set in a config file is overridden.

### How?

Remove the default value for the CLI parameter, and only rely on `DEFAULT_COMMAND_CONFIG`, which is the source of truth for default config parameters.

### Review checklist

### DIFF
--- a/src/commands/synthetics/__tests__/cli.test.ts
+++ b/src/commands/synthetics/__tests__/cli.test.ts
@@ -4,7 +4,7 @@ import {createCommand} from '../../../helpers/__tests__/fixtures'
 import * as ciUtils from '../../../helpers/utils'
 
 import * as api from '../api'
-import {UserConfigOverride} from '../interfaces'
+import {RunTestsCommandConfig, UploadApplicationCommandConfig, UserConfigOverride} from '../interfaces'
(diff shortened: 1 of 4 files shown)


src/commands/synthetics/__tests__/cli.test.ts
Drarig29, line 1: I edited the unit tests to be future-proof by using the actual `RunTestsCommandConfig` and `UploadApplicationCommandConfig` types.

src/commands/synthetics/__tests__/cli.test.ts
```diff
@@ -81,15 +81,17 @@ describe('run-test', () => {
...
         },
         locations: [],
         pollingTimeout: 1,
-        proxy: {protocol: 'https'},
+        proxy: {
+          protocol: 'https',
+        },
         publicIds: ['ran-dom-id'],
-        selectiveRerun: false,
+        selectiveRerun: true,
```
Drarig29, line 94: Use the non-default value of `selectiveRerun`, to have proper testing.

PR #1161, opened 2024-01-23
Support custom tags and metrics for GH jobs

andrea-mosk:
### What and why?

Adding support for custom tags at job level for GitHub by relying on the `GITHUB_JOB` environment variables. In case the [name](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#name) property is used when defining jobs, customers will need to additionally expose the `DD_GITHUB_JOB_NAME` pointing to that name.

### How?

### Review checklist

### DIFF
--- a/src/commands/metric/README.md
+++ b/src/commands/metric/README.md
@@ -28,7 +28,7 @@ datadog-ci metric --level job --metrics binary.size:1024
 
 ### Supported providers
 
-The metric command only works for the following CI providers: [Buildkite, CircleCI, GitHub, GitLab, AzurePipelines, Jenkins]. If used in
+The metric command only works for the following CI providers: Buildkite, CircleCI, GitHub, GitLab, Azure Pipelines and Jenkins. If used in
 any other provider it will fail.
 
 ### End-to-end testing process
--- a/src/commands/metric/README.md
+++ b/src/commands/metric/README.md
@@ -28,10 +28,8 @@ datadog-ci metric --level job --metrics binary.size:1024
 
 ### Supported providers
 
-The metric command only works for the following CI providers: [Buildkite, CircleCI, GitHub, GitLab]. If used in
-any other provider it will fail. Note that for GitHub actions only the level `pipeline` is supported. If the
-command is invoked in GitHub actions with level `job` it will exit with status code 1 and return an
-error.
+The metric command only works for the following CI providers: [Buildkite, CircleCI, GitHub, GitLab, AzurePipelines, Jenkins]. If used in
+any other provider it will fail.
 
 ### End-to-end testing process
 
--- a/src/commands/tag/README.md
+++ b/src/commands/tag/README.md
@@ -29,10 +29,8 @@ datadog-ci tag --level job --tags "go.version:`go version`"
 
 ### Supported providers
 
-The tag command only works for the following CI providers: [Buildkite, CircleCI, GitHub, GitLab]. If used in
-any other provider it will fail. Note that for GitHub actions only the level `pipeline` is supported. If the
-command is invoked in GitHub actions with level `job` it will exit with status code 1 and return an
-error.
+The tag command only works for the following CI providers: [Buildkite, CircleCI, GitHub, GitLab, AzurePipelines, Jenkins]. If used in
+any other provider it will fail.
 
 ### End-to-end testing process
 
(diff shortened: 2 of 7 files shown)


src/helpers/ci.ts
```diff
@@ -48,6 +48,10 @@ export const PROVIDER_TO_DISPLAY_NAME = {
   buddy: 'Buddy',
 }
 
+// DD_GITHUB_JOB_NAME is an override that is required for adding custom tags and metrics
+// to GHA jobs if the 'name' property is used. It's ok for it to be missing in case the name property is not used.
```
Drarig29, line 52: ❓  It's this kind of 'name' that you are talking about, right?
https://github.com/DataDog/datadog-ci/blob/c0a0cc657a47682106d9fbf691b8c282ea0404e0/.github/workflows/ci.yml#L7-L14

Are you saying that, in the end, the job name is totally optional? i.e. that having no `name` and no `DD_GITHUB_JOB_NAME` is acceptable?

anmarchenko (approved):

src/helpers/ci.ts
AdrianLC: The name is in `GITHUB_JOB` this is an optional workaround for when they are providing a custom name like in your example

AdrianLC (approved):

src/helpers/ci.ts
andrea-mosk: Correct. The `GITHUB_JOB` environment variable will contain `build-and-test` both when a name is specified and when it isn't. In order to support the case in which is specified we will need customers to specify another env var to point to the custom name

Drarig29: Oh ok, I see. Just want to point out that there is the same issue when using a job `matrix`: although GitHub's UI shows the right name (with [`${{ matrix.version }}`](https://github.com/DataDog/datadog-ci/blob/c0a0cc657a47682106d9fbf691b8c282ea0404e0/.github/workflows/ci.yml#L12) between parenthesis, see screenshot), the `GITHUB_JOB` stays the same.

Would be worth to document that `DD_GITHUB_JOB_NAME` is a good use for that as well:
```yaml
env:
  DD_GITHUB_JOB_NAME: "Build and test (${{ matrix.version }})"
```

image

brett0000FF (approved):

src/commands/metric/README.md
```diff
@@ -28,10 +28,8 @@ datadog-ci metric --level job --metrics binary.size:1024
 
 ### Supported providers
 
-The metric command only works for the following CI providers: [Buildkite, CircleCI, GitHub, GitLab]. If used in
-any other provider it will fail. Note that for GitHub actions only the level `pipeline` is supported. If the
-command is invoked in GitHub actions with level `job` it will exit with status code 1 and return an
-error.
+The metric command only works for the following CI providers: [Buildkite, CircleCI, GitHub, GitLab, AzurePipelines, Jenkins]. If used in
```
Drarig29, line 31: nit:
```suggestion
The metric command only works for the following CI providers: Buildkite, CircleCI, GitHub, GitLab, Azure Pipelines and Jenkins. If used in
```

Drarig29 (approved):
LGTM (just one nit)

src/commands/tag/README.md
```diff
@@ -29,10 +29,8 @@ datadog-ci tag --level job --tags "go.version:`go version`"
 
 ### Supported providers
 
-The tag command only works for the following CI providers: [Buildkite, CircleCI, GitHub, GitLab]. If used in
-any other provider it will fail. Note that for GitHub actions only the level `pipeline` is supported. If the
-command is invoked in GitHub actions with level `job` it will exit with status code 1 and return an
-error.
+The tag command only works for the following CI providers: [Buildkite, CircleCI, GitHub, GitLab, AzurePipelines, Jenkins]. If used in
```
Drarig29, line 32: sorry, I didn't add the suggestion on this one 😟
```suggestion
The tag command only works for the following CI providers: Buildkite, CircleCI, GitHub, GitLab, Azure Pipelines and Jenkins. If used in
```

src/helpers/ci.ts
andrea-mosk: Oh, this is a great spot thanks. I will make sure to include this in the docs I'll create

src/commands/tag/README.md
andrea-mosk: And I didn't remember there were two very similar 😂 Committed!

PR #1170, opened 2024-02-05
docs(sourcemaps): fix minified path prefix URL example on E2E tests

Gummiees:
### What and why?

Copying the code under "[End-to-end testing process](https://github.com/DataDog/datadog-ci/tree/master/src/commands/sourcemaps#end-to-end-testing-process)" does not work and throws this error:

```
❌ --minified-path-prefix should either be an URL (such as "http://example.com/static") or an absolute path starting with a / such as "/static"
```

That is because the `--minified-path-prefix` URL parameter is missing `:`.

### How?

Just updating docs.

### Review checklist


amortemousque (approved):

rtrieu (approved):
Hi @Gummiees, thank you so much for this submission!

PR #1172, opened 2024-02-05
[SYNTH-12059] Trailing slash in test suite id leads to `[] Found test "undefined"`

teodor2312:
### What and why?

In cases where we would have a "/" at the end of the url in a `Test Config` or a cli command like
`datadog-ci synthetics run-tests --public-id ''`
We would get a `[] Found test "undefined"` message. Neither of these are desired behaviours.

For the first case we want the test to be correctly identified. And for the 2nd we want to raise an error. 

### How?

This PR is addressing both by improving the way we find the public_id in a url and in case we don't find it we raise an error.

### Review checklist


Drarig29 (approved):
LGTM

PR #1181, opened 2024-02-09
[SYNTH-12557] Add missing mobileApplicationVersion  configurations for Bitrise integrations

teodor2312:
### What and why?

For the Bitrise integrations we need two configurations that are mentioned in datadog-ci but not really implemented

### How?

Adding the first of these 2 configurations, `mobileApplicationVersion`, to be able to configured through Global Config, env variables, cli param and Test Config. 
We agreed that we want to align all the configurations to be able to be configured in those 4 ways in this doc - https://datadoghq.atlassian.net/wiki/spaces/SYN/pages/3378446383/Configuration+parameters+and+their+usages
so this is a good opportunity to do this and create an example for the rest.

### Review checklist


src/commands/synthetics/interfaces.ts
```diff
@@ -381,6 +381,7 @@ export interface TriggerConfig {
   config: UserConfigOverride
   id: string
   suite?: string
+  mobileApplicationVersion?: string
```
teodor2312, line 384: What do you think about putting this one here? 
I was taking inspiration from our talks about `mobileApplicationVersionFilePath` and how it wasn't supposed to be actually part of the config but of the same level as `id`. 
However i'm not sure if it's better to have it here or in a separate class that this one merged with 🤔

src/commands/synthetics/run-tests-lib.ts
```diff
@@ -68,6 +68,13 @@ export const executeTests = async (
     }
   }
 
+  if (config.mobileApplicationVersion && config.mobileApplicationVersionFilePath) {
+    throw new CiError(
+      'INVALID_CONFIG',
+      'Both mobileApplicationVersion and mobileApplicationVersionFilePath cannot be set at the same time'
+    )
+  }
+
```
teodor2312, line 77: from reading the summaries from the RFC and i think you also mentioned something about it @Drarig29, think these should be mutually exclusive 🤔 
Also what do you think about where i've put this logic? Is there a better place for it?

src/commands/synthetics/README.md
```diff
@@ -133,6 +133,7 @@ For example:
     "followRedirects": true,
     "headers": {"<NEW_HEADER>": "<NEW_VALUE>"},
     "locations": ["aws:us-west-1"],
+    "mobileApplicationVersion": "01234567-8888-9999-abcd-efffffffffff",
```
Drarig29, line 136: **nit** (was probably forgotten before)
```suggestion
    "mobileApplicationVersion": "01234567-8888-9999-abcd-efffffffffff",
    "mobileApplicationVersionFilePath": "path/to/application.apk",
```

src/commands/synthetics/utils/public.ts
```diff
@@ -54,8 +54,8 @@ import {
 } from './internal'
 
 const POLLING_INTERVAL = 5000 // In ms
-const PUBLIC_ID_REGEX = /\b[a-z0-9]{3}-[a-z0-9]{3}-[a-z0-9]{3}\b/
 const TEMPLATE_REGEX = /{{\s*([^{}]*?)\s*}}/g
+export const PUBLIC_ID_REGEX = /\b[a-z0-9]{3}-[a-z0-9]{3}-[a-z0-9]{3}\b/
```
Drarig29, line 58: Anything you export here will be exported publicly, and I would rather export a function to extract a publicId from a string, than exporting a regex constant.

src/commands/synthetics/__tests__/cli.test.ts
```diff
@@ -124,6 +130,9 @@ describe('run-test', () => {
         tunnel: true,
         variableStrings: ['key=value'],
       }
+      const defaultTestOverrides: UserConfigOverride = {
```
Drarig29, line 133: +1 for the name :grin:

Drarig29 (commented):
LGTM, except one nit for `PUBLIC_ID_REGEX` - and let's wait for an answer for a better description of `mobile_application_version`

src/commands/synthetics/utils/public.ts
teodor2312: ah shit okay will probably pull it out then 👍

src/commands/synthetics/__tests__/cli.test.ts
teodor2312: Yeah decided to make it easier for myself for later 😂

urseberry (approved):

src/commands/synthetics/run-tests-command.ts
```diff
@@ -103,6 +103,9 @@ export class RunTestsCommand extends Command {
   private files = Option.Array('-f,--files', {
     description: `Glob pattern to detect Synthetic test ${$2('configuration files')}}.`,
   })
+  private mobileApplicationVersion = Option.String('--mobileApplicationVersion', {
+    description: 'Override the default mobile application version to test a different version within Datadog. ',
+  })
```
Drarig29, line 108:
```suggestion
  private mobileApplicationVersion = Option.String('--mobileApplicationVersion', {
    description: 'Override the default mobile application version to test a different version within Datadog.',
  })
```

src/commands/synthetics/utils/public.ts
```diff
@@ -281,6 +281,8 @@ const getBatch = async (api: APIHelper, trigger: Trigger): Promise<Batch> => {
 
 const getTestByPublicId = (id: string, tests: Test[]): Test => tests.find((t) => t.public_id === id)!
 
+export const normalizePublicId = (id: string) => id.match(PUBLIC_ID_REGEX)?.[0]
```
Drarig29, line 284: I thought we had an ESLint rule about that, but for `export`ed functions, we always want to have an explicit return type

src/commands/synthetics/run-tests-lib.ts
```diff
@@ -230,7 +231,7 @@ export const getTestsList = async (
     .map((suite) =>
       suite.content.tests.map((test) => ({
         config: overrideTestConfig(test),
-        id: test.id,
+        id: normalizePublicId(test.id) || '',
```
Drarig29, line 234: **nit** since your returned value cannot be an empty string (only possibly `undefined`) the `??` operator is better suited
```suggestion
        id: normalizePublicId(test.id) ?? '',
```

teodor2312: I can't remember if i've ever used this operator before 😂

src/commands/synthetics/utils/public.ts
teodor2312: ah wait never mind i thing i was doing it wrong 😂

Drarig29 (approved):
LGTM

PR #1196, opened 2024-02-15
[SYNTH-12557] Add missing deviceIds configurations for Bitrise integrations

teodor2312:
### What and why?

For the Bitrise integrations we need two configurations that are mentioned in datadog-ci but not really implemented

### How?

Adding the 2nd of these 2 configurations, `deviceIds`, to be able to configured through Global Config, env variables, cli param and Test Config. 
We agreed that we want to align all the configurations to be able to be configured in those 4 ways in this doc - https://datadoghq.atlassian.net/wiki/spaces/SYN/pages/3378446383/Configuration+parameters+and+their+usages
so this is a good opportunity to do this and create an example for the rest.

### Review checklist


Drarig29 (approved):
LGTM

PR #1197, opened 2024-02-15
Keep envvars consistent

TophrC-dd:
### What and why?
Currently `datadog-ci` currently looks between two variable for the datadog site. 
`DATADOG_SITE` and `DD_SITE` except it doesn't always look at `DD_SITE` for some commands. 
This can be frustrating especially where most documentation states to use `DD_SITE` which is heavily documented on the serverless documentation. 
This PR is an attempt to be consistent in configuring the site so no matter which variable is used it will work across every command. 

### How?
I have made modifications to the helper api module that implements a new exported constant called `datadogSite` which check both `DATADOG_SITE` and `DD_SITE`. 
I also ensured that this order is kept if a customer preferred `DATADOG_SITE` that would be the first variable used. 
However If neither environment variables are set then the default shall be `DATADOG_SITE_US1` which is `datadoghq.com`. 

I have also made updates to the commands to use this new export whereas before it would try to read either `DATADOG_SITE` or `DD_SITE` and sometimes both. 

### Review checklist
Because this make a change across all commands the current tests will validate that it is working properly.


src/commands/gate/utils.ts
```diff
@@ -1,18 +1,15 @@
+import {datadogSite} from '../../helpers/api'
 import {getCommonAppBaseURL} from '../../helpers/app'
 
 export const getBaseUrl = () => {
-  const site = process.env.DD_SITE || 'datadoghq.com'
+  const site = process.env.DATADOG_SITE || process.env.DD_SITE || 'datadoghq.com'
```
purple4reina, line 5: I'm guessing this one should also be changed to use the imported value?

TophrC-dd: Oh I missed this, I orginally was going to target the git-metadata command as this was the command that inspired me to make the code change. I wanted to be a bit more ambitious after. 

I will fix this.

src/helpers/api.ts
```diff
@@ -1,13 +1,11 @@
 import {DATADOG_SITE_US1} from '../constants'
 
+export const datadogSite = process.env.DATADOG_SITE || process.env.DD_SITE || DATADOG_SITE_US1
```
duncanista, line 3: I wonder if this should be uppercase, since it's a constant widely used?

```suggestion
export const DATADOG_SITE = process.env.DATADOG_SITE || process.env.DD_SITE || DATADOG_SITE_US1
```

Drarig29: Since we have `getBaseIntakeUrl()` in the file, I would prefer to have a `getDatadogSite()` function instead. Otherwise it may not be clear that this "constant" is tied to an environment _variable_

src/commands/dora/api.ts
```diff
@@ -1,11 +1,11 @@
 import type {AxiosPromise, AxiosRequestConfig, AxiosResponse} from 'axios'
 
+import {getDatadogSite} from '../../helpers/api'
 import {getRequestBuilder} from '../../helpers/utils'
 
 import {DeploymentEvent} from './interfaces'
 
-export const datadogSite = process.env.DD_SITE || 'datadoghq.com'
-export const apiUrl = `https://api.${datadogSite}`
+export const apiUrl = `https://api.${getDatadogSite}`
```
Drarig29, line 8: I'm impressed this hasn't made any unit test fail 😅 
```suggestion
export const apiUrl = `https://api.${getDatadogSite()}`
```

src/commands/gate/utils.ts
```diff
@@ -1,18 +1,14 @@
...
+  return getCommonAppBaseURL(getDatadogSite(), subdomain)
 }
 
 export const getBaseIntakeUrl = () => {
-  if (process.env.DD_SITE) {
-    return `https://quality-gates.${process.env.DD_SITE}`
-  }
-
-  return 'https://quality-gates.datadoghq.com'
+  return `https://quality-gates.${getDatadogSite}`
```
Drarig29, line 11: same here

src/helpers/api.ts
```diff
@@ -1,13 +1,17 @@
...
  * Get the base intake URL for a service. If the `DD_SITE` or `DATADOG_SITE` environment
  * variables are not defined, use the default site (US1).
  */
 export const getBaseIntakeUrl = (intake: string) => {
-  if (process.env.DATADOG_SITE || process.env.DD_SITE) {
-    return `https://${intake}.${process.env.DATADOG_SITE || process.env.DD_SITE}`
-  }
-
-  return `https://${intake}.${DATADOG_SITE_US1}`
+  return `https://${intake}.${getDatadogSite}`
```
Drarig29, line 16: and here

src/commands/dora/api.ts
Drarig29: We could add a test in the `describe('getApiHelper` of each command ensuring that the endpoint is the expected one

TophrC-dd: yeah absolutely!

src/commands/git-metadata/upload.ts
```diff
@@ -216,7 +217,7 @@ export class UploadCommand extends Command {
   private getApiRequestBuilder(apiKey: string): RequestBuilder {
     return getRequestBuilder({
       apiKey,
-      baseUrl: 'https://' + apiHost,
+      baseUrl: getBaseIntakeUrl('api'),
```
Drarig29, line 220: I don't think we should call the API endpoint an intake, those are 2 different things

Drarig29:
```suggestion
      baseUrl: `https://api.${getDatadogSite()}`,
```

src/helpers/api.ts
```diff
@@ -1,13 +1,17 @@
 import {DATADOG_SITE_US1} from '../constants'
 
+/**
+ *  Get the site domain by reading DATADOG_SITE or DD_SITE
+ *  Otherwise we go with US1 as default
+ */
```
Drarig29, line 6: **nit** To align with the comment below:
```suggestion
/**
 * Get Datadog site's domain. If the `DD_SITE` or `DATADOG_SITE` environment
 * variables are not defined, use the default site (US1). 
 */
```

Drarig29 (approved):
LGTM

rpelliard (approved):

src/commands/react-native/upload.ts
```diff
@@ -132,7 +133,6 @@ export class UploadCommand extends Command {
       this.config,
       {
         apiKey: process.env.DATADOG_API_KEY,
-        datadogSite: process.env.DATADOG_SITE,
```
louiszawadzki, line 135: This creates a small regression in the case where the site is both specified in the environment and the `datadog-ci.json` file with different values.

Before the order of priority used to be: env > config file > default value ('datadoghq.com')
Now since the site from env is put in the default value it will just be: config file  > env.

It's only a tiny case but that also introduces a different logic for api key and site variables, which can make debugging harder.

I would suggest the following change here (we can either keep or remove the change above):
```suggestion
        datadogSite: process.env.DATADOG_SITE || process.env.DD_SITE,
```

src/commands/flutter-symbols/upload.ts
```diff
@@ -139,7 +140,6 @@ export class UploadCommand extends Command {
       this.config,
       {
         apiKey: process.env.DATADOG_API_KEY,
-        datadogSite: process.env.DATADOG_SITE,
```
louiszawadzki, line 142: This creates a small regression in the case where the site is both specified in the environment and the `datadog-ci.json` file with different values.

Before the order of priority used to be: env > config file > default value ('datadoghq.com')
Now since the site from env is put in the default value it will just be: config file  > env.

It's only a tiny case but that also introduces a different logic for api key and site variables, which can make debugging harder.

I would suggest the following change here (we can either keep or remove the change above):
```suggestion
        datadogSite: process.env.DATADOG_SITE || process.env.DD_SITE,
```

src/commands/dsyms/upload.ts
```diff
@@ -77,7 +78,6 @@ export class UploadCommand extends Command {
       this.config,
       {
         apiKey: process.env.DATADOG_API_KEY,
-        datadogSite: process.env.DATADOG_SITE,
```
louiszawadzki, line 80: This creates a small regression in the case where the site is both specified in the environment and the `datadog-ci.json` file with different values.

Before the order of priority used to be: env > config file > default value ('datadoghq.com')
Now since the site from env is put in the default value it will just be: config file  > env.

It's only a tiny case but that also introduces a different logic for api key and site variables, which can make debugging harder.

I would suggest the following change here (we can either keep or remove the change above):
```suggestion
        datadogSite: process.env.DATADOG_SITE || process.env.DD_SITE,
```

louiszawadzki (changes requested):
This introduces a tiny regression for mobile rum upload commands since we can either get the site from env variables or a configuration file.
More importantly it introduces different behaviours for different env variables.

src/commands/git-metadata/upload.ts
```diff
@@ -216,7 +217,7 @@ export class UploadCommand extends Command {
   private getApiRequestBuilder(apiKey: string): RequestBuilder {
     return getRequestBuilder({
       apiKey,
-      baseUrl: 'https://' + apiHost,
+      baseUrl: `https://api.${getDatadogSite()}`,
```
rpelliard, line 220: nit: Small suggestion for consistency
```suggestion
      baseUrl: `https://api.${this.config.datadogSite}`,
```

Could you also update the line 225 below to use `this.config.datadogSite` instead ? 🙏

dastrong (approved):
Consistency nit, could we update `src/commands/sarif/utils.ts` to use this util?

BenoitZugmeyer (approved):
PR looks good

PR #1199, opened 2024-02-17
[STAL-1340] Add link to results

juli1:
### What and why?

Add a link to the results page to direct the user to our product page on datadog.

### How?

We are adding a link to the code-analysis dashboard to check the results.

### SARIF

```
...
✅ Uploaded 1 files in 0.336 seconds.
ℹ️ Results available on https://app.datad0g.com/ci/code-analysis
=================================================================================================
```

### SBOM

**Prod**
```
File /Users/julien.delange/datadog-ci-sbom.json successfully uploaded in 381 ms
Upload finished, please check results on https://app.datadoghq.com/ci/code-analysis
```

**Staging**

```
File /Users/julien.delange/datadog-ci-sbom.json successfully uploaded in 366 ms
Upload finished, results available on https://app.datad0g.com/ci/code-analysis
```

### DIFF
--- a/src/commands/sarif/renderer.ts
+++ b/src/commands/sarif/renderer.ts
@@ -4,6 +4,8 @@ import chalk from 'chalk'
 
 import {SpanTags} from '../../helpers/interfaces'
 
+import {getBaseUrl} from '../junit/utils'
+
 import {Payload} from './interfaces'
 
 const ICONS = {
@@ -45,6 +47,7 @@ export const renderSuccessfulCommand = (
 ) => {
   let fullStr = ''
   fullStr += chalk.green(`${ICONS.SUCCESS} Uploaded ${fileCount} files in ${duration} seconds.\n`)
+  fullStr += chalk.green(`${ICONS.INFO}  Results available on ${getBaseUrl()}ci/code-analysis\n`)
   fullStr += chalk.green(
     '=================================================================================================\n'
   )
--- a/src/commands/sbom/upload.ts
+++ b/src/commands/sbom/upload.ts
@@ -9,6 +9,8 @@ import {Command, Option} from 'clipanion'
 
 import {getSpanTags} from '../../helpers/tags'
 
+import {getBaseUrl} from '../junit/utils'
+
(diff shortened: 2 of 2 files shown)


src/commands/sarif/renderer.ts
```diff
@@ -45,6 +47,7 @@ export const renderSuccessfulCommand = (
 ) => {
   let fullStr = ''
   fullStr += chalk.green(`${ICONS.SUCCESS} Uploaded ${fileCount} files in ${duration} seconds.\n`)
+  fullStr += chalk.green(`${ICONS.INFO}  Results available on ${getBaseUrl()}ci/code-analysis\n`)
```
Drarig29, line 50: Could we make the link more specific with some query params? For example, since we give `--service` to the command, the link could at least filter on the service.

Example for JUnit uploads: https://github.com/DataDog/datadog-ci/blob/4e722da522276da8bf1089087d984dee8b74b837/src/commands/junit/renderer.ts#L46-L69

juli1: I can and thought about it. The reason I did not is because the page is still significantly changing and I wanted to prevent linking to a page that has URL params that may change quickly in the near future. But if you think this is necessary, I can change this.

juli1: Looked again at this and this is definitely something I do not want to do.

1. First, the results page for static analysis/sca are changing a lot. We risk to expose an outdated URL in the future.
2. The URL parameters use the repository id and calculating the repository ID from the CLI will lead to 404. We already have multiple issues relatd to repository ID and I really do not want to add more.

If we want to handle this correctly, we should let the frontend/backend accepts a repository URL and have the backend transforms to a consistent repository id but that is def outside the scope of this PR.

Drarig29: Ok 👌

Drarig29 (approved):

dastrong (approved):

dastrong (approved):
Approving. Only nit is that adding that results might take a second or two to appear, but otherwise it's good.

PR #1211, opened 2024-02-23
[CIVIS-9031] Deprecate 'metrics' naming in favour of 'measures' in junit cmd

ManuelPalenzuelaDD:
This PR implements the same thing as the reverted https://github.com/DataDog/datadog-ci/pull/1203, but with the regression fixed

### Review checklist

- [x] Tested in staging

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -100,6 +100,18 @@ jobs:
           DD_API_KEY: ${{ secrets.DD_API_KEY_CI_VISIBILITY }}
           DD_APP_KEY: ${{ secrets.DD_APP_KEY_CI_VISIBILITY }}
           DD_SERVICE: datadog-ci-e2e-tests-junit
+      - name: Run junit upload test (metrics)
+        run: yarn datadog-ci junit upload --service=datadog-ci-e2e-tests-junit artifacts/e2e/junit-reports --metrics key2:40 --report-metrics key2:40
+        env:
+          DD_API_KEY: ${{ secrets.DD_API_KEY_CI_VISIBILITY }}
+      - name: Check that test data can be queried (metrics)
+        run: |
+          yarn add @datadog/datadog-api-client
+          yarn check-junit-upload '@key2:40'
+        env:
+          DD_API_KEY: ${{ secrets.DD_API_KEY_CI_VISIBILITY }}
+          DD_APP_KEY: ${{ secrets.DD_APP_KEY_CI_VISIBILITY }}
+          DD_SERVICE: datadog-ci-e2e-tests-junit
       - name: Run sarif upload test
         run: yarn datadog-ci sarif upload --service=datadog-ci-e2e-tests-sarif artifacts/e2e/sarif-reports
         env:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -97,7 +97,7 @@ jobs:
           yarn add @datadog/datadog-api-client
           yarn check-junit-upload
         env:
-          EXTRA_QUERY_FILTER: key1:20
+          EXTRA_QUERY_FILTER: '@key1:20'
           DD_API_KEY: ${{ secrets.DD_API_KEY_CI_VISIBILITY }}
           DD_APP_KEY: ${{ secrets.DD_APP_KEY_CI_VISIBILITY }}
           DD_SERVICE: datadog-ci-e2e-tests-junit
@@ -110,7 +110,7 @@ jobs:
           yarn add @datadog/datadog-api-client
           yarn check-junit-upload
         env:
-          EXTRA_QUERY_FILTER: key2:40
+          EXTRA_QUERY_FILTER: '@key2:40'
           DD_API_KEY: ${{ secrets.DD_API_KEY_CI_VISIBILITY }}
           DD_APP_KEY: ${{ secrets.DD_APP_KEY_CI_VISIBILITY }}
           DD_SERVICE: datadog-ci-e2e-tests-junit
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,14 +88,26 @@ jobs:
         env:
           DATADOG_API_KEY: ${{ secrets.DATADOG_API_KEY_E2E }}
           DATADOG_APP_KEY: ${{ secrets.DATADOG_APP_KEY_E2E }}
-      - name: Run junit upload test
+      - name: Run junit upload test (test level with measures)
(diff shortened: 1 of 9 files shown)


hestonhoffman (approved):

juan-fernandez (changes requested):
LGTM, but could we add e2e tests so that we're protected against future regressions? 

We check junit uploads in https://github.com/DataDog/datadog-ci/blob/master/.github/workflows/ci.yml#L92 and https://github.com/DataDog/datadog-ci/blob/master/bin/check-junit-upload.js. 

Maybe we can add `--report-measures` and `--measures` to the `junit` command and query the API with those tags

bin/check-junit-upload.js
```diff
@@ -44,12 +47,13 @@ async function checkJunitUpload () {
...
-    console.log(`❌ Failed check: the API did not return any test for the given filter.`)
+    console.log(`❌ Failed check: the API did not return any ${testLevel}s for the given filter.`)
     process.exit(1)
   }
 }
 
-checkJunitUpload()
\ No newline at end of file
+checkJunitUpload("test", process.argv[2] || '')
+checkJunitUpload("session", process.argv[2] || '')
```
juan-fernandez, line 59: are we actually using `testLevel` filter somewhere? It seems we only use for logs: could we add it to the `filterQuery`?

juan-fernandez: I think it'd also be more readable to use an environment variable for `extraFilter` rather than + `process.argv[2]`. 

We could set an `env` option in the workflow file, something like `EXTRA_QUERY_FILTER`: 
```
env: 
  ...
  EXTRA_QUERY_FILTER: @key2:40
```
and then use 
```
checkJunitUpload("session", process.env.EXTRA_QUERY_FILTER)
```

ManuelPalenzuelaDD: > are we actually using testLevel filter somewhere? It seems we only use for logs: could we add it to the filterQuery?
Nice catch, somehow forgot to do that!

bin/check-junit-upload.js
```diff
@@ -27,12 +31,12 @@ function waitFor (waitSeconds) {
...
 } 
 
-async function checkJunitUpload () {
+async function checkJunitUpload (testLevel, extraFilter) {
   let numAttempts = 0
   let isSuccess = false
   let data = []
   while (numAttempts++ < MAX_NUM_ATTEMPTS && !isSuccess) {
-    data = await getTestData()
+    data = await getTestData(extraFilter + ` test_level:${testLevel}`)
```
Drarig29, line 39:
```suggestion
    data = await getTestData(`test_level:${testLevel} ${extraFilter}`)
```

.github/workflows/ci.yml
```diff
@@ -88,15 +88,29 @@ jobs:
         env:
           DATADOG_API_KEY: ${{ secrets.DATADOG_API_KEY_E2E }}
           DATADOG_APP_KEY: ${{ secrets.DATADOG_APP_KEY_E2E }}
-      - name: Run junit upload test
-        run: yarn datadog-ci junit upload --service=datadog-ci-e2e-tests-junit artifacts/e2e/junit-reports
+      - name: Run junit upload test (measures)
+        run: yarn datadog-ci junit upload --service=datadog-ci-e2e-tests-junit artifacts/e2e/junit-reports --measures key1:20 --report-measures key1:20
```
Drarig29, line 92: Would it make sense to have different measures on the session and test levels? This way we can test them separately! You can then change `EXTRA_QUERY_FILTER` into `EXTRA_SESSION_QUERY_FILTER` and `EXTRA_TEST_QUERY_FILTER`

bin/check-junit-upload.js
```diff
@@ -44,12 +48,13 @@ async function checkJunitUpload () {
...
-    console.log(`❌ Failed check: the API did not return any test for the given filter.`)
+    console.log(`❌ Failed check: the API did not return any ${testLevel}s for the given filter.`)
     process.exit(1)
   }
 }
 
-checkJunitUpload()
\ No newline at end of file
+checkJunitUpload("test", process.env.EXTRA_QUERY_FILTER)
+checkJunitUpload("session", process.env.EXTRA_QUERY_FILTER)
```
Drarig29, line 60: Nice!

Drarig29 (commented):
@ManuelPalenzuelaDD thanks for remaking the PR!

Can you point out what exactly is the difference between this new PR and the faulty one (https://github.com/DataDog/datadog-ci/pull/1203)? What broke exactly in the old one?

.github/workflows/ci.yml
juan-fernandez: two queries are made for two test levels, so even if the measures are the same, if one of the test levels (session or test) are not assigning the measures correctly, we'll detect it.

src/commands/junit/api.ts
```diff
@@ -33,13 +33,13 @@ export const uploadJUnitXML = (request: (args: AxiosRequestConfig) => AxiosPromi
 
   const reportTagsAndMetrics: Record<string, any> = {
     tags: jUnitXML.reportTags,
-    metrics: jUnitXML.reportMetrics,
+    metrics: jUnitXML.reportMeasures,
```
juan-fernandez, line 36: could we add a quick comment on top of `reportTagsAndMetrics` saying something like:

```javascript
// We can't change `metrics` to `measures` because the backend only accepts `metrics`
```

This is so that we can make sense of the name in the future

juan-fernandez (approved):
LGTM, thanks for the e2e tests

src/commands/junit/api.ts
ManuelPalenzuelaDD: Good idea!

.github/workflows/ci.yml
ManuelPalenzuelaDD: I think this would be testing the same functionality but I like your idea of having different measures! Ty :)

ManuelPalenzuelaDD:
> Can you point out what exactly is the difference between this new PR and the faulty one (https://github.com/DataDog/datadog-ci/pull/1203)? What broke exactly in the old one?

Hi @Drarig29 , thank you for taking a look at this. The problem with the old PR is that we were [renaming a field](https://github.com/DataDog/datadog-ci/pull/1203/files#diff-912743e760d20d6462cea4b81a0b0f71e6d86be0698eb0312f38fc4ff6f9b633R36) that was used in the body of an api call. This was done by mistake and the api didn't recognize that field name.

Drarig29 (approved):
LGTM

PR #1213, opened 2024-02-28
[git-metadata] Make exported function upload to gitdb

rpelliard:
Currently, the `datadog-ci git-metadata upload` command both:
- synchronises GitDB
- uploads a payload with tracked files to the legacy sourcemap intake

But the `uploadGitCommitSha` function that is exported and used in the serverless plugin (here: https://github.com/DataDog/serverless-plugin-datadog/blob/51b7844efd6224d1a4fb5c1215453aac3038f915/src/index.ts#L302) only upload to the sourcemap intake.

We want all upload tooling to synchronize GitDB, and eventually deprecate the sourcemap store for Git metadata.

This PR updates the exported function to also synchronize GitDB.

### DIFF
--- a/src/commands/git-metadata/library.ts
+++ b/src/commands/git-metadata/library.ts
@@ -1,10 +1,14 @@
+import {SimpleGit} from 'simple-git'
+
 import {newApiKeyValidator} from '../../helpers/apikey'
 import {RequestBuilder} from '../../helpers/interfaces'
+import {Logger, LogLevel} from '../../helpers/logger'
 import {upload, UploadOptions, UploadStatus} from '../../helpers/upload'
 import {getRequestBuilder, filterAndFormatGithubRemote} from '../../helpers/utils'
 import {version} from '../../helpers/version'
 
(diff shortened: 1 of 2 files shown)


src/commands/git-metadata/library.ts
```diff
@@ -33,21 +37,41 @@ export const getGitCommitInfo = async (filterAndFormatGitRepoUrl = true): Promis
   return [gitRemote ?? '', payload.hash]
 }
 
-// UploadGitCommitHash uploads local git metadata and returns the current [repositoryURL, commitHash].
+// uploadGitCommitHash uploads local git metadata and returns the current [repositoryURL, commitHash].
```
samarcan, line 40: why are you lowercasing the first letter? was it wrong?

samarcan (approved):

src/commands/git-metadata/library.ts
rpelliard: It's just the comment, so I'm updating it to match the function name :)

PR #1214, opened 2024-02-28
[SYNTH-13000] Fix bug where Test Config was not being used when a public_id was already defined

teodor2312:
### What and why?

Currently the logic that fetches the Test Config and uses to override the previous configurations only gets executed if publicIds were not defined before that point through the Global Level Configuration (Global Config, Env Variables, CLI params). It should happen always instead, so fix that.

### How?

Moved out the code that fetches the Test Config out of that if statement and did some small refactoring. Also added tests to make sure that we account for that edge case

### Review checklist

### DIFF
--- a/src/commands/synthetics/__tests__/run-tests-lib.test.ts
+++ b/src/commands/synthetics/__tests__/run-tests-lib.test.ts
@@ -516,16 +516,17 @@ describe('run-test', () => {
     })
   })
 
-  describe('getTestsList', () => {
+  describe('getFinalVersionOfTestsFromCi', () => {
     beforeEach(() => {
       jest.restoreAllMocks()
     })
 
+    const startUrl = 'fakeUrl'
     const conf1 = {
-      tests: [{config: {}, id: 'abc-def-ghi'}],
+      tests: [{config: {deviceIds: ['chrome.laptop_large']}, id: 'abc-def-ghi'}],
     }
     const conf2 = {
-      tests: [{config: {}, id: 'jkl-mno-pqr'}],
+      tests: [{config: {startUrl: 'someOtherFakeUrl'}, id: 'jkl-mno-pqr'}],
     }
     const fakeSuites = [
       {
@@ -537,7 +538,6 @@ describe('run-test', () => {
         name: 'Suite 2',
       },
     ]
-    const startUrl = 'fakeUrl'
     const fakeApi = {
       searchTests: () => ({
         tests: [
@@ -548,22 +548,54 @@ describe('run-test', () => {
       }),
     } as any
 
-    test('should find all tests and extend global config', async () => {
+    test('should extend global config and execute all tests from Test Config when no publicIds were defined', async () => {
       jest.spyOn(utils, 'getSuites').mockImplementation((() => fakeSuites) as any)
       const configOverride = {startUrl}
 
       await expect(
-        runTests.getTestsList(fakeApi, {...ciConfig, global: configOverride}, mockReporter)
+        runTests.getFinalVersionOfTestsFromCi(
+          fakeApi,
+          {...ciConfig, global: configOverride, locations: ['aws:ap-northeast-1']},
+          mockReporter
+        )
       ).resolves.toEqual([
         {
-          config: {startUrl},
+          config: {deviceIds: ['chrome.laptop_large'], startUrl, locations: ['aws:ap-northeast-1']},
           id: 'abc-def-ghi',
           suite: 'Suite 1',
         },
         {
-          config: {startUrl},
+          config: {startUrl: 'someOtherFakeUrl', locations: ['aws:ap-northeast-1']},
           id: 'jkl-mno-pqr',
           suite: 'Suite 2',
+        },
+      ])
+    })
+
+    test('should override and execute only publicIds that were defined in the global config', async () => {
+      jest.spyOn(utils, 'getSuites').mockImplementation((() => fakeSuites) as any)
+      const configOverride = {startUrl}
+
+      await expect(
+        runTests.getFinalVersionOfTestsFromCi(
+          fakeApi,
+          {
+            ...ciConfig,
+            global: configOverride,
+            locations: ['aws:ap-northeast-1'],
+            publicIds: ['abc-def-ghi', '123-456-789'],
+          },
+          mockReporter
+        )
+      ).resolves.toEqual([
+        {
+          config: {deviceIds: ['chrome.laptop_large'], startUrl, locations: ['aws:ap-northeast-1']},
+          id: 'abc-def-ghi',
+          suite: 'Suite 1',
+        },
+        {
+          config: {startUrl, locations: ['aws:ap-northeast-1']},
+          id: '123-456-789',
         },
       ])
     })
@@ -574,16 +606,34 @@ describe('run-test', () => {
       const searchQuery = 'fake search'
 
       await expect(
-        runTests.getTestsList(
+        runTests.getFinalVersionOfTestsFromCi(
           fakeApi,
-          {...ciConfig, global: configOverride, testSearchQuery: searchQuery},
+          {...ciConfig, global: configOverride, locations: ['aws:ap-northeast-1'], testSearchQuery: searchQuery},
+          mockReporter
+        )
+      ).resolves.toEqual([
+        {
+          config: {startUrl, locations: ['aws:ap-northeast-1']},
+          id: 'stu-vwx-yza',
+          suite: 'Query: fake search',
+        },
+      ])
+    })
+
+    test('should not use testSearchQeury if global config has defined public_ids', async () => {
+      const configOverride = {startUrl}
+      const searchQuery = 'fake search'
+
+      await expect(
+        runTests.getFinalVersionOfTestsFromCi(
+          fakeApi,
+          {...ciConfig, global: configOverride, publicIds: ['abc-def-ghi'], testSearchQuery: searchQuery},
           mockReporter
         )
       ).resolves.toEqual([
         {
           config: {startUrl},
-          id: 'stu-vwx-yza',
-          suite: 'Query: fake search',
+          id: 'abc-def-ghi',
         },
       ])
     })
@@ -597,7 +647,7 @@ describe('run-test', () => {
 
       const searchQuery = 'fake search'
 
-      await runTests.getTestsList(apiHelper, {...ciConfig, testSearchQuery: searchQuery}, mockReporter)
+      await runTests.getFinalVersionOfTestsFromCi(apiHelper, {...ciConfig, testSearchQuery: searchQuery}, mockReporter)
       expect(mockReporter.error).toHaveBeenCalledWith(
         `More than ${MAX_TESTS_TO_TRIGGER} tests returned by search query, only the first ${MAX_TESTS_TO_TRIGGER} will be fetched.\n`
       )
@@ -608,7 +658,7 @@ describe('run-test', () => {
       const configOverride = {startUrl}
       const files = ['new glob', 'another one']
 
-      await runTests.getTestsList(fakeApi, {...ciConfig, global: configOverride, files}, mockReporter)
+      await runTests.getFinalVersionOfTestsFromCi(fakeApi, {...ciConfig, global: configOverride, files}, mockReporter)
       expect(getSuitesMock).toHaveBeenCalledTimes(2)
       expect(getSuitesMock).toHaveBeenCalledWith('new glob', mockReporter)
       expect(getSuitesMock).toHaveBeenCalledWith('another one', mockReporter)
@@ -619,15 +669,15 @@ describe('run-test', () => {
       const configOverride = {startUrl}
       const files: string[] = []
 
-      const tests = await runTests.getTestsList(
+      const tests = await runTests.getFinalVersionOfTestsFromCi(
         fakeApi,
         {...ciConfig, global: configOverride, files},
         mockReporter,
         fakeSuites
       )
       expect(tests).toEqual([
-        {config: {startUrl}, id: conf1.tests[0].id, suite: fakeSuites[0].name},
-        {config: {startUrl}, id: conf2.tests[0].id, suite: fakeSuites[1].name},
+        {config: {deviceIds: ['chrome.laptop_large'], startUrl}, id: conf1.tests[0].id, suite: fakeSuites[0].name},
+        {config: {startUrl: 'someOtherFakeUrl'}, id: conf2.tests[0].id, suite: fakeSuites[1].name},
       ])
       expect(getSuitesMock).not.toHaveBeenCalled()
     })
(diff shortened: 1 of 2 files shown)


src/commands/synthetics/run-tests-lib.ts
```diff
@@ -192,52 +168,117 @@ const getTestListBySearchQuery = async (
   }))
 }
 
-export const getTestsList = async (
+export const getFinalVersionOfTestsFromCi = async (
   api: APIHelper,
   config: RunTestsCommandConfig,
   reporter: MainReporter,
-  suites: Suite[] = []
+  suites?: Suite[]
```
jakepruitt, line 175: Why change this signature?

src/commands/synthetics/run-tests-lib.ts
```diff
@@ -192,52 +168,117 @@ const getTestListBySearchQuery = async (
...
 ) => {
-  // If "testSearchQuery" is provided, always default to running it.
-  if (config.testSearchQuery) {
-    const testsToTriggerBySearchQuery = await getTestListBySearchQuery(api, config.global, config.testSearchQuery)
+  let testConfig: TriggerConfig[]
```
jakepruitt, line 177:
```suggestion
  let testConfigs: TriggerConfig[]
```

src/commands/synthetics/run-tests-lib.ts
```diff
@@ -192,52 +168,117 @@ const getTestListBySearchQuery = async (
...
+  if (testsToTriggerBySearchQuery.length > MAX_TESTS_TO_TRIGGER) {
+    reporter.error(
+      `More than ${MAX_TESTS_TO_TRIGGER} tests returned by search query, only the first ${MAX_TESTS_TO_TRIGGER} will be fetched.\n`
+    )
   }
 
+  return testsToTriggerBySearchQuery
+}
+
+const getTestConfig = async (
```
jakepruitt, line 245:
```suggestion
const getTestConfigs = async (
```

src/commands/synthetics/run-tests-lib.ts
jakepruitt: Or (as you have in the other function) `testsFromTestConfigs`

src/commands/synthetics/run-tests-lib.ts
```diff
@@ -192,52 +168,117 @@ const getTestListBySearchQuery = async (
   }))
 }
 
-export const getTestsList = async (
+export const getFinalVersionOfTestsFromCi = async (
```
jakepruitt, line 171: I appreciate the specificity, but why `FromCi` here? Maybe a `getFinalVersionOfTests` would be sufficient?

jakepruitt (approved):
Looks good, barring some naming nits

src/commands/synthetics/__tests__/run-tests-lib.test.ts
```diff
@@ -548,46 +548,96 @@ describe('run-test', () => {
...
         {
-          config: {startUrl},
+          config: {startUrl, locations: ['aws:ap-northeast-1']},
           id: 'stu-vwx-yza',
           suite: 'Query: fake search',
         },
       ])
     })
 
+    test('should not use testSearchQeury if global config has defined public_ids', async () => {
```
etnbrd, line 623: nit:
```suggestion
    test('should not use testSearchQuery if global config has defined public_ids', async () => {
```

src/commands/synthetics/run-tests-lib.ts
etnbrd, line 171: nit, I would suggest to use the name we are going to use in the config files: `test config override`, what do you think?
```suggestion
export const getTestConfigOverrideList = async (
```

src/commands/synthetics/run-tests-lib.ts
```diff
@@ -192,52 +168,117 @@ const getTestListBySearchQuery = async (
...
-      ...configFromEnvironment,
-      ...test.config,
-    })
-
-  const testsToTrigger = suites
+  const testsFromTestConfig = suites
     .map((suite) =>
       suite.content.tests.map((test) => ({
-        config: overrideTestConfig(test),
+        config: {...test.config},
```
etnbrd, line 259: Do we need this spread here? Can't we just pass `test.config`?
```suggestion
        config: test.config,
```

src/commands/synthetics/run-tests-lib.ts
etnbrd, line 245: The notion of `TestConfig` is getting a bit confusing. What do you think of this naming?
```suggestion
const getTestConfigOverrideFromTestFiles = async (
```

src/commands/synthetics/run-tests-lib.ts
```diff
@@ -192,52 +168,117 @@ const getTestListBySearchQuery = async (
...
+  if (!testsToTrigger.length && config.testSearchQuery) {
+    try {
+      testsToTrigger = await getTestsFromSearchQuery(api, config, reporter)
+      testsToTrigger = testsToTrigger.map((test) => ({
+        ...test,
+        config: {
+          ...configFromEnvironment,
+          ...test.config,
+        },
+      }))
```
etnbrd, line 198: This part is a bit confusing to me, especially given the now duplicated overrides (between `config.global` and `configFromEnvironment`, between `configFromEnvironment` and `test.config`, and between `config.global`, `configFromEnvironment` and `test.config`). Could we try to merge these three overrides together?

The following is probably over-simplistic, and missing a lot of edge cases I couldn't foresee, but hopefully it gets the idea across, and we can talk about it.

```typescript
// grab all the test config overrides from different sources
const defaultTestConfigOverride = config.global
const testConfigOverridesFromEnv = config.locations?.length ? {locations: config.locations} : {}
const testConfigOverridesFromTestFiles = getTestConfigOverridesFromTestFiles(config, reporter, suites)

// grab the list of tests to trigger from command line, config file, search query or test config files
const testsToTrigger =
  config.publicIds ??
  config.testSearchQuery && await getTestsFromSearchQuery(api, config, reporter) ??
  testConfigOverridesFromTestFiles.map(({id}) => id)

// Create the overrides required for the list of tests to trigger
const testConfigOverrides = testsToTrigger.map(id => ({
  id,
  config: {
    ...defaultTestConfigOverride,
    ...testConfigOverridesFromEnv,
    ...testConfigOverridesFromTestFiles.find(testConfigOverrideFromTestFile => testConfigOverrideFromTestFile.id === id) ?? {}
  }
})
```

etnbrd (commented):
Thanks a lot for pushing this through, I appreciate the effort to improve this code base 🙏
I left a big comment trying to help make this code a bit less confusing (at least to me). Happy to talk about it to help you implement that further, or let me know if I missed something.
Thanks 🙇

src/commands/synthetics/run-tests-lib.ts
teodor2312: Suits is only used in `getTestConfig` so i moved the `Suite[] = []` to there 
https://github.com/DataDog/datadog-ci/pull/1214/files#diff-d3548a697d009bf2222ed857f083b39afd505e05f34c1b3686686d7d05c6ff81R248

and made the signature of this function match the one it is called from

src/commands/synthetics/run-tests-lib.ts
teodor2312: artefact from earlier when i was spreading something else in there as well 👍 good catch

src/commands/synthetics/run-tests-lib.ts
teodor2312: I like it, there's a card later on (https://datadoghq.atlassian.net/browse/SYNTH-12972) to rename `config` from the `TestConfig` to `testOverrides`, i think this renaming might be better as part of that work though 🤔 will add it as a DoD 👍 
For now i would prefer to keep the consistency 👍

src/commands/synthetics/run-tests-lib.ts
teodor2312: will put it as `testConfigs` everywhere i think to keep it consistent 👍

src/commands/synthetics/run-tests-lib.ts
teodor2312: Hmm i had the `FromCi` bit in there since later on we get more configuration information from the test details in Datadog. So after this function the test won't be changed by configuration from the ci anymore but they will be enriched some more from datadog later on 🤔 don't know if that distinction is enough to keep the `FromCi` though

etnbrd: I would lean toward the wording `TestOverride` rather than `FromCI`: this is the synthetics `RunTestsLib` which could be called outside of CI, strictly speaking.

src/commands/synthetics/run-tests-lib.ts
```diff
@@ -52,37 +52,22 @@ export const executeTests = async (
...
-    } catch (error) {
-      throw new CriticalError(
-        isForbiddenError(error) ? 'AUTHORIZATION_ERROR' : 'UNAVAILABLE_TEST_CONFIG',
-        error.message
-      )
-    }
+  try {
+    triggerConfigs = await getTriggerConfigs(api, config, reporter, suites)
+  } catch (error) {
+    throw new CriticalError(isForbiddenError(error) ? 'AUTHORIZATION_ERROR' : 'UNAVAILABLE_TEST_CONFIG', error.message)
```
etnbrd, line 67: I love that this error handling is at the same level as the one for `NO_TESTS_TO_RUN` 🙌

src/commands/synthetics/run-tests-lib.ts
```diff
@@ -191,53 +176,86 @@ const getTestListBySearchQuery = async (
...
-  if (config.testSearchQuery) {
-    const testsToTriggerBySearchQuery = await getTestListBySearchQuery(api, config.global, config.testSearchQuery)
-
-    if (testsToTriggerBySearchQuery.length > MAX_TESTS_TO_TRIGGER) {
-      reporter.error(
-        `More than ${MAX_TESTS_TO_TRIGGER} tests returned by search query, only the first ${MAX_TESTS_TO_TRIGGER} will be fetched.\n`
-      )
+  suites?: Suite[]
+): Promise<TriggerConfig[]> => {
+  const defaultTestConfigOverrides = config.global
```
etnbrd, line 185:
```suggestion
  // Grab the test config overrides from all the sources: default test config override test files containing specific test config override , env variable, and cli params
  const defaultTestConfigOverrides = config.global
```

src/commands/synthetics/run-tests-lib.ts
```diff
@@ -191,53 +176,86 @@ const getTestListBySearchQuery = async (
...
+  // TODO: Clean up locations as part of SYNTH-12989
+  const testConfigOverridesFromEnv = config.locations?.length ? {locations: config.locations} : {}
+  const testsFromSearchQuery = config.testSearchQuery ? await getTestsFromSearchQuery(api, config, reporter) : []
+  const testsFromTestConfigs = await getTestConfigs(config, reporter, suites)
+
```
etnbrd, line 190: Let's move this line to another paragraph, with a comment:
```suggestion
  const testsFromTestConfigs = await getTestConfigs(config, reporter, suites)

  // Grab the test defined from the search query. Their config will contain the suite name, and the search query itself.
  const testsFromSearchQuery = config.testSearchQuery ? await getTestsFromSearchQuery(api, config, reporter) : []

```

Also, for later, we might want to split this `getTestsFromSearchQuery` function into two:
- `getTestIdsFromSearchQuery` returning the list of public ids
- `getAdditionalPropertiesFromSearchQuery` returning the suite name, and the search query, or `{}` if no search query is defined.
This would avoid the need for the find later on, and be more straightforward: `const testFromSearchQuery = testsFromSearchQuery.find((test) => test.id === id)`

src/commands/synthetics/run-tests-lib.ts
```diff
@@ -191,53 +176,86 @@ const getTestListBySearchQuery = async (
...
+  // Grab the list of publicIds of tests to trigger from config file/env variables/CLI params, search query or test config files
+  const testIdsFromCli = config.publicIds
+  const testIdsFromSearchQuery = testsFromSearchQuery.map(({id}) => id)
+  const testIdsFromTestConfigs = testsFromTestConfigs.map(({id}) => id)
+
+  const testIdsToTrigger =
```
etnbrd, line 196: Let's add some comments
```suggestion
  // Pick the list of tests to run from the first source defining it, in order of precedence
  const testIdsToTrigger =
```

src/commands/synthetics/run-tests-lib.ts
```diff
@@ -191,53 +176,86 @@ const getTestListBySearchQuery = async (
...
+    testIdsFromCli.length !== 0
+      ? testIdsFromCli
+      : testIdsFromSearchQuery.length !== 0
+      ? testIdsFromSearchQuery
+      : testIdsFromTestConfigs.length !== 0
+      ? testIdsFromTestConfigs
+      : []
```
etnbrd, line 203: After giving it some thought, we might have been going in the wrong direction with this last change.
If the search query is defined but returns no tests, should it:
- continue with the test configs?
- return an empty list to throw the 'no tests to run' error later?
It might more intuitive for it to be the last case, so we can really use `undefined` to mean that the list is not defined, and `[]` to mean that the list is defined but contain no tests, which will lead to an error.

In the latter case, we might want to add these modifications:
```suggestion
    testIdsFromCli ??  testIdsFromSearchQuery ?? testIdsFromTestConfigs ?? []
```

```diff
- const testsFromSearchQuery = config.testSearchQuery ? await getTestsFromSearchQuery(api, config, reporter) : []
+ const testsFromSearchQuery = config.testSearchQuery ? await getTestsFromSearchQuery(api, config, reporter)
```

```diff
- const testFromSearchQuery = testsFromSearchQuery.find((test) => test.id === id)
+ const testFromSearchQuery = testsFromSearchQuery?.find((test) => test.id === id)
```

src/commands/synthetics/run-tests-lib.ts
```diff
@@ -191,53 +176,86 @@ const getTestListBySearchQuery = async (
...
-        config: overrideTestConfig(test),
+        config: test.config,
         id: normalizePublicId(test.id) ?? '',
         suite: suite.name,
       }))
     )
     .reduce((acc, suiteTests) => acc.concat(suiteTests), [])
 
-  return testsToTrigger
+  return TestConfigs
```
etnbrd, line 258: In typescript, we use capitalised name only for types and classes (which are a kind of type):
```suggestion
  return testConfigs
```

src/commands/synthetics/run-tests-lib.ts
```diff
@@ -191,53 +176,86 @@ const getTestListBySearchQuery = async (
...
+  return triggerConfigs
+}
+
+const getTestsFromSearchQuery = async (api: APIHelper, config: RunTestsCommandConfig, reporter: MainReporter) => {
+  const testsToTriggerBySearchQuery = await getTestListBySearchQuery(api, config.global, config.testSearchQuery || '')
+
+  if (testsToTriggerBySearchQuery.length > MAX_TESTS_TO_TRIGGER) {
+    reporter.error(
+      `More than ${MAX_TESTS_TO_TRIGGER} tests returned by search query, only the first ${MAX_TESTS_TO_TRIGGER} will be fetched.\n`
+    )
```
etnbrd, line 231: Let's not change anything in this PR, but I wonder if this is necessary, since it's handled as well [here](https://github.com/DataDog/datadog-ci/blob/17e3c7f6e3a1726477db9b78c85f762010049314/src/commands/synthetics/utils/public.ts#L714).
Maybe we could simplify this later?

etnbrd (approved):
Awesome work! 🙌 
I left a few comments, mostly non-blocking nits.
I'm not sure about the behaviour to adopt when the search query returns no tests.

src/commands/synthetics/run-tests-lib.ts
teodor2312: ah yup must have mistyped it (ba-dum-ts) :D

src/commands/synthetics/run-tests-lib.ts
teodor2312: If you give a wrong `--public-id` in the CLI, so no such test was found, we return `no tests to run` so it does make sense that if you give a `--search` that returns nothing for it to also give `no tests to run`, which would be different than not having a `--search` defined at all in the first place, yes 🤔

teodor2312: We got together and found a nice way to do it 👍

src/commands/synthetics/run-tests-lib.ts
teodor2312: In the end went with `getTriggerConfigs` as the result of this function is referred as `triggerConfigs` later one also it describes well what the function does

PR #1216, opened 2024-03-04
[CIVIS] Add contains deployment tag to deployment mark command

rodrigo-roca:
### What and why?

The deployment mark command is basically a wrapper on top of the tag command. This PR changes this "wrapper". It adds the tag `ci.contains_deployment: true` to pipeline level. After these changes the command will:
* One request to attach tags to the job (with whatever the user wants to attach to the deployment such as service or env). No changes here
* One request to attach tags to the pipeline (for now with only `contains_deployment`). This is the new behavior

I have seen that when both tag commands succeed they print `Tags sent` twice now (since they are 2 commands). This should not be a problem right? Or should we make a "silent" option for the tag command?

### How?

### Review checklist


juan-fernandez (approved):
> I have seen that when both tag commands succeed they print Tags sent twice now (since they are 2 commands). This should not be a problem right? Or should we make a "silent" option for the tag command?

it's probably a good idea, as I don't think it's a great experience to see the message twice (it leaks an implementation detail to the user). The command is still in beta, right? We can do changes later on. 

Also: we might want to consider e2e tests at some point.

PR #1239, opened 2024-03-18
[SYNTH-13307] Add retries for 429

teodor2312:
### What and why?

Sometimes we get a Too Many Requests Error (429) on some of the endpoints we call in datadog-ci. We decided that we want to add some retries for those cases to counteract us hitting a particularly high load.
We're doing this since what the user would most probably do after getting the failure is just retry again so saving them that one extra step.

Currently it will retry max 3 times with a 5sec delay between retries, which also has some jitter.

### How?

Refactored our retry policy, to not take a function as part of its parameters, in this way simplifying it a bit and then added the case for 429s with a new type of delay for the jitter.

Then added tests 👍 

### Review checklist

### DIFF
--- a/src/commands/synthetics/__tests__/api.test.ts
+++ b/src/commands/synthetics/__tests__/api.test.ts
@@ -108,6 +108,7 @@ describe('dd-api', () => {
         makeApiRequest: () => api.getBatch('batch-id'),
         name: 'get batch' as const,
         shouldBeRetriedOn404: true,
+        shouldBeRetriedOn429: true,
         shouldBeRetriedOn5xx: true,
       },
       {
@@ -115,30 +116,35 @@ describe('dd-api', () => {
           api.getMobileApplicationPresignedURLs('applicationId', 1025, MOBILE_PRESIGNED_UPLOAD_PARTS),
         name: 'get presigned url' as const,
         shouldBeRetriedOn404: false,
+        shouldBeRetriedOn429: false,
         shouldBeRetriedOn5xx: true,
       },
       {
         makeApiRequest: () => api.getTunnelPresignedURL(['test-id']),
         name: 'get presigned url' as const,
         shouldBeRetriedOn404: false,
+        shouldBeRetriedOn429: false,
         shouldBeRetriedOn5xx: true,
       },
       {
         makeApiRequest: () => api.getTest('public-id'),
         name: 'get test' as const,
         shouldBeRetriedOn404: false,
+        shouldBeRetriedOn429: true,
         shouldBeRetriedOn5xx: true,
       },
       {
         makeApiRequest: () => api.pollResults(['result-id']),
         name: 'poll results' as const,
         shouldBeRetriedOn404: true,
+        shouldBeRetriedOn429: true,
         shouldBeRetriedOn5xx: true,
       },
       {
         makeApiRequest: () => api.searchTests('search query'),
         name: 'search tests' as const,
         shouldBeRetriedOn404: false,
+        shouldBeRetriedOn429: false,
         shouldBeRetriedOn5xx: true,
       },
       {
@@ -146,19 +152,21 @@ describe('dd-api', () => {
           api.triggerTests({tests: [{public_id: '123-456-789', executionRule: ExecutionRule.NON_BLOCKING}]}),
         name: 'trigger tests' as const,
         shouldBeRetriedOn404: false,
+        shouldBeRetriedOn429: true,
         shouldBeRetriedOn5xx: true,
       },
       {
         makeApiRequest: () => api.getSyntheticsOrgSettings(),
         name: 'get settings' as const,
         shouldBeRetriedOn404: false,
+        shouldBeRetriedOn429: false,
         shouldBeRetriedOn5xx: true,
       },
     ]
 
     test.each(cases)(
-      'should retry "$name" request (HTTP 404: $shouldBeRetriedOn404, HTTP 5xx: $shouldBeRetriedOn5xx)',
-      async ({makeApiRequest, shouldBeRetriedOn404, shouldBeRetriedOn5xx}) => {
+      'should retry "$name" request (HTTP 404: $shouldBeRetriedOn404, HTTP 429: $shouldBeRetriedOn429, HTTP 5xx: $shouldBeRetriedOn5xx)',
+      async ({makeApiRequest, shouldBeRetriedOn404, shouldBeRetriedOn429, shouldBeRetriedOn5xx}) => {
         const serverError = new Error('Server Error') as AxiosError
 
         const requestMock = jest.fn().mockImplementation(() => {
@@ -174,6 +182,18 @@ describe('dd-api', () => {
           await expect(requestPromise).rejects.toThrow('Server Error')
 
           expect(requestMock).toHaveBeenCalledTimes(shouldBeRetriedOn404 ? MAX_ATTEMPTS : MIN_ATTEMPTS)
+        }
+
+        requestMock.mockClear()
+
+        {
+          serverError.response = {status: 429} as AxiosResponse
+
+          const requestPromise = makeApiRequest()
+          await fastForwardRetries()
+          await expect(requestPromise).rejects.toThrow('Server Error')
+
+          expect(requestMock).toHaveBeenCalledTimes(shouldBeRetriedOn429 ? MAX_ATTEMPTS : MIN_ATTEMPTS)
         }
 
         requestMock.mockClear()
--- a/src/commands/synthetics/__tests__/api.test.ts
+++ b/src/commands/synthetics/__tests__/api.test.ts
@@ -438,26 +438,26 @@ describe('getApiHelper', () => {
 
 describe('formatBackendErrors', () => {
   test('backend error - no error', () => {
-    const backendError = getAxiosHttpError(500, {errors: []})
+    const backendError: AxiosError<any, any> = getAxiosHttpError(500, {errors: []})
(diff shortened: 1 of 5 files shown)


src/commands/synthetics/api.ts
```diff
@@ -259,27 +265,43 @@ const createMobileVersion = (request: (args: AxiosRequestConfig) => AxiosPromise
...
+  error: Error,
+  optionalRetries: OptionalRetries = {retryOn404: false, retryOn429: false}
+) => {
+  // Retry on 5xx
   if (retries < MAX_RETRIES && isNodeError(error)) {
     return LARGE_DELAY_BETWEEN_RETRIES
   }
 
-  if (retries < MAX_RETRIES && is5xxError(error)) {
+  if (retries < MAX_RETRIES && is5xxError(error as AxiosError<unknown, any> | EndpointError)) {
```
teodor2312, line 285: I'm not sure if there's not a clear way to do this 🤔

src/commands/synthetics/api.ts
```diff
@@ -259,27 +265,43 @@ const createMobileVersion = (request: (args: AxiosRequestConfig) => AxiosPromise
   return resp.data
 }
 
-type RetryPolicy = (retries: number, error: AxiosError) => number | undefined
+const retryWithJitter = (delay: number = DELAY_FOR_TOO_MANY_REQUESTS) => delay + Math.floor(Math.random() * delay)
+
+export type OptionalRetries = {
```
Drarig29, line 270: **nit** the object is singular
```suggestion
export type RetryPolicy = {
```

src/commands/synthetics/api.ts
```diff
@@ -259,27 +265,43 @@ const createMobileVersion = (request: (args: AxiosRequestConfig) => AxiosPromise
...
+  retryOn404?: boolean | undefined
+  retryOn429?: boolean | undefined
+}
 
-const retryOn5xxErrors: RetryPolicy = (retries, error) => {
-  // Retry on Node.js errors for both retry policies.
+export const determineRetryDelay = (
+  retries: number,
+  error: Error,
+  optionalRetries: OptionalRetries = {retryOn404: false, retryOn429: false}
```
Drarig29, line 278:
```suggestion
  retryPolicy: RetryPolicy = {retryOn404: false, retryOn429: false}
```

src/commands/synthetics/api.ts
```diff
@@ -259,27 +265,43 @@ const createMobileVersion = (request: (args: AxiosRequestConfig) => AxiosPromise
...
+) => {
+  // Retry on 5xx
```
Drarig29, line 280:
```suggestion
  // Retry on Node.js errors
```

src/commands/synthetics/api.ts
Drarig29, line 285:
```suggestion
  // Always retry on 5xx
  if (retries < MAX_RETRIES && is5xxError(error as AxiosError<unknown, any> | EndpointError)) {
```

Drarig29: Yes there is. Maybe `is5xxError()` should take either a `Error` or `unknown` and just return a boolean?

src/commands/synthetics/api.ts
teodor2312: For this one i want to express in some way that these are optional 🤔 and i feel like it's less of a policy and more configuration for one 🤔

src/commands/synthetics/api.ts
```diff
@@ -27,6 +27,7 @@ import {ciTriggerApp, getDatadogHost, retry} from './utils/public'
 const MAX_RETRIES = 3
 const DELAY_BETWEEN_RETRIES = 500 // In ms
 const LARGE_DELAY_BETWEEN_RETRIES = 1000 // In ms
+const DELAY_FOR_TOO_MANY_REQUESTS = 5000 // In ms (5s). Could be changed to the header returned for 429s later on
```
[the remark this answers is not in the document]
teodor2312: Will create a ticket 👍

src/commands/synthetics/api.ts
teodor2312: Tried it, but in the end we can't cuz later one we try to do `error.response` when trying to get the http status and `Property 'response' does not exist on type 'Error'`

teodor2312: Also i was using the 5xx as an example but we do the same casting for 404 and 429

src/commands/synthetics/api.ts
Drarig29: You have 2 levels of optionality, and IMO none of them deserves to have "optional" in the name:
- The 1st level is that you can omit the parameter altogether: `determineRetryDelay(retries, error)`
  - For this level, the "optionality" is conveyed by the type, and that's enough.
- The 2nd level is that each property in the object is optional... but it's actually only a boolean (which is optional for simplicity). In the end it's only answering "do you want to retry or not?".

Also, IMO there is a naming conflict between `retries: number` and `optionalRetries: OptionalRetries`. Why would adding an "optional" prefix make `: number` become `: OptionalRetries`?

---

As you said, you are _configuring_ the behavior for retries given the status code. So this could be named "RetryConfiguration"... but I think we use "configuration" far too much already.

And to come back to my suggestion, your refactor is exactly changing one implementation of retry policy with another implementation of retry policy. See [this](https://docs.servicenow.com/bundle/tokyo-application-development/page/administer/flow-designer/concept/retry-policy.html) and [this](https://www.webhooks.io/docs/relay/retry-policies/).

teodor2312: hmm okay those are good points 🤔

src/commands/synthetics/api.ts
```diff
@@ -27,7 +27,8 @@ import {ciTriggerApp, getDatadogHost, retry} from './utils/public'
 const MAX_RETRIES = 3
 const DELAY_BETWEEN_RETRIES = 500 // In ms
 const LARGE_DELAY_BETWEEN_RETRIES = 1000 // In ms
-const DELAY_FOR_TOO_MANY_REQUESTS = 5000 // In ms (5s). Could be changed to the header returned for 429s later on
+// SYNTH-13709: Use the `Retry-After` header.
+const DELAY_BETWEEN_429_RETRIES = 5000 // In ms (5s). Could be changed to the header returned for 429s later on
```
Drarig29, line 31: **nit**
- Either add comments everywhere to say `500ms, 1s, 5s` or only keep `In ms` everywhere, but stay consistent.
- And "Could be changed to the header returned for 429s later on" is redundant: the comment above with `SYNTH-13709` says it all. If you want you can update it to "Use the `Retry-After` header returned for 429s"
```suggestion
const DELAY_BETWEEN_429_RETRIES = 5000 // In ms
```

src/commands/synthetics/api.ts
```diff
@@ -259,50 +266,69 @@ const createMobileVersion = (request: (args: AxiosRequestConfig) => AxiosPromise
...
 }
 
-const getErrorHttpStatus = (error: AxiosError | EndpointError) =>
-  'status' in error ? error.status : error.response?.status
+const getErrorHttpStatus = (error: Error): number | undefined =>
+  'status' in error
+    ? error.status
+    : 'response' in error && 'status' in (error.response as any)
+    ? (error.response as any)?.status
+    : undefined
```
Drarig29, line 307: You don't need any `any`: TypeScript generates a "pseudo-type" for you 😁 
```suggestion
const getErrorHttpStatus = (error: Error): number | undefined =>
  'status' in error
    ? error.status
    : 'response' in error && 'status' in error.response
    ? error.response.status
    : undefined
```

src/commands/synthetics/api.ts
teodor2312: Interesting i was sure i removed that 🤔

src/commands/synthetics/api.ts
teodor2312: I was hoping so as well, but it really doesn't like it without it 😅

teodor2312: Yeah nah it doesn't like it :D 
https://github.com/DataDog/datadog-ci/actions/runs/8357462396/job/22876652674?pr=1239

Drarig29: Ha, yes it's because we are missing the type of `error.status`, `error.response`, and `error.response.status`.

Drarig29: Then you can do something like that instead:

```ts
const isEndpointError = (error: Error): error is EndpointError => error instanceof EndpointError

const getErrorHttpStatus = (error: Error): number | undefined =>
  isAxiosError(error) ? error.response?.status : isEndpointError(error) ? error.status : undefined
```

which requires the following imports:
```ts
import type {AxiosError, AxiosPromise, AxiosRequestConfig} from 'axios'

import {isAxiosError} from 'axios'
```

teodor2312: I like how clean it looks like this, but it doesn't work 😅 tests fail since it doesn't do the correct number of retries. Will try and see if i can massage it into working 👍

src/commands/synthetics/__tests__/api.test.ts
```diff
@@ -418,26 +438,26 @@ describe('getApiHelper', () => {
 
 describe('formatBackendErrors', () => {
   test('backend error - no error', () => {
-    const backendError = getAxiosHttpError(500, {errors: []})
+    const backendError: AxiosError<any, any> = getAxiosHttpError(500, {errors: []})
```
Drarig29: Let's make `getAxiosHttpError()` return `AxiosError<any, any>` instead

Drarig29 (approved):
LGTM

PR #1240, opened 2024-03-18
[SYNTH-12516][Mobile] Add new validation and custom reporter to mobile app uploads

dd-spells:
### What and why?

**Note: This PR will not be merged right away. We're "staging" this PR to more easily test the `app-upload-validator` service. We'll likely wait for `web-ui` to change first (it also will need to adjust the way it uploads apps) before merging changes to the datadog-ci tool.**

This PR changes the way we upload applications. There are a few changes.
1. We now call `completeMultipartMobileApplicationUpload` with a `validateMode`, as well as optional `newVersionParams`. These additions change the way the backend endpoint works, and requires the caller (`datadog-ci`) to use the returned `job_id` to poll for validation results.
2. We implement the polling endpoint `pollMobileApplicationUploadResponse`, and call it for 5 minutes every second (numbers up for debate).
3. We implement a CLI "UI" (a new `reporter` `AppUploadReporter`) which will report on the progress of the uploads.
4. We refactor how "temporary app" uploads work as part of running ci tests. When people run ci tests and override the `mobileApplicationVersionFilePath` (basically targeting a local version of their app), we need to upload these apps. There could be multiple depending on how many tests they are running and how many different apps are associated with those tests. Because we want to be able to show "progress" during the uploading, we first need to gather how many apps are to be uploaded, and then, as we upload, render progress updates. We use a new class `AppUploadCache` for this purpose.

Here are successes and failures for both a normal `UploadApplicationCommand` and a `run-tests` with temporary local apps:
UploadApplicationCommand - success:
Screenshot 2024-03-19 at 10 21 03 AM

UploadApplicationCommand - failure:
Screenshot 2024-03-19 at 10 20 40 AM

run-tests - success:
Screenshot 2024-03-19 at 10 20 17 AM

run-tests - failure:
Screenshot 2024-03-19 at 10 24 47 AM

Spinning (in progress):
Screenshot 2024-03-19 at 10 25 27 AM

### Review checklist


src/commands/synthetics/mobile.ts
```diff
@@ -70,114 +79,129 @@ export const uploadMobileApplications = async (
...
+    return this.cache[appPath][appId]
   }
 
-  uploadedApplicationByPath[applicationPathToUpload].push({
-    applicationId: testApplicationId,
-    fileName,
-  })
+  public setFileName(appPath: string, appId: string, fileName: string): void {
+    this.cache[appPath][appId] = fileName
+  }
```
kevin-pietsch, line 177: `getFileName` makes sense in the upload context, but it wasn't immediately obvious to me what was happening here. 

What do you think of naming like `(g|s)etUploadedAppFileName`? Little verbose, but I feel like it eases the grok a little -- that these are set when and only when the app has been uploaded successfully

kevin-pietsch (approved):
This is great 🚀 

Left a bunch of style nits, consider them all optional.

I ran temp-app test (just one test) and upload app and everything looks and behaves as I expected 👍 

Let's get a review from CI too

src/commands/synthetics/utils/public.ts
```diff
@@ -726,31 +727,8 @@ export const getTestsToTrigger = async (
...
-          test,
-          userConfigOverride,
-          overriddenConfig,
-          uploadedApplicationByPath
-        )
-      } catch (e) {
-        throw new CriticalError('UPLOAD_MOBILE_APPLICATION_TESTS_FAILED', e.message)
-      }
-    }
+  if (testsAndConfigsOverride.some((item) => 'test' in item && item.test.type === 'mobile')) {
```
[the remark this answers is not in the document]
Drarig29: That's not possible because `item` is a type union (`TestNotFound | TestSkipped | TestWithOverride`), and the `test` property is only there in 1 case. :grin: I think the current code is ok 👌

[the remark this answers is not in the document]
kevin-pietsch: But that's exactly the point of the `?.` operator -- so you can concisely dereference possibly non-existent fields.

[Here's a simple example](https://www.typescriptlang.org/play?#code/C4TwDgpgBAygrgIwGYEsIBsAmUC8UDeAUFCVAMYD2F6AXFAHZwC2CEAToQL6GiRQCCuAsVL0IAQ2B1GLdgBoRJJiABiaLHXjJ1mBd17QAQkKKkGEqQ2as2enuGgBhE4qgB3Cm+nX5XKPb4AFQgAZ2AhQQAfKGNox0JKejCoS2DkvHwxSToAVjkoZTUMTDp8Smo6AEZObkSQ6ggAOnQKAHMACmAAfkbCnR7y9ABKIA)

```
type Subfield = {
    cool: number
}
type A = {
    neat: number,
    myField: Subfield,
}
type B = {
    neat: number,
}
type C = {
    wow: number,
} 
type Test = A | B | C
const t: Test = {neat: 5, myField: {cool: 1}}
console.log(t?.myField?.cool)
```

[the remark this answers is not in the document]
kevin-pietsch: Oh no! You're right 😆 -- I unintentionally manipulated the type system. This fails once you wipe out the fact that `t` is an `A`

https://www.typescriptlang.org/play?#code/C4TwDgpgBAygrgIwGYEsIBsAmUC8UDeAUFCVAMYD2F6AXFAHZwC2CEAToQL6GiRQCCuAsVL0IAQ2B1GLdgBoRJJiABiaLHXjJ1mBd17QAQkKKkGEqQ2as2enuGgBhE4qgB3Cm+nX5XKPb4AFQgAZ2AhQQAfKGNox0JKejCoS2DkvHwxSToAVjkoZTUMTDp8Smo6AEZOTihxELr6EASKJOoIADp0CgBzAApgDsKdAH4O8vQASiA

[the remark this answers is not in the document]
Drarig29: No this won't work 😄. Your example only works because TypeScript inferred `t` as `A`:

[Your example](https://www.typescriptlang.org/play?#code/C4TwDgpgBAygrgIwGYEsIBsAmUC8UDeAUFCVAMYD2F6AXFAHZwC2CEAToQL6GiRQCCuAsVL0IAQ2B1GLdgBoRJJiABiaLHXjJ1mBd17QAQkKKkGEqQ2as2enuGgBhE4qgB3Cm+nX5XKPb4AFQgAZ2AhQQAfKGNox0JKejCoS2DkvHwxSToAVjkoZTUMTDp8Smo6AEZObkSQ6ggAOnQKAHMACmAAfkbCnR7y9ABKQgB6UbNJqAA9LqA):

image

[versus when using `as Test`](https://www.typescriptlang.org/play?#code/C4TwDgpgBAygrgIwGYEsIBsAmUC8UDeAUFCVAMYD2F6AXFAHZwC2CEAToQL6GiRQCCuAsVL0IAQ2B1GLdgBoRJJiABiaLHXjJ1mBd17QAQkKKkGEqQ2as2enuGgBhE4qgB3Cm+nX5XKPb4AFQgAZ2AhQQAfKGNox0JKejCocLx8MUk6AFY5KGU1DEw6fEpqOgBGTk4ocRCoYLCEiiTqCAA6dAoAcwAKYAB+NvydQdL0AEpCAHopszmoAD1+oA), to prevent TypeScript from inferring `t` as `A`:

image

Now you get the error which @dd-spells was having (legitimately):

```ts
Property 'myField' does not exist on type 'Test'.
  Property 'myField' does not exist on type 'B'.(2339)
```

src/commands/synthetics/mobile.ts
```diff
@@ -70,114 +79,129 @@ export const uploadMobileApplications = async (
...
+export class AppUploadCache {
+  private cache: {[applicationFilePath: string]: {[applicationId: string]: string | undefined}} = {}
+
+  public setAppCacheKeys(
+    triggerConfigs: TriggerConfig[],
+    testsAndConfigsOverride: (TestNotFound | TestSkipped | TestWithOverride)[]
+  ): void {
+    for (const [index, item] of testsAndConfigsOverride.entries()) {
+      if ('test' in item && item.test.type === 'mobile' && !('errorMessage' in item)) {
+        const appId = item.test.options.mobileApplication!.applicationId
```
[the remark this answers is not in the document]
Drarig29: Instead of possibly giving non-mobile tests to mobile-specific logic, you can define a new type in `interfaces.ts` which contains everything you need:

```ts
export type MobileTestWithOverride = TestWithOverride & {
  test: Test & {
    type: 'mobile'
    options: {
      mobileApplication: MobileApplication
    }
  }
}
```

Then in `internal.ts`, define a new type-guard function:

```ts
export const isMobileTestWithOverride = (
  item: TestNotFound | TestSkipped | TestWithOverride
): item is MobileTestWithOverride =>
  'test' in item && item.test.type === 'mobile' && !!item.test.options.mobileApplication
```

And now, your [`testsAndConfigsOverride.some((item) => 'test' in item && item.test.type === 'mobile')`](https://github.com/DataDog/datadog-ci/blob/fc4c876568d834b8c0996408e7ff80721adc36a7/src/commands/synthetics/utils/public.ts#L730-L732) can become something like this (`uploadMobileApplicationsAndOverrideConfigs` will need a early return if the array is empty):

```ts
await uploadMobileApplicationsAndOverrideConfigs(
  api,
  triggerConfigs,
  testsAndConfigsOverride.filter(isMobileTestWithOverride),
  appUploadReporter
)
```

Thanks to this, you can replace all the `TestNotFound | TestSkipped | TestWithOverride` in `mobile.ts` with `MobileTestWithOverride`, since you don't care about uploading "not found" or "skipped" tests. This will remove some if conditions and non-null assertions 😁

src/commands/synthetics/reporters/appUpload.ts
Drarig29, line 1: Can you rename this file and move it in a mobile folder? e.g. `reporters/mobile/app-upload.ts`

src/commands/synthetics/reporters/appUpload.ts
```diff
@@ -0,0 +1,70 @@
...
+      text,
+    })
+    this.spinner.start()
+  }
+
+  public reportSuccess = (appendLineBreak = false): void => {
+    this.endRendering()
+    this.write(
+      `${ICONS.SUCCESS} Uploaded applications in ${Math.round((Date.now() - this.startTime) / 1000)}s${
+        appendLineBreak ? '\n' : ''
```
Drarig29, line 39: Why not always append a line break? 🤔

src/commands/synthetics/__tests__/fixtures.ts
```diff
@@ -569,16 +573,18 @@ export const getTestPayload = (override?: Partial<TestPayload>) => ({
...
-  version_name: 'test version',
-  created_at: '22-09-2022',
-  ...override,
-})
+export const getTestAndConfigOverride = (appId: string): TestWithOverride => {
+  const testWithOverride = {test: getMobileTest(), overriddenConfig: getTestPayload()}
+  testWithOverride.test.options.mobileApplication!.applicationId = appId
+
+  return testWithOverride
+}
```
Drarig29, line 581: **nit** I find it weird to call this function `getTestAndConfigOverride()` when it's actually made for mobile unit tests. Can you rename it, and avoid using a non-null assertion by adding an optional `appId = 'mobileAppUuid'` argument to `getMobileTest()`?
```suggestion
export const getMobileTestWithOverride = (appId: string): MobileTestWithOverride => {
  return {
    test: getMobileTest('abc-def-ghi', appId),
    overriddenConfig: getTestPayload(),
  }
}
```

src/commands/synthetics/__tests__/fixtures.ts
```diff
@@ -569,16 +573,18 @@ export const getTestPayload = (override?: Partial<TestPayload>) => ({
...
+
+export const getTriggerConfig = (appPath?: string, appVersion?: string): TriggerConfig => {
```
Drarig29, line 583: **nit** same nit about naming
```suggestion
export const getMobileTriggerConfig = (appPath?: string, appVersion?: string): TriggerConfig => {
```

src/commands/synthetics/run-tests-lib.ts
```diff
@@ -45,6 +46,7 @@ type ExecuteOptions = {
 
 export const executeTests = async (
   reporter: MainReporter,
+  appUploadReporter: AppUploadReporter,
```
Drarig29, line 49: I'd like to avoid changing this function's signature at all costs, because it would be a breaking change for the folks who use `@datadog/datadog-ci` programmatically, as a library.

And since this app upload reporter is short-lived and only used for Mobile tests, I think we should construct it directly in `uploadMobileApplicationsAndOverrideConfigs()` instead of passing it around. WDYT?

src/commands/synthetics/reporters/appUpload.ts
Drarig29: For example, this doesn't look too good IMO (I don't know if `appendLineBreak = true` would have fixed it in this context though):
 
image

Drarig29 (commented):
Great work! Mostly nits

Drarig29 (approved):
LGTM (just one nit)

Thanks for addressing all my comments ❤️

PR #1242, opened 2024-03-19
[dep] Bump `upload-artifact` github action

Drarig29:
### What and why?

Fixes these warnings:

image

### How?

Bumping github actions.

### Review checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,7 +50,7 @@ jobs:
       - run: cp -r .github/workflows/e2e artifacts/
       - uses: actions/upload-artifact@v4
         with:
-          name: artifacts
+          name: artifacts-${{ matrix.version }}
           path: artifacts/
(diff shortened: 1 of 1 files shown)


juan-fernandez (approved):

teodor2312 (approved):
LGTM, but out of curiosity how can you test this that the change is working 🤔

Drarig29:
@teodor2312 it failed previously, and the CI failed (see the first commit). And I downloaded the artifacts to ensure we were keeping the right behavior:

- For the `artifacts` folder, we want to have each a `.tgz` for datadog-ci in each Node.js version ✅ 
- For the `sarif-datadog-ci.sarif`, it's just for us to manually download [in case the linting fails](https://github.com/DataDog/datadog-ci/blob/bee8ce79912669c0783b7a5415e2150c14a8a51b/.github/workflows/ci.yml#L25-L36), so it's OK for it to be overwritten (it's the old behavior)

Issue #1243, opened 2024-03-19
components generated by cyclonedx/cyclonedx-npm are not detected properly


omercnet:
https://github.com/DataDog/datadog-ci/blob/fb039260b321e753096c5e1515ed4bc3b4b95c4f/src/commands/sbom/language.ts#L9

cyclonedx/cyclonedx-npm creates an sbom with the `pkg` prefix in `purl` and not `bom-ref`

PR #1244, opened 2024-03-19
fix(sbom): detect npm from purl

omercnet:
### What and why?

fixes https://github.com/DataDog/datadog-ci/issues/1243

### How?

If npm does not exist in `bom-ref` also check `purl`

### Review checklist


src/commands/sbom/language.ts
```diff
@@ -6,7 +6,7 @@ export const getLanguageFromComponent = (component: any): DependencyLanguage | u
   const componentName = component['name']
 
   if (component['bom-ref']) {
-    if (component['bom-ref'].includes('pkg:npm')) {
+    if (component['bom-ref'].includes('pkg:npm') || component['purl'].includes('pkg:npm')) {
```
juli1, line 9: `component['purl']` can be undefined here. We should check this is defined.

src/commands/sbom/__tests__/fixtures/sbom.1.5.ok.json
```diff
@@ -117,7 +117,7 @@
       ]
     },
     {
-      "bom-ref": "pkg:npm/%40graphql-typed-document-node/core@3.1.1",
+      "bom-ref": "@graphql-typed-document-node/core@3.1.1",
```
juli1, line 120: Please do put change this file. If you want to add another test, please add another file.

juli1 (changes requested):

src/commands/sbom/__tests__/fixtures/sbom.1.5.ok.json
omercnet: there are no specific tests for this use case of language detection, the test just ensures the deps were parsed and counts them

creating a new file with just this patch and copying the exact same test seems redundant and excessive - but lmkwyt

src/commands/sbom/language.ts
omercnet: it is not checked in the rest of the cases, should I fix those as well?

src/commands/sbom/__tests__/fixtures/sbom.1.5.ok.json
juli1: Thanks for your comment. This file (sbom.1.5.json) is what our internal tooling returns (returns from [osv-scanner](https://github.com/DataDog/osv-scanner)). We should not change it since it's used by [our tests](https://github.com/DataDog/datadog-ci/blob/master/src/commands/sbom/__tests__/payload.test.ts) and test our internal tools. If you want to test another logic, please
1. Add another file
2. Test this file to check this file.

Thank you,

src/commands/sbom/language.ts
juli1: Wow, this is a good point 😅 we should have checked then.

Do not worry for now, let's keep it like this - if we have an issue with this key, we will add a check later.

juli1 (approved):
Thank you so much for addressing the comments 🙏

omercnet:
@dastrong can you take a look?

dastrong (approved):

PR #1252, opened 2024-03-21
[synthetics] Move non-exported utils to `batch.ts`

Drarig29:
### What and why?

- https://github.com/DataDog/datadog-ci/pull/1252 (← **you are here**)
- https://github.com/DataDog/datadog-ci/pull/1251

This PR is a noop. All moved functions were **NOT** exported.

### How?

Reduce the `utils/public.ts` file size by moving all non-exported and batch-related functions to `batch.ts`.

### Review checklist

### DIFF
--- a/src/commands/synthetics/batch.ts
+++ b/src/commands/synthetics/batch.ts
@@ -1,0 +1,191 @@
+import deepExtend from 'deep-extend'
+
+import {APIHelper, EndpointError, formatBackendErrors} from './api'
+import {Batch, MainReporter, PollResultMap, Result, ResultDisplayInfo, ResultInBatch, Test, Trigger} from './interfaces'
+import {isResultInBatchSkippedBySelectiveRerun, getResultIdOrLinkedResultId} from './utils/internal'
+import {wait, getAppBaseURL, hasResultPassed} from './utils/public'
+
+const POLLING_INTERVAL = 5000 // In ms
+
+export const waitForBatchToFinish = async (
+  api: APIHelper,
+  maxPollingTimeout: number,
+  trigger: Trigger,
+  resultDisplayInfo: ResultDisplayInfo,
+  reporter: MainReporter
+): Promise<Result[]> => {
+  const maxPollingDate = Date.now() + maxPollingTimeout
+  const emittedResultIndexes = new Set<number>()
+
+  while (true) {
+    const batch = await getBatch(api, trigger)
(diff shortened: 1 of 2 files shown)


teodor2312 (approved):
LGTM - copy functions from one place to another 👍 

Can you tell me what's the logic behind the ordering of the functions in the new file?

Drarig29:
> Can you tell me what's the logic behind the ordering of the functions in the new file?

@teodor2312 I didn't really think about it 😅 I only took care of having the only exported function at the very top

teodor2312:
> > Can you tell me what's the logic behind the ordering of the functions in the new file?
> 
> @teodor2312 I didn't really think about it 😅 I only took care of having the only exported function at the very top

In that case is there a preferred style that we would order them in for this repo? And can we order them accordingly :D 
I would personally go exports on top private on bottom and within each alphabetically

Drarig29:
There isn't any preferred style, and I'm never a fan of alphabetical order. Go to definition is the king, and I prefer having small utility functions at the very bottom: [I moved `getTestByPublicId()` at the bottom of the file in the next PR](https://github.com/DataDog/datadog-ci/pull/1251/files#diff-d3eb5816e581934b4e13d75612e98ce6d15b9aab40a830501474449efca415a0R268-R273)

jakepruitt (approved):
Looks great!

jakepruitt:
Actually proposition: What about moving all batch-related logic to `batch.ts` and then exposing those functions by importing them into `public.ts` before exporting? That way people all batch-related logic goes in `batch` and the choice to make it public or private is determined by that `public` file?

Drarig29:
@jakepruitt I don't see much more batch-related logic left in `utils/public.ts` except `waitForResults()` maybe, but it requires other utils like `getAppBaseURL()` or `isTestSupportedByTunnel()`, which _are_ actual utils IMO and not batch-related, so they should stay in `utils/public.ts`.

But if we keep those 2 functions in `utils/public.ts`, then `batch.ts` will have to import from `utils/public.ts`, creating a circular reference.

I don't see much value in adding indirection by re-exporting things from `batch.ts` in `utils/public.ts`. And if they are in `batch.ts`, it's because they aren't utils (a.k.a. small reusable functions).

So I would rather make a breaking change at some point and stop exporting any batch-related logic, since I'm not sure we really want to export this.

jakepruitt:
Sounds good! Like that logic

PR #1255, opened 2024-03-21
[SYNTH-12914] Include `?batch_id` in result links

Drarig29:
### What and why?

- https://github.com/DataDog/datadog-ci/pull/1254
- https://github.com/DataDog/datadog-ci/pull/1255  (← **you are here**)

The frontend will now provide more information when the result isn't found yet, but known to be in-progress in the batch. For this, we change `?from_ci=true` to `?batch_id=<batchId>` in the results link.

#### Before
image

#### After
image

### How?

Since the lines were becoming really long, I added a line break. The `⎋` icon represents an external link, so I kept it next to the result link.

### Review checklist


hestonhoffman (approved):

hakanb:
@Drarig29 Looks good but I have one concern. If we remove `from_ci=true`, I believe we lose the ability to track the # of times these links are clicked, which I think is useful. Would it be possible to keep `from_ci=true` in the URL? What do you think?

jakepruitt (commented):
This looks good from my end, though let's check with @hakanb about the analytics use case

etnbrd (approved):
LGTM 👍

PR #1259, opened 2024-03-25
[SYNTH-13724] Differentiate batch timeout and local deadline

Drarig29:
### What and why?

The batch timeout (aka. `pollingTimeout`) is enforced **by our backend**, so only the backend should be the source of truth for `datadog-ci` showing the message "The batch timed out before receiving the result".

In case the backend doesn't respect the `pollingTimeout`, `datadog-ci` has a **safety mechanism** to stop polling. And as this is a safety mechanism, we should show something different when it happens.

In order to be resilient to latency, the safety mechanism has a margin delay of 3 polling cycles, i.e. 15 seconds. This can be seen as "retrying" 3 times.

### How?

> [!NOTE]  
> 💡 Note for the reviewer: don't hesitate to suggest another name for `SAFE_DEADLINE_REACHED`, and/or another error message.

In case this safety mechanism is used, we do report timed out results to reporters: they will be displayed in the console output, and marked as timed out in JUnit reports - but all of that with a different error message:

> Datadog did not answer on time.

Once the reporting is done, a critical error is thrown.

### Review checklist


src/commands/synthetics/utils/public.ts
```diff
@@ -885,6 +885,9 @@ export const reportCiError = (error: CiError, reporter: MainReporter) => {
     case 'POLL_RESULTS_FAILED':
       reporter.error(`\n${chalk.bgRed.bold(' ERROR: unable to poll test results ')}\n${error.message}\n\n`)
       break
+    case 'SAFE_DEADLINE_REACHED':
+      reporter.error(`\n${chalk.bgRed.bold(' ERROR: safe deadline reached ')}\n${error.message}\n\n`)
```
etnbrd, line 889: I'm not sure "safe deadline" is very explicit outside of the code base, could we rephrase that to something more explicit, like "the batch didn't timeout after the expected timeout period"?

src/commands/synthetics/batch.ts
```diff
@@ -57,14 +58,16 @@ export const waitForBatchToFinish = async (
       reporter
     )
 
-    reportResults(resultsToReport, pollResultMap, resultDisplayInfo, hasBatchExceededMaxPollingDate, reporter)
+    reportResults(resultsToReport, pollResultMap, resultDisplayInfo, safeDeadlineReached, reporter)
 
     oldIncompleteResultIds = incompleteResultIds
 
+    if (safeDeadlineReached) {
+      throw new CriticalError('SAFE_DEADLINE_REACHED', 'Datadog did not answer on time.')
```
etnbrd, line 66: This error message seems a bit misleading: the backend was responding, and it's the batch that was not timing out whereas it was expected to, isn't it?
I would rephrase that with a unified error message, such as "the batch didn't timeout after the expected timeout period".

src/commands/synthetics/batch.ts
```diff
@@ -211,7 +211,10 @@ const getResultFromBatch = (
 
   const pollResult = pollResultMap[resultInBatch.result_id]
 
-  if (hasTimedOut) {
+  if (safeDeadlineReached) {
+    pollResult.result.failure = {code: 'SAFE_DEADLINE_REACHED', message: 'Datadog did not answer on time.'}
```
etnbrd, line 215: I see that this "Datadog did not answer on time" message is duplicated between two places. Is it possible to have these messages in a single place?

src/commands/synthetics/utils/public.ts
etnbrd: I think it'd help to have some reference to the batch, and the fact that the batch timeout is unexpectedly running away. What do you think about `UNEXPECTED_BATCH_RUNAWAY`?
But naming is hard and in the code base, we can always add comments to explain what are the concepts behind these names, so I don't really have a strong opinion about that.

etnbrd (approved):
LGTM 👍 
Thanks for taking these review into account 🙇

src/commands/synthetics/utils/public.ts
Drarig29: Oh, maybe `MISSED_BATCH_TIMEOUT`

Drarig29: Per discussion, we went with `BATCH_TIMEOUT_RUNAWAY`. The concept of running away exists in other systems: https://www.ibm.com/docs/en/psfa/7.2.1?topic=reference-runaway-query-notification

PR #1270, opened 2024-04-01
[ci-visibility] Add silent option to tag command

rodrigo-roca:
### What and why?

This PR introduces the `--silent` option to the `tag` command. This option, as the README will state, it prevents the tag command from writing to stdout and stderr. One of its uses will be to hide implementation details for the `datadog-ci deployment mark` beta command. That command runs the tag command twice and we do not want the second time to output anything as it would be redundant.

### How?

### Review checklist


juan-fernandez (approved):
don't you want to use this option in the `mark` command? We could add that in this PR I guess

juan-fernandez (approved):

hestonhoffman (approved):

PR #1271, opened 2024-04-02
[SYNTH-13771] Fix "No tests to run" not being shown anymore

Drarig29:
### What and why?

- https://github.com/DataDog/datadog-ci/pull/1271 (← **you are here**)
- https://github.com/DataDog/datadog-ci/pull/1272

Fixes a bug introduced in https://github.com/DataDog/datadog-ci/pull/1214 (released in `2.32.1`)

For better unit testing, this PR first moves some non-exported code to a new file named `test.ts`, which contains all S8S-test-related logic.

> [!NOTE]  
> 💡 Better reviewed commit by commit.

### How?

The main change of this PR is: 50a262c13af8fea8e7c64cdfcfe77f5f92371351

### Review checklist

### DIFF
--- a/src/commands/synthetics/test.ts
+++ b/src/commands/synthetics/test.ts
@@ -1,7 +1,6 @@
 import chalk from 'chalk'
 
 import {APIHelper, EndpointError, formatBackendErrors, isNotFoundError} from './api'
-import {CiError} from './errors'
 import {MainReporter, RunTestsCommandConfig, Suite, Test, TriggerConfig, UserConfigOverride} from './interfaces'
 import {MAX_TESTS_TO_TRIGGER} from './run-tests-command'
 import {getSuites, normalizePublicId} from './utils/public'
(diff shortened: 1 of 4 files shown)


src/commands/synthetics/test.ts
```diff
@@ -0,0 +1,82 @@
...
+      ...(await api.getTest(id)),
+      suite,
+    }
+
+    return {test}
+  } catch (error) {
+    if (isNotFoundError(error)) {
+      const errorMessage = formatBackendErrors(error)
+
+      return {errorMessage: `[${chalk.bold.dim(id)}] ${chalk.yellow.bold('Test not found')}: ${errorMessage}`}
```
etnbrd, line 63: Would it be possible to throw, and catch the error higher up?
I'd like to avoid using `chalk` outside of the terminal writing reporters.

etnbrd (approved):
LGTM 👍 
Left one nit about not using `chalk`.

src/commands/synthetics/test.ts
Drarig29: This is a big scope creep IMO. This `getTest()` is used in `getTestAndOverrideConfig()` which has a reporter, but we would still have to format the message with `chalk` in `public.ts` and give it to the reporter which is just moving the problem.

One way to solve this would be to have the logic to `reportCiError()` inside reporters, which would then use `chalk` or not. But this is totally out of scope.

PR #1272, opened 2024-04-02
[synthetics] Improve warning for too many tests to trigger 

Drarig29:
### What and why?

- https://github.com/DataDog/datadog-ci/pull/1271
- https://github.com/DataDog/datadog-ci/pull/1272 (← **you are here**)

This PR merges 2 not-so-useful warnings into a single more-useful warning.

#### Before

The 2 warnings were almost duplicates, and the most important fact is that _only 100 tests_ will be **TRIGGERED**. The fact that datadog-ci will FETCH only 100 tests isn't important.

Also, we never say how many were returned _exactly_, just "more than 100".

image

#### After

image

### How?

The `getTestsList()` function used to be the one reporting the "will be fetched" warning, and https://github.com/DataDog/datadog-ci/pull/1214 changed it to be reported by `getTriggerConfigs()` instead.

The "were fetched" one didn't change: it's still reported by `getTestsToTrigger()`.

My proposition is to only keep the "were fetched" one, and rephrase it into "will be triggered".

See https://github.com/DataDog/datadog-ci/pull/1214/files#r1508534127.

### Review checklist

### DIFF
--- a/src/commands/synthetics/test.ts
+++ b/src/commands/synthetics/test.ts
@@ -1,7 +1,6 @@
 import chalk from 'chalk'
 
 import {APIHelper, EndpointError, formatBackendErrors, isNotFoundError} from './api'
-import {CiError} from './errors'
 import {MainReporter, RunTestsCommandConfig, Suite, Test, TriggerConfig, UserConfigOverride} from './interfaces'
 import {MAX_TESTS_TO_TRIGGER} from './run-tests-command'
 import {getSuites, normalizePublicId} from './utils/public'
@@ -41,10 +40,6 @@ export const getTestsFromSearchQuery = async (
     reporter.error(
       `More than ${MAX_TESTS_TO_TRIGGER} tests returned by search query, only the first ${MAX_TESTS_TO_TRIGGER} will be fetched.\n`
     )
-  }
-
-  if (testsToTriggerBySearchQuery.length === 0) {
-    throw new CiError('NO_TESTS_TO_RUN')
   }
 
   return testsToTriggerBySearchQuery
--- a/src/commands/synthetics/test.ts
+++ b/src/commands/synthetics/test.ts
@@ -30,9 +30,14 @@ export const getTestConfigs = async (
 
 export const getTestsFromSearchQuery = async (
   api: APIHelper,
-  config: Required<Pick<RunTestsCommandConfig, 'global' | 'testSearchQuery'>>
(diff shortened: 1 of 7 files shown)


src/commands/synthetics/utils/public.ts
```diff
@@ -507,11 +507,12 @@ export const getTestsToTrigger = async (
   const errorMessages: string[] = []
   // When too many tests are triggered, if fetched from a search query: simply trim them and show a warning,
   // otherwise: retrieve them and fail later if still exceeding without skipped/missing tests.
-  if (triggerConfigs.length > MAX_TESTS_TO_TRIGGER && triggerFromSearch) {
+  if (triggerFromSearch && triggerConfigs.length > MAX_TESTS_TO_TRIGGER) {
```
etnbrd, line 510: Doesn't it mean that if more than 100 tests are triggered from test files, it won't show the warning, and more importantly, it'll trigger all of the tests?

src/commands/synthetics/test.ts
```diff
@@ -31,18 +30,17 @@ export const getTestConfigs = async (
 
 export const getTestsFromSearchQuery = async (
   api: APIHelper,
-  config: RunTestsCommandConfig,
-  reporter: MainReporter
+  config: Required<Pick<RunTestsCommandConfig, 'global' | 'testSearchQuery'>>
 ) => {
-  const testsToTriggerBySearchQuery = await getTestListBySearchQuery(api, config.global, config.testSearchQuery || '')
+  const {global: globalConfigOverride, testSearchQuery} = config
```
etnbrd, line 35: nit: we could move the test on `testSearchQuery` inside this function to avoid this slightly more complex type here, and the need for the user-defined guard `hasTestSearchQuery`.

```suggestion
  if (!config.testSearchQuery) {
    return []
  }
  
  const {global: globalConfigOverride, testSearchQuery} = config
```

src/commands/synthetics/utils/public.ts
Drarig29: No. The comment in the line above says that if more than 100 tests are triggered _from test files_, we will throw an error:

https://github.com/DataDog/datadog-ci/blob/aa2aca22eb480f9c0d045c8a4724bc59ed7451af/src/commands/synthetics/utils/public.ts#L579-L584

src/commands/synthetics/test.ts
Drarig29: Very true

PR #1273, opened 2024-04-02
enable diff-aware and non-diff-aware

juli1:
### What and why?

We want to enable the static analyzer with diff-aware and not diff-aware to test the results on a live, public repo

### How?

Have two jobs to upload results for this repository:
1. One with diff-aware
2. One without diff-aware

This way we can compare the output in the datadog UI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -219,7 +219,7 @@ jobs:
           dd_app_key: ${{ secrets.DATADOG_APP_KEY_MAIN_ACCOUNT }}
           dd_api_key: ${{ secrets.DATADOG_API_KEY_MAIN_ACCOUNT }}
           dd_service: 'datadog-ci'
-          dd_env: 'ci-no-dq'
+          dd_env: 'ci-no-da'
           cpu_count: 2
           sca_enabled: true
           diff_aware: false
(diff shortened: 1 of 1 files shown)


dastrong (approved):

.github/workflows/ci.yml
```diff
@@ -203,8 +203,27 @@ jobs:
...
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Run Datadog static analyzer
+        id: datadog-static-analysis
+        uses: DataDog/datadog-static-analyzer-github-action@main
+        with:
+          dd_app_key: ${{ secrets.DATADOG_APP_KEY_MAIN_ACCOUNT }}
+          dd_api_key: ${{ secrets.DATADOG_API_KEY_MAIN_ACCOUNT }}
+          dd_service: 'datadog-ci'
+          dd_env: 'ci-no-dq'
```
Drarig29, line 222: ❓ Is the "q" in `ci-no-dq` a typo?

Drarig29 (approved):
LGTM, just one question

.github/workflows/ci.yml
juli1: Fixed. It's is for `ci-no-da` (`no-diff-aware`)

Drarig29: Yep, I figured it was very likely to be azerty vs qwerty 😁

PR #1287, opened 2024-04-17
[git upload] Improvements to git upload and change junit git upload default 

juan-fernandez:
### What and why?

* Change the way we upload git metadata:
  * Before unshallowing, we'll get the latest commits, do a request to `search_commits` and check whether the commits we have are already in our backend. 
  * If they are all, we won't even attempt unshallowing.
 
* Change `junit upload`'s `--skip-git-metadata-upload` default from `true` to `false`: this means that git upload will happen unless specified otherwise. 

### How?

* Change `uploadToGitDB`
* Update unit tests.

### Review checklist

### DIFF
--- a/src/commands/git-metadata/__tests__/gitdb.test.ts
+++ b/src/commands/git-metadata/__tests__/gitdb.test.ts
@@ -516,6 +516,82 @@ describe('gitdb', () => {
     mocks.expectCalls()
   })
 
+  test('should not unshallow repository if backend has all commits already', async () => {
+    const mocks = new MockAll({
+      getConfig: [
+        {
+          input: 'clone.defaultRemoteName',
+          output: defaultRemoteNameNotConfigured,
+        },
+      ],
+      fetch: [],
+      getRemotes: [
+        {
+          input: undefined,
+          output: [{name: 'origin', refs: {push: 'https://github.com/DataDog/datadog-ci'}}],
+        },
+      ],
+      log: [
+        {
+          input: ['-n 1000', '--since="1 month ago"'],
+          output: {
+            all: [
+              {
+                hash: '87ce64f636853fbebc05edfcefe9cccc28a7968b',
+              },
+              {
+                hash: 'cc424c261da5e261b76d982d5d361a023556e2aa',
+              },
+            ],
+          },
+        },
+      ],
+      raw: [],
+      revparse: [],
+      version: [],
+      execSync: [],
+      axios: [
+        {
+          input: {
+            url: '/api/v2/git/repository/search_commits',
+            data: {
+              meta: {
+                repository_url: 'https://github.com/DataDog/datadog-ci',
+              },
+              data: [
+                {
+                  id: '87ce64f636853fbebc05edfcefe9cccc28a7968b',
+                  type: 'commit',
+                },
+                {
+                  id: 'cc424c261da5e261b76d982d5d361a023556e2aa',
+                  type: 'commit',
+                },
+              ],
+            },
+          },
+          output: {
+            data: {
+              data: [
+                {
+                  id: '87ce64f636853fbebc05edfcefe9cccc28a7968b',
+                  type: 'commit',
+                },
+                {
+                  id: 'cc424c261da5e261b76d982d5d361a023556e2aa',
+                  type: 'commit',
+                },
+              ],
+            },
+          },
+        },
+      ],
+    })
+    const upload = uploadToGitDB(logger, request, mocks.simpleGit as any, false)
+    await expect(upload).resolves.toBe(undefined)
+    mocks.expectCalls()
+  })
+
   test('should unshallow repository if the local HEAD is a commit not pushed to the remote', async () => {
     const mocks = new MockAll({
       getConfig: [
--- a/src/commands/git-metadata/__tests__/gitdb.test.ts
+++ b/src/commands/git-metadata/__tests__/gitdb.test.ts
@@ -161,7 +161,7 @@ describe('gitdb', () => {
       this.execSync = jest.spyOn(child_process, 'execSync').mockImplementation(() => '') as jest.Mock
       this.axios = jest.spyOn(axios, 'create').mockImplementation(() => ((_: any) => {}) as any) as jest.Mock
 
-      const initMockWithParams = <I, O>(mock: jest.Mock, params: MockParam<I, O>[], promise: boolean) => {
+      const initMockWithParams = <I, O>(mock: jest.Mock, params: MockParam<I, O>[], promise: boolean, name = '') => {
         params.forEach((param) => {
           if (param.output instanceof Error) {
             mock = mock.mockImplementationOnce((..._: any) => {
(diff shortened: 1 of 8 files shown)


src/commands/git-metadata/__tests__/gitdb.test.ts
```diff
@@ -177,23 +177,40 @@ describe('gitdb', () => {
...
+            expect(mock.mock.calls).toHaveLength(params.length)
+            params.forEach((param, i) => {
+              if (param.input !== undefined) {
+                expect(mock.mock.calls[i][0]).toStrictEqual(param.input)
+              }
+            })
+          } catch (e) {
+            // To make it easier to debug the tests
+            // eslint-disable-next-line
+            console.log('Error in', name, 'mock')
```
juan-fernandez, line 190: this makes it slightly easier to know what mock you need to inspect

src/commands/git-metadata/__tests__/gitdb.test.ts
```diff
@@ -283,17 +300,54 @@ describe('gitdb', () => {
...
+                  type: 'commit',
+                },
+                {
+                  id: 'cc424c261da5e261b76d982d5d361a023556e2aa',
+                  type: 'commit',
+                },
+              ],
+            },
+          },
+          output: {data: {data: []}},
```
juan-fernandez, line 348: ℹ️ now we request `search_commits` before checking whether it's a shallow clone

src/commands/git-metadata/__tests__/gitdb.test.ts
```diff
@@ -344,17 +398,54 @@ describe('gitdb', () => {
...
```
juan-fernandez, line 446: ℹ️ now we request `search_commits` before checking whether it's a shallow clone

src/commands/git-metadata/__tests__/gitdb.test.ts
```diff
@@ -376,20 +467,131 @@ describe('gitdb', () => {
...
```
juan-fernandez, line 510: ℹ️ now we request `search_commits` before checking whether it's a shallow clone

src/commands/git-metadata/__tests__/gitdb.test.ts
```diff
@@ -376,20 +467,131 @@ describe('gitdb', () => {
...
+              {
+                hash: 'cc424c261da5e261b76d982d5d361a023556e2aa',
+              },
+            ],
+          },
+        },
+      ],
+      raw: [],
+      revparse: [],
+      version: [],
```
juan-fernandez, line 551: ℹ️ no revparse or version checks because we don't even attempt to unshallow (or check shallowness)

src/commands/git-metadata/__tests__/gitdb.test.ts
```diff
@@ -588,7 +863,7 @@ describe('gitdb', () => {
         },
       ],
       revparse: [{input: '--is-shallow-repository', output: 'false'}],
-      version: [],
+      version: [{input: undefined, output: newGitVersion}],
```
juan-fernandez, line 866: ℹ️ this is changed because I've changed the way we unshallow: we first check version. Let me know if you disagree with the change

src/commands/git-metadata/__tests__/gitdb.test.ts
```diff
@@ -1153,21 +1428,8 @@ describe('gitdb', () => {
...
-            '--filter=blob:none',
-            '--since="1 month ago"',
-            '^87ce64f636853fbebc05edfcefe9cccc28a7968b',
-            '^cc424c261da5e261b76d982d5d361a023556e2aa',
-          ],
-          output: '\n',
-        },
-      ],
-      revparse: [{input: '--is-shallow-repository', output: 'false'}],
+      raw: [],
```
juan-fernandez, line 1431: ℹ️ we now don't run `rev-list` if there are no commits to include, which I think makes more sense. I might be missing something though

src/commands/git-metadata/gitdb.ts
```diff
@@ -118,15 +155,16 @@ const getLatestLocalCommits = async (git: simpleGit.SimpleGit) => {
...
-    return
-  }
+const isShallowRepository = async (git: simpleGit.SimpleGit) => {
   const gitversion = String(await git.version())
   if (lte(gitversion, '2.27.0')) {
-    return
+    return false
   }
+
+  return (await git.revparse('--is-shallow-repository')) === 'true'
```
juan-fernandez, line 164: ⚠️ additionally change: we now check version first

src/commands/git-metadata/__tests__/gitdb.test.ts
juan-fernandez: before this change: we run `getObjectsToUpload` (the rev-list command) even if `commitsToInclude` was empty. And only then in `generatePackFilesForCommits` we would bail out if `objectsToUpload` was an empty list. 

❓ Can `git rev-list` return something if list of input commits is empty? If the answer is yes, I need to change this

juan-fernandez: this, btw, is also in `dd-trace-js`, so I understand why it'd be here 😄 . If running `git rev-list` with no input commits turns out to be useless (it looks this way testing locally), it'll have to be removed from `dd-trace-js` too.

romainkomorndatadog (approved):

src/commands/git-metadata/gitdb.ts
```diff
@@ -62,14 +50,63 @@ export const uploadToGitDB = async (
...
+
+  // If there are no commits to include, it means the backend already has all the commits.
+  if (commitsToInclude.length === 0) {
+    return
+  }
+  // If there are commits to include and the repository is shallow, we need to repeat the process after unshallowing
+  const isShallow = await isShallowRepository(git)
+  if (isShallow) {
+    await unshallowRepository(log, git)
+    const getCommitsAfterUnshallowing = await getCommitsToInclude(log, request, git, repoURL)
```
drodriguezhdez, line 105: Should we check the repo `isShallow` before invoking to the backend the first time? 

If I understood correctly, with the current approach if in the first iteration the found local commits is an incomplete subset, the backend might returned that all the commits are known, but.. in case we were unshallow the repo before, the local commits subset might be bigger (?), and therefore, the backend might return that some commits are missing. 
Also, it looks we're doing the work twice. 

Assuming that `isShallow` method is performant enough, should we do it always in order to reduce the back and force with the backend and have a more accurate set of commits store in the backend?

I've seen we're changing the tests explicitly to implement this use case. Just throwing a challenge here, because if looks unexpected to me.

Am I missing something?

src/commands/git-metadata/gitdb.ts
```diff
@@ -99,7 +136,7 @@ export const uploadToGitDB = async (
       return
     }
     log.debug(`Uploading packfiles...`)
-    await uploadPackfiles(log, request, repoURL, latestCommits[0], packfiles)
+    await uploadPackfiles(log, request, repoURL, headCommit, packfiles)
```
drodriguezhdez, line 139: unrelated to this PR, but I've observed that in the `getObjectsToUpload` method ([code](https://github.com/DataDog/datadog-ci/blob/master/src/commands/git-metadata/gitdb.ts#L234-L243)), we're using the `commitsToExclude`. 

Would it be enough only use the `commitsToInclude`? Now we're doing:

```  
return {
    commitsToInclude: latestCommits.filter((x) => !commitsToExclude.includes(x)),
    commitsToExclude,
    headCommit: latestCommits[0],
  }
```

Should improve the performance of the command execution if we remove the `commitsToExclude`?

drodriguezhdez (commented):
Thanks for the change!. I've dropped a couple of comments, especially one for the case with shallow repositories.

src/commands/git-metadata/gitdb.ts
juan-fernandez: > Assuming that isShallow method is performant enough

It's not, that's why we're doing this (indeed weird) dance.

The assumption here is that even if the clone is shallow, you'll at least have _one_ commit, and that'll be the last one. If the backend does have this last commit, we assume that it has the rest, so there's no need for unshallowing. If it doesn't we indeed need to unshallow and check again (so getting latest commit and requesting `search_commit` again).

src/commands/git-metadata/gitdb.ts
juan-fernandez: but we need both, right? Do you mean to just use `latestCommits` and `commitsToExclude` and build `commitsToInclude` on the spot? That could be done but there would be no performance gains. Maybe it'd be clearer, but I'm not sure either

drodriguezhdez (approved):
LGTM

rpelliard (approved):

PR #1293, opened 2024-04-25
[SYNTH-12971] Make `executeTests` backwards compatible

teodor2312:
### What and why?

We made changes recently of renaming the `global` field in the Global Test Config to `defaultTestOverrides` but created a bug for the publicly exposed `executeTests` function. This PR fixes that.

The issue was that we had to have both `global` and `defaultTestOverrides` always present as they weren't marked as optional in the `RunTestsCommandConfig` object.

### How?

We've fixed the problem by creating a separate object called `LegacyRunTestsCommandConfig` to hold the original form of the object containing `global` and made `RunTestsCommandConfig` only contain `defaultTestOverrides` then changed every usage of an object of type `RunTestsCommandConfig` to `LegacyRunTestsCommandConfig | RunTestsCommandConfig`

### Review checklist


src/commands/synthetics/interfaces.ts
```diff
@@ -436,14 +436,31 @@ export interface APIHelperConfig {
...
+  locations: string[]
+  mobileApplicationVersionFilePath?: string
+  pollingTimeout: number
+  publicIds: string[]
+  selectiveRerun: boolean
+  subdomain: string
+  testSearchQuery?: string
+  tunnel: boolean
+  variableStrings: string[]
+}
```
Drarig29, line 456: Can you change `LegacyRunTestsCommandConfig` to `extends Omit<>` from the other interface?

src/commands/synthetics/__tests__/run-tests-lib.test.ts
```diff
@@ -31,6 +31,59 @@ describe('run-test', () => {
...
+          pollingTimeout: 2 * 60 * 1000,
+          proxy: {protocol: 'http'},
+          publicIds: [],
+          selectiveRerun: false,
+          subdomain: 'app',
+          tunnel: false,
+          variableStrings: [],
+        })
+      ).rejects.toThrow(new CiError('NO_TESTS_TO_RUN'))
+    })
```
Drarig29, line 86: missing new line
```suggestion
    })

```

Drarig29 (approved):
~~LGTM~~ **Edit**: see comment below

Drarig29 (changes requested):
People still using code like `const config: synthetics.RunTestsCommandConfig = ...` will have TS complain with:

```ts
error TS2322: Type '{ apiKey: string; appKey: string; configPath: string; datadogSite: string; failOnCriticalErrors: false; failOnMissingTests: false; failOnTimeout: true; files: string[]; global: {}; locations: never[]; ... 6 more ...; variableStrings: never[]; }' is not assignable to type 'RunTestsCommandConfig'.
  Object literal may only specify known properties, and 'global' does not exist in type 'RunTestsCommandConfig'.

12   global: {},
     ~~~~~~~~~~
```

src/commands/synthetics/run-tests-command.ts
```diff
@@ -207,9 +208,12 @@ export class RunTestsCommand extends Command {
     }
 
     // Use global only if defaultTestOverrides does not exist
-    // SYNTH-12989: Clean up deprecated `global` in favor of `defaultTestOverrides`
-    if (Object.keys(this.config.global).length !== 0 && Object.keys(this.config.defaultTestOverrides).length === 0) {
-      this.config.defaultTestOverrides = {...this.config.global}
+    // TODO SYNTH-12989: Clean up deprecated `global` in favor of `defaultTestOverrides`
+    if (isUsingGlobal(this.config, this.reporter)) {
```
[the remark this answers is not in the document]
teodor2312: It is yes :D i messaged you that in slack :D i'm considering if i shouldn't just remove the `isUsingGlobal` function completely and going back to the if statements i already had before 🤔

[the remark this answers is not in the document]
teodor2312: yeah i think i'll do that 🤔

src/commands/synthetics/run-tests-lib.ts
```diff
@@ -1,6 +1,7 @@
 import {getProxyAgent} from '../../helpers/utils'
 
 import {APIHelper, getApiHelper, isForbiddenError} from './api'
+import {replaceGlobalWithDefaultTestOverrides} from './compatability'
```
Drarig29, line 4:
```suggestion
import {replaceGlobalWithDefaultTestOverrides} from './compatibility'
```

teodor2312: 🤦 oops will change it :D

Drarig29 (approved):
LGTM

PR #1294, opened 2024-04-26
[chores] Remove old unused files

Drarig29:
### What and why?

This PR removes:
- An old tslint config file
- 2 files which were added in https://github.com/DataDog/datadog-ci/pull/20, but never used (we never asked to sign CLA for this repository)

### How?

Remove files

### Review checklist

- ~~[ ] Feature or bugfix MUST have appropriate tests (unit, integration)~~


bin/prepend_license_header.sh
duncanista, line 1: Would it make sense to add the license header to the repo tho?

duncanista (approved):

bin/prepend_license_header.sh
Drarig29: We don't do it in other projects (at least in `dd-trace-js`), so let's not introduce it since we never had it in the first place

PR #1298, opened 2024-05-03
[SYNTH-12976] Ignore default polling timeout in override count

Drarig29:
### What and why?

The CLI output was showing `(1 config override)` for every test, and that was misleading since the user didn't set any custom polling timeout.

image

### How?

Do not count `pollingTimeout` if it's set to the default value.

### Review checklist


src/commands/synthetics/__tests__/utils/public.test.ts
```diff
@@ -652,6 +652,22 @@ describe('utils', () => {
     })
   })
 
+  describe('getTestOverridesCount', () => {
```
teodor2312, line 655: 👌

teodor2312 (approved):

Issue #1261, opened 2024-03-26
Add support for measures on trace command


Djiit:
### Feature request description

I'd like to be able to add measures using `--measures` (as per the current `--tags` behavior) with the `trace` command.

### Solution

Add support for measures on trace command

### Command

trace

PR #1300, opened 2024-05-09
Add measures for custom commands

edrevo:
### What and why?

Add measure support to custom commands. Fixes https://github.com/DataDog/datadog-ci/issues/1261

### How?

Send new field to backend.

### Review checklist

### DIFF
--- a/src/commands/trace/trace.ts
+++ b/src/commands/trace/trace.ts
@@ -30,8 +30,8 @@ export class TraceCommand extends Command {
         'datadog-ci trace --name "Say Hello" -- echo "Hello World"',
       ],
       [
-        'Trace a command with name "Say Hello" and a extra tags and report to Datadog',
-        'datadog-ci trace --name "Say Hello" --tags key1:value1 --tags key2:value2 -- echo "Hello World"',
+        'Trace a command with name "Say Hello", extra tags and measures and report to Datadog',
+        'datadog-ci trace --name "Say Hello" --tags key1:value1 --tags key2:value2 --measured key1:3.5 -- echo "Hello World"',
       ],
       [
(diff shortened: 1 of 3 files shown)


src/commands/trace/trace.ts
```diff
@@ -101,6 +104,7 @@ export class TraceCommand extends Command {
           error_message: stderr,
           exit_code: exitCode,
           is_error: exitCode !== 0,
+          measures,
```
juan-fernandez, line 107: are we OK with this sending null measures? If the user sets a string measure `parseFloat` would return `NaN` which is serialized as `null` by `JSON.stringify`: 
```
> JSON.stringify({ a: parseFloat('a')})
'{"a":null}'
```

juan-fernandez (approved):

src/commands/trace/trace.ts
edrevo: I tried that and it ends up as a 0 after all the processing. I'd say it's an ok-ish behavior, but maybe it's better to fail the command? What do you think?

Drarig29 (approved):
LGTM

src/commands/trace/trace.ts
juan-fernandez: I don't think it's worth failing the command, but a `0` doesn't seem correct either. Maybe we can do: 

```javascript
Object.fromEntries(
    Object.entries(cliMeasures)
        .map(([key, value]) => [key, parseFloat(value)])
        .filter(([key, value]) => {
            if (isNaN(value)) {
                logger.debug(`Value for ${key} is not a number`);
                return false;
            }
            return true;
        }),
);
```
WDYT? Maybe even without the logger.

juan-fernandez (approved):

PR #1301, opened 2024-05-09
[SYNTH-14344] Do not require etag from app part upload

kevin-pietsch:
### What and why?

Do not require the etag header be returned when uploading mobile application parts. Azure strictly does not return the etag header.

### How?

Just default to empty string for etag field when completing upload. The backend for azure expects this behavior

### Review checklist

### DIFF
--- a/src/commands/synthetics/api.ts
+++ b/src/commands/synthetics/api.ts
@@ -220,7 +220,7 @@ const uploadMobileApplicationPart = (request: (args: AxiosRequestConfig) => Axio
       request
     )
 
-    // Azure part-upload does not return ETag -- set to empty string
+    // Azure part-upload does not return ETag headers, so our backend ignores it for Azure
     const quotedEtag = isAzureUrl(presignedUrl) ? '' : (resp.headers.etag as string)
 
     return {
--- a/src/commands/synthetics/api.ts
+++ b/src/commands/synthetics/api.ts
@@ -221,9 +221,7 @@ const uploadMobileApplicationPart = (request: (args: AxiosRequestConfig) => Axio
     )
 
     // Azure part-upload does not return ETag -- set to empty string
-    const quotedEtag = isAzureUrl(presignedUrl)
-        ? ''
-        : resp.headers.etag as string
+    const quotedEtag = isAzureUrl(presignedUrl) ? '' : (resp.headers.etag as string)
 
     return {
       ETag: quotedEtag.replace(/"/g, ''),
@@ -373,7 +371,6 @@ export const getApiHelper = (config: APIHelperConfig): APIHelper => {
     baseUrl: getDatadogHost({useIntake: false, apiVersion: 'v1', config}),
     proxyOpts: config.proxy,
   })
-
 }
 
 const isAzureUrl = (presignedUrl: string) => {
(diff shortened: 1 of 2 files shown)


teodor2312 (approved):
LGTM

teodor2312:
Hey i noticed that there are tests for this tag is it possible to add a test case for empty string?

https://github.com/DataDog/datadog-ci/blob/a4041c4fca346122fa54dd67dad3d93d4f1731b9/src/commands/synthetics/__tests__/api.test.ts#L261-L290

Drarig29 (changes requested):
The `ETag` should uniquely represent the request resource for caching (see [docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag)). Defaulting it to an empty string means that all resources will be seen as the same, which could cause caching issues.

It would be better to not return an `ETag` property at all if it's falsy (empty string included).

kevin-pietsch:
> The `ETag` should uniquely represent the request resource for caching (see [docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag)). Defaulting it to an empty string means that all resources will be seen as the same, which could cause caching issues.
> 
> It would be better to not return an `ETag` property at all if it's falsy (empty string included).

@Drarig29 This is not setting any `ETag` headers to empty string -- we don't actually return any ETags, this is related to the behavior of the different cloud providers. We pass the ETag value from provider up to dogweb backend. In particular we return it because AWS S3 uses it as part of their object integrity check: https://docs.aws.amazon.com/AmazonS3/latest/userguide/checking-object-integrity.html

So right now this isn't used for caching in any meaningful way and we are not setting any ETag headers with it, so I don't think any concerns there.

I do agree that it would be better to have it explicitly optional in this code it was just significantly more painful as we need to change the API schema to allow that in dogweb, and I'd argue the value is fairly limited. 

But I can make the change -- I'll update here after modifying dogweb.

kevin-pietsch:
> Hey i noticed that there are tests for this tag is it possible to add a test case for empty string?
> 
> https://github.com/DataDog/datadog-ci/blob/a4041c4fca346122fa54dd67dad3d93d4f1731b9/src/commands/synthetics/__tests__/api.test.ts#L261-L290

I think this will be impacted by making it explicitly optional (discussed [https://github.com/DataDog/datadog-ci/pull/1301#issuecomment-2117863639](here)) -- but will add an explicit test with the update for present vs. non-present ETag field

Drarig29:
Discussed it offline, and we agreed it was OK to keep the empty string, and just add a unit test about it.

src/commands/synthetics/api.ts
```diff
@@ -220,7 +220,8 @@ const uploadMobileApplicationPart = (request: (args: AxiosRequestConfig) => Axio
       request
     )
 
-    const quotedEtag = resp.headers.etag as string
+    // Azure part-upload does not return ETag -- set to empty string
```
Drarig29, line 223: I still think it's useful to say _why_ is an empty string OK in our very specific case
```suggestion
    // Azure part-upload does not return ETag headers, and our backend ignores it for Azure
```

Drarig29 (approved):
LGTM

PR #1306, opened 2024-05-21
[SYNTH-12972] Change `config` to `testOverrides`

teodor2312:
### What and why?

Renaming `config` to `testOverrides` for the Test Config to disambiguate it all other usages of the word config throughout datadog-ci. 
We've moved to a intermediary state where we're still backwards compatible .A ticket for the eventual clean up has been created and linked in the appropriate places

### How?

Changed the interface in a backwards compatible way and change all occurrences of it again making sure it was backwards compatible

### Review checklist


src/commands/synthetics/__tests__/mobile.test.ts
```diff
@@ -21,412 +21,414 @@ import {
...
-            md5: 'QCi9PCxLLuyHmU0aRshoeA==',
-            partNumber: 1,
-          },
-        ],
-      })
-    } finally {
-      // always clean up created tmpdir
-      fs.rmSync(tmpdir, {recursive: true, force: true})
-    }
+describe('mobile tests', () => {
```
teodor2312, line 24: This is the only new line added to this file (well and the closing brackets on the last line) everything else is just indented. The fix for the failing test here was fixed in the `fixtures.ts` above

src/commands/synthetics/interfaces.ts
```diff
@@ -9,7 +9,7 @@ export interface MainReporter {
   log(log: string): void
   error(error: string): void
   initErrors(errors: string[]): void
-  testTrigger(test: Test, testId: string, executionRule: ExecutionRule, config: UserConfigOverride): void
+  testTrigger(test: Test, testId: string, executionRule: ExecutionRule, testOverrides: UserConfigOverride): void
```
teodor2312, line 12: Since this MainReporter is in `interfaces.ts` it's publicly exported but as far as i understand even then changing the name here should still be backwards compatible 🤔

src/commands/synthetics/reporters/default.ts
```diff
@@ -441,14 +441,14 @@ export class DefaultReporter implements MainReporter {
     test: Pick<Test, 'name'>,
     testId: string,
     executionRule: ExecutionRule,
-    config: UserConfigOverride
+    testOverrides: UserConfigOverride
```
teodor2312, line 444: This file is not publicly exported so changing the name here is not a breaking change

src/commands/synthetics/utils/internal.ts
```diff
@@ -16,10 +16,10 @@ import {getStrictestExecutionRule, isResultSkippedBySelectiveRerun} from './publ
 
 export const getOverriddenExecutionRule = (
   test?: Test,
-  configOverride?: UserConfigOverride
+  testOverrides?: UserConfigOverride
```
teodor2312, line 19: Not publicly exported

src/commands/synthetics/utils/public.ts
```diff
@@ -78,24 +79,29 @@ export const getOverriddenConfig = (
   test: Test,
   publicId: string,
   reporter: MainReporter,
-  config?: UserConfigOverride
+  // TODO SYNTH-12989: Clean up deprecated `config` in favor of `testOverrides`
+  config?: UserConfigOverride,
+  testOverrides?: UserConfigOverride
```
teodor2312, line 84: This whole file is publicly exported so had to set up compatibility stuff for each function

src/commands/synthetics/interfaces.ts
```diff
@@ -400,7 +400,10 @@ export interface BasicAuthCredentials {
   username: string
 }
 export interface TriggerConfig {
-  config: UserConfigOverride
+  // TODO SYNTH-12989: Clean up deprecated `config` in favor of `testOverrides`
+  /** @deprecated This property is deprecated, please use `testOverrides` instead. */
+  config?: UserConfigOverride
+  testOverrides?: UserConfigOverride
```
teodor2312, line 406: Did it the same way as we did it for `global`

src/commands/synthetics/__tests__/mobile.test.ts
Drarig29: What's the benefit of this parent `describe` block?

src/commands/synthetics/compatibility.ts
```diff
@@ -24,3 +24,18 @@ export const replaceGlobalWithDefaultTestOverrides = (
...
 }
+
+export const replaceConfigWithTestOverrides = (
+  config?: UserConfigOverride,
+  testOverrides?: UserConfigOverride
+): UserConfigOverride => {
+  const isConfigUsed = Object.keys(config ?? {}).length !== 0
+  const isTestOverridesUsed = Object.keys(testOverrides ?? {}).length !== 0
+
+  // If both config and testOverrides exist use testOverrides
```
Drarig29, line 35: **nit** add a comma to make it breathe 😁 (same above)
```suggestion
  // If both config and testOverrides exist, use testOverrides
```

src/commands/synthetics/utils/public.ts
Drarig29: This doesn't make sense: the content of the object stays the same, and the type as well. Since you would only change the name of the parameter, this is not a breaking change from TypeScript's and even JavaScript's point of view.

This function can stay untouched. Only our usages of it should be updated, and you already did it 👍

src/commands/synthetics/compatibility.ts
```diff
@@ -24,3 +24,18 @@ export const replaceGlobalWithDefaultTestOverrides = (
...
+  if (isConfigUsed && !isTestOverridesUsed) {
+    return {...config}
```
Drarig29, line 37: No need to do a copy, and it hides the function's purpose

```suggestion
    return config
```

src/commands/synthetics/test.ts
```diff
@@ -15,10 +16,21 @@ export const getTestConfigs = async (
...
 
+  // TODO SYNTH-12989: Clean up deprecated `config` in favor of `testOverrides`
+  const isUsingConfig = suites.some((suite) =>
+    suite.content.tests.some((test) => Object.keys(test.config ?? {}).length > 0)
+  )
+  if (isUsingConfig) {
+    reporter?.error(
+      "The 'config' property is deprecated. Please use 'testOverrides' instead.\nIf both 'config' and 'testOverrides' properties exist, 'testOverrides' is used!\n"
+    )
+  }
```
Drarig29, line 27: I understand that you can't log the warning in `replaceConfigWithTestOverrides()` because it's in a loop. But I'd like that to have everything about compatibility (logic, and generated warnings) in a single file.

You could move this into a `warnIfDeprecatedConfigUsed()`:

```suggestion
  warnIfDeprecatedConfigUsed(...)
```

src/commands/synthetics/compatibility.ts
teodor2312: > What's the benefit of this parent describe block?

Being able to run all of the test in the file all at once

teodor2312:
@Drarig29 

> What's the benefit of this parent describe block?

Being able to run all of the test in the file all at once

src/commands/synthetics/compatibility.ts
teodor2312: Is there a reason not to make a copy? This way i feel like i separate more config from testOverrides. If config gets changed in some weird way it won't impact testOverrides 

Also what do you mean by it hides the function's purpose 🤔

drichards-87 (approved):

teodor2312:
> This doesn't make sense: the content of the object stays the same, and the type as well. Since you would only change the name of the parameter, this is not a breaking change from TypeScript's and even JavaScript's point of view.
>
> This function can stay untouched. Only our usages of it should be updated, and you already did it 👍

Ah right, because the position of the parameter remains the same, right?

src/commands/synthetics/test.ts
teodor2312: Yeah i originally had a function in compatibility.ts doing that, but then moved some stuff around and i didn't like how it looked. I guess in this last iteration it is possible to extract it again though 👍

src/commands/synthetics/__tests__/mobile.test.ts
Drarig29: > Being able to run all of the test in the file all at once

@teodor2312 you can already run `yarn test src/commands/synthetics/__tests__/mobile.test.ts` to do exactly that 😁 

But it's true that we have one in other files. However, I think that it's superfluous to add the word "tests" in the name of the describe block, and we usually just use the file name:

```suggestion
describe('mobile', () => {
```

src/commands/synthetics/compatibility.ts
teodor2312: okay i see 🤔

src/commands/synthetics/utils/public.ts
```diff
@@ -78,24 +79,25 @@ export const getOverriddenConfig = (
   test: Test,
   publicId: string,
   reporter: MainReporter,
-  config?: UserConfigOverride
+  // TODO SYNTH-12989: Clean up deprecated `config` in favor of `testOverrides`
```
Drarig29, line 82:
```suggestion
```

src/commands/synthetics/utils/public.ts
```diff
@@ -459,19 +462,23 @@ export const getTestAndOverrideConfig = async (
...
+  const testResult = await getTest(api, {id: normalizedId, suite})
   if ('errorMessage' in testResult) {
     summary.testsNotFound.add(normalizedId)
 
     return {errorMessage: testResult.errorMessage}
   }
 
   const {test} = testResult
-  const overriddenConfig = getOverriddenConfig(test, normalizedId, reporter, config)
+  // TODO SYNTH-12989: Clean up deprecated `config` in favor of `testOverrides`
```
Drarig29, line 476:
```suggestion
```

Drarig29 (approved):
LGTM

src/commands/synthetics/__tests__/mobile.test.ts
teodor2312: I run tests from inside my VS Code though i and i have a little `Run | Debug` thing, which is what i need the big describe for, but yes also there's usually a big describe in other tests as well :D

image

PR #1316, opened 2024-05-27
[SYNTH-12985] Add properties to env variables

AntoineDona:
### What and why?

Many configuration properties were not configurable through environment variables:  `configPath`, `failOnCriticalErrors`, `failOnMissingTests`,`failOnTimeout`,`files`,`jUnitReport`,`latest`,`mobileApplicationId`,`publicIds`,`selectiveRerun`,`testSearchQuery`,`tunnel` and `versionName`. This PR adds support for configuring these properties through environment variables, all prefixed with `DATADOG_SYNTHETICS_`.

### How?

We fixed this problem by adding the the missing environment variables. In this PR, we also:

- Added `jUnitReport` property to the `Global Config`
- Added `testSearchQuery` to the `DEFAULT_COMMAND_CONFIG` as it was missing
- Created a utility function `toBoolean` to convert environment variable strings to boolean values, while handling edge cases (undefined values).

### Review checklist

### DIFF
--- a/src/commands/synthetics/run-tests-command.ts
+++ b/src/commands/synthetics/run-tests-command.ts
@@ -144,7 +144,6 @@ export class RunTestsCommand extends Command {
   private config: RunTestsCommandConfig = JSON.parse(JSON.stringify(DEFAULT_COMMAND_CONFIG)) // Deep copy to avoid mutation
 
   public async execute() {
-
     try {
       await this.resolveConfig()
     } catch (error) {
--- a/src/commands/synthetics/run-tests-command.ts
+++ b/src/commands/synthetics/run-tests-command.ts
@@ -12,6 +12,7 @@ import {MainReporter, Reporter, Result, RunTestsCommandConfig, Summary} from './
 import {DefaultReporter} from './reporters/default'
 import {JUnitReporter} from './reporters/junit'
 import {executeTests} from './run-tests-lib'
+import {toBoolean} from './utils/internal'
 import {
   getExitReason,
   getOrgSettings,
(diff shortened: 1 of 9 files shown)


src/commands/synthetics/run-tests-command.ts
```diff
@@ -195,10 +198,22 @@ export class RunTestsCommand extends Command {
   private async resolveConfig() {
     // Defaults < file < ENV < CLI
 
+    let overrideConfigPath = this.config.configPath
```
jakepruitt, line 201: I think this will never be defined, since the config file can't specify the path to... itself. I believe we want to start with the env config as the baseline, then put the CLI on top of it.

src/commands/synthetics/upload-application-command.ts
```diff
@@ -106,7 +107,11 @@ export class UploadApplicationCommand extends Command {
       removeUndefinedValues({
         apiKey: process.env.DATADOG_API_KEY,
         appKey: process.env.DATADOG_APP_KEY,
+        configPath: process.env.DATADOG_SYNTHETICS_CONFIG_PATH,
```
jakepruitt, line 110: I believe you'll need to do a similar trick above -- we want to take the `process.env.DATADOG_SYNTHETICS_CONFIG_PATH` env var into consideration when calling `resolveConfigFromFile`.

src/commands/synthetics/run-tests-command.ts
```diff
@@ -158,6 +154,13 @@ export class RunTestsCommand extends Command {
       return 1
     }
 
+    const reporters: Reporter[] = [new DefaultReporter(this)]
+    this.reporter = getReporter(reporters)
+
+    if (this.config.jUnitReport) {
+      reporters.push(new JUnitReporter(this))
```
jakepruitt, line 161: I think the `JUnitReporter` expects there to be a top-level `jUnitReport` available. We should change this behavior to rely on `this.config.jUnitReporter` instead

src/commands/synthetics/upload-application-command.ts
AntoineDona: oh right, forgot about this one !

kayayarai (approved):

PR #1325, opened 2024-05-31
Add step function context to nested step function calls

agocs:
### What and why?
Adds the step function context to the state definition when a step function calls another stepfunction. This is gated behind the `--propagate-upstream-trace` or `--merge-step-function-and-lambda-traces` flags.

### How?

### Review checklist

### DIFF
--- a/src/commands/stepfunctions/helpers.ts
+++ b/src/commands/stepfunctions/helpers.ts
@@ -189,13 +189,14 @@ export type StepType = {
 export type ParametersType = {
   'Payload.$'?: string
   FunctionName?: string
+  StateMachineArn?: string
   TableName?: string
   Input?: {
     'CONTEXT.$'?: string
   }
 }
 
-function injectContextForLambdaFunctions(step: StepType, context: BaseContext, stepName: string) {
+const injectContextForLambdaFunctions = (step: StepType, context: BaseContext, stepName: string): boolean => {
   if (shouldUpdateStepForTracesMerging(step)) {
     addTraceContextToLambdaParameters(step)
 
@@ -211,7 +212,7 @@ function injectContextForLambdaFunctions(step: StepType, context: BaseContext, s
   return false
 }
 
-function injectContextForStepFunctions(step: StepType) {
+export const injectContextForStepFunctions = (step: StepType): boolean => {
   if (shouldUpdateStepForStepFunctionContextInjection(step)) {
     addTraceContextToStepFunctionParameters(step)
 
--- a/src/commands/stepfunctions/helpers.ts
+++ b/src/commands/stepfunctions/helpers.ts
@@ -117,7 +117,7 @@ export const addTraceContextToLambdaParameters = ({Parameters}: StepType): void 
 
 export const addTraceContextToStepFunctionParameters = ({Parameters}: StepType): void => {
   if (Parameters) {
-    if (!Parameters.Input){
+    if (!Parameters.Input) {
       Parameters.Input = {}
     }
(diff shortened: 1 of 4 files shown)


src/commands/stepfunctions/instrument.ts
```diff
@@ -350,15 +351,10 @@ export class InstrumentStepFunctionsCommand extends Command {
         hasChanges = true
       }
 
-      if (this.mergeStepFunctionAndLambdaTraces) {
+      if (this.mergeStepFunctionAndLambdaTraces || this.propagateUpstreamTrace) {
```
nine5two7, line 354: So, if either one of the flag is used, the nested SF definition is overwritten, and there are 3 possible case
Lambda: Yes, Nested: Yes
Lambda: No, Nested: No
Lambda: No, Nested: Yes.
-------
BTW: Do we really need to have an extra flag to support the case of Lambda No, Nested Yes? What about we reuse the current flag "mergeStepFunctionAndLambdaTraces", and once it is used by the customer, both definitions of Lambda and nested SF are modified.  @sumedha.

agocs: I'm fine reusing `mergeStepFunctionAndLambdaTraces` for both. We discussed this in #serverlesstraceandspanlinking, but didn't really come to a consensus. 

> Chris Agocs
>   12:25
> @sumedha @kimi What do we think of --trace-nested-step-function-invocations as a flag to pass to datadog ci to toggle passing context between nested step function invocations? (edited) 
> 
> Chris Agocs
>   14:04
> @sumedha, the other option is, do we want to make it one flag that controls both Stepfunction -> Lambda and Stepfunction -> Stepfunction? Thoughts? (edited) 
> 
> 
> Sumedha Mehta
>   14:53
> like --propogate-upstream-trace ? (edited) 
> 14:54
> I don't see why a customer would want to turn it on and off per-service
> 
> 
> Kimi Wu
>   14:58
> Are you proposing a new flag to replace the old one? I think we still need to keep the old one working if that's the suggestion though.
> 
> 
> Chris Agocs
>   15:09
> I'd keep the old flag as well to preserve backward compatibility. I think it would be safe for --merge-step-function-and-lambda-traces and --propogate-upstream-trace to do the same thing, and turn on trace propagation from stepfunction->lambda and stepfunction->stepfunction (edited)

nine5two7 (approved):
LGTM

src/commands/stepfunctions/helpers.ts
```diff
@@ -142,6 +143,34 @@ export const shouldUpdateStepForTracesMerging = (step: StepType): boolean => {
...
 }
 
+// Truth table
+// Input                    | Expected
+// -------------------------|---------
+// Empty object             | true
+// undefined                | true
+// not object               | false
+// object without CONTEXT.$ | true
+// object with CONTEXT.$    | false
```
Drarig29, line 153: Why have a truth table when the function is unit tested?

Drarig29: Can you have 1 test case per line in this truth table, and remove the truth table?

agocs: Due to the way Stepfunctions are defined, the Input can be any valid JSON value. We really only care about the case where Input is {"CONTEXT.$":{...}}, so that's how the type is defined. Instead of changing it to any, I'd rather leave a note to future developers explaining what this function returns, and cover it with test cases as best I can.

I added a test case I missed, that was possible to add without changing the type of Input.

PR #1328, opened 2024-06-03
Explore .dot directories when uploading elf symbols

nsavoire:
### What and why?

Debug symbols present in a `.debug` directory were ignored by `elf-symbols upload` command.

### How?

Add `dot: true` option to `glob` to ensure that `.dot` directories/files are matched.

Also display more information when uploading elf symbol file (arch + Go/GNU buildID + file hash).

### Review checklist


sanchda (approved):
Absolutely true, good catch.  I can't comment on how this patch fits into the larger repo, but approving from the native side as this is indeed a known issue with the location of split debuginfo.

Drarig29 (approved):

PR #1331, opened 2024-06-07
[PROF-9872] Improve elf-symbols upload command

nsavoire:
### What and why?

* `symbols-location` is now a positional argument
* Allow multiple input locations
* Ignore unreadable directories
    `glob` throws an error when it encounter an unreadable directory.
    Pass `strict: false`, to make `glob` ignore these errors and then later
    recheck if directories are readable and emit a warning if not.
    Top-level directory being unreadable results in a an error.

### Review checklist

### DIFF
--- a/src/commands/elf-symbols/__tests__/upload.test.ts
+++ b/src/commands/elf-symbols/__tests__/upload.test.ts
@@ -38,20 +38,20 @@ describe('elf-symbols upload', () => {
   }
 
   describe('parameter validation', () => {
-    test('fails if symbols-location is blank', async () => {
+    test('fails if symbols locations is empty', async () => {
       const {exitCode, context} = await runCommand((cmd) => {
-        cmd['symbolsLocation'] = ''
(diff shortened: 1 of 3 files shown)


src/commands/elf-symbols/__tests__/upload.test.ts
```diff
@@ -98,28 +100,38 @@ describe('elf-symbols upload', () => {
       const command = createCommand(UploadCommand)
       await expect(command['getElfSymbolFiles'](`${fixtureDir}/go_x86_64_only_go_build_id`)).rejects.toThrow()
     })
+
+    test('should no throw an error when a directory (except top-level) is not readable', async () => {
```
Drarig29, line 104:
```suggestion
    test('should not throw an error when a directory (except top-level) is not readable', async () => {
```

Drarig29 (approved):
LGTM

r1viollet (approved):
LGTM
Thanks for the fix and improvements.

PR #1333, opened 2024-06-10
[ci] Add build job for Ubuntu ARM

gjulianm:
### What and why?

This PR adds a Github job that creates standalone binaries for ARM64 

### How?

Using the Github runners in ARM, I duplicated the `build-binary-ubuntu` job to generate ARM64 binaries.

### Review checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -205,7 +205,7 @@ jobs:
       - run: yarn install --immutable
       - run: yarn build
       - name: Create standalone binary
-        run: yarn dist-standalone -t node14-macos-x64 -o datadog-ci_darwin-x64
+        run: yarn dist-standalone -t node14-macos-arm64 -o datadog-ci_darwin-arm64
       - name: Remove dist and src folder to check that binary can stand alone
         run: |
           rm -rf dist
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -147,6 +147,28 @@ jobs:
       - name: Test generated standalone binary
         run: yarn dist-standalone:test
 
+  standalone-binary-test-ubuntu-arm:
+    name: Test standalone binary in ubuntu
+    runs-on:
+      group: ARM LINUX SHARED
+      labels: arm-4core-linux
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install node
+        uses: actions/setup-node@v4
+        with:
+          node-version: 20.11.0 # Use newer version to build the standalone binary
+      - run: yarn install --immutable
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -147,6 +147,30 @@ jobs:
       - name: Test generated standalone binary
         run: yarn dist-standalone:test
 
+  standalone-binary-test-ubuntu-arm:
+    name: Test standalone binary in ARM ubuntu
+    runs-on:
+      group: ARM LINUX SHARED
+      labels: arm-4core-linux
```
Drarig29, line 154: Apparently, [other Datadog repos](https://github.com/search?q=org%3ADataDog+%22arm-4core-linux%22&type=code) use the following syntax:
```suggestion
    runs-on: arm-4core-linux
```

gjulianm: I'm using this syntax as it's the one the folks at #build-support recommended ([docs](https://datadoghq.atlassian.net/wiki/spaces/DEVX/pages/3790274750/Github+Actions)), in theory this should be more robust against changes

gjulianm:
By the way, based on [this job result](https://github.com/DataDog/datadog-ci/actions/runs/9450981908/job/26030865507?pr=1333) it seems like the macOS runners are running ARM64 too, but the naming of the binary implies it's x64.

.github/workflows/ci.yml
Drarig29: Thanks, I wasn't aware of these docs 👍

Drarig29:
@gjulianm oh, thanks. The naming used to be correct, but it became incorrect recently because `macos-latest` went from x64 to arm64 (see https://github.com/DataDog/datadog-ci/pull/1291 for a related fix).

To bring back a x64 macOS binary, we'd need another job - on `macos-latest-large` (see [docs](https://github.com/actions/runner-images)).

But since nobody complained, I guess the macOS binary isn't used, so let's just align the naming and keep arm64 only.

gjulianm:
> @gjulianm oh, thanks. The naming used to be correct, but it became incorrect recently because `macos-latest` went from x64 to arm64 (see #1291 for a related fix).
> 
> To bring back a x64 macOS binary, we'd need another job - on `macos-latest-large` (see [docs](https://github.com/actions/runner-images)).
> 
> But since nobody complained, I guess the macOS binary isn't used, so let's just align the naming and keep arm64 only.

Do you want me to do it in this PR? It's a quick change so it's no problem

Drarig29:
@gjulianm cool, then I'm not against it! Thanks 🙇

PR #1339, opened 2024-06-13
Jward/unity symbols

fuzzybinary:
### What and why?

Add Android `.
### How?

### Review checklist


src/commands/unity-symbols/README.md
```diff
@@ -28,20 +28,27 @@ To override the full URL for the intake endpoint, define the `DATADOG_SOURCEMAP_
...
+```bash
+datadog unity-symbols upload --android
 ```
 
 | Parameter | Condition | Description |
 |-----------|-----------|-------------|
-| `--symbols-location`  | Optional  | The location of of your dSYMs, `build_id` and `LineNumberMappings.json` file.  Defaults to `datadogSymbols`. |
+| `--ios` or `--android` | One is required | Specify the platform we are uploading symbols for, iOS or Android |
+| `--variant` | Optional, Android Only | Specify the variant that was built. Defaults to `release` |
+| `--symbols-location`  | Optional, iOS Only | The location of of your dSYMs, `build_id` and `LineNumberMappings.json` file.  Defaults to `datadogSymbols`. |
```
alai97, line 49:
```suggestion
| `--variant` | Optional, Android only | Specify the variant that was built. Defaults to `release`. |
| `--symbols-location`  | Optional, iOS only | The location of your dSYMs, `build_id`, and `LineNumberMappings.json` file. Defaults to `datadogSymbols`. |
```

src/commands/unity-symbols/__tests__/fixtures/androidSymbols/README.md
```diff
@@ -0,0 +1,5 @@
+# NOTE
+
+This fixture does contain actual symbol files from a build of the Unity Sample app, but not all of them,
+as the `il2cpp` so file is very large. These are actual symbol files so we can test that the
```
alai97, line 4:
```suggestion
as the `il2cpp` so file is very large. These are actual symbol files so Datadog can test that the
```

alai97 (commented):
Looks good, two small suggestions.

src/commands/unity-symbols/README.md
fuzzybinary: 👍 Though actually you've pointed out a mistake. `--symbols-locaiton` is no longer iOS only

alai97 (approved):

marco-saia-datadog (approved):
Looks good 👍 

**Non-blocking suggestion**:

In order to avoid using non-null assertion operators, and having to rely on the fact that `verifyParameters()` will make sure that the variables are correctly set, I would provide some dedicated getters (while also keeping the `verifyParameters` logic where necessary).

Consider this example code:

```typescript
const ENV_IOS = false;
const ENV_ANDROID = false;
const ENV_SYMBOLS_PATH = '';

type UploadPlatform = 'ios' | 'android';

class TestClass {
    // User Vars
    private ios: boolean = ENV_IOS;
    private android: boolean = ENV_ANDROID;
    private symbolsPath: string | undefined = ENV_SYMBOLS_PATH;

    // Properties
    private get uploadPlatform(): UploadPlatform {
        if (this.ios && this.android) {
            throw Error("THROW AN ERROR");
        }

        if (!this.ios && !this.android) {
            throw Error("THROW AN ERROR");
        }

        return this.ios ? 'ios' : 'android';
    }

    private get symbolsLocation(): string {
        const symbolsPath = this.symbolsPath;

        if (symbolsPath !== undefined) {
            return symbolsPath;
        }

        switch (this.uploadPlatform) {
            case 'ios':
                return './datadogSymbols';
            case 'android':
                return './unityLibrary/symbols';
        }
    }
}
```

These properties can also be made lazy, for example:

```typescript
    private _symbolsLocation: string | undefined;
    private get symbolsLocation(): string {
        if (this._symbolsLocation !== undefined) {
            return this._symbolsLocation;
        }

        let path: string;

        if (this.symbolsPath !== undefined) {
            path = this.symbolsPath;
        } else {
            path = this.uploadPlatform == 'ios' ? './datadogSymbols' : './unityLibrary/symbols';
        }

        this._symbolsLocation = path;

        return path;
    }
```

fuzzybinary:
It's not a bad suggestion, but I'd have to research whether it's compatible with the framework we're using for command line parameter parsing. I think I would have to provide getters / setters for those properties because of how that framework works.

Drarig29:
@marco-saia-datadog non-null assertions are used a lot in that codebase because of how Clipanion (the framework) works (entrypoints are class methods, and not constructors), and I'm afraid having getters like this would make us deviate from using framework features - like validation and default values - and make it harder to have an homogeneous codebase.

I would rather use non-null assertions for now, and make a card to tackle it in the whole codebase. They are easy to find: we have an ESLint rule that warns us about them already 😁

src/commands/unity-symbols/renderer.ts
```diff
@@ -72,8 +72,14 @@ export const renderCommandSummary = (statuses: UploadStatus[], duration: number,
   return output.join('\n') + '\n'
 }
 
+export const renderMustSupplyPlatform = () =>
+  chalk.red(`${ICONS.FAILED} Error: must supply either iOS or Android as your platform.\n`)
```
Drarig29, line 76:
```suggestion
  chalk.red(`${ICONS.FAILED} Error: Must supply either iOS or Android as your platform.\n`)
```

Drarig29 (approved):
LGTM, just one comment

PR #1285, opened 2024-04-16
[SYNTH-12971] Rename global to defaultTestOverrides

teodor2312:
### What and why?

Renaming `global` field in the Global Config to `defaultTestOverrides`. This is done becase having it named global is confusing. We've moved to a intermediary state where we're still backwards compatible and a ticket for the eventual clean up has been created and linked in the appropriate places

### How?

Created a new field called `defaultTestOverrides` and if it's not defined but `global` is we copy over to it whatever is in `global`. Otherwise we just use `defaultTestOverrides` as is. The rest of the code has been transitioned to use `defaultTestOverrides`

### Review checklist

- [x] Feature or bugfix MUST have appropriate tests (unit, integration)

PR #1340, opened 2024-06-14
[synthetics] Parameterize unit tests for deprecated usages

Drarig29:
### What and why?

We deprecated `global` and `config` in favor of `defaultTestOverrides` and `testOverrides` (#1285 and #1306), and we still want to unit test the old usage.

Instead of duplicating some unit tests, we are using `test.each()` to test the deprecated and current behaviors.

### How?

Use `test.each()` and computed object properties

### Review checklist

### DIFF
--- a/src/commands/synthetics/__tests__/run-tests-lib.test.ts
+++ b/src/commands/synthetics/__tests__/run-tests-lib.test.ts
@@ -7,7 +7,7 @@ import {CiError, CriticalCiErrorCode, CriticalError} from '../errors'
 import {ExecutionRule, RunTestsCommandConfig, Summary, UserConfigOverride} from '../interfaces'
 import {DefaultReporter} from '../reporters/default'
 import {JUnitReporter} from '../reporters/junit'
-import {MAX_TESTS_TO_TRIGGER} from '../run-tests-command'
+import * as appUploadReporterModule from '../reporters/mobile/app-upload'
 import * as runTests from '../run-tests-lib'
 import {Tunnel} from '../tunnel'
 import * as utils from '../utils/public'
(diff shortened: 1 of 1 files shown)


src/commands/synthetics/__tests__/run-tests-lib.test.ts
```diff
@@ -85,47 +102,50 @@ describe('run-test', () => {
...
+            ...ciConfig,
+            [defaultTestOverrides]: userConfigOverride,
+            publicIds: ['aaa-aaa-aaa', 'bbb-bbb-bbb'],
+          })
+        ).rejects.toThrow()
+        expect(getTestsToTriggersMock).toHaveBeenCalledWith(
+          apiHelper,
+          expect.arrayContaining([
+            expect.objectContaining({id: 'aaa-aaa-aaa', testOverrides: userConfigOverride}),
+            expect.objectContaining({id: 'bbb-bbb-bbb', testOverrides: userConfigOverride}),
```
Drarig29, line 136: As you can see, we expect the object to contain `testOverrides` even in the deprecated case because of our backwards compatibility translation layer.

AntoineDona (approved):

src/commands/synthetics/__tests__/run-tests-lib.test.ts
```diff
@@ -390,6 +418,11 @@ describe('run-test', () => {
       const testStream = fs.createReadStream('/dev/null')
       jest.spyOn(fs, 'createReadStream').mockReturnValue(testStream)
 
+      const {AppUploadReporter} = jest.requireActual<typeof appUploadReporterModule>('../reporters/mobile/app-upload')
+      jest
+        .spyOn(appUploadReporterModule, 'AppUploadReporter')
+        .mockImplementation(() => new AppUploadReporter({stdout: {write: jest.fn()}} as any))
+
```
teodor2312, line 425: Just out of curiosity why is this part needed?

teodor2312 (approved):
LGTM

src/commands/synthetics/__tests__/run-tests-lib.test.ts
Drarig29: Sorry, didn't see the comment. It's unrelated - just to avoid having logs printed in the output by some reporters which aren't mocked

PR #1347, opened 2024-06-18
[STAL-1727] Hard-coding source to be `ci`

modernplumbing:
### What and why?
We'd like to collect metrics on the source of scans, the potential values being `ci` and `cwp`. `local` has such negligible value that lumping them in with `ci` is tolerable. This PR hardcodes that all scans triggered from `datadog-ci` are coming from `ci` and not `cwp`.

### DIFF
--- a/src/commands/sbom/types.ts
+++ b/src/commands/sbom/types.ts
@@ -108,5 +108,6 @@ export interface ScaRequest {
   dependencies: Dependency[]
   service: string
   env: string
+  scan_source: 'ci'
   tags: Record<string, string>
 }
--- a/src/commands/sbom/payload.ts
+++ b/src/commands/sbom/payload.ts
@@ -176,5 +176,6 @@ export const generatePayload = (
     dependencies,
     service,
     env,
+    scan_source: 'ci',
   }
 }


src/commands/sbom/types.ts
```diff
@@ -108,5 +108,6 @@ export interface ScaRequest {
   dependencies: Dependency[]
   service: string
   env: string
+  source: 'ci'
```
juli1, line 111: There is already the source of the scan identifier in the header of the request to the API. This change should not be done in the API but rather sent by CWP. See [this file](https://github.com/DataDog/dd-source/blob/main/domains/static-analysis/apps/apis/static-analysis-sca-api/internal/services/evp_sbom.go#L21).

Drarig29:
@modernplumbing you'll want to merge `master` in your PR to unblock the CI

src/commands/sbom/types.ts
modernplumbing: this is great info, just what I needed, thanks!

modernplumbing:
closing because I got this intel:

> There is already the source of the scan identifier in the header of the request to the API. This change should not be done in the CLI but rather in the API that already knows the source of the request. See [this file](https://github.com/DataDog/dd-source/blob/main/domains/static-analysis/apps/apis/static-analysis-sca-api/internal/services/evp_sbom.go#L21).

PR #1332, opened 2024-06-10
[SYNTH-12981] Fix usage of locations

AntoineDona:
### What and why?

-   (← **you are here**)
- #1330
- #1327 
- #1320 

In a previous PRs, we implemented the new `--override` CLI parameter that allows to configure testOverrides properties through the **CLI** and the **ENV variables**. We didn't add support for all of them though, as some were more complex to implement than others. This PR adds CLI and ENV support for the `locations` property.

### How?

- Add override from CLI for `locations`
- Renamed `DATADOG_SYNTHETICS_LOCATIONS` to `DATADOG_SYNTHETICS_OVERRIDE_LOCATIONS` as it's now part of the overrides
- Moved `locations` from the root of the config to `defaultTestOverrides`
- Implemented retrocompatibility functions

### Review checklist

- [x] Feature or bugfix MUST have appropriate tests (unit, integration)

PR #1330, opened 2024-06-05
[SYNTH-12984] Add  complex overrides to `--override` CLI argument

AntoineDona:
### What and why?

- #1330 (← **you are here**)
- #1327 
- #1320 

In a previous PRs, we implemented the new `--override` CLI parameter that allows to configure testOverrides properties through the **CLI** and the **ENV variables**. We didn't add support for all of them though, as some were more complex to implement than others. This PR adds CLI and ENV support for the following properties: `basicAuth`, `cookies`, `deviceIds` and `headers`. 

Here is a brief example of how the new CLI override options should be used:

```
datadog-ci synthetics run-tests \
  --public-id aaa-bbb-ccc \
  --override basicAuth.username='username'
  --override basicAuth.password='Sup3r_S3cur3_P@ssw0rd'
  --override cookies='cookie1;cookie2;cookie3'
  --override deviceIds='dev1;dev2;dev3'
  --override headers.Authorization='Bearer token'
  --override headers.FANCY_HEADER='nice'
```

And here is how the ENV variables should be passed:

```
export DATADOG_SYNTHETICS_OVERRIDE_BASIC_AUTH_PASSWORD=Sup3r_S3cur3_P@ssw0rd
export DATADOG_SYNTHETICS_OVERRIDE_BASIC_AUTH_USERNAME=username
export DATADOG_SYNTHETICS_OVERRIDE_COOKIES="cookie1;cookie2;cookie3"
export DATADOG_SYNTHETICS_OVERRIDE_DEVICE_IDS="dev1;dev2;dev3"
export DATADOG_SYNTHETICS_OVERRIDE_HEADERS='{
       "Content-Type": "application/json", 
        "Authorization": "Bearer token"
}'
```

### How?

- Implemented the possibility to configure these properties through ENV variables and CLI.
- Added a `toStringObject` parser, to parse the headers from ENV

### Review checklist

- [x] Feature or bugfix MUST have appropriate tests (unit, integration)

PR #1327, opened 2024-06-03
add override cli argument

AntoineDona:
### What and why?

We wanted to add an option for testOverrides properties to be configured through the `--override` CLI parameter, which is a clipanion command array. This PR adds CLI support for the following properties: `allowInsecureCertificates`, `body`, `bodyType`, `defaultStepTimeout`, `executionRule`, `followRedirects`,  `resourceUrlSubstitutionRegexes`, `retry`, `startUrl`, `startUrlSubstitutionRegex` and `testTimeout`. Here is a brief example of how the new CLI option should be used:

```
datadog-ci synthetics run-tests \
  --public-id aaa-bbb-ccc \
  --override testTimeout=999 \
  --override startUrlSubstitutionRegex="s/prod/staging/" \
  --override allowInsecureCertificates=true
```

### How?

- Created a new `--override` CLI option
- Added a validator and parser to process every parameter passed through the `--override` option

### Review checklist

- [x] Feature or bugfix MUST have appropriate tests (unit, integration)

PR #1320, opened 2024-05-28
[SYNTH-12983] Add testOverrides properties to env variables

AntoineDona:
### What and why?

We wanted to add an option for testOverrides properties to be configured through ENV variable, starting with `DATADOG_SYNTHETICS_OVERRIDE_`. This PR adds ENV variable for the following properties: `allowInsecureCertificates`, `body`, `bodyType`, `defaultStepTimeout`, `executionRule`, `followRedirects`, `pollingTimeout`, `resourceUrlSubstitutionRegexes`, `retry`, `startUrl`, `startUrlSubstitutionRegex` and `testTimeout`

### How?

- Added the variables listed above as ENV variables
- Created a utility function toNumber to convert environment variables to numbers, while handling edge cases (undefined values).

### Review checklist

- [x] Feature or bugfix MUST have appropriate tests (unit, integration)

PR #1354, opened 2024-06-19
[SYNTH-12982] Fix the usage of variableStrings

AntoineDona:
### What and why?

- #1354 (← **you are here**)
- #1332
- #1330
- #1327 
- #1320 

In a previous PRs, we implemented the new `--override` CLI parameter that allows to configure testOverrides properties through the **CLI** and the **ENV variables**. We didn't add support for all of them though, as some were more complex to implement than others. 

This PR adds CLI and ENV support for the `variables` property, and removes the deprecated `variableStrings` that was used to get variables from the CLI.

### How?

- Change override from CLI for `variables` from `--variables` to `--override variables.NAME=VALUE`
- Add new `DATADOG_SYNTHETICS_OVERRIDE_VARIABLES` ENV variable
- No need for retrocompatibility functions, as `this.config.variableStrings` was never used.

### Review checklist


src/commands/synthetics/utils/internal.ts
```diff
@@ -253,11 +255,12 @@ export const validateAndParseOverrides = (overrides: string[] | undefined): Accu
...
 
-        // Convert to {[key: string]: string}
+        // Convert to StringObject
         case 'headers':
+        case 'variables':
           if (subKey) {
-            acc['headers'] = acc['headers'] ?? {}
-            acc['headers'][subKey] = value
+            acc[key] = acc[key] ?? {}
+            ;(acc[key] as StringObject)[subKey] = value
```
teodor2312, line 263: Why is there a `;` at the start of the line here 🤔

src/commands/synthetics/run-tests-command.ts
```diff
@@ -299,6 +301,7 @@ export class RunTestsCommand extends Command {
         startUrl: process.env.DATADOG_SYNTHETICS_OVERRIDE_START_URL,
         startUrlSubstitutionRegex: process.env.DATADOG_SYNTHETICS_OVERRIDE_START_URL_SUBSTITUTION_REGEX,
         testTimeout: toNumber(process.env.DATADOG_SYNTHETICS_OVERRIDE_TEST_TIMEOUT),
+        variables: toStringObject(process.env.DATADOG_SYNTHETICS_OVERRIDE_VARIABLES),
```
teodor2312, line 304: Can you please add a test case for the env variable. I think this one is the correct test - https://github.com/DataDog/datadog-ci/pull/1354/files#diff-fed723dc96fbca46fa0fc4add02e5a092a935c4778780668acc07e993bc916deR60

src/commands/synthetics/run-tests-command.ts
```diff
@@ -366,7 +369,9 @@ export class RunTestsCommand extends Command {
         startUrl: validatedOverrides.startUrl,
         startUrlSubstitutionRegex: validatedOverrides.startUrlSubstitutionRegex,
         testTimeout: validatedOverrides.testTimeout,
-        variables: parseVariablesFromCli(this.variableStrings, (log) => this.reporter.log(log)),
+        // TODO SYNTH-12989: Clean up deprecated `variableStrings` in favor of `variables` in `defaultTestOverrides`.
+        variables:
+          validatedOverrides.variables ?? parseVariablesFromCli(this.variableStrings, (log) => this.reporter.log(log)),
```
teodor2312, line 374: Can you also please add a test for checking the priority behaviour here.

teodor2312: Hey and as we discussed also one for checking the old behaviour still works as well

teodor2312: and also for the override priority :D 
here - https://github.com/DataDog/datadog-ci/pull/1354/files#diff-fed723dc96fbca46fa0fc4add02e5a092a935c4778780668acc07e993bc916deR375
and here - https://github.com/DataDog/datadog-ci/pull/1354/files#diff-fed723dc96fbca46fa0fc4add02e5a092a935c4778780668acc07e993bc916deR415

i'm okay if you end up creating a card for this one as it will require adding cases not just for variables but also all the other parameters we've added to env and cli 👍

rtrieu (approved):
Looks good!

src/commands/synthetics/utils/internal.ts
AntoineDona: Otherwise you get this issue: `This expression is not callable. Type '{}' has no call signatures.` and it explicitly requires a semi-colon. The yarn format moves it to the start of the next line.

src/commands/synthetics/__tests__/cli.test.ts
```diff
@@ -3,6 +3,8 @@ import {Cli} from 'clipanion/lib/advanced'
 import {createCommand} from '../../../helpers/__tests__/fixtures'
 import * as ciUtils from '../../../helpers/utils'
 
+import {DATADOG_API_KEY_REG_EXP} from '../../lambda/constants'
```
Drarig29, line 6: Let's not import from `lambda`, and move their constant to the general `src/constants.ts` file. Their team will be notified for a review automatically

src/commands/synthetics/run-tests-command.ts
```diff
@@ -143,6 +144,7 @@ export class RunTestsCommand extends Command {
   private tunnel = Option.Boolean('-t,--tunnel', {
     description: `Use the ${$3('Continuous Testing Tunnel')} to execute your test batch.`,
   })
+  // TODO SYNTH-12989: Clean up deprecated `variableStrings` in favor of `variables` in `defaultTestOverrides`.
   private variableStrings = Option.Array('-v,--variable', {description: 'Pass a variable override.'})
```
Drarig29, line 148: Can you do the same as [here](https://github.com/DataDog/datadog-ci/pull/1356/files#diff-083f09b898f286ae51bdc48ee28d591f3a567f3a3e295a19d14ab19e009f0d9eR134-R139)?

Namely:
- mark it as `@deprecated`
- change its description to a deprecation notice describing the new way

src/commands/synthetics/utils/internal.ts
```diff
@@ -90,7 +90,7 @@ export const toExecutionRule = (env: string | undefined): ExecutionRule | undefi
   return undefined
 }
 
-export const toStringObject = (env: string | undefined): {[key: string]: string} | undefined => {
+export const toStringObject = (env: string | undefined): StringObject | undefined => {
```
Drarig29, line 93: I'd call this a `StringMap` 😁

Anything in JS is an object, so "string object" is a pleonasm

src/commands/synthetics/utils/internal.ts
```diff
@@ -253,11 +255,12 @@ export const validateAndParseOverrides = (overrides: string[] | undefined): Accu
...
           break
 
-        // Convert to {[key: string]: string}
+        // Convert to StringObject
         case 'headers':
+        case 'variables':
           if (subKey) {
-            acc['headers'] = acc['headers'] ?? {}
-            acc['headers'][subKey] = value
+            const stringObject: StringObject = acc[key] = acc[key] ?? {} // acc[key] could be undefined, because of potential js Proxy
```
Drarig29, line 262: > because of potential js Proxy

Did this happen? Also, I'm not sure what you mean - and is it the only possibility of `acc[key]` to be `undefined`?

src/commands/synthetics/utils/internal.ts
teodor2312: TIL - the meaning of pleonasm :D

src/commands/synthetics/__tests__/cli.test.ts
AntoineDona: oups, it's actually not being used haha, gotta love auto import 😅

src/commands/synthetics/utils/internal.ts
AntoineDona: No but if you don't do that, ts still thinks `acc[key]` can be undefined and won't compile...

Drarig29: Which is good - I don't think you need the comment then, I was confused about it 😉

src/commands/synthetics/utils/internal.ts
```diff
@@ -90,7 +90,7 @@ export const toExecutionRule = (env: string | undefined): ExecutionRule | undefi
   return undefined
 }
 
-export const toStringObject = (env: string | undefined): {[key: string]: string} | undefined => {
+export const toStringObject = (env: string | undefined): StringMap | undefined => {
```
Drarig29, line 93:
```suggestion
export const toStringMap = (env: string | undefined): StringMap | undefined => {
```

src/commands/synthetics/utils/internal.ts
```diff
@@ -106,13 +106,16 @@ export const toStringObject = (env: string | undefined): {[key: string]: string}
...
 
-      return parsed as {[key: string]: string}
+      return parsed as StringMap
     }
   } catch (error) {
     return undefined
   }
 }
 
+// create type for {[key: string]: string}
```
Drarig29, line 116:
```suggestion
```

src/commands/synthetics/utils/internal.ts
```diff
@@ -253,11 +255,12 @@ export const validateAndParseOverrides = (overrides: string[] | undefined): Accu
           }
           break
 
-        // Convert to {[key: string]: string}
+        // Convert to StringObject
```
Drarig29, line 258:
```suggestion
        // Convert to StringMap
```

src/commands/synthetics/utils/internal.ts
```diff
@@ -253,11 +255,12 @@ export const validateAndParseOverrides = (overrides: string[] | undefined): Accu
...
         case 'headers':
+        case 'variables':
           if (subKey) {
-            acc['headers'] = acc['headers'] ?? {}
-            acc['headers'][subKey] = value
+            const stringMap: StringMap = (acc[key] = acc[key] ?? {}) // acc[key] could be undefined, because of potential js Proxy
+            stringMap[subKey] = value
```
Drarig29, line 263: The first line aligns better with the other cases, and the second line tells TS that `acc[key]` cannot truly be undefined in that case.

```suggestion
            acc[key] = acc[key] ?? {}
            ;(acc[key] as StringMap)[subKey] = value
```

src/commands/synthetics/utils/internal.ts
```diff
@@ -255,12 +254,12 @@ export const validateAndParseOverrides = (overrides: string[] | undefined): Accu
...
           break
 
-        // Convert to StringObject
+        // Convert to StringMap
         case 'headers':
         case 'variables':
           if (subKey) {
-            const stringMap: StringMap = (acc[key] = acc[key] ?? {}) // acc[key] could be undefined, because of potential js Proxy
-            stringMap[subKey] = value
+            acc[key] = acc[key] ?? {} // acc[key] could be undefined, because of potential js Proxy
```
Drarig29, line 261: Now that this line is (almost) the same as the others, I don't understand why you are talking about "potential js Proxy" only on that line. Why could a `Proxy` be used, and how?

AntoineDona: Oups forgot to remove the comment 😁

Drarig29 (approved):
LGTM

teodor2312 (approved):

PR #1356, opened 2024-06-20
[SYNTH-10422] Rename `pollingTimeout` to `batchTimeout`

Drarig29:
### What and why?

This PR introduces a new `--batchTimeout` CLI parameter and `batchTimeout` global configuration option.

The `pollingTimeout` ones are deprecated, but will still be used if `batchTimeout` isn't set. If both are set, `batchTimeout` is used.

Setting `pollingTimeout` in test overrides is also deprecated, and **has no equivalent** with `batchTimeout`.

### How?

Add deprecation warnings and backwards compatibility functions.

### Review checklist


buraizu (approved):
Thanks, doc updates LGTM

src/commands/synthetics/__tests__/cli.test.ts
```diff
@@ -122,7 +122,6 @@ describe('run-test', () => {
           followRedirects: toBoolean(overrideEnv.DATADOG_SYNTHETICS_OVERRIDE_FOLLOW_REDIRECTS),
           headers: toStringObject(overrideEnv.DATADOG_SYNTHETICS_OVERRIDE_HEADERS),
           mobileApplicationVersion: overrideEnv.DATADOG_SYNTHETICS_OVERRIDE_MOBILE_APPLICATION_VERSION,
-          pollingTimeout: DEFAULT_POLLING_TIMEOUT,
```
teodor2312, line 125: Should we add a case for `DATADOG_SYNTHETICS_BATCH_TIMEOUT` in `overrideEnv` above and here in the root level?

Drarig29: Good point, thanks 👍

src/commands/synthetics/__tests__/cli.test.ts
```diff
@@ -191,12 +191,13 @@ describe('run-test', () => {
...
-          pollingTimeout: 2,
+          pollingTimeout: 3,
           mobileApplicationVersion: '00000000-0000-0000-0000-000000000000',
           mobileApplicationVersionFilePath: './path/to/application.apk',
         },
-        // TODO SYNTH-12989: Clean up `locations` that should only be part of the testOverrides
+        // TODO SYNTH-12989: Clean up `locations` that should only be part of test overrides
         locations: [],
+        // TODO SYNTH-12989: Clean up `pollingTimeout` in favor of `batchTimeout`
         pollingTimeout: 1,
```
teodor2312, line 204: Should the value here not be 4? It is 4 in the `config-with-all-keys.json` and this test should just be picking up the config from that file, no? 🤔

Drarig29: No, because `batchTimeout` takes the precedence over anything (the deprecated root-level `pollingTimeout`, and the deprecated `pollingTimeout` test override in both `global` and `defaultTestOverrides`):

https://github.com/DataDog/datadog-ci/blob/3a716defe86f3973b695a3826389acebb7662293/src/commands/synthetics/__tests__/config-fixtures/config-with-all-keys.json#L4

src/commands/synthetics/__tests__/cli.test.ts
```diff
@@ -347,6 +354,58 @@ describe('run-test', () => {
...
+          locations: ['us-east-1'],
+          mobileApplicationVersion: '00000000-0000-0000-0000-000000000000',
+          mobileApplicationVersionFilePath: './path/to/application.apk',
+          retry: {
+            count: 5,
+            interval: 42,
+          },
+          startUrl: 'startUrl',
+          startUrlSubstitutionRegex: 'startUrlSubstitutionRegex',
+          testTimeout: 42,
```
teodor2312, line 405: You'll also need a `pollingTimeout` here 👍 ~This would also show that the nested `pollingTimeout` does not get updated like the one at the root~ (I saw you show this in the next test already)

teodor2312 (approved):
👍

src/commands/synthetics/__tests__/cli.test.ts
```diff
@@ -254,7 +258,8 @@ describe('run-test', () => {
         headers: {'Content-Type': 'application/json', Authorization: 'Bearer token'},
         locations: ['us-east-1'],
         mobileApplicationVersion: '00000000-0000-0000-0000-000000000000',
-        pollingTimeout: 42,
+        // TODO SYNTH-12989: Clean up `pollingTimeout` from `defaultTestOverrides`
+        pollingTimeout: 3,
```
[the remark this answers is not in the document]
Drarig29: should be solved now

AntoineDona (commented):
Nice!

PR #1357, opened 2024-06-27
[SYNTH-9539] Report non-final attempts of a result

Drarig29:
### What and why?

Every 5 second when the batch is fetched, if a non-final attempt is detected, it's now reported. Before this PR, we used to only report the final attempt.

**Visual:**

image

### How?

Following a recent backend change, all properties of a batch result are updated with non-final attempts (e.g. `retries`, `duration`, `result_id`), and the `status` stays `in_progress` for backwards compatibility.

The main change of this PR is changing from `emittedResultIndexes` to `emittedResultIds`, so that we can report multiple results per result index (i.e. each attempt).

Note that **if the retries are too quick** (less than the polling cycle of 5 seconds), they won't be detected and reported by datadog-ci. For that reason, we may have a single ${{\color{Green}{\textsf{  Passed\ (attempt 3 of 3)\}}}}\$ result, without the attempts 1 and 2 (which brings more information than before anyway).

### Review checklist

### DIFF
--- a/src/commands/synthetics/__tests__/fixtures.ts
+++ b/src/commands/synthetics/__tests__/fixtures.ts
@@ -38,6 +38,7 @@ import {
   TriggerConfig,
   MobileTestWithOverride,
   BaseResultInBatch,
+  ResultInBatchSkippedBySelectiveRerun,
 } from '../interfaces'
 import {AppUploadReporter} from '../reporters/mobile/app-upload'
 import {createInitialSummary} from '../utils/public'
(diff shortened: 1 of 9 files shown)


src/commands/synthetics/__tests__/fixtures.ts
```diff
@@ -504,17 +509,57 @@ export const getResults = (resultsFixtures: ResultFixtures[]): Result[] => {
...
-      test_public_id: 'pid',
-      timed_out: false,
+export const getInProgressResultInBatch = (): BaseResultInBatch => {
+  return {
+    execution_rule: ExecutionRule.BLOCKING,
+    location: mockLocation.name,
+    result_id: 'rid',
+    // eslint-disable-next-line no-null/no-null
+    retries: null,
+    status: 'in_progress',
```
teodor2312, line 519:
```suggestion
    status: 'in_progress',
```

PR #1360, opened 2024-07-01
[SYNTH-15171] Fix multiple test configs with same id bug

AntoineDona:
### What and why?

We discovered a bug where having multiple test configs with the same test id would result in only the first occurence being used multiple time. This was caused by the usage of the `find` function in `getTriggerConfigs` in `run-tests-lib.ts:202`. Check out the [ticket](https://datadoghq.atlassian.net/browse/SYNTH-15171) for more context on how the bug was discovered and how to reproduce it.

### How?

- Use `findIndex` and `splice` to remove a test from `testsFromTestConfigs` once it has been found.
- Add test coverage for this issue.

### Review checklist


teodor2312 (approved):
LGTM

PR #1361, opened 2024-07-03
[gate evaluate] Add `options.pull_request_sha` to the `datadog-ci gate evaluate` payload

juan-fernandez:
### What and why?

Whenever a github action is triggered through a `pull_request` or `pull_request_target`, the `GITHUB_SHA` env var does not point to the head of the branch but to a merge commit that GH does on the spot. 

After this change, in addition to `GITHUB_SHA` we also grab `git.commit.head_sha` by reading `GITHUB_EVENT_PATH`, which is, according to [github](https://docs.github.com/en/actions/learn-github-actions/variables):

> The path to the file on the runner that contains the full event webhook payload. For example, /github/workflow/event.json.

This new tag is also passed as part of `options.pull_request_sha` in the `gate evaluate` payload. 

### How?

Read file in `process.env.GITHUB_EVENT_PATH` and grab `event.pull_request.head.sha`

### Review checklist

### DIFF
--- a/src/helpers/utils.ts
+++ b/src/helpers/utils.ts
@@ -1,5 +1,5 @@
 import {exec} from 'child_process'
-import fs, {existsSync} from 'fs'
+import fs, {existsSync, readFileSync} from 'fs'
 import {promisify} from 'util'
 
 import type {SpanTag, SpanTags} from './interfaces'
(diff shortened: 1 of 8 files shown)


juan-fernandez:
sry, moved this to ready for review accidentally. It's still a WIP

src/helpers/utils.ts
```diff
@@ -412,3 +412,24 @@ export const execute = (cmd: string, cwd?: string): Promise<{stderr: string; std
...
+export const getGitHeadShaFromGitHubWebhookPayload = () => {
+  if (!process.env.GITHUB_EVENT_PATH) {
+    return ''
+  }
+  try {
+    const parsedContents = JSON.parse(readFileSync(process.env.GITHUB_EVENT_PATH, 'utf8')) as GitHubWebhookPayload
+
+    return parsedContents.pull_request.head.sha
+  } catch (e) {
+    return ''
```
drodriguezhdez, line 433: Can we send a log message to indicate we could not parse the payload?

drodriguezhdez (approved):

src/helpers/utils.ts
juan-fernandez: added in https://github.com/DataDog/datadog-ci/pull/1361/commits/5efc53ad6c5ab8d10e189f290870971f0e9fcbdc

anmarchenko (approved):

src/helpers/utils.ts
etnbrd, line 433: Why not be explicit and return `undefined`?

etnbrd (approved):
LGTM 👍

src/helpers/utils.ts
juan-fernandez: yeah, that makes sense. We'll need to add some changes later on so I'll do it then

PR #1363, opened 2024-07-04
Fix download link for windows binaries

TobiasBreuer:
### What and why?

Using the powershell command as given in the readme, leads to a 404 response page from github.
This is due to the fact, that the URL must not contain the *.exe* extension.

As this is somewhat of the smallest possible contribution I thought it would be okay to just send a PR for the fix instead of having a discussion on it beforehand 😉


kayayarai (commented):
This change is fine from a docs perspective

Drarig29 (approved):

PR #1367, opened 2024-07-09
[SYNTH-13898] Inform when selective re-run is rate-limited

Drarig29:
### What and why?

When the internal query that the selective re-run feature does is rate-limited, the endpoint still returns a 200, and re-runs all tests (ignoring the selective re-run).

In that case, the backend now returns a `selective_rerun_rate_limited` property. This PR adds a log to inform when that happens.

Demo with a rate-limit on staging:
image

### How?

- Update the interface and add a log.

### Review checklist

### DIFF
--- a/src/commands/synthetics/compatibility.ts
+++ b/src/commands/synthetics/compatibility.ts
@@ -50,7 +50,7 @@ export const replacePollingTimeoutWithBatchTimeout = (
 ): number | undefined => {
   // At this point, `global` should already have been moved to `defaultTestOverrides`
   const pollingTimeout = pollingTimeoutCliParam ?? config.defaultTestOverrides?.pollingTimeout ?? config.pollingTimeout
-  const isPollingTimeoutUsed = pollingTimeout !== DEFAULT_POLLING_TIMEOUT
+  const isPollingTimeoutUsed = pollingTimeout !== undefined && pollingTimeout !== DEFAULT_POLLING_TIMEOUT
 
   if (isPollingTimeoutUsed && warnDeprecatedPollingTimeout) {
     reporter.error(
--- a/src/commands/synthetics/interfaces.ts
+++ b/src/commands/synthetics/interfaces.ts
@@ -320,6 +320,7 @@ export interface LocationsMapping {
 export interface Trigger {
   batch_id: string
   locations: Location[]
+  selective_rerun_rate_limited?: boolean
 }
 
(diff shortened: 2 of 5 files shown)


teodor2312:
> internal query that the selective re-run feature does

Just out of curiosity can you tell me what this is?

Drarig29:
@teodor2312 this is a query to find the previous batches on the same commit

src/commands/synthetics/compatibility.ts
```diff
@@ -50,7 +50,7 @@ export const replacePollingTimeoutWithBatchTimeout = (
 ): number | undefined => {
   // At this point, `global` should already have been moved to `defaultTestOverrides`
   const pollingTimeout = pollingTimeoutCliParam ?? config.defaultTestOverrides?.pollingTimeout ?? config.pollingTimeout
-  const isPollingTimeoutUsed = pollingTimeout !== DEFAULT_POLLING_TIMEOUT
+  const isPollingTimeoutUsed = pollingTimeout !== undefined && pollingTimeout !== DEFAULT_POLLING_TIMEOUT
```
teodor2312, line 53: Is this change related to the selective re-run rate limiting or just something you noticed needing fixing?

src/commands/synthetics/interfaces.ts
```diff
@@ -320,6 +320,7 @@ export interface LocationsMapping {
 export interface Trigger {
   batch_id: string
   locations: Location[]
+  selective_rerun_rate_limited?: boolean
```
teodor2312, line 323: So does the backend return this information then? 🤔

teodor2312 (approved):
Some questions but LGTM

src/commands/synthetics/compatibility.ts
Drarig29: It's unrelated 😬 I noticed this when updating the unit tests on that PR - it shouldn't possible in the real world because `config.pollingTimeout` always defaults to `DEFAULT_POLLING_TIMEOUT` but we can keep this extra safety check

src/commands/synthetics/interfaces.ts
Drarig29: From the PR description:
> In that case, the backend now returns a `selective_rerun_rate_limited` property.

Yes, that's a backend PR that I did for the same Jira card

PR #1368, opened 2024-07-09
[SYNTH-12978] Clarify use of global configuration file in README

teodor2312:
### What and why?

Changing the README to make it clear that the run-tests and upload-application commands use the same global config. This was not clear before.

Also updated the example for the global config file to actually have all of the possible thing that you can configure, so people can actually see how they can use the global config

### How?

Made changes to the README and added a new example global config file

### Review checklist

### DIFF
--- a/src/commands/synthetics/README.md
+++ b/src/commands/synthetics/README.md
@@ -540,7 +540,7 @@ Additional helpful documentation, links, and articles:
 [6]: https://www.datadoghq.com/blog/datadog-github-action-synthetics-ci-visibility/
 [7]: https://docs.datadoghq.com/continuous_testing/cicd_integrations/
 [8]: https://docs.datadoghq.com/continuous_testing/explorer/
-[9]: https://github.com/DataDog/datadog-ci/blob/master/.github/workflows/e2e/global.config.json
+[9]: https://github.com/DataDog/datadog-ci/blob/master/.github/workflows/e2e/example-global-config.json
 [10]: https://docs.datadoghq.com/mobile_app_testing/
 
 <!--
--- a/src/commands/synthetics/README.md
+++ b/src/commands/synthetics/README.md
@@ -54,7 +54,7 @@ To setup the client, your Datadog API and application keys need to be configured
 
 ### Global configuration file options
 
-Create a JSON configuration file on your system. Specify the path to the file using the `--config` flag [when launching your tests](#run-tests). If you don't specify a file path, Datadog looks for a file with the default filename of `datadog-ci.json`.
+Using a global configuration file is one of the ways to configure datadog-ci. To do so, create a JSON configuration file on your system. Specify the path to the file using the `--config` flag [when launching your tests](#run-tests). If you don't specify a file path, Datadog looks for a file with the default filename of `datadog-ci.json`.
 
 See below for the list of advanced options in the global configuration file. For an example configuration file, see this [`example-global-config.json` file][9].
 
@@ -202,8 +202,8 @@ export DATADOG_SYNTHETICS_LOCATIONS="aws:us-east-1;aws:us-east-2"
 
 ### API
 
-By default, `datadog-ci` runs at the root of the working directory and looks for `{,!(node_modules)/**/}*.synthetics.json` files (every file ending with `.synthetics.json`, except for those in the `node_modules` folder) to find a Test Config file.
-The default value for the Global Configuration file is `datadog-ci.json`. If you name you global config file that you don't need to use the `--config` flag, the file will be automatically picked up.
+By default, `datadog-ci` runs at the root of the working directory and looks for `{,!(node_modules)/**/}*.synthetics.json` files (every file ending with `.synthetics.json`, except for those in the `node_modules` folder) to find a [test configuration file](#test-files).
+The default file name for the [global configuration file](#global-configuration-file-options) is `datadog-ci.json`. If you use this name for your global configuration file, you may omit the `--config` flag.
 
 For example:
 
@@ -512,7 +512,7 @@ You can also pass these options in a configuration file:
 }
 ```
 
-They can also be added the same global configuration file as used for the run-tests command.
+These options can also be added to the same global configuration file used for the run-tests command.
 
 Pass this config file to the command with the `--config` flag:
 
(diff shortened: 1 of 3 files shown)


src/commands/synthetics/README.md
```diff
@@ -202,7 +202,8 @@ export DATADOG_SYNTHETICS_LOCATIONS="aws:us-east-1;aws:us-east-2"
 
 ### API
 
-By default, `datadog-ci` runs at the root of the working directory and finds `{,!(node_modules)/**/}*.synthetics.json` files (every file ending with `.synthetics.json`, except for those in the `node_modules` folder). The tool loads `datadog-ci.json`, which can be overridden through the `--config` argument.
+By default, `datadog-ci` runs at the root of the working directory and looks for `{,!(node_modules)/**/}*.synthetics.json` files (every file ending with `.synthetics.json`, except for those in the `node_modules` folder) to find a Test Config file.
+The default value for the Global Configuration file is `datadog-ci.json`. If you name you global config file that you don't need to use the `--config` flag, the file will be automatically picked up.
```
buraizu, line 206:
```suggestion
The default value for the Global Configuration file is `datadog-ci.json`. If you use this name for your global config file, you don't need to use the `--config` flag, as the file is automatically picked up.
```

src/commands/synthetics/README.md
```diff
@@ -509,11 +512,16 @@ You can also pass these options in a configuration file:
...
+They can also be added the same global configuration file as used for the run-tests command.
+
+Pass this config file to the command with the `--config` flag:
+
 ```bash
-datadog-ci synthetics upload-application --config global.config.json
+datadog-ci synthetics upload-application --config global-config.json
 ```
 
+The default value for the Global Configuration file is `datadog-ci.json`. If you name you global config file that you don't need to use the `--config` flag, the file will be automatically picked up.
```
buraizu, line 523:
```suggestion
The default value for the Global Configuration file is `datadog-ci.json`. If you use this name for your global config file, you don't need to use the `--config` flag, as the file is automatically picked up.
```

src/commands/synthetics/README.md
```diff
@@ -509,11 +512,16 @@ You can also pass these options in a configuration file:
 }
 ```
 
-And pass it to the command with the `--config` flag:
+They can also be added the same global configuration file as used for the run-tests command.
```
buraizu, line 515:
```suggestion
These options can also be added to the same global configuration file used for the run-tests command.
```

buraizu (approved):
Thanks, approving with some minor doc updates requested

src/commands/synthetics/README.md
```diff
@@ -54,9 +54,9 @@ To setup the client, your Datadog API and application keys need to be configured
 
 ### Global configuration file options
 
-When you run your tests, use the `--config` flag on the command line to specify the path to the global configuration file.
+Create a JSON configuration file on your system. Specify the path to the file using the `--config` flag [when launching your tests](#run-tests). If you don't specify a file path, Datadog looks for a file with the default filename of `datadog-ci.json`.
```
Drarig29, line 57: I see this text is exactly the same as line 53. Could you change that bullet point `3.`  to only redirect to the `#global-configuration-file-options` section?

Also, I think the paragraph could use a little introduction, instead of starting with "Create a JSON configuration file" out of nowhere

src/commands/synthetics/README.md
```diff
@@ -202,7 +202,8 @@ export DATADOG_SYNTHETICS_LOCATIONS="aws:us-east-1;aws:us-east-2"
 
 ### API
 
-By default, `datadog-ci` runs at the root of the working directory and finds `{,!(node_modules)/**/}*.synthetics.json` files (every file ending with `.synthetics.json`, except for those in the `node_modules` folder). The tool loads `datadog-ci.json`, which can be overridden through the `--config` argument.
+By default, `datadog-ci` runs at the root of the working directory and looks for `{,!(node_modules)/**/}*.synthetics.json` files (every file ending with `.synthetics.json`, except for those in the `node_modules` folder) to find a Test Config file.
```
Drarig29, line 205: "Test Config" file is what we use internally, for short
```suggestion
By default, `datadog-ci` runs at the root of the working directory and looks for `{,!(node_modules)/**/}*.synthetics.json` files (every file ending with `.synthetics.json`, except for those in the `node_modules` folder) to find a [test configuration file](#test-files).
```

src/commands/synthetics/README.md
Drarig29: here would be my suggestion if we are really going to keep that sentence - though I think we are just explaining the concept of a default value 😅 
```suggestion
The default file name for the [global configuration file](#global-configuration-file-options) is `datadog-ci.json`. If you use this name for your global configuration file, you may omit the `--config` flag.
```

src/commands/synthetics/README.md
Drarig29: ditto

src/commands/synthetics/README.md
teodor2312: Ah yeah this is a side effect of splitting the changes into multiple PRs 😅 on my other branch i've reordered the points and slightly changed them, removing this part from point `3.` 

https://github.com/DataDog/datadog-ci/pull/1372/files#diff-5a0fda3a46d1c105f78a54a8d87902bdb9886b1739928f72abc7cefc23a2f86bR38-R62

src/commands/synthetics/README.md
```diff
@@ -202,7 +202,8 @@ export DATADOG_SYNTHETICS_LOCATIONS="aws:us-east-1;aws:us-east-2"
 
 ### API
 
-By default, `datadog-ci` runs at the root of the working directory and finds `{,!(node_modules)/**/}*.synthetics.json` files (every file ending with `.synthetics.json`, except for those in the `node_modules` folder). The tool loads `datadog-ci.json`, which can be overridden through the `--config` argument.
+By default, `datadog-ci` runs at the root of the working directory and looks for `{,!(node_modules)/**/}*.synthetics.json` files (every file ending with `.synthetics.json`, except for those in the `node_modules` folder) to find a [test configuration file](#test-files).
+The default file name for the [global configuration file](#global-configuration-file-options) is `datadog-ci.json`. If you use this name for your global configuration file, you may omit the `--config` flag.
```
Drarig29, line 206: @buraizu do you prefer "file name" or "filename"? My suggestion here is just to align with the spelling in other places of that file
```suggestion
The default filename for the [global configuration file](#global-configuration-file-options) is `datadog-ci.json`. If you use this name for your global configuration file, you may omit the `--config` flag.
```

Drarig29:
@buraizu asking for a re-review because I suggested other edits

src/commands/synthetics/README.md
buraizu: @Drarig29 I see that both styles are used about equally in the documentation currently, even within this page, so either should be fine.

src/commands/synthetics/README.md
buraizu: @Drarig29 thanks, I agree that your suggestion is better in this case (if the decision is to keep the sentence).

Drarig29 (approved):

PR #1373, opened 2024-07-11
[SYNTH-15150] Add the rest of the missing test cases

AntoineDona:
### What and why?

- #1373 (← **you are here**)
- #1365

In a previous PR, we added missing test cases for the parameters override precedence.

> When aligning the datadog-ci variables, we realised that a lot of test cases were missing to check the parameters override precedence. This PR adds a part of the missing test cases by refactoring override from config file < ENV < CLI into smaller tests ...

In this PR, we add the rest of the missing test cases, that are: `config file < test file`, `ENV < test file ` and `CLI < test file`

### How?

- Add `config file < test file` overriding test
- Add `ENV < test file ` overriding test
- Add `CLI < test file` overriding test
- Remove the now useless `parameters override precedence: global < ENV < test file` test

### Review checklist

### DIFF
--- a/src/commands/synthetics/__tests__/cli.test.ts
+++ b/src/commands/synthetics/__tests__/cli.test.ts
@@ -1020,18 +1020,53 @@ describe('run-test', () => {
         }
         expect(command['config']).toEqual(expectedCLIOverrideResult)
       })
-    })
 
-    test('parameters override precedence: global < ENV < test file', async () => {
+      const overrideTestConfig = {
+        allowInsecureCertificates: false,
+        basicAuth: {
+          password: 'password-test-file',
+          username: 'username-test-file',
+        },
+        body: 'a body from test file',
+        bodyType: 'bodyType from test file',
+        cookies: {
+          value: 'test-file1=value1;test-file2=value2;',
+          append: false,
+        },
+        defaultStepTimeout: 31,
+        deviceIds: [
+          'chrome.laptop_large_from_test_file',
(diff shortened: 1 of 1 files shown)


src/commands/synthetics/__tests__/cli.test.ts
```diff
@@ -1047,80 +1080,155 @@ describe('run-test', () => {
...
-      expect(await command.execute()).toBe(0)
-      expect(getTestsToTriggerMock).toHaveBeenNthCalledWith(
-        3,
-        ...getExpectedTestsToTriggerArguments({
-          locations: ['aws:us-east-3', 'aws:us-east-4'],
-          mobileApplicationVersionFilePath: './path/to/application_global.apk',
-          pollingTimeout: DEFAULT_POLLING_TIMEOUT,
-        })
-      )
+        process.env = testOverrideEnv
```
etnbrd, line 1145: Maybe let's not reassign `process.env` to avoid removing any existing unrelated env variable?
```suggestion
        Object.assign(process.env, testOverrideEnv)
```

etnbrd (approved):
LGTM

src/commands/synthetics/__tests__/cli.test.ts
AntoineDona: I'm not sure, we do this on multiple occurrences in other tests (on line 102, 163,  708, 949) and it seems fine... And for the scope of the test, I guess we actually want to only have those variables and remove the unrelated ones?

PR #1377, opened 2024-07-18
[SYNTH-14523] Only auto-discover with default glob when nothing given

Drarig29:
### What and why?

This PR fixes a regression introduced by #1214:

- The user has tests A, B and C defined in configuration files, with specific overrides that **are important** ⚠️.
- The user relies on the auto-discovery:
  - They are omitting the `--files` argument, thus defaulting to `{,!(node_modules)/**/}*.synthetics.json`
- Just to test something locally, they want to use `--public-id A` to "focus" on test A and only run that test.

#### Before #1214

Passing a `--public-id` would disable the auto-discover → the specific overrides **would not be loaded** → the test wouldn't behave as expected ❌ 

#### After #1214

Passing a `--public-id` now keeps the auto-discover → **config files containing the specific overrides are discovered** → test A is run and configured properly ✅ 

#### The problem

Sadly, this came with a big performance impact when running `datadog-ci` from the user's home directory with just a `--public-id` parameter, because the CLI is traversing the whole disk (from the home directory).

### How?

Go back to the previous behavior, unless `--files` is set explicitly:

- Only auto-discover with the default glob if the user doesn't give any clue what to run.
- If they give `publicIds` only, we shouldn't auto-discover for performance reasons. (old behavior)
- If they give `publicIds` and `files`, then only the given `publicIds` are run and the test config files are applied on them. (keeping the intended fix from [#1214](https://github.com/DataDog/datadog-ci/pull/1214))

### Review checklist

### DIFF
--- a/src/commands/synthetics/__tests__/run-tests-lib.test.ts
+++ b/src/commands/synthetics/__tests__/run-tests-lib.test.ts
@@ -767,12 +767,11 @@ describe('run-test', () => {
         )
       ).resolves.toEqual([
         {
-          testOverrides: {deviceIds: ['chrome.laptop_large'], startUrl, locations: ['aws:ap-northeast-1']},
+          testOverrides: {startUrl, locations},
           id: 'abc-def-ghi',
-          suite: 'Suite 1',
         },
         {
-          testOverrides: {startUrl, locations: ['aws:ap-northeast-1']},
+          testOverrides: {startUrl, locations},
           id: '123-456-789',
         },
       ])
--- a/src/commands/synthetics/run-tests-lib.ts
+++ b/src/commands/synthetics/run-tests-lib.ts
@@ -23,7 +23,7 @@ import {
 } from './interfaces'
 import {DefaultReporter, getTunnelReporter} from './reporters/default'
 import {JUnitReporter} from './reporters/junit'
-import {DEFAULT_BATCH_TIMEOUT, DEFAULT_COMMAND_CONFIG} from './run-tests-command'
+import {DEFAULT_BATCH_TIMEOUT, DEFAULT_COMMAND_CONFIG, DEFAULT_TEST_CONFIG_FILES_GLOB} from './run-tests-command'
 import {getTestConfigs, getTestsFromSearchQuery} from './test'
 import {Tunnel} from './tunnel'
 import {
@@ -250,11 +250,6 @@ export const executeWithDetails = async (
   const localConfig = {
     ...DEFAULT_COMMAND_CONFIG,
     ...runConfig,
-  }
-
-  // We don't want to have default globs in case suites are given.
-  if (!runConfig.files && suites?.length) {
-    localConfig.files = []
   }
 
   // Handle reporters for the run.
--- a/src/commands/synthetics/test.ts
+++ b/src/commands/synthetics/test.ts
@@ -7,6 +7,7 @@ import {
   warnIfDeprecatedPollingTimeoutUsed,
 } from './compatibility'
 import {MainReporter, RunTestsCommandConfig, Suite, Test, TriggerConfig} from './interfaces'
+import {DEFAULT_TEST_CONFIG_FILES_GLOB} from './run-tests-command'
 import {getSuites, normalizePublicId} from './utils/public'
 
 export const getTestConfigs = async (
@@ -14,7 +15,16 @@ export const getTestConfigs = async (
   reporter: MainReporter,
   suites: Suite[] = []
 ): Promise<TriggerConfig[]> => {
-  const suitesFromFiles = (await Promise.all(config.files.map((glob: string) => getSuites(glob, reporter))))
+  const files = [...config.files]
+
+  // Only auto-discover with the default glob if the user doesn't give any clue what to run.
+  // If they give publicIds (only), they might be in their home directory so we shouldn't auto-discover for performance reasons.
+  // If they give publicIds and files, then the test overrides from those test files will be applied to the given publicIds.
+  if (config.publicIds.length === 0 && files.length === 0 && suites.length === 0) {
+    files.push(DEFAULT_TEST_CONFIG_FILES_GLOB)
+  }
+
+  const suitesFromFiles = (await Promise.all(files.map((glob: string) => getSuites(glob, reporter))))
     .reduce((acc, val) => acc.concat(val), [])
     .filter((suite) => !!suite.content.tests)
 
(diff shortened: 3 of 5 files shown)


src/commands/synthetics/run-tests-lib.ts
```diff
@@ -23,7 +23,7 @@ import {
 } from './interfaces'
 import {DefaultReporter, getTunnelReporter} from './reporters/default'
 import {JUnitReporter} from './reporters/junit'
-import {DEFAULT_BATCH_TIMEOUT, DEFAULT_COMMAND_CONFIG} from './run-tests-command'
+import {DEFAULT_BATCH_TIMEOUT, DEFAULT_COMMAND_CONFIG, DEFAULT_TEST_CONFIG_FILES_GLOB} from './run-tests-command'
```
teodor2312, line 26: I don't see `DEFAULT_TEST_CONFIG_FILES_GLOB` being now used in this file, so i assume this is a leftover from a previous change 🤔 

```suggestion
import {DEFAULT_BATCH_TIMEOUT, DEFAULT_COMMAND_CONFIG} from './run-tests-command'
```

src/commands/synthetics/run-tests-lib.ts
```diff
@@ -252,11 +252,6 @@ export const executeWithDetails = async (
     ...runConfig,
   }
 
-  // We don't want to have default globs in case suites are given.
-  if (!runConfig.files && suites?.length) {
-    localConfig.files = []
-  }
```
teodor2312, line 258: We remove this because we don't have a default glob anymore right? 🤔

src/commands/synthetics/__tests__/run-tests-lib.test.ts
```diff
@@ -767,12 +767,11 @@ describe('run-test', () => {
         )
       ).resolves.toEqual([
         {
-          testOverrides: {deviceIds: ['chrome.laptop_large'], startUrl, locations: ['aws:ap-northeast-1']},
+          testOverrides: {startUrl, locations},
           id: 'abc-def-ghi',
-          suite: 'Suite 1',
```
teodor2312, line 772: Why are we removing the suite and the deviceIds?

src/commands/synthetics/test.ts
```diff
@@ -7,14 +7,24 @@ import {
...
   config: RunTestsCommandConfig,
   reporter: MainReporter,
   suites: Suite[] = []
 ): Promise<TriggerConfig[]> => {
-  const suitesFromFiles = (await Promise.all(config.files.map((glob: string) => getSuites(glob, reporter))))
+  const files = [...config.files]
+
+  // Only auto-discover with the default glob if the user doesn't give any clue what to run.
+  // If they give publicIds (only), they might be in their home directory so we shouldn't auto-discover for performance reasons.
+  // If they give publicIds and files, then only the given publicIds are run and the test config files are applied on them.
```
teodor2312, line 22: Can you please add info about this detail in usage in the README and also a test case so if someone in the future changes it again by mistake they can see and know that this behaviour is on purpose 🙇

teodor2312 (approved):
Thanks for fixing this :)

teodor2312 (changes requested):
Putting Request changes to undo my approval and signal that we might want to consider if we should also add not having a search query as a predicate for auto discovery as per this conversation with Hakan - https://dd.slack.com/archives/C04P926D96Z/p1721836534366369?thread_ts=1719518396.430959&cid=C04P926D96Z

src/commands/synthetics/__tests__/run-tests-lib.test.ts
Drarig29: Because we aren't reading test configuration files anymore when `files` is not passed explicitly. So in this unit test, we lose any information coming from `fakeSuites`.

src/commands/synthetics/run-tests-lib.ts
Drarig29: Yep, indeed 👍

src/commands/synthetics/test.ts
Drarig29: Added one test case, the rest is already covered

teodor2312, line 22: Small nit: since we updated the if underneath we also need to update the comments here

src/commands/synthetics/README.md
```diff
@@ -682,7 +682,14 @@ These files take precedence over global configuration files, environment variabl
...
 
-By default, `datadog-ci` runs at the root of the working directory and looks for `{,!(node_modules)/**/}*.synthetics.json` files (every file ending with `.synthetics.json`, except for those in the `node_modules` folder) to find a test configuration file. This can be manually configured with the [`files` parameter](#files).
+To determine which tests to run, one or more of those options may be passed to `datadog-ci`:
+- The [`files` option](#files)
+- The [`publicIds` option](#publicids)
+- The [`testSearchQuery` option](#testsearchquery)
+
+If none of these options is passed, `datadog-ci` auto-discovers test configuration files with the `{,!(node_modules)/**/}*.synthetics.json` glob pattern (every file ending with `.synthetics.json`, except for those in the `node_modules` folder).
+
+**Note**: The file search starts from the current working directory, so it may be slow if the command is run from a big directory, like a home folder. For better performance, use the options above to specify the tests, or refine the glob pattern with the [`files` option](#files).
```
teodor2312, line 692:
```suggestion
**Note**: The file search starts from the current working directory, so it may be slow if the command is run from a big directory, like a home folder. For better performance, use the options above to specify the tests, or refine the path using the glob pattern with the [`files` option](#files).
```
nit: just to specify that narrowing down the path is the thing that would actually fix the issue with slowness.

Also can we add a sentence for the interaction between ids from `--public-id` or `--search` and the ids in a test config

teodor2312 (approved):
Thanks for fixing this again :)

src/commands/synthetics/README.md
Drarig29: Not really, they could change it in a different way: for example, changing `**` to `*` would help. That's why I didn't want to be too specific, and using the word "refine" only means "make it more specific".

I'll try to find a better wording

Drarig29: > Also can we add a sentence for the interaction between ids from `--public-id` or `--search` and the ids in a test config

I don't think it's useful to document this as it's an internal behavior, and hopefully we made the behavior good enough so that the user doesn't have to think about it

michaelcretzman (approved):

PR #1386, opened 2024-07-24
[SYNTH-15356] Wrong case with `--override` fails silently

KarimDatadog:
### What and why?

`--override` keys are case sensitive, and a case typo leads to silent failure.

Example image, courtesy of @Drarig29 :
image

So we use Levenshtein distance to predict a corrective suggestion instead of failing silently.

⚠️ **This PR is large since we add `fast-levenshtein` npm package, which changed the yarn.lock significantly** ⚠️ 

### How?

- Use `fast-levenshtein` npm package for Levenshtein algo.
- Create an array of all possible `--override` keys to apply Levenshtein algo comparing user input to each key.
- If no valid key is found, calculate Levenshtein distance : distance > 5 ? Invalid key error with no suggestion, otherwise suggest a correction.

### Review checklist

### DIFF
--- a/src/commands/synthetics/__tests__/utils/internal.test.ts
+++ b/src/commands/synthetics/__tests__/utils/internal.test.ts
@@ -216,6 +216,28 @@ describe('utils', () => {
         const overrides = ['defaultStepTimeout=notANumber']
         expect(() => validateAndParseOverrides(overrides)).toThrow('Invalid number value: notANumber')
       })
+
+      it('should suggest correction for invalid case of valid keys', () => {
+        let overrides = ['startURL=blah']
+        expect(() => validateAndParseOverrides(overrides)).toThrow('Invalid key: startURL. Did you mean: startUrl?')
+
+        // resourceUrlSubstitutionRegexe is missing the plural "s"
+        overrides = ['resourceUrlSubstitutionRegexe=blah']
+        expect(() => validateAndParseOverrides(overrides)).toThrow(
+          'Invalid key: resourceUrlSubstitutionRegexe. Did you mean: resourceUrlSubstitutionRegexes?'
+        )
+
+        overrides = ['startUrlSubstitution=blah']
+        expect(() => validateAndParseOverrides(overrides)).toThrow(
+          'Invalid key: startUrlSubstitution. Did you mean: startUrlSubstitutionRegex?'
+        )
+
+        overrides = ['startUrlS=blah']
+        expect(() => validateAndParseOverrides(overrides)).toThrow('Invalid key: startUrlS. Did you mean: startUrl?')
+
+        overrides = ['startUrlSub=blah']
+        expect(() => validateAndParseOverrides(overrides)).toThrow('Invalid key: startUrlSub. Did you mean: startUrl?')
+      })
     })
   })
 })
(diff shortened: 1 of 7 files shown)


src/commands/synthetics/__tests__/utils/internal.test.ts
```diff
@@ -216,6 +216,28 @@ describe('utils', () => {
...
+        expect(() => validateAndParseOverrides(overrides)).toThrow(
+          'Invalid key: startUrlSubstitution. Did you mean: startUrlSubstitutionRegex?'
+        )
+
+        overrides = ['startUrlS=blah']
+        expect(() => validateAndParseOverrides(overrides)).toThrow('Invalid key: startUrlS. Did you mean: startUrl?')
+
+        overrides = ['startUrlSub=blah']
+        expect(() => validateAndParseOverrides(overrides)).toThrow('Invalid key: startUrlSub. Did you mean: startUrl?')
+      })
```
etnbrd, line 240: Could you add some tests of keys that are not extrapolated? I'm curious about the arbitrary value of 5 for the threshold.

src/commands/synthetics/utils/internal.ts
```diff
@@ -140,6 +142,26 @@ type AccumulatorBaseConfigOverride = Omit<
   cookies?: Partial<Exclude<UserConfigOverride['cookies'], string>>
 }
 type AccumulatorBaseConfigOverrideKey = keyof AccumulatorBaseConfigOverride
+const allOverrideKeys = [
```
etnbrd, line 145: Let's make so that this list is required to be updated when we add an override key.
```suggestion
const allOverrideKeys: AccumulatorBaseConfigOverrideKey[] = [
```

I tried to make it throw a compilation type error when we remove an override key as well, but it's getting too complex for this.

etnbrd (approved):
LGTM modulo a couple of small comments.
Thanks for this, I'm sure it'll be very helpful!

AntoineDona (approved):
Nice!

AntoineDona (commented):
Oh actually, I still don't get any error message when running the cli... `datadog-ci` just errors silently... I think it's due to [this part of the code](https://github.com/DataDog/datadog-ci/blob/master/src/commands/synthetics/run-tests-command.ts#L181)...

```
  public async execute() {
    try {
      await this.resolveConfig()
    } catch (error) {
      if (error instanceof CiError) {
        reportCiError(error, this.reporter)
      }

      return 1
    }
```

I guess we should change it?

src/commands/synthetics/__tests__/utils/internal.test.ts
KarimDatadog: The user could end up writing some gibberish, to which I don't think we should suggest anything. 
In my design POV, we should suggest when we are almost sure we can predict the right word.

> Could you add some tests of keys that are not extrapolated?
sure thing!

src/commands/synthetics/run-tests-command.ts
```diff
@@ -171,19 +171,19 @@ export class RunTestsCommand extends Command {
   private config: RunTestsCommandConfig = JSON.parse(JSON.stringify(DEFAULT_COMMAND_CONFIG)) // Deep copy to avoid mutation
 
   public async execute() {
+    const reporters: Reporter[] = [new DefaultReporter(this)]
```
AntoineDona, line 174: @Drarig29 Was there a reason why the reporter was instantiated after this try-catch?

Drarig29: No, this must have been a mistake... thanks for catching it 👍

PR #1387, opened 2024-07-24
[ci-visibility] Fix bug on trace command URL

rodrigo-roca:
### What and why?

The URL needs to have the API prefix and it does not.

### How?

### Review checklist

### DIFF
--- a/src/commands/trace/trace.ts
+++ b/src/commands/trace/trace.ts
@@ -152,7 +152,7 @@ export class TraceCommand extends Command {
   private getBaseIntakeUrl() {
     const site = process.env.DATADOG_SITE || process.env.DD_SITE || 'datadoghq.com'
 
-    return `https://${site}`
+    return `https://api.${site}`
   }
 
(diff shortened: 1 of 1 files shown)


src/commands/trace/trace.ts
```diff
@@ -152,7 +152,7 @@ export class TraceCommand extends Command {
   private getBaseIntakeUrl() {
     const site = process.env.DATADOG_SITE || process.env.DD_SITE || 'datadoghq.com'
 
-    return `https://${site}`
+    return `https://api.${site}`
```
romainkomorndatadog, line 155: Should this get some kind of test?

romainkomorndatadog (approved):

src/commands/trace/trace.ts
rodrigo-roca: As a general rule I see that we do not really create unit tests for testing URLs themselves. Usually the developer runs manually some end to end tests

PR #1396, opened 2024-07-31
[SYNTH-15675] Fix bug where timed out retries are not reported

AntoineDona:
### What and why?

We recently discovered a bug where timed out fast retries were not reported as fast retries. This lead to `executeTests` exiting without the reporter spinner being stopped.

This was due to the fact that `getResultsToReport` didn’t handle still non-final results after the end of a batch, resulting in an empty `resultsToReport` array. This PR patches this issue.

### How?

- Add the still non-final fast retry results to the `residualResults` 

### Review checklist

### DIFF
--- a/src/commands/synthetics/batch.ts
+++ b/src/commands/synthetics/batch.ts
@@ -113,6 +113,8 @@ const getResultsToReport = (
     .filter((r) => isResultInBatchSkippedBySelectiveRerun(r) || !incompleteResultIds.has(r.result_id))
     .concat(newlyCompleteResults)
 
+  const resultIdsToReport = new Set(getResultIds(resultsToReport))
+
   if (shouldContinuePolling) {
     return resultsToReport
   }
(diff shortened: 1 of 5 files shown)


src/commands/synthetics/batch.ts
```diff
@@ -113,14 +115,21 @@ const getResultsToReport = (
...
+      (emittedResultIds.has(r.result_id) && r.timed_out)
   )
 
   const errors: string[] = []
   for (const result of residualResults) {
     if (!result.timed_out) {
       errors.push(`The full information for result ${result.result_id} was incomplete at the end of the batch.`)
+    } else if (emittedResultIds.has(result.result_id) && result.timed_out) {
+      errors.push(`A fast retry was expected for initial result ${result.result_id}, but the batch timed out.`)
+      ;(result as ResultToReport).timedOutRetry = true
```
etnbrd, line 132: Could we find a way to avoid using `as`?

src/commands/synthetics/reporters/default.ts
```diff
@@ -275,6 +275,10 @@ const getAttemptSuffix = (passed: boolean, retries: number, test: Test) => {
     return ''
   }
 
+  if (timedOutRetry) {
+    return ` (attempt ${currentAttempt + 1} of ${maxAttempts})`
```
etnbrd, line 279: I'm not sure to understand this equation: `currentAttempt + 1`, given that `currentAttempt` is already `retries + 1`.
There could be several fast retries timing out, no? So it wouldn't be `+ 1` but `+ n`, no?

Drarig29: We don't need `+ n` because we are talking about the retry that was expected directly after the current attempt, not _all_ of the future retries.

src/commands/synthetics/batch.ts
```diff
@@ -113,14 +115,21 @@ const getResultsToReport = (
   // Residual results are either:
   //  - Still in progress (from the batch POV): they were never emitted.
   //  - Or still incomplete (from the poll results POV): report them with their incomplete data and a warning.
+  //  - Or still non-final: a fast retry was expected, show a warning.
```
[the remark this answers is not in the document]
Drarig29: - Incomplete means "we couldn't get the full data thanks to the poll results endpoint"
- Non-final means that the result is failed, and retries are expected. A final result is the final attempt.

PR #1404, opened 2024-08-01
Add SARIF validation

juli1:
### What and why?

We want to check if the SARIF file has error that will prevent us from processing the file in the backend.

### How?

Checking that the file has the appropriate key and values before sending it to our backend.

### Review checklist

### Test

Screenshot 2024-08-01 at 2 36 46 PM

### DIFF
--- a/src/commands/sarif/upload.ts
+++ b/src/commands/sarif/upload.ts
@@ -179,16 +179,14 @@ export class UploadSarifReportCommand extends Command {
 
       const validationErrorMessage = validateSarif(sarifReport)
       if (validationErrorMessage) {
-        this.context.stdout.write(renderInvalidFile(sarifReport, validationErrorMessage))
+        this.context.stdout.write(renderInvalidFile(sarifReport, [validationErrorMessage]))
 
         return false
       }
 
       const potentialErrors = checkForError(sarifReport)
       if (potentialErrors.length > 0) {
-        for (const potentialError of potentialErrors) {
-          this.context.stdout.write(renderInvalidFile(sarifReport, potentialError))
-        }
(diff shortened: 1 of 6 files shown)


src/commands/sarif/validation.ts
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  return undefined
+}
+
+/**
+ * Functions that looks for errors specific to how Datadog processes SARIF file. This way, we
+ * show the error directly to the user instead of uploading a file we cannot process in our backend.
+ *
+ * The function returns undefined if everything is good, otherwise, it returns the error to show.
```
dastrong, line 42: We should support multiple errors not just one

juli1: Why? The validation works the same way. We should keep it simple.

src/commands/sarif/upload.ts
```diff
@@ -210,6 +184,15 @@ export class UploadSarifReportCommand extends Command {
         return false
       }
 
+      const potentialErrors = checkForError(sarifReport)
+      if (potentialErrors.length > 0) {
+        for (const potentialError of potentialErrors) {
+          this.context.stdout.write(renderInvalidFile(sarifReport, potentialError))
```
dastrong, line 190: Please review the `renderInvalidFile` function.

It should be called once or else we're going to be repeating things. We should change it to accept a `[]string` instead and then loop over the errors here (https://github.com/DataDog/datadog-ci/blob/1e909b1b537a143a749b3f808dcb1b2010c21ba5/src/commands/sarif/renderer.ts#L22)

dastrong (approved):

PR #1406, opened 2024-08-06
await calls to generatePackfiles

denbeigh2000:
### What and why?
This change adds a missing `await` to the fallback case of `generatePackFilesForCommits`.
It is more-or-less the same fix as #813, but adds it to the fallback case as well as the primary.

I'm making this because I found regressions in checkout times in our CI agents after hitting the "fallback to CWD" edge case, and it seems possible that this may be due to an orphaned `git pack-objects` that is running longer than expected.

### How?

Before, we did not await the completion of this asynchronous method, and after this change, we will.

### Review checklist

I didn't do this (and neither did #813), but I believe you would unconditionally want to `await` this condition.

### DIFF
--- a/src/commands/git-metadata/gitdb.ts
+++ b/src/commands/git-metadata/gitdb.ts
@@ -320,7 +320,7 @@ const generatePackFilesForCommits = async (log: Logger, commits: string[]): Prom
     log.warn(`Failed generation of packfiles in tmpdir: ${err}`)
     log.warn(`Generating them in ${process.cwd()} instead`)
 
-    return generatePackfiles(process.cwd())
+    return await generatePackfiles(process.cwd())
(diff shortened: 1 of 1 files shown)


etnbrd (commented):
Contrary to https://github.com/DataDog/datadog-ci/pull/813, this call is in the `catch` clause. To my knowledge, adding `await` won't change anything as the rejected promise won't be caught by the `try` clause and will just bubble up anyway. Maybe I'm missing something? Could you elaborate on the expected change?

Drarig29:
Hi @denbeigh2000, I agree with @etnbrd and I don't think this change will fix your issue.

Can you send the logs you have when "hitting the fallback to CWD edge case"?

Issue #1415, opened 2024-08-08
Dependency vulnerability in @grpc/grpc-js


reggieWayflyer:
### Bug description

datadog-ci@2.41.0 has a security vulnerability within originating from `@google-cloud/logging` and `@google-cloud/run` dependency `@grpc/grpc-js`. 

result from `pnpm why @grpc/grpc-js` 

```
dependencies:
@datadog/datadog-ci 2.41.0
├─┬ @google-cloud/logging 11.0.0
│ └─┬ google-gax 4.3.2
│   └── @grpc/grpc-js 1.10.6
└─┬ @google-cloud/run 1.2.0
  └─┬ google-gax 4.3.2
    └── @grpc/grpc-js 1.10.6
```

### Describe what you expected

Resolve the issue by bumping the appropriate @google-cloud/*  dependencies.

### Steps to reproduce the issue

For any node project with @datadog/datadog-ci installed with a version of 2.40.0 ~ 2.41.0

### Command

None

Drarig29:
Hi! This will be fixed in the next patch. Thanks for reporting 🙇

PR #1418, opened 2024-08-12
[dep] Align `package.json` with `yarn.lock` for GCP packages

Drarig29:
### What and why?

Closes https://github.com/DataDog/datadog-ci/issues/1415.

This PR doesn't change the pinned versions in `yarn.lock`: it only aligns the `package.json` file to require versions greater or equal to what we defined in the `yarn.lock`.

Because we didn't update `package.json` last time, some customers are left with older versions of `@google-cloud/logging` and `@google-cloud/run`, which are depending on `@grpc/grpc-js@>=1.10.0, <1.10.9`, resulting in a vulnerability: https://github.com/DataDog/datadog-ci/security/dependabot/41.

### How?

Align `package.json` with `yarn.lock` for the 2 GCP packages that we use.

### Review checklist


yarn.lock
```diff
@@ -2108,7 +2108,7 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@google-cloud/logging@npm:^11.0.0":
+"@google-cloud/logging@npm:^11.1.0":
   version: 11.1.0
```
Drarig29, line 2112: As you can see, our pinned version doesn't change.

yarn.lock
```diff
@@ -2155,7 +2155,7 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@google-cloud/run@npm:^1.0.2":
+"@google-cloud/run@npm:^1.4.0":
   version: 1.4.0
```
Drarig29, line 2159: same here

AntoineDona (approved):

PR #1419, opened 2024-08-13
[SYNTH-15833] Use batch as source of truth for `hasResultPassed()`

Drarig29:
### What and why?

This PR fixes a bug where results without the complete information (server result from the polling endpoint) are reported as passed by default, even when the batch says the result failed.

### How?

Change the source of truth for `hasResultPassed()` to be the `ResultInBatch`, instead of the `ServerResult`.

### Review checklist

### DIFF
--- a/src/commands/synthetics/__tests__/utils/public.test.ts
+++ b/src/commands/synthetics/__tests__/utils/public.test.ts
@@ -81,6 +81,7 @@ import {
   getBatch,
   getBrowserServerResult,
   getFailedResultInBatch,
+  getIncompleteServerResult,
   getInProgressResultInBatch,
   getPassedResultInBatch,
   getResults,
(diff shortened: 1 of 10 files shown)


AntoineDona (approved):
Nice!

src/commands/synthetics/__tests__/utils/public.test.ts
```diff
@@ -1306,10 +1313,15 @@ describe('utils', () => {
       // Result 3 was available instantly
       expect(mockReporter.resultEnd).toHaveBeenNthCalledWith(2, {...result, resultId: 'rid-3'}, MOCK_BASE_URL, 'bid')
 
-      // Result 1 never became available
+      // Result 1 never became available (but the backend says it did not pass)
```
Drarig29, line 1316:
```suggestion
      // Result 1 never became available (but the batch says it did not pass)
```

PR #1423, opened 2024-08-14
[STAL-2668] Remove app key requirement

dastrong:
### What and why?

We removed the requirement for an application key to be required to upload an SBOM.

### How?

Removes the application key from the code and from the command documentation.

### Review checklist

### DIFF
--- a/src/commands/sbom/README.md
+++ b/src/commands/sbom/README.md
@@ -24,7 +24,6 @@ datadog-ci sbom upload <path/to/sbom.json>
 The following environment variables must be defined:
 
  - `DD_SITE`: the [Datadog site](https://docs.datadoghq.com/getting_started/site/#access-the-datadog-site)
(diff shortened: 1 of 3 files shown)


michaelcretzman (approved):

amaanq (approved):

duncanista (approved):

Drarig29 (approved):

Drarig29:
@dastrong is the CI failing because of a prod deployment that isn't done yet?

juli1 (changes requested):
The API requires a `ValidAPIUser` and needs an APP key. We should not remove it. It can be scoped.

juli1 (commented):
Correction - this seems to be okay according to the RFC. We should really have the backend pieces ready and have it tested before rolling it out. The CI/CD pipeline should pass before merging.

PR #1425, opened 2024-08-23
Show error message when payload is invalid

juli1:
### What and why?

1. We should have only one path passed as parameter
2. We should show an error if we fail to generate the payload

### How?

Ensure we have only one path passed as parameter
Show an error and exit if we cannot generate the payload

### DIFF
--- a/src/commands/sbom/renderer.ts
+++ b/src/commands/sbom/renderer.ts
@@ -21,8 +21,11 @@ export const renderInvalidFile = (sbomReport: string) => {
 
 export const renderInvalidPayload = (sbomReport: string) => {
   const reportPath = `[${chalk.bold.dim(sbomReport)}]`
+  let fullStr = ''
+  fullStr += chalk.red(`Cannot generate payload for file ${reportPath}.\n`)
+  fullStr += chalk.red(`Make sure you run the command inside a git repository\n`)
 
-  return chalk.red(`Cannot generate payload for file ${reportPath}.\n`)
+  return fullStr
 }
 
 export const renderMissingSpan = (errorMessage: string) => {
--- a/src/commands/sbom/renderer.ts
+++ b/src/commands/sbom/renderer.ts
@@ -65,9 +65,9 @@ export const renderFailedUpload = (sbomReport: string, error: any) => {
 
 export const renderUploading = (sbomReport: string): string => `Uploading SBOM report in ${sbomReport}\n`
 
-export const renderSuccessfulCommand = (fileCount: number, duration: number) => {
+export const renderSuccessfulCommand = (duration: number) => {
   let fullStr = ''
-  fullStr += chalk.green(`${ICONS.SUCCESS} Uploaded ${fileCount} files in ${duration} seconds.\n`)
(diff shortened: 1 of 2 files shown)


src/commands/sbom/upload.ts
```diff
@@ -93,53 +93,58 @@ export class UploadSbomCommand extends Command {
 
     const validator: Ajv = getValidator()
 
+    if (this.basePaths.length !== 1) {
```
Drarig29, line 96: I think it would make sense to change `this.basePaths` to `private basePath = Option.String()`.

Basically, [`Option.Rest()`](https://mael.dev/clipanion/docs/api/option/#optionrest) is defining **_multiple_ positional arguments**, and `required: 1` means "at least one value". You could add `arity: 1` to require "exactly one value", but `this.basePaths` would remain an array.

So I suggest you use [`Option.String()`](https://mael.dev/clipanion/docs/api/option/#optionstring-positional) _without any parameters_, which defines a single positional argument.

juli1: Thanks I did not know that. Updated the PR based on your recommendations.

src/commands/sbom/upload.ts
```diff
@@ -148,7 +143,7 @@ export class UploadSbomCommand extends Command {
     }
 
     const uploadTimeMs = (Date.now() - startTimeMs) / 1000
-    this.context.stdout.write(renderSuccessfulCommand(this.basePaths.length, uploadTimeMs))
+    this.context.stdout.write(renderSuccessfulCommand(this.basePath.length, uploadTimeMs))
```
Drarig29, line 146: `this.basePath.length` would be the length of the string 😬

Drarig29: You probably need to rewrite a little the `renderSuccessfulCommand()` function too

Drarig29: @juli1 I'm not sure why you marked this as resolved? 🤔

juli1: The commit was still pending on my laptop! Should be resolved now for real!

Drarig29: oh, sorry 😁

Drarig29 (approved):
LGTM

src/commands/sbom/renderer.ts
```diff
@@ -21,8 +21,11 @@ export const renderInvalidFile = (sbomReport: string) => {
 
 export const renderInvalidPayload = (sbomReport: string) => {
   const reportPath = `[${chalk.bold.dim(sbomReport)}]`
+  let fullStr = ''
+  fullStr += chalk.red(`Cannot generate payload for file ${reportPath}.\n`)
+  fullStr += chalk.red(`Make sure you run the command inside a git repository and the SBOM file is valid\n`)
 
-  return chalk.red(`Cannot generate payload for file ${reportPath}.\n`)
+  return fullStr
```
amaanq, line 28: I think you can just do something like 
```js
  return chalk.red(
    `Cannot generate payload for file ${reportPath}.\n` +
    `Ensure that you run the command inside a git repository, and that the SBOM file is valid.\n`
  )
```

Unfortunately GH wont let me make suggestions when there's a deletion 😑

amaanq (approved):

PR #1426, opened 2024-08-23
[ci-visibility] Add `--tags-file` and `--measures-file` options

rodrigo-roca:
### What and why?

Adds the option `--tags-file` for the `tag` command and the `--measures-file` for the `measure` command. Also improves error handling in the `tag` command (by displaying the error details and not only the status code). Those options allow to send multiple tags in bulk using a JSON file. The reason we are adding this is that adding large amounts of tags is very tedious to do through the command line only. 

### How?

The user can provide a file path to a `.json` file and we will read all the tags / measures inside of it. 
* For the `tag` command, we will convert booleans and numbers to strings, emitting a warning if it is a number since it should be reported using the `measure` command instead. 
* For the `measure` command, everything that is not a number will be ignored.

### Review checklist


hestonhoffman (approved):

src/commands/tag/README.md
```diff
@@ -19,9 +19,24 @@ datadog-ci tag --level job --tags "go.version:`go version`"
   span for the currently running job.
 - `--no-fail` (default: `false`) will prevent the tag command from failing if there are issues submitting the data.
 - `--tags` is an array of key value pairs of the shape `key:value`. This will be the tags added to the pipeline or job span.
-  The resulting dictionary will be merged with whatever is in the `DD_TAGS` environment variable. If a `key` appears both in `--tags` and `DD_TAGS`, whatever value is in `DD_TAGS` will take precedence.
+  The resulting dictionary will be merged with whatever is in the `DD_TAGS` environment variable and in the `--tags-file` argument.
+- `--tags-file` is a file path to a JSON file that contains the tags in the form:
+  ```json
+  {
+    "key1": "value1",
+    "key2": "value2"
```
juan-fernandez, line 27: should we mention that all tags here should be strings?

juan-fernandez: or that it should be a flat map?

juan-fernandez (approved):

src/commands/tag/README.md
rodrigo-roca: Good idea let me let me clarify that in the readme

PR #1434, opened 2024-09-03
check rule existence in extensions

juli1:
### What and why?

We check that the SARIF file is valid to check that all rules are defined. We currently check in the tool section and not in the extensions. When processing a SARIF file that does have rules in extensions, it fails.

### How?

Detect rules in extensions

### DIFF
--- a/src/commands/sarif/validation.ts
+++ b/src/commands/sarif/validation.ts
@@ -58,6 +58,18 @@ export const checkForError = (filePath: string): string[] => {
           }
         }
       }
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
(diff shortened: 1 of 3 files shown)


amaanq (approved):

src/commands/sarif/validation.ts
```diff
@@ -58,6 +58,18 @@ export const checkForError = (filePath: string): string[] => {
           }
         }
       }
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
```
Drarig29, line 61: That's not ideal, but the whole function does that so I'll let it pass. It could be better typed in the future 😁

Drarig29 (approved):

PR #1443, opened 2024-09-10
Print warning if traces merging is skipped for Lambda in Step Function

lym953:
### Background
The context injection for Lambda in Step Function requires the Lambda function requires the Lambda definition to be like
```
    "Lambda Invoke": {
      ...
      "Parameters": {
        ...
        "Payload.$": "$"
      },
```

If the user has a custom `Payload.$` field, context injection will be skipped, but no error/warning message will be printed.

### What

Print a warning message if context injection is skipped if:
1. `Parameters` field doesn't exist for the Lambda function, or
2. there's a custom `Payload.$` field

### Why

To make it transparent to users what won't work.

### How?

Print the messages

#### Missing Parameters field
I wasn't able to test this because Parameters field is required when editing the State Machine from AWS Management Console. Should we remove this if-check?
Screenshot 2024-09-10 at 5 43 44 PM

Update: We are going to keep this if-check as @agocs suggested:
> This if statement prevents datadog-ci from crashing if parameters is missing for whatever reason.

#### Custom Payload field
Message printed as expected:
image

#### Happy case
`datadog-ci stepfunctions instrument` finishes with no error

### Review checklist

### DIFF
--- a/src/commands/stepfunctions/helpers.ts
+++ b/src/commands/stepfunctions/helpers.ts
@@ -125,20 +125,35 @@ export const addTraceContextToStepFunctionParameters = ({Parameters}: StepType):
   }
 }
 
-export const shouldUpdateStepForTracesMerging = (step: StepType): boolean => {
+export const shouldUpdateStepForTracesMerging = (step: StepType, context: BaseContext, stepName: string): boolean => {
   // is default lambda api
   if (step.Resource === 'arn:aws:states:::lambda:invoke') {
     if (!step.Parameters) {
+      context.stdout
+        .write(`[Warn] Step ${stepName} does not have a Parameters field. Step Functions Context Object injection \
+skipped. Your Step Functions trace will not be merged with downstream Lambda traces. To manually merge these traces, \
+check out https://docs.datadoghq.com/serverless/step_functions/troubleshooting/\n`)
+
       return false
     }
+
     // payload field not set
     if (!step.Parameters.hasOwnProperty('Payload.$')) {
       return true
     }
+
     // default payload
     if (step.Parameters['Payload.$'] === '$') {
       return true
     }
+
+    // custom payload
+    context.stdout
+      .write(`[Warn] Step ${stepName} has a custom Payload field. Step Functions Context Object injection skipped. \
+Your Step Functions trace will not be merged with downstream Lambda traces. To manually merge these traces, \
+check out https://docs.datadoghq.com/serverless/step_functions/troubleshooting/\n`)
+
+    return false
   }
 
   return false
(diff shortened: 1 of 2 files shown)


src/commands/stepfunctions/helpers.ts
```diff
@@ -125,20 +125,29 @@ export const addTraceContextToStepFunctionParameters = ({Parameters}: StepType):
   }
 }
 
-export const shouldUpdateStepForTracesMerging = (step: StepType): boolean => {
+export const shouldUpdateStepForTracesMerging = (step: StepType, context: BaseContext, stepName: string): boolean => {
   // is default lambda api
   if (step.Resource === 'arn:aws:states:::lambda:invoke') {
     if (!step.Parameters) {
```
agocs, line 131: This if statement prevents datadog-ci from crashing if parameters is missing for whatever reason. That happens before the validation you screenshotted. I think we should handle invalid configuration gracefully. Warning at this step is the right move.

agocs (commented):
There are a couple other cases we should consider, I think:

- The customer has a `Payload` object (not `Payload.$`). I'm not 100% sure what the `.$` does in Cloudformation templates or the Stepfunction DSL, but the outcome seems to be that it turned into `"Payload"` when the step function is executed.

- If the customer has `Payload.$` or `Payload`, should we merge our payload with theirs? Might be a broader discussion.

agocs:
Oh, here's the answer. Giving an object a name that ends with `.$` means "evaluate the value of this object, rename the object to strip off the ending `.$`, and set the value of this object to be the evaluated value`.

https://states-language.net/#appendix-b:~:text=%7B%22output%22%3A%20%22aaff4a450a104cd177d28d18d7485e8cae074b7%22%20%7D-,States.JsonMerge,-Use%20the%C2%A0

src/commands/stepfunctions/helpers.ts
lym953: Make sense. Let's keep it then.

lym953:
> There are a couple other cases we should consider, I think:
> 
> * The customer has a `Payload` object (not `Payload.$`). I'm not 100% sure what the `.$` does in Cloudformation templates or the Stepfunction DSL, but the outcome seems to be that it turned into `"Payload"` when the step function is executed.
> * If the customer has `Payload.$` or `Payload`, should we merge our payload with theirs? Might be a broader discussion.

Good points! I'll do them in separate PRs to keep this PR minimal.

> * If the customer has `Payload.$` or `Payload`, should we merge our payload with theirs? Might be a broader discussion.

I think yes.
1. The customer can indeed have custom `Payload`,  e.g. https://github.com/DataDog/serverless-plugin-datadog/issues/517
2. If we want to support Datadog instrumentation for this case, we will have to merge the payload.

kimi-p (approved):

lym953:
I found a bug: if traces merging is already set up and instrumentation command is run again, the warning will also be printed, which is wrong. I will fix this in a separate PR.

PR #1447, opened 2024-09-12
Print warning if traces merging from xfn to xfn is skipped

lym953:
### Why?
Let's say Step Function A executes Step Function B. If B:
1. doesn't have a `Parameters` field, or
2. its `Parameters.Input` field is not a JSON object
3. its `Parameters.Input` field has a custom `CONTEXT.$` field

traces merging will be skipped, but no warning message will be printed.

### What
Print a warning message in these cases to let users know what won't work.

#### 1. Missing Parameters field

I wasn't able to test this because `Parameters` field is required when editing the State Machine from AWS Management Console.

Screenshot 2024-09-12 at 12 51 26 PM

##### Steps
1. Change the Step Function execution step to be like:
```
    "Step Functions StartExecution": {
      "Type": "Task",
      "Resource": "arn:aws:states:::states:startExecution",
      "Parameters": {
        "StateMachineArn": "...",
        "Input": "Input!"
      },
      "End": true
    }
```
2. Run `datadog-ci stepfunctions instrument` command

##### Result
A warning message is printed as expected:
image

#### 3. custom `CONTEXT.$`
image

#### 4. Happy case

`datadog-ci stepfunctions instrument` finishes with no error

### Review checklist

### DIFF
--- a/src/commands/stepfunctions/__tests__/helpers.test.ts
+++ b/src/commands/stepfunctions/__tests__/helpers.test.ts
@@ -1,4 +1,5 @@
 import {DescribeStateMachineCommandOutput, LogLevel} from '@aws-sdk/client-sfn'
+import {BaseContext} from 'clipanion'
 
 import {
   buildArn,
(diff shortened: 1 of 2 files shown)


kimi-p (approved):
lgtm

lym953:
I found a problem: if traces merging is already set up and instrumentation command is run again, the warning will also be printed, which is wrong. I will fix this in a separate PR.

PR #1451, opened 2024-09-12
Skip xfn->xfn context injection if custom Payload is used without $

lym953:
### Problem
If a Lambda function in a State Machine has a custom `Payload` defined using `"Payload": ...`, e.g.,
```
    "Step Functions StartExecution": {
      "Type": "Task",
      "Resource": "arn:aws:states:::states:startExecution",
      "Parameters": {
        "StateMachineArn": "xxx",
        "Input": {
          "CONTEXT": "Context!"
        }
      },
      "End": true
    }
```
then context injection for the Step Function will fail with the error:
> [Error] Invalid State Machine Definition: 'SCHEMA_VALIDATION_FAILED: Cannot have both the field 'CONTEXT.$' and the field 'CONTEXT' at the same time at /States/Step Functions StartExecution/Parameters'. Failed to inject context into lambda functions' payload of arn:aws:states:sa-east-1:425362996713:stateMachine:YimingTestExpressStateMachine2 

This is because our instrumentation code will add a field `"CONTEXT.$": ...`, which duplicates with `"CONTEXT": ...`. Right now we only check for custom `"CONTEXT.$": ...` (with `$`, i.e. using JSONPath) but don't check for custom `"CONTEXT": ...`.

### What

Also skip context injection when a custom `"CONTEXT": ...` field is used (without `$`)

#### Automated testing
Pass the added test, which would fail without the changes on `helpers.ts`

#### Manual testing

Steps:
1. change the `CONTEXT` field of a Lambda function to what's shown in "Problem" section
2. run `datadog-ci stepfunctions instrument`

Result: a warning is printed as expected
image

### Notes

1. Thanks @agocs for bringing this up in https://github.com/DataDog/datadog-ci/pull/1443#pullrequestreview-2293937693
2. The behavior will be changed soon: we will soon allow custom `CONTEXT` field as long as it doesn't contain `Execution` or `State` field inside. But I want to first merge this PR to make the existing behavior more clear, so the changes in future PRs will be more clear so the PRs will be easier to review.

### Review checklist

### DIFF
--- a/src/commands/stepfunctions/helpers.ts
+++ b/src/commands/stepfunctions/helpers.ts
@@ -138,7 +138,7 @@ check out https://docs.datadoghq.com/serverless/step_functions/troubleshooting/\
     }
 
     // payload field not set
-    if (!step.Parameters.hasOwnProperty('Payload.$')) {
+    if (!step.Parameters.hasOwnProperty('Payload.$') && !step.Parameters.hasOwnProperty('Payload')) {
       return true
     }
 
@@ -228,6 +228,7 @@ export type StepType = {
 
 export type ParametersType = {
   'Payload.$'?: string
+  Payload?: string
   FunctionName?: string
   StateMachineArn?: string
   TableName?: string
--- a/src/commands/stepfunctions/helpers.ts
+++ b/src/commands/stepfunctions/helpers.ts
@@ -196,7 +196,7 @@ merge these traces, check out https://docs.datadoghq.com/serverless/step_functio
       return false
     }
 
-    if (!step.Parameters.Input['CONTEXT.$']) {
+    if (!step.Parameters.Input['CONTEXT.$'] && !step.Parameters.Input['CONTEXT']) {
       return true
     }
 
@@ -234,6 +234,7 @@ export type ParametersType = {
   TableName?: string
   Input?: {
     'CONTEXT.$'?: string
+    CONTEXT?: string
   }
 }
 
--- a/src/commands/stepfunctions/__tests__/helpers.test.ts
+++ b/src/commands/stepfunctions/__tests__/helpers.test.ts
@@ -68,6 +68,32 @@ describe('stepfunctions command helpers tests', () => {
       expect(shouldUpdateStepForTracesMerging(step, context, 'Lambda Invoke')).toBeTruthy()
     })
 
+    test('custom payload field not using JsonPath expression', () => {
+      const step: StepType = {
+        Type: 'Task',
+        Resource: 'arn:aws:states:::lambda:invoke',
+        Parameters: {
+          FunctionName: 'arn:aws:lambda:sa-east-1:425362991234:function:unit-test-lambda-function',
+          Payload: '{"action": "service/delete_customer"}',
+        },
+        End: true,
+      }
+      expect(shouldUpdateStepForTracesMerging(step, context, 'Lambda Invoke')).toBeFalsy()
+    })
+
+    test('custom payload field using JsonPath expression', () => {
+      const step: StepType = {
+        Type: 'Task',
+        Resource: 'arn:aws:states:::lambda:invoke',
+        Parameters: {
+          FunctionName: 'arn:aws:lambda:sa-east-1:425362991234:function:unit-test-lambda-function',
+          'Payload.$': '{"customer.$": "$.customer"}',
+        },
+        End: true,
+      }
+      expect(shouldUpdateStepForTracesMerging(step, context, 'Lambda Invoke')).toBeFalsy()
+    })
+
     test('none-lambda step should not be updated', () => {
       const step: StepType = {
         Type: 'Task',
--- a/src/commands/stepfunctions/__tests__/helpers.test.ts
+++ b/src/commands/stepfunctions/__tests__/helpers.test.ts
@@ -242,13 +242,28 @@ describe('stepfunctions command helpers tests', () => {
       ).toBeTruthy()
     })
 
-    test('is false when Input is an object that contains a CONTEXT key', () => {
+    test('is false when Input is an object that contains a CONTEXT key using JSONPath expression', () => {
       const step: StepType = {
         Type: 'Task',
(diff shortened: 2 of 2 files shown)


src/commands/stepfunctions/helpers.ts
```diff
@@ -228,11 +228,13 @@ export type StepType = {
 
 export type ParametersType = {
   'Payload.$'?: string
+  Payload?: string
   FunctionName?: string
   StateMachineArn?: string
   TableName?: string
   Input?: {
     'CONTEXT.$'?: string
+    CONTEXT?: string
```
avedmala, line 237: I was wondering if it's possible and if there's a plan to rename to "Payload" for naming consistency between Lambda and SFN cases

Just curious if it can be added down the line

avedmala (approved):
LGTM 🚢 

left a question that's non blocking

src/commands/stepfunctions/helpers.ts
lym953: I think a better way is to encapsulate only those fields we need into a field, maybe called `DDTracingContext`. Details:

Right now we are passing all the 11 fields from `$$`:
```
{
  "Execution": {
    "Id": "arn:aws:states:sa-east-1:425362996713:express:YimingTestExpressStateMachine2:a5f1a67a-dd92-401e-ae32-d270d68a1741:f2450599-6c08-4717-a91c-e0b41fd340a0",
    "Input": {},
    "Name": "a5f1a67a-dd92-401e-ae32-d270d68a1741",
    "RoleArn": "arn:aws:iam::425362996713:role/service-role/StepFunctions-YimingTestExpressStateMachine2-role-gydji1d2u",
    "StartTime": "2024-09-13T16:43:32.537Z",
    "RedriveCount": 0
  },
  "StateMachine": {
    "Id": "arn:aws:states:sa-east-1:425362996713:stateMachine:YimingTestExpressStateMachine2",
    "Name": "YimingTestExpressStateMachine2"
  },
  "State": {
    "Name": "Lambda Invoke",
    "EnteredTime": "2024-09-13T16:43:33.543Z",
    "RetryCount": 0
  }
}
```
and the downstream is probably using only a small subset of them. Passing too many field will impact user experience when they check the log. For example, their custom fields can be hard to find from the long JSON string.

I'm thinking of creating:
```
"DDTracingContext": {
  "ExecutionId": "arn:aws:states:sa-east-1:425362996713:express:YimingTestExpressStateMachine2:a5f1a67a-dd92-401e-ae32-d270d68a1741:f2450599-6c08-4717-a91c-e0b41fd340a0",
  "StateName": "Lambda Invoke",
  "StateEnteredTime": "2024-09-13T16:43:33.543Z"
}
```
(assuming only these three fields are what we need. We can add more if needed.)
and add it to
- `Payload` of Lambda
- `Parameters.Input` for Step Function

Does this sound good to you?

avedmala: I really like this idea, way cleaner!

avedmala: Just realized another point, I think this change will also bring some performance wins!

Our current implementation merges the entire Execution which means the Input also gets copied and **duplicated**. We currently truncate the actual Input/Output fields to 10K but this can lead to a case where a large Input can live under the Execution at its full length 😨

kimi-p: Good suggestion! I think we can use `DDTraceContext` as other repos do (not `tracing`). 
https://github.com/search?q=org%3ADataDog+ddtracecontext&type=code

I think there are a couple of things to consider:

1. We'll need to update ALL tracer layers for this change. Another change to logs-backend should be straightforward.
2. If AWS put more information into the context object and we want to utilize them (e.g. `Execution.RedriveCount` and `Execution.RedriveTime`), we'll need to update the schema in all tracers & logs-backend and ask customers to update their SF definition again. Although new information added probably only happen every couple of years, but it still seems a lot to ask for our customers.
3. When we inject the whole SF context object into payload, the customers may already expect these fields to be in their lambda for a year now. So, this might be a breaking change for few customers. But this is less of a concern IMO.

### SF Context Object
https://docs.aws.amazon.com/step-functions/latest/dg/input-output-contextobject.html#contextobject-format
```
{
    "Execution": {
        "Id": "String",
        "Input": {},
        "Name": "String",
        "RoleArn": "String",
        "StartTime": "Format: ISO 8601",
        "RedriveCount": Number,
        "RedriveTime": "Format: ISO 8601"
    },
    "State": {
        "EnteredTime": "Format: ISO 8601",
        "Name": "String",
        "RetryCount": Number
    },
    "StateMachine": {
        "Id": "String",
        "Name": "String"
    },
    "Task": {
        "Token": "String"
    }
}
```

lym953: @kimi-p Thanks for the comment!
> I think we can use `DDTraceContext` as other repos do (not `tracing`).
https://github.com/search?q=org%3ADataDog+ddtracecontext&type=code

Is that a different thing? For example, in your link, the `DDTraceContext` in `datadog-lambda-java` contains `traceID`, `parentID` and `samplingPriority`, not `ExecutionId`, `StateName` and `StateEnteredTime` discussed here. If that's a different thing, I think we need to pick a name that's very different from `DDTraceContext` to avoid confusion.

> If AWS put more information into the context object and we want to utilize them (e.g. `Execution.RedriveCount` and `Execution.RedriveTime`), we'll need to update the schema in all tracers & logs-backend and ask customers to update their SF definition again. Although new information added probably only happen every couple of years, it still seems a lot to ask for our customers.

So here's the tradeoff:
|  | Maintainability (for us) | Overhead (for users) | Performance | Debugging experience (for users) |
| - | - | - | - | - |
| Current approach | Good.  | Good (low). No trouble to update the version | Bad. Passes unnecessary fields | Bad. Too many fields. |
| Proposed approach | Bad. need to make tracer keep backward compatibility for a long time until all clients switch to the new version | Bad (high). Need to update the version when asked by us. | Good. Passes less data | Good. Fewer fields. |

Which option would you go for? Is there a dimension that's particularly important?

PR #1453, opened 2024-09-16
Propagating CycloneDx information

marcwieserdev:
### What and why?

The goal of this PR is to proagate new information coming from CycloneDX sboms, such as libraries package manager, is a library direct or not, files and dependencies between components.

### How?

This is done by updating the CycloneDX SBOM to SCARequest payload. Please note it will be more convenient to review it commit by commit as they have been split by propagation type + 1 refacto at first to prepare file propagation

### Review checklist


juli1 (approved):
We should probably make sure we align on https://github.com/DataDog/dd-source/pull/129608 before landing.

PR #1455, opened 2024-09-16
Refactor step function code by flipping if-condition

lym953:
### What
Refactors some code by changing:
```
if (<condition>) {
  // do a lot of things
}
return false;
```
to
```
if (!<condition>) {
  return false;
}
// do a lot of things
```

## Why
Make code more readable in that:
- the "if" block is shorter so the reader won't need to take time to find where the `}` is
- it's more clear that we don't cover the case when `<condition>` is false

## Testing
This is a simple change, so just `npm test`

### Review checklist


avedmala (approved):
LGTM 🚢 

definitely makes it a lot more readable!

PR #1456, opened 2024-09-16
fix: print a better message if Step Function context injection is already set up

lym953:
### Problem
https://github.com/DataDog/datadog-ci/pull/1443 and https://github.com/DataDog/datadog-ci/pull/1447 introduce a bug that when context injection is already set up for a Lambda or a Step Function in a Step Function, a warning will be printed, saying context injection won't work, which is wrong:
image
image

### What
If context injection is already set up, i.e.
- a Lambda has `"Payload.$": "States.JsonMerge($$, $, false)"`
- a Step Function has `"CONTEXT.$": "States.JsonMerge($$, $, false)"`

then print a better error message:
> Context injection is already set up. Skipping context injection.

### Testing
Run `datadog-ci stepfunctions instrument` command twice. The second time, these messages were printed:
image

### Review checklist

### DIFF
--- a/src/commands/stepfunctions/helpers.ts
+++ b/src/commands/stepfunctions/helpers.ts
@@ -150,6 +150,13 @@ check out https://docs.datadoghq.com/serverless/step_functions/troubleshooting/\
     return true
   }
 
+  // context injection is already set up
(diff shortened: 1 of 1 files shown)


avedmala (approved):
LGTM 🚢 

much better user experience!

PR #1457, opened 2024-09-16
Refactor xfn Lambda injection code my merging two functions

lym953:
### What
Currently the `injectContextForLambdaFunctions()` function roughly looks like this:
```
const injectContextForLambdaFunctions = (...): boolean => {
  if (shouldUpdateStepForTracesMerging()) {
    addTraceContextToLambdaParameters()
    return true
  }
  return false
}
```
It calls `shouldUpdateStepForTracesMerging()`.

This PR merges the two functions by moving the code of `shouldUpdateStepForTracesMerging()` into `injectContextForLambdaFunctions()`.

# Why?
We will soon have different instrumentation strategies depending on the existing state of Lambda definition, so `shouldUpdateStepForTracesMerging()` which has a boolean return value will no longer work.

Design doc: [Fixing Step Function Instrumentation](https://docs.google.com/document/d/18YpNVN6reCjA-dq6U1Tfs2MyLxcVnjO_N8Uy9Gm_BGM/edit#bookmark=id.pczh1xpu3h9y)

### Testing

Passed the touched automated tests

### Review checklist

### DIFF
--- a/src/commands/stepfunctions/helpers.ts
+++ b/src/commands/stepfunctions/helpers.ts
@@ -110,10 +110,8 @@ export const injectContextIntoTasks = async (
   }
 }
 
-export const addTraceContextToLambdaParameters = ({Parameters}: StepType): void => {
-  if (Parameters) {
-    Parameters[`Payload.$`] = 'States.JsonMerge($$, $, false)'
-  }
+export const addTraceContextToLambdaParameters = (Parameters: ParametersType): void => {
+  Parameters[`Payload.$`] = 'States.JsonMerge($$, $, false)'
 }
 
 export const addTraceContextToStepFunctionParameters = ({Parameters}: StepType): void => {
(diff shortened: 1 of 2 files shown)


src/commands/stepfunctions/helpers.ts
```diff
@@ -252,19 +209,58 @@ export type ParametersType = {
...
+      .write(`[Warn] Step ${stepName} does not have a Parameters field. Step Functions Context Object injection \
+skipped. Your Step Functions trace will not be merged with downstream Lambda traces. To manually merge these traces, \
+check out https://docs.datadoghq.com/serverless/step_functions/troubleshooting/\n`)
+
+    return false
+  }
+
+  // payload field not set
+  if (!step.Parameters.hasOwnProperty('Payload.$') && !step.Parameters.hasOwnProperty('Payload')) {
+    addTraceContextToLambdaParameters(step.Parameters)
```
lym953, line 239: Adding `addTraceContextToLambdaParameters()` before every `return true`

PR #1460, opened 2024-09-16
Inject context for xfn->Lambda when no Payload exists (case 1)

lym953:
### What and why?

In Step Machine definition, if a Lambda function doesn't have `Payload` field, right now we are injecting context using
```
"Payload.$": "States.JsonMerge($$, $, false)"
```

This PR changes it to:
```
"Payload.$": "$$['Execution', 'State']"
```
which passes less information.

This is case #1 in this design doc: [Fixing Step Function Instrumentation](https://docs.google.com/document/d/18YpNVN6reCjA-dq6U1Tfs2MyLxcVnjO_N8Uy9Gm_BGM/edit)

#### Automated testing
Pass the touched tests

##### Steps
1. Make the Lambda Execution step have no `Payload` field
2. Build the package: `yarn prepack`
3. Instrument the Step Function using `./dist/cli.js stepfunctions instrument`
4. Execute the State Machine

##### Result
1. `"Payload.$": "$$['Execution', 'State']"` was added to Lambda definition
2. TODO: question to reviewers: What product should we check to ensure context injection works this way?

### Review checklist

### DIFF
--- a/src/commands/stepfunctions/__tests__/helpers.test.ts
+++ b/src/commands/stepfunctions/__tests__/helpers.test.ts
@@ -53,7 +53,7 @@ describe('stepfunctions command helpers tests', () => {
         End: true,
       }
       expect(injectContextForLambdaFunctions(step, context, 'Lambda Invoke')).toBeTruthy()
-      expect(step.Parameters?.['Payload.$']).toEqual('States.JsonMerge($$, $, false)')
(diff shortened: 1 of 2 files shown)


avedmala (approved):
<strike>
LGTM 🚢
</strike>

avedmala (changes requested):
I think we should hold off on removing `StateMachine` from the schema for now

avedmala (approved):
thanks for updating!

PR #1462, opened 2024-09-18
Handle sfn->Lambda context injection when Payload is used without $ (case 2)

lym953:
### What and why?

This PR handles Step Function context injection into Lambda when `Payload` (not `Payload.$`) is used in the Lambda execution step.

This is case # 2 in this design doc: [Fixing Step Function Instrumentation](https://docs.google.com/document/d/18YpNVN6reCjA-dq6U1Tfs2MyLxcVnjO_N8Uy9Gm_BGM/edit)

Sub-cases:

#### Case 2.1: Already instrumented
```
    "Lambda Invoke": {
      "Parameters": {
        "Payload": {
          "Execution.$": "$$.Execution",
          "State.$": "$$.State"
        },
      },
    },
```
Then just print a message.

#### Case 2.2: Already has Execution or State field
If the payload already has any of `Execution.$`, `Execution`, `State.$` or `State`, e.g.
```
    "Lambda Invoke": {
      "Parameters": {
        "Payload": {
          "State": {
            "Name": 'Lambda Invoke'
          }
        },
      },
    },
```
and it's not case 2.1, then print a warning and skip context injection

#### Case 2.3: no Execution or State field
i.e. it's not Case 2.1 or 2.2, e.g.
```
    "Lambda Invoke": {
      "Parameters": {
        "Payload": {
          "CustomerId": 42
        },
      },
    },
```
then just add the two fields:
```
    "Lambda Invoke": {
      "Parameters": {
        "Payload": {
          "CustomerId": 42,
          "Execution.$": "$$.Execution",
          "State.$": "$$.State"
        },
      },
    },
```
### Testing

Passed the added tests

### Review checklist

### DIFF
--- a/src/commands/stepfunctions/helpers.ts
+++ b/src/commands/stepfunctions/helpers.ts
@@ -193,9 +193,18 @@ export type StepType = {
   End?: boolean
 }
 
+export type PayloadObject = {
+  'Execution.$'?: any
+  Execution?: any
+  'State.$'?: any
+  State?: any
+  'StateMachine.$'?: any
+  StateMachine?: any
+}
+
 export type ParametersType = {
   'Payload.$'?: string
-  Payload?: string
+  Payload?: string | PayloadObject
   FunctionName?: string
   StateMachineArn?: string
   TableName?: string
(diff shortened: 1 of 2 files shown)


src/commands/stepfunctions/helpers.ts
```diff
@@ -237,13 +246,43 @@ check out https://docs.datadoghq.com/serverless/step_functions/troubleshooting/\
...
+      context.stdout
+        .write(`[Warn] Step ${stepName}'s Payload field is not a JSON object. Step Functions Context Object \
 injection skipped. Your Step Functions trace will not be merged with downstream Lambda traces. To manually \
 merge these traces, check out https://docs.datadoghq.com/serverless/step_functions/troubleshooting/\n`)
 
-    return false
+      return false
+    } else {
+      const payload = step.Parameters.Payload
+      if (payload['Execution.$'] === '$$.Execution' && payload['State.$'] === '$$.State') {
```
avedmala, line 260: Can we also check for `StateMachine` in this statement as the current layers require it to be set

avedmala (approved):
One minor change but LGTM otherwise, thank you!!

src/commands/stepfunctions/helpers.ts
lym953: Good catch!

PR #1463, opened 2024-09-18
Update documentation for xfn->Lambda context injection

lym953:
### What and why?

Updates documentation to clarify:
- different cases and their expected behavior
- which case each code section is addressing
- which case each test is addressing

### Review checklist


avedmala (approved):
nit: we don't mention `StateMachine` in some areas like Case 2.1 and 2.3

LGTM otherwise, thank you!

PR #1467, opened 2024-09-23
Refactor sfn->sfn context injection code by merging two functions (2)

lym953:
### What
Merge `addTraceContextToStepFunctionParameters()` into its callsites, and removing the extra `if` checks.

### Why
Depending on the original state `Input`, we will have different instrumentation strategies, so a single `addTraceContextToStepFunctionParameters()` will no longer work.

### Testing
Run `npm test` to pass existing tests

### Review checklist


avedmala (approved):
🚢

PR #1468, opened 2024-09-23
Support sfn->sfn context injection case 3

lym953:
### What and why?
Skip the context injection in Step Function execution step if the `CONTEXT` field exists and it's not an object.

This is case # 2 in this design doc: [Fixing Step Function Instrumentation](https://docs.google.com/document/d/18YpNVN6reCjA-dq6U1Tfs2MyLxcVnjO_N8Uy9Gm_BGM/edit)

### Review checklist


avedmala (approved):
lgtm 🚢 

also just to clarify this is addressing case 3 right?

PR #1471, opened 2024-09-26
Revamp SFN->SFN context injection

lym953:
### What and why?
Revamps Step Function -> Step Function context injection:
1. Change `States.JsonMerge($$, $, false)` to `$$['Execution', 'State', 'StateMachine']`, so the unnecessary `Task` won't be passed
2. Revamp the case numbers. Currently the case numbers (1, 2.1, 2.2, 2.3, 3, 4.1, 4.2, 4.3) are in sync with the case numbers for SFN->Lambda context injection. However the cases for SFN->SFN is simpler because we are using a custom field name `CONTEXT`, which users are unlikely to use, so we can just not handle the rare case the `CONTEXT` field is set.
3. Add test cases
4. Add doc

Design doc: [Fixing Step Function Instrumentation](https://docs.google.com/document/d/18YpNVN6reCjA-dq6U1Tfs2MyLxcVnjO_N8Uy9Gm_BGM/edit#bookmark=id.15flqi5dkv9i)

### Testing
Passed existing and new tests.

### Review checklist

### DIFF
--- a/src/commands/stepfunctions/__tests__/helpers.test.ts
+++ b/src/commands/stepfunctions/__tests__/helpers.test.ts
@@ -303,7 +303,7 @@ describe('stepfunctions command helpers tests', () => {
 
       const changed = injectContextForStepFunctions(step, context, 'Step Functions StartExecution')
       expect(changed).toBeTruthy()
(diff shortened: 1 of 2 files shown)


avedmala (approved):
lgtm 🚢

PR #1473, opened 2024-10-02
Read pull_request and pull_request_target event info from GHA

juan-fernandez:
### What and why?

GitHub does not run workflows in the head of the branch about to be merged whenever the workflow is triggered through `pull_request` or `pull_request_target` events. It runs them in a merge commit. This is the one that we add to `git.commit.sha`. This is sometimes confusing because this commit does not exist in the history of the branch. What we did:

* Notice when this happens
* When it does, add a new set of tags:
  * `git.commit.head_sha`, which is the head SHA of the feature branch (branch to be merged).
  * `git.pull_request.base_branch_sha`, which is the head SHA of the base branch (branch you're merging into)
  * `git.pull_request.base_branch`, which is the name of the base branch 

### How?

Detect if we're in a `pull_request` or `pull_request_target` event by checking `GITHUB_BASE_REF`, and if we are, attempt to read the `event.json` payload made available by GitHub in `GITHUB_EVENT_PATH`. 

### Review checklist


Drarig29 (approved):
The changed helpers look good to me (from Synthetics CT)

PR #1479, opened 2024-10-22
Add option to pass config shas manually on `correlate` command

rodrigo-roca:
### What and why?

The `correlate` command tries to automatically get the affected commit SHAs ([docs](https://docs.datadoghq.com/continuous_delivery/deployments/argocd#correlate-deployments-with-ci-pipelines)). However, in some setups such as generating a PR and merging it from the CI, the command is not able to properly make the correlation.

This PR adds an option to the correlate command so that customers can manually select which git commit SHAs should be correlated. 

### How?

Adding an array option to the command.

### Review checklist

### DIFF
--- a/src/commands/deployment/README.md
+++ b/src/commands/deployment/README.md
@@ -42,7 +42,7 @@ datadog-ci deployment correlate --provider argocd
 
 - `--provider` (**required**): the CD provider name. Currently, the only supported CD provider is `argocd`.
 - `--config-repo`: configuration repository URL where the kubernetes manifests are stored. If empty, the command tries to get it using the git command `git ls-remote --get-url`.
-- `--config-shas`: a list of the Git commit SHAs of the configuration repository. If empty, the command tries to get all unpushed commits using a `git log` command.
+- `--config-shas`: a list of the Git commit SHAs of the configuration repository. If empty, the command tries to get all local commits using a `git log` command.
 - `--dry-run` (default: `false`): prevents the command from sending any data to Datadog. All the other checks are still performed.
 
 ### Environment variables
--- a/src/commands/deployment/README.md
+++ b/src/commands/deployment/README.md
@@ -56,6 +56,7 @@ datadog-ci deployment correlate --provider argocd
 
 - `--provider` (**required**): the CD provider name. Currently, the only supported CD provider is `argocd`.
 - `--config-repo`: configuration repository URL where the kubernetes manifests are stored. If empty, the command tries to get it using the git command `git ls-remote --get-url`.
+- `--config-shas`: a list of the Git commit SHAs of the configuration repository. If empty, the command tries to get all unpushed commits using a `git log` command.
 - `--dry-run` (default: `false`): prevents the command from sending any data to Datadog. All the other checks are still performed.
 
 ### Environment variables
--- a/src/commands/deployment/README.md
+++ b/src/commands/deployment/README.md
@@ -33,21 +33,7 @@ datadog-ci deployment mark --env prod --service payment-service --revision v1.1.
 The `correlate` command connects a GitOps deployment with the CI pipeline of the application repository. Once they are connected, you can see in Datadog's UI which pipeline
 triggered a deployment, and which deployments were triggered by a pipeline.
 
-**Important**: This command does not work for every setup. Refer to [when to call the command](#When-to-call-the-command) for more details.
-
-#### When to call the command
-
-In order for the command to work properly, it needs to be called between when the configuration changes are committed and when they are pushed to the configuration repository (where the Kubernetes manifests are). The flow should be similar to the following:
-
-1. Make the changes to the configuration (for example, update a image tag).
-2. Run `git commit -m "update kubernetes configuration"`.
-3. Run `datadog-ci deployment correlate --provider <cd_provider>` (refer to the command syntax below).
-4. Run `git push`.
-
-If you are using [argo cd image updater][3], this command does not work since it relies on making the changes using `git commit`. 
-
-Again, these steps need to happen in your CI since the end goal of this command is to correlate the pipeline doing the configuration changes
-with the associated deployments.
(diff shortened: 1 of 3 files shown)


src/commands/deployment/correlate.ts
```diff
@@ -76,45 +77,45 @@ export class DeploymentCorrelateCommand extends Command {
...
-    let localCommitShas: string[]
-    if (this.configurationRepo) {
-      localCommitShas = await gitLocalCommitShas(git, currentBranch)
-    } else {
-      ;[this.configurationRepo, localCommitShas] = await Promise.all([
-        gitRepositoryURL(git),
-        gitLocalCommitShas(git, currentBranch),
-      ])
+    if (!this.configurationShas) {
+      const currentBranch = await gitCurrentBranch(git)
```
andrea-mosk, line 91: Can we log something here? e.g. Retrieving current commit from Git configuration or something that makes more sense 😬

andrea-mosk (commented):
Overall LGTM, but we need to add docs for this

src/commands/deployment/README.md
```diff
@@ -56,6 +56,7 @@ datadog-ci deployment correlate --provider argocd
 
 - `--provider` (**required**): the CD provider name. Currently, the only supported CD provider is `argocd`.
 - `--config-repo`: configuration repository URL where the kubernetes manifests are stored. If empty, the command tries to get it using the git command `git ls-remote --get-url`.
+- `--config-shas`: a list of the Git commit SHAs of the configuration repository. If empty, the command tries to get all unpushed commits using a `git log` command.
```
andrea-mosk, line 59: We also need to take a look at the rest of the docs. For example above we state:

> In order for the command to work properly, it needs to be called between when the configuration changes are committed and when they are pushed to the configuration repository (where the Kubernetes manifests are).

This is now one of the use cases we will support. If config-shas are not specified (and config-repo as well??), they need to execute it at a specific point. We should adapt the docs to mention that commits detection only works for the commit-push use case if the command is executed in the middle. For other use cases, they should explicitly define the shas (and repo?)

src/commands/deployment/README.md
```diff
@@ -33,21 +33,7 @@ datadog-ci deployment mark --env prod --service payment-service --revision v1.1.
...
-1. Make the changes to the configuration (for example, update a image tag).
-2. Run `git commit -m "update kubernetes configuration"`.
-3. Run `datadog-ci deployment correlate --provider <cd_provider>` (refer to the command syntax below).
-4. Run `git push`.
-
-If you are using [argo cd image updater][3], this command does not work since it relies on making the changes using `git commit`. 
-
-Again, these steps need to happen in your CI since the end goal of this command is to correlate the pipeline doing the configuration changes
-with the associated deployments.
+**Important**: This command does not work for every setup. Refer to the [documentation][3] for more details.
```
rodrigo-roca, line 36: This information was duplicated, we will refer the documentation instead.

jhgilbert (approved):

Drarig29 (approved):

src/commands/deployment/correlate.ts
```diff
@@ -76,45 +83,46 @@ export class DeploymentCorrelateCommand extends Command {
...
+        return 1
+      }
+      this.configurationShas = await gitLocalCommitShas(git, currentBranch)
     }
 
-    if (this.configurationRepo === undefined || this.configurationRepo === '') {
-      this.logger.error('Could not retrieve repository URL, check out a repository or provide it with --config-repo')
+    if (this.configurationShas.length === 0) {
+      this.logger.error(
+        'Could not retrieve commit SHAs, make commits and then call this command or provide them with --config-shas'
```
juan-fernandez, line 109: I think this is clearer
```suggestion
        'Could not retrieve commit SHAs, commit changes and then call this command or provide them with --config-shas'
```

src/commands/deployment/correlate.ts
```diff
@@ -25,11 +25,18 @@ export class DeploymentCorrelateCommand extends Command {
       This command will correlate the pipeline with a GitOps CD deployment.\n
       See README for additional details.
     `,
-    examples: [['Correlate an Argo CD deployment', 'datadog-ci deployment correlate --provider argocd']],
+    examples: [
+      ['Correlate an Argo CD deployment', 'datadog-ci deployment correlate --provider argocd'],
+      [
+        'Correlate ArgoCD deployment manually',
+        'datadog-ci deployment correlate --provider argocd --config-repo https://github.com/my-manifests-repo --config-shas 92eb0db6926aaf51b9fb223895b6d8d1c0ff1ff4',
```
juan-fernandez, line 32: could we add an example where multiple shas are passed? or modify this one to have multiple. Some devs might not know how to pass multiple SHAs (is it by separating them by comma? Is it passing `--config-shas` multiple times?)

src/commands/deployment/__tests__/correlate.test.ts
```diff
@@ -56,15 +68,21 @@ describe('execute', () => {
       CI_PIPELINE_ID: '1',
       CI_JOB_ID: '1',
     }
-    const {context, code} = await runCLI(['--provider', 'argocd', '--dry-run'], envVars)
+    const {context, code} = await runCLI(
+      ['--provider', 'argocd', '--config-shas', 'abcdef', '--config-shas', 'fedcba', '--dry-run'],
```
juan-fernandez, line 72: do we have a test where `--config-shas` is not passed but we _can_ retrieve the commits?

juan-fernandez (approved):
LGTM, mostly nitpicks

src/commands/deployment/__tests__/correlate.test.ts
rodrigo-roca: Not at the moment. Is there an easy way to mock the git client? Otherwise the test would not work because we try to get the local commits that were not pushed.

PR #1483, opened 2024-10-28
[SYNTH-16456] Make server result optional in `Result`

Drarig29:
### What and why?

The source of truth for a result is the batch and not the server result (which we poll with the `/poll_results` endpoint).

The server result corresponding to a batch result may happen to never be available, and in that case datadog-ci should fall back to the source of truth (the batch) instead of failing.

When such thing happens for a result, a warning log is printed (`The information for result {resultId} of test {publicId} was incomplete at the end of the batch.`) and the `Result.result` property will be `undefined`.

### How?

- Make `Result.result` optional
- Update fixtures and tests

### Review checklist

### DIFF
--- a/src/commands/synthetics/utils/internal.ts
+++ b/src/commands/synthetics/utils/internal.ts
@@ -8,6 +8,7 @@ import {
   ResultInBatch,
   ResultInBatchSkippedBySelectiveRerun,
   RetryConfig,
+  ServerResult,
   Test,
   TestNotFound,
   TestSkipped,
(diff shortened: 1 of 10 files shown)


src/commands/synthetics/utils/internal.ts
```diff
@@ -48,10 +49,20 @@ export const hasResultPassed = (
...
-export const hasResult = (result: Result): result is BaseResult => {
+/**
+ * Whether the result is of type {@link BaseResult}, i.e. it wasn't skipped.
+ */
+export const isBaseResult = (result: Result): result is BaseResult => {
   return !isResultSkippedBySelectiveRerun(result)
 }
 
+/**
+ * Whether the result has a defined {@link BaseResult.result} property.
```
etnbrd, line 60:
```suggestion
 * Whether the result has a defined {@link BaseResult.result} property.
 * The result property could be undefined if it's not available when polling for it, which is a known bug.
```

etnbrd (approved):
Thanks for adding the comments! 🙇 
Added some precision on the comment for `hasDefinedResult` about missing `result`.
Otherwise, LGTM!

PR #1485, opened 2024-10-29
[SDCI] Remove correlate command flaky test

rodrigo-roca:
### What and why?

The tests fails if you make some commits and don't push them making it a flaky test since it depends on the state of the runner in order for the test to pass.

### How?

We will remove the test for now. We may want to look into how to mock the git client.

### Review checklist


Drarig29 (approved):
Thanks!

PR #1487, opened 2024-10-31
Remove git checks in SBOM payload generation

tanishq-dubey:
### What and why?
~~STAL-3069~~

- ~~If all required git tags are not in our analysis, we returned a generic failure message on payload upload. This made debugging difficult for users. We are now returning the specific git tags that are missing to make debugging easier.~~

### How?

~~This change adds a new error message renderer for missing git fields, and changes the tag requirement logic to share it between the error renderer and the `generatePayload` check~~

Must of this ticket was completed in #1489, this PR finishes up the error handling in the payload.

### Review checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@datadog/datadog-ci",
-  "version": "2.45.0",
+  "version": "2.46.0",
   "description": "Run datadog actions from the CI.",
   "repository": "https://github.com/DataDog/datadog-ci",
   "license": "Apache-2.0",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@datadog/datadog-ci",
-  "version": "2.46.0",
+  "version": "2.45.0",
   "description": "Run datadog actions from the CI.",
   "repository": "https://github.com/DataDog/datadog-ci",
   "license": "Apache-2.0",
--- a/src/commands/sbom/payload.ts
+++ b/src/commands/sbom/payload.ts
@@ -12,7 +12,7 @@ import {
   GIT_SHA,
 } from '../../helpers/tags'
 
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@datadog/datadog-ci",
-  "version": "2.45.0",
+  "version": "2.46.0",
```
dastrong, line 3: This won't be a breaking change, so we might want to bump the patch version only?

Drarig29: @tanishq-dubey can you undo this change in your PR? Bumping the version in `package.json` is part of the release process and should be done in its own PR

src/commands/sbom/payload.ts
```diff
@@ -138,15 +126,15 @@ export const generatePayload = (
   return {
     id: crypto.randomUUID(),
     commit: {
-      author_name: tags[GIT_COMMIT_AUTHOR_NAME],
-      author_email: tags[GIT_COMMIT_AUTHOR_EMAIL],
-      committer_name: tags[GIT_COMMIT_COMMITTER_NAME],
-      committer_email: tags[GIT_COMMIT_COMMITTER_EMAIL],
-      sha: tags[GIT_SHA],
-      branch: tags[GIT_BRANCH],
+      author_name: tags[GIT_COMMIT_AUTHOR_NAME] ?? '',
```
dastrong, line 129: Nit: could we add a comment to indicate that these tags have a value and are checked earlier?

dastrong (approved):
None blocking nit

src/commands/sbom/payload.ts
```diff
@@ -135,18 +123,20 @@ export const generatePayload = (
     }
   }
 
+  // The tags ditionary is validated prior to generatePayload being called
```
Drarig29, line 126:
```suggestion
  // The tags dictionary is validated prior to generatePayload being called
```

vitor-de-araujo (approved):

PR #1494, opened 2024-11-18
[SYNTH-16911] Trigger batch with LTDs

Drarig29:
### What and why?

This PR adds support for loading LTDs and sending them to the trigger endpoint.

### How?

- Add `localTestDefinition` to `.synthetics.json` files

### Review checklist

### DIFF
--- a/src/commands/synthetics/interfaces.ts
+++ b/src/commands/synthetics/interfaces.ts
@@ -232,7 +232,7 @@ export interface Step {
   }[]
 }
 
-export interface ServerTest {
+export interface LocalTestDefinition {
   config: {
     assertions: Assertion[]
     request: {
@@ -247,42 +247,35 @@ export interface ServerTest {
     steps?: {subtype: string}[] // For multistep API tests
     variables: string[]
   }
-  created_at: string
-  created_by: User
   locations: string[]
   message: string
-  modified_at: string
-  modified_by: User
-  monitor_id: number
   name: string
   options: {
     ci?: {
       executionRule: ExecutionRule
     }
     device_ids?: string[]
-    min_failure_duration: number
-    min_location_failed: number
     mobileApplication?: MobileApplication
-    tick_every: number
     retry?: {
       count?: number
     }
   }
-  overall_state: number
-  overall_state_modified: string
-  public_id: string
-  status: string
-  stepCount: number
+  /** Can be used to link to an existing remote test. */
+  public_id?: string
   subtype: string
   tags: string[]
   type: string
 }
 
-export interface Test extends ServerTest {
-  suite?: string
+export interface ServerTest extends LocalTestDefinition {
+  monitor_id: number
+  status: 'live' | 'paused'
+  public_id: string
 }
 
-export type LocalTestDefinition = Omit<ServerTest, 'public_id'>
+export type Test = (ServerTest | LocalTestDefinition) & {
+  suite?: string
+}
 
 export interface Assertion {
   actual: string | number | Date | {[key: string]: any}
@@ -309,13 +302,6 @@ export enum Operator {
   doesNotMatch = 'doesNotMatch',
   validatesJSONPath = 'validatesJSONPath',
   validatesXPath = 'validatesXPath',
-}
-
-export interface User {
-  email: string
-  handle: string
-  id: number
-  name: string
 }
 
 export interface Location {
@@ -440,8 +426,6 @@ export interface RemoteTriggerConfig {
   suite?: string
 }
 export interface LocalTriggerConfig extends Omit<RemoteTriggerConfig, 'id'> {
-  /** Can be used to link to an existing remote test. */
-  id?: string
   localTestDefinition: LocalTestDefinition
 }
 export type TriggerConfig = RemoteTriggerConfig | LocalTriggerConfig
--- a/src/commands/synthetics/interfaces.ts
+++ b/src/commands/synthetics/interfaces.ts
@@ -384,14 +384,14 @@ export interface Payload {
   options?: BatchOptions
 }
 
-interface BaseTestPayload extends ServerConfigOverride {
+export interface BaseTestPayload extends ServerConfigOverride {
   executionRule?: ExecutionRule
 }
-interface LocalTestPayload extends BaseTestPayload {
-  local_test_definition?: LocalTestDefinition
+export interface LocalTestPayload extends BaseTestPayload {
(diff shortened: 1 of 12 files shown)


src/commands/synthetics/interfaces.ts
```diff
@@ -427,14 +430,21 @@ export interface BasicAuthCredentials {
...
+export interface RemoteTriggerConfig {
   // TODO SYNTH-12989: Clean up deprecated `config` in favor of `testOverrides`
   /** @deprecated This property is deprecated, please use `testOverrides` instead. */
   config?: UserConfigOverride
   testOverrides?: UserConfigOverride
   id: string
   suite?: string
 }
+export interface LocalTriggerConfig extends Omit<RemoteTriggerConfig, 'id'> {
+  /** Can be used to link to an existing remote test. */
```
etnbrd, line 443: Would the id be a public id? If so, do we need it at the root? Otherwise, the `LocalTestDefinition` should allow for an optional `publicId`, no?
The backend is expected to return an error if a public id is sent outside of `localTestDefinition`, to avoid any confusion between ephemeral tests and main tests.

src/commands/synthetics/interfaces.ts
```diff
@@ -282,6 +282,8 @@ export interface Test extends ServerTest {
   suite?: string
 }
 
+export type LocalTestDefinition = Omit<ServerTest, 'public_id'>
```
etnbrd, line 285: Let's remove all the irrelevant fields. From the [RFC](https://docs.google.com/document/d/16AwC0moUzRYjXfSB0SrTte3fhvenwbhqXCjiXIViNmQ/edit?pli=1&tab=t.0#heading=h.pf3svqg4q2o8), that's:
`message`, `monitor_id`, `options.tick_every`, `options.min_failure_duration`, `options.min_location_failed`, `options.monitor_name`, `options.monitor_priority`, `options.monitor_options`, `options.scheduling`, `status`, `tags`.

Given that it might be quite convoluted to remove all the `options.`, it might be preferable to start from an LTD type, and build the Server type around it, no?

src/commands/synthetics/test.ts
```diff
@@ -35,12 +36,26 @@ export const getTestConfigs = async (
...
+            id: test.id && (normalizePublicId(test.id) ?? ''),
+            localTestDefinition: test.localTestDefinition,
+          }
+        }
+
+        return {
+          ...common,
+          id: normalizePublicId(test.id) ?? '',
+        }
+      })
```
etnbrd, line 58: Following that an ltd trigger might not need an id:
```suggestion
        return {
          ...common,
          ...(
            isLocalTriggerConfig(test)
              ? { localTestDefinition: test.localTestDefinition }
              : { id: normalizePublicId(test.id) }
          )
        }
```

src/commands/synthetics/utils/public.ts
```diff
@@ -453,21 +453,41 @@ export const getReporter = (reporters: Reporter[]): MainReporter => ({
...
+      throw new CriticalError(
+        'INVALID_CONFIG',
+        'Linking local test definitions to an existing remote test is currently mandatory.'
+      )
+    }
+
+    return {
+      test: {
+        ...triggerConfig.localTestDefinition,
+        public_id: triggerConfig.id,
```
etnbrd, line 475: The public id should already be present in the `localTestDefinition`, no?

src/commands/synthetics/utils/public.ts
```diff
@@ -453,21 +453,41 @@ export const getReporter = (reporters: Reporter[]): MainReporter => ({
...
+      },
+      overriddenConfig: {
+        ...testOverrides,
+        public_id: triggerConfig.id,
+        local_test_definition: triggerConfig.localTestDefinition,
```
etnbrd, line 480: Do we need the ltd in the `overridenConfig`?

src/commands/synthetics/utils/public.ts
```diff
@@ -453,21 +453,41 @@ export const getReporter = (reporters: Reporter[]): MainReporter => ({
...
 
-  if (!normalizedId) {
-    throw new CriticalError('INVALID_CONFIG', `No valid public ID found in: \`${id}\``)
+  if (isLocalTriggerConfig(triggerConfig)) {
+    if (!triggerConfig.id) {
+      throw new CriticalError(
+        'INVALID_CONFIG',
+        'Linking local test definitions to an existing remote test is currently mandatory.'
+      )
+    }
```
etnbrd, line 470: What would be the consequences of removing this?
I know that we don't support LTD without a public id for now, but it might be easier for us to keep this enforced only in the backend, so there are no `datadog-ci` version with this requirement.
Can we keep track of the expected results just with their result id?

src/commands/synthetics/utils/public.ts
Drarig29: In my previous assumption, I didn't want it

src/commands/synthetics/utils/public.ts
Drarig29: Yes that's where it should be: the test overrides are flat at the root of the payload, at the same level as `public_id` and `local_test_definition`. I'll try to rename `overriddenConfig` to "payload something" because it's confusing

Drarig29: > I'll try to rename overriddenConfig to "payload something" because it's confusing

forget that

src/commands/synthetics/utils/public.ts
Drarig29: There are 3 places that require the public ID: starting tunnels, and both reporters (default and JUnit). For now, I'm replacing the public ID with `local`, so that we still have something to show. We will iterate on that

etnbrd (approved):
LGTM, thanks!
Left a couple of comments to improve the types.

src/commands/synthetics/interfaces.ts
```diff
@@ -244,41 +244,36 @@ export interface ServerTest {
...
+    steps?: {subtype: string}[] // For multistep API tests
     variables: string[]
   }
-  created_at: string
-  created_by: User
   locations: string[]
   message: string
-  modified_at: string
-  modified_by: User
-  monitor_id: number
```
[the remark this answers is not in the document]
Drarig29: For `modified_at`, `modified_by` and alike yes. For the `monitor_id`, because it only makes sense (and is used) in `ServerTest`

PR #1501, opened 2024-12-06
add a fips option to all existing commands

etnbrd:
### What and why?

This PR enables the `fips` and `fips-ignore-error`  options:

The `fips` option enables FIPS at the node level. It still requires an OpenSSL provider supporting fips.
The default OpenSSL provider in OpenSSL version 3 supports FIPS, but is not FIPS compliant. To be FIPS compliant, a validated provider must be installed on the host system: https://github.com/openssl/openssl/blob/master/README-FIPS.md

FIPS is not supported below node 16. In this case, `datadog-ci` will throw an error and exit. To ignore this error and continue the execution anyway, you can provide the `fips-ignore-error` option.

With version 17 and onward, node is compatible with OpenSSL 3, which supports FIPS. However there is no way for `datadog-ci` to enforce the OpenSSL provider is FIPS *compliant*, even though enabling the `fips` option should ensure it's only using cyphers compatible with FIPS.

> A cryptographic module is only FIPS validated after it has gone through the complex FIPS 140 validation process. As this process takes a very long time, it is not possible to validate every minor release of OpenSSL.

#### `fips`
ENV variable: DATADOG_FIPS=True
CLI param: `--fips`

#### `fips-ignore-error`
ENV variable: DATADOG_FIPS_IGNORE_ERROR=True
CLI param: `--fips-ignore-error`

### How?

These new options are added to all of the existing commands. For each of these command, an `enableFips` function is called before anything else in the execution.

### Review checklist

### DIFF
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -56,7 +56,7 @@ jobs:
       - name: Bundle library
         run: yarn build
       - name: Create standalone binary
-        run: yarn dist-standalone -t node14-linux-x64 -o datadog-ci_linux-x64
+        run: yarn dist-standalone -t node18-linux-x64 -o datadog-ci_linux-x64
       - name: Remove dist folder to check that binary can stand alone
         run: |
           rm -rf dist
@@ -96,7 +96,7 @@ jobs:
       - name: Bundle library
         run: yarn build
       - name: Create standalone binary
-        run: yarn dist-standalone -t node14-linux-arm64 -o datadog-ci_linux-arm64
+        run: yarn dist-standalone -t node18-linux-arm64 -o datadog-ci_linux-arm64
       - name: Remove dist folder to check that binary can stand alone
         run: |
           rm -rf dist
@@ -132,7 +132,7 @@ jobs:
       - name: Bundle library
         run: yarn build:win
       - name: Create standalone binary
-        run: yarn dist-standalone -t node14-win-x64 -o datadog-ci_win-x64
+        run: yarn dist-standalone -t node18-win-x64 -o datadog-ci_win-x64
       - name: Remove dist folder to check that binary can stand alone
         run: |
           rm dist -r
@@ -170,7 +170,7 @@ jobs:
       - name: Bundle library
         run: yarn build
       - name: Create standalone binary
-        run: yarn dist-standalone -t node14-macos-x64 -o datadog-ci_darwin-x64
+        run: yarn dist-standalone -t node18-macos-x64 -o datadog-ci_darwin-x64
       - name: Remove dist folder to check that binary can stand alone
         run: |
           rm -rf dist
@@ -206,7 +206,7 @@ jobs:
       - name: Bundle library
         run: yarn build
       - name: Create standalone binary
-        run: yarn dist-standalone -t node14-macos-arm64 -o datadog-ci_darwin-arm64
+        run: yarn dist-standalone -t node18-macos-arm64 -o datadog-ci_darwin-arm64
       - name: Remove dist folder to check that binary can stand alone
         run: |
           rm -rf dist
--- a/README.md
+++ b/README.md
@@ -123,7 +123,7 @@ The `fips` option allows `datadog-ci` to use a FIPS cryptographic module provide
 **Note**: `datadog-ci` cannot assert if such a provider is available, and doesn't throw any error if the provider is not FIPS validated.
 
 Node.js versions below 17 are incompatible with OpenSSL 3, which provides FIPS support.
-If you are using a Node.js version below 17, enabling the `fips` option will cause the command to throw an error.
+If you are using a Node.js version below 17, enabling the `fips` option causes the command to throw an error.
 The option `fips-ignore-error` ignores this error.
 The released `datadog-ci` binary now uses Node.js version 18 to be compatible with OpenSSL 3.
 
--- a/README.md
+++ b/README.md
@@ -115,6 +115,31 @@ The following are **beta** commands, you can enable them with with `DD_BETA_COMM
 #### `gate`
 - `evaluate`: Evaluate [Quality Gates](src/commands/gate) rules in Datadog. [📚](https://docs.datadoghq.com/quality_gates/)
 
+
+### FIPS support
+
+The `fips` option allow `datadog-ci` to use a FIPS cryptographic module provider, if the OpenSSL library installed on the host system provides it.
+`datadog-ci` cannot assert if such a provider is available, and won't throw any error in case the provider is not FIPS compliant.
+
+However, node version below 17 are incompatible with OpenSSL 3 which provides FIPS support.
+In this case, enabling the `fips` option will cause the command to throw an error.
+The option `fips-ignore-error` is meant to ignore this error.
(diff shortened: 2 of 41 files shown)


anmarchenko (approved):

.github/workflows/publish-release.yml
```diff
@@ -56,7 +56,7 @@ jobs:
       - name: Bundle library
         run: yarn build
       - name: Create standalone binary
-        run: yarn dist-standalone -t node14-linux-x64 -o datadog-ci_linux-x64
+        run: yarn dist-standalone -t node18-linux-x64 -o datadog-ci_linux-x64
```
Drarig29, line 59: For future reference, this will increase the binary size:

| File name               | Size before | Size after |
| ----------------------- | ----------- | ---------- |
| `datadog-ci_win-x64`      | 53.5 MB     | 60 MB      |
| `datadog-ci_darwin-arm64` | 61 MB       | 68 MB      |
| `datadog-ci_darwin-x64`   | 65.7 MB     | 73 MB      |
| `datadog-ci_linux-arm64`  | 58.3 MB     | 66 MB      |
| `datadog-ci_linux-x64`    | 59.6 MB     | 68 MB      |

README.md
```diff
@@ -115,6 +115,31 @@ The following are **beta** commands, you can enable them with with `DD_BETA_COMM
 #### `gate`
 - `evaluate`: Evaluate [Quality Gates](src/commands/gate) rules in Datadog. [📚](https://docs.datadoghq.com/quality_gates/)
 
+
+### FIPS support
+
+The `fips` option allow `datadog-ci` to use a FIPS cryptographic module provider, if the OpenSSL library installed on the host system provides it.
```
Drarig29, line 121:
```suggestion
The `fips` option allows `datadog-ci` to use a FIPS cryptographic module provider, if the OpenSSL library installed on the host system provides it.
```

README.md
```diff
@@ -115,6 +115,31 @@ The following are **beta** commands, you can enable them with with `DD_BETA_COMM
...
+`datadog-ci` cannot assert if such a provider is available, and won't throw any error in case the provider is not FIPS compliant.
```
Drarig29, line 122:
```suggestion
Note that `datadog-ci` cannot assert if such a provider is available, and won't throw any error in case the provider is not FIPS compliant.
```

README.md
```diff
@@ -115,6 +115,31 @@ The following are **beta** commands, you can enable them with with `DD_BETA_COMM
...
+
+However, node version below 17 are incompatible with OpenSSL 3 which provides FIPS support.
+In this case, enabling the `fips` option will cause the command to throw an error.
+The option `fips-ignore-error` is meant to ignore this error.
```
Drarig29, line 126:
```suggestion
Node.js versions below 17 are incompatible with OpenSSL 3 which provides FIPS support.
In this case, enabling the `fips` option will cause the command to throw an error.
The option `fips-ignore-error` is meant to ignore this error.
```

README.md
```diff
@@ -115,6 +115,31 @@ The following are **beta** commands, you can enable them with with `DD_BETA_COMM
...
+
+#### `fips`
+Enable FIPS support by datadog-ci, given a FIPS provider is installed on the host system.
+If no FIPS provider is installed, no error are raised.
+
+ENV variable: DATADOG_FIPS=True
+CLI param: `--fips`
+
+#### `fips-ignore-error`
+Ignore node errors in case FIPS cannot be enabled on the host system.
```
Drarig29, line 136:
```suggestion
Ignore Node.js errors in case FIPS cannot be enabled on the host system.
```

README.md
Drarig29: > in case the provider is not FIPS compliant

Did you mean FIPS validated?

src/helpers/fips.ts
```diff
@@ -0,0 +1,16 @@
+import crypto from 'crypto'
+
+export const enableFips = (fips: boolean, ignoreError = false) => {
+  try {
+    // FIX-ME The mark command is calling tag.execute function, and when doing so, the fips option
+    // is not reduced to a boolean, but remains a clipanion option, mistakenly enabling fips.
+    // eslint-disable-next-line @typescript-eslint/no-unnecessary-boolean-literal-compare
```
Drarig29, line 7: The current approach to fix this is to use setters (like [explained here](https://github.com/DataDog/datadog-ci/blob/70afd0ecee4933251b045cb90c71edeb02a394a5/src/commands/tag/tag.ts#L50-L56)).

Maybe in the future we'll want some "reset" function that takes the `this` of the parent command, to inherit its reduced options.

But let's not do that in your PR 👌

Drarig29 (approved):
LGTM, just a few nits on docs

src/helpers/fips.ts
etnbrd: My understanding is that this is caused by the confusion between the command and the functions which are executed when calling the command.

We would have two valid way of calling a command from another one:
- calling the command is done through CLI, or through `this.cli.run(['my-command', ...arguments])`" documented [here](https://mael.dev/clipanion/docs/api/cli#clirun)
- calling the function behind the command is done entirely in js, and providing a valid config object, e.g. `sendTags(ciEnv, level, provider, tags)`.

But instead of these two valid approaches, we are using an in-between, which can't initialise the command properly because it's not wrapped by Clipanion runtime, and it can't pass the expected options either, since it's not the plain js function expecting arguments.

Drarig29: Indeed, both would be better Clipanion-y ways to do it 😁

README.md
etnbrd: Yes, thanks for catching that! 🙇

README.md
```diff
@@ -118,25 +118,28 @@ The following are **beta** commands, you can enable them with with `DD_BETA_COMM
...
-`datadog-ci` cannot assert if such a provider is available, and won't throw any error in case the provider is not FIPS compliant.
+The `fips` option allows `datadog-ci` to use a FIPS cryptographic module provider if the OpenSSL library installed on the host system provides it.
 
-However, node version below 17 are incompatible with OpenSSL 3 which provides FIPS support.
-In this case, enabling the `fips` option will cause the command to throw an error.
-The option `fips-ignore-error` is meant to ignore this error.
+**Note**: `datadog-ci` cannot assert if such a provider is available, and doesn't throw any error if the provider is not FIPS validated.
+
+Node.js versions below 17 are incompatible with OpenSSL 3, which provides FIPS support.
+If you are using a Node.js version below 17, enabling the `fips` option will cause the command to throw an error.
```
neko-dd, line 126:
```suggestion
If you are using a Node.js version below 17, enabling the `fips` option causes the command to throw an error.
```

neko-dd (approved):

README.md
etnbrd: Thanks for catching that! 🙇

tanishq-dubey (approved):

Letiste (approved):

ncreated (approved):
👌

nsavoire (approved):

clifordshelton (approved):
Looks good for Serverless

PR #1504, opened 2024-12-13
Allow for legacy lambda step function steps

avangelillo:
### Commit
Allow context injection for legacy step function steps using lambda. Parameterized many of the existing unit tests to use a legacy and standard definition of the step to assert that the behavior for each of them is the same.

JIRA: https://datadoghq.atlassian.net/browse/SVLS-5638

### What and why?

From the [JIRA](https://datadoghq.atlassian.net/browse/SVLS-5638):

> A Step Function State Machine can invoke a Lambda function or a nested State Machine. The outer state machine, inner state machine and the Lambda function can all have their own traces.
> 
> To merge their traces (so that, for example, [their flame graph can be combined](https://private-user-images.githubusercontent.com/10097700/387345806-202ed91a-f034-42e9-83c8-12e82fa101e2.png?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MzMxNzM5NTksIm5iZiI6MTczMzE3MzY1OSwicGF0aCI6Ii8xMDA5NzcwMC8zODczNDU4MDYtMjAyZWQ5MWEtZjAzNC00MmU5LTgzYzgtMTJlODJmYTEwMWUyLnBuZz9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDEyMDIlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQxMjAyVDIxMDczOVomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTliNDU1Y2MwNGEyNTk5ZWQ5ZGIxZDc0NGE4NTE0ZDE0MGYxYzI3M2IyZDA2ZjBhZjQ5NmJlYmU2MzFkMDU5MGUmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0In0.1FWfi9JYhk5jiuXa_mKpfuxIuvmpFsg52EwZ-jHLcJo)), we need to alter the Payload field of the Lambda Invoke state and the Input field of the Step Function Execution state, so the necessary fields can be passed from the outer state machine into the inner state machine and the Lambda function. We sometimes call this “context injection”.

This PR allows context injection to happen on step functions using the legacy definitions for lambdas.

### How?

There was already a check to not add context when the step function step was using the legacy definition.  I removed that check and adjusted the validation that the step was for a lambda.

### Review checklist

- [ x ] Feature or bugfix MUST have appropriate tests (unit, integration)
  - Unit tests added for each of the existing ones to assert they work on either format
  - I also manually created a stepfunction and instrumented it.  You can find the [stepfunction in datadog here](https://dd.datad0g.com/functions?search=statemachinename%3Amystatemachine-cm389ejtp&cloud=aws&entity_view=step_functions&fromUser=true&panel_end=1734113465299&panel_paused=false&panel_start=1734099065299&panel_tab=executions&shouldShowLegend=true&sort=time&sp=%5B%7B%22p%22%3A%7B%22entityId%22%3A%22aws-step-functions%2Barn%3Aaws%3Astates%3Asa-east-1%3A425362996713%3Astatemachine%3Amystatemachine-cm389ejtp%22%7D%2C%22i%22%3A%22step-function-panel%22%7D%5D&start=1734099112213&end=1734113512213&paused=false).
  - Screenshot of datadog:
Screenshot 2024-12-13 at 1 13 11 PM

  
### Open questions I have
1. Paramaterizing the tests to make all of the tests assert both formats work felt like a good idea at the start, but kind of seemed like overkill once I was more familiar with things.  Should I keep them in or what tests do people think make sense?
2. Are there any general PR things I missed when creating this?


src/commands/stepfunctions/helpers.ts
```diff
@@ -159,18 +159,8 @@ export type ParametersType = {
...
-          More details can be found on https://docs.aws.amazon.com/step-functions/latest/dg/connect-lambda.html \n`
-    )
-
-    return false
-  }
-
-  // not default lambda api
-  if (step.Resource !== 'arn:aws:states:::lambda:invoke') {
+  // not default lambda api or legacy lambda definition
+  if (step.Resource !== 'arn:aws:states:::lambda:invoke' && !step.Resource?.startsWith('arn:aws:lambda')) {
```
clifordshelton, line 163: There's also the [callback invocation pattern](https://docs.aws.amazon.com/step-functions/latest/dg/connect-to-resource.html#connect-wait-token) for invoking the lambda which probably still won't work (If you're thinking of handling that in a separate PR, that's fine too). This is a case where the value for `step.Resource` will be `arn:aws:states:::lambda:invoke.waitForTaskToken`

clifordshelton (approved):
Looks good! 👍

clifordshelton:
>  ### Open questions I have
> 1. Paramaterizing the tests to make all of the tests assert both formats work felt like a good idea at the start, but kind of seemed like overkill once I was more familiar with things. Should I keep them in or what tests do people think make sense?
> 2. Are there any general PR things I missed when creating this?
1. The tests you've added seem fine. As long as we're having some test coverage, whatever setup makes our lives as developers easier is fine. If at some time we hit the point where parameterizing the tests will save us time and effort, we can switch over.
2. This PR looks great. I don't think there's any other general things that're missing 👍

Drarig29 (approved):
LGTM

PR #1499, opened 2024-12-04
[K9VULN-1175] Validate purl

juli1:
## What problem are you trying to solve?

We want to prevent customer from uploading invalid requests (for example, with invalid `purl`).

## Solution

Filter the data and prevent uploading payload with invalid package url.

## Testing

Added one unit test.

Running locally

```
$ yarn launch sbom upload ./src/commands/sbom/__tests__/fixtures/sbom-invalid-purl.json             
invalid purl invalid purl: Invalid purl: missing required "pkg" scheme component
❌ Invalid SBOM report file [./src/commands/sbom/__tests__/fixtures/sbom-invalid-purl.json].
The report is not a valid SBOM or is not compliant with our json schema.
```

PR #1507, opened 2024-12-18
fix(sbom): don't iterate over an undefined object

amaanq:
### What and why?

In https://github.com/DataDog/datadog-ci/pull/1499, we added checks in the `sbom` subcommand to validate `PURL`s. The problem is, in that code, the call to `validateFileAgainstToolRequirements` was hoisted out of the condition gated by `validateSbomFileAgainstSchema` evaluating to false, meaning that validating the file would always be called. However, in this function, we fetch the sbom's `components` and iterate over it, always assuming it exists, but this is not true. Components are an optional field in the SBOM, and so we have to check if it's actually there.

### How?

This PR adds a check to early-return in `validateSbomFileAgainstSchema` if `components` is `undefined`.

### Review checklist

A test was added to test the abscence of the components object in an sbom, but while doing so, I noticed that a prior test int here checked that this was *invalid*, but specified the wrong file name (file ended in `...-component.json`, test looked for `...-components.json`). So I've gone ahead and removed that file, and made the test return true.


juli1 (approved):

amaanq:
This is specific to the `sbom upload` command, and it *is* affecting users currently, so I'm going to go ahead and merge this.

PR #1512, opened 2024-12-24
[K9VULN-2465] Add validation of package names

juli1:
## What problem are you trying to solve?

Clients are sending us package names for Python with invalid characters. We want to start filtering these names. They are filtered in the API.

## Solution

Warn the user about invalid package names for Python. This is not removing for the payload as this is done by the API. It warns the user about a malformed payload.

## Testing

Added units tests

Tried locally with a SBOM and a library containing the name `cloud picke` (with a space).

Screenshot 2024-12-23 at 8 18 53 PM

### DIFF
--- a/src/commands/sbom/renderer.ts
+++ b/src/commands/sbom/renderer.ts
@@ -1,6 +1,8 @@
 import chalk from 'chalk'
 
 import {getBaseUrl} from '../junit/utils'
+import {Dependency} from './types'
+import {validateDependency, validateDependencyName} from './validation'
 
 const ICONS = {
   FAILED: '❌',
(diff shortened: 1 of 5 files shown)


src/commands/sbom/renderer.ts
```diff
@@ -74,3 +77,15 @@ export const renderSuccessfulCommand = (duration: number) => {
...
+  let ret = ''
+
+  for (const dep of dependencies) {
+    if (!validateDependencyName(dep)) {
+      ret += `invalid dependency name ${dep.name}\n`
+    }
+  }
+
+  return ret
+}
```
dastrong, line 91: Could we prettify this information similar to what we do in `renderMissingTags`, so it's easier to notice for end-users?

```suggestion
export const renderPayloadWarning = (dependencies: Dependency[]): string => {
  const errors = []

  for (const dep of dependencies) {
    if (!validateDependencyName(dep)) {
      errors.push(chalk.red(` - invalid dependency name ${dep.name}\n`))
    }
  }

  if (errors.length === 0) {
    return ''
  }

  let baseHeading = 'The following issues were detected:\n'

  return baseHeading += errors.join('')
}
```

dastrong (approved):
Left one code suggestion that I think we should make, but won't block on it.

PR #1515, opened 2025-01-02
[SYNTH-17698] Fix local test definition variable case

AntoineDona:
### What and why?

Fixes the case of the variable containing the local test definition. Before it was in Camel Case but it should have been in snake case.

Note: There are still some instances of the Camel Case version, but it is only for variables defined in typescript as it is required by the linter.

### How?

- Change `localTestDefinition` to `local_test_definition`

### Review checklist


Alexis-Jacob-datadog (approved):
LGTM

PR #1521, opened 2025-01-09
[SYNTH-17719] Command to deploy LTDs

Drarig29:
### What and why?

This PR adds a command to deploy LTDs as main tests. For now, it only supports deploying the LTDs from a single test file.

It also focuses on API tests only. Browser tests will be supported later.

### How?

To deploy all LTDs inside the given file:
```bash
yarn launch synthetics deploy-tests -f path/to/single-test-file.synthetics.json
```

To deploy some LTDs inside the given file:
```bash
yarn launch synthetics deploy-tests -f path/to/single-test-file.synthetics.json -p 123-456-789
```

### Review checklist

### DIFF
--- a/src/commands/synthetics/deploy-tests-lib.ts
+++ b/src/commands/synthetics/deploy-tests-lib.ts
@@ -28,12 +28,13 @@ export const deployTests = async (reporter: MainReporter, config: DeployTestsCom
   })
 
   for (const localTestDefinition of localTestDefinitionsToDeploy) {
+    // SYNTH-18434: public ID should be made required
     const publicId = localTestDefinition.local_test_definition.public_id!
 
     try {
       await deployLocalTestDefinition(api, localTestDefinition)
 
-      // SYNTH-17840: the edit test endpoint should return a version in the response, so we can print it in the logs and see it in version history
+      // SYNTH-18527: the edit test endpoint should return a version in the response, so we can print it in the logs and see it in version history
       const baseUrl = getCommonAppBaseURL(config.datadogSite, config.subdomain)
       const testLink = `${baseUrl}synthetics/details/${publicId}`
 
@@ -46,8 +47,6 @@ export const deployTests = async (reporter: MainReporter, config: DeployTestsCom
       )
     }
   }
-
-  reporter.log(`Deployed local test definitions defined in ${config.files[0]}\n`)
 }
 
 const deployLocalTestDefinition = async (api: APIHelper, test: LocalTriggerConfig): Promise<void> => {
@@ -56,7 +55,7 @@ const deployLocalTestDefinition = async (api: APIHelper, test: LocalTriggerConfi
 
   const existingRemoteTest = await api.getTest(publicId)
 
-  // XXX: We should detect noop updates in the backend
+  // SYNTH-18528: the client should not have to handle the merge for partial update
   const newRemoteTest = removeUnsupportedEditTestFields({
     ...existingRemoteTest,
     ...test.local_test_definition,
--- a/src/commands/synthetics/deploy-tests-lib.ts
+++ b/src/commands/synthetics/deploy-tests-lib.ts
@@ -1,8 +1,7 @@
 import {getCommonAppBaseURL} from '../../helpers/app'
 
 import {APIHelper, EndpointError, formatBackendErrors, getApiHelper} from './api'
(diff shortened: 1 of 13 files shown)


src/commands/synthetics/deploy-tests-lib.ts
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  if (config.publicIds.length > 0) {
+    testConfigFromFile.tests = testConfigFromFile.tests.filter(
+      (test) =>
+        isLocalTriggerConfig(test) &&
+        test.local_test_definition.public_id &&
+        config.publicIds.includes(test.local_test_definition.public_id)
+    )
+  }
```
etnbrd, line 25: Maybe let's keep the same behavior as the `run-tests` command: if no public id is provided, it deploy everything?

src/commands/synthetics/deploy-tests-lib.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  const publicIds = new Set(
+    testConfigFromFile.tests.flatMap((test) =>
+      isLocalTriggerConfig(test) && test.local_test_definition.public_id ? [test.local_test_definition.public_id] : []
+    )
+  )
```
etnbrd, line 31: Why do you need a `flatMap` and a `Set` here? What different between `publicIds` and `testConfigFromFile.tests`?

src/commands/synthetics/deploy-tests-lib.ts
```diff
@@ -0,0 +1,93 @@
...
+  const api = getApiHelper(config)
+
+  reporter.log('Deploying tests...\n\n')
+
+  const testConfigFromFile: TestConfig = {
+    tests: await getTestConfigs(config, reporter),
+  }
+
+  if (config.publicIds.length > 0) {
+    testConfigFromFile.tests = testConfigFromFile.tests.filter(
```
etnbrd, line 19: Why the mutation?

src/commands/synthetics/deploy-tests-lib.ts
```diff
@@ -0,0 +1,93 @@
...
+    ...test.local_test_definition,
+    config: {
+      ...existingRemoteTest.config,
+      ...test.local_test_definition.config,
+    },
+    options: {
+      ...existingRemoteTest.options,
+      ...test.local_test_definition.options,
+    },
+  })
```
etnbrd, line 74: Why can't we let the backend do the merge?

etnbrd (approved):
I'm ok to merge this to work incrementally, but I left a few comments to be addressed.

src/commands/synthetics/deploy-tests-lib.ts
Drarig29: I wanted to, but the edit endpoint **requires** some properties that are not in LTDs (e.g. `options.tick_every`).

Basically that endpoint doesn't fully support partial updates 😕

src/commands/synthetics/deploy-tests-lib.ts
Drarig29: I wanted to follow the exact same logic [as here](https://github.com/DataDog/datadog-ci/blob/82bbf9832259a05efdc9b711bf4496bfb02d36a1/src/commands/synthetics/multilocator.ts#L91) where we iterate over unique `publicIds`, then use `findUniqueLocalTestDefinition()` to enforce the public ID to be unique in the file.

But it's true that we actually don't need it, since `findUniqueLocalTestDefinition()` will enforce the uniqueness already.

src/commands/synthetics/deploy-tests-lib.ts
Drarig29: When you say "it deploys everything", do you mean everything in all the `.synthetics.json` files, or everything in the single test file?

I wanted to align with the other PRs we made about LTDs to only support a single test file for now

etnbrd: I think we should align on the run-test command, and do the action on everything. If `-f` is provided, then the single file is used, otherwise, do the action on all the files that are available.

src/commands/synthetics/deploy-tests-lib.ts
etnbrd: Can we update the endpoint then? 😁

Drarig29: Created [a card](https://datadoghq.atlassian.net/browse/SYNTH-18528) 👌

PR #1523, opened 2025-01-13
[SYNTH-17721] Import LTD from a main test

teodor2312:
### What and why?

Adding a new command for Synthetics that would allow users to download the definitions of a synthetics test from Datadog into a local file called a Local Test Definition (LTD)

### How?

Adding code to support this new command

You can test it with the following command
```
yarn launch synthetics import-tests --public-id nhr-q5e-44u --public-id mdx-5fv-4cj -f import_tests.json
```

### Review checklist


src/commands/synthetics/import-tests-command.ts
```diff
@@ -0,0 +1,140 @@
...
+    description: 'Import the Main Test Definition from a Datadog scheduled tests as a Local Test Definitions.',
+    details: `
+      This command imports a Main Test Definition from a Datadog scheduled tests as a Local Test Definitions to be used in local development.
+    `,
+    examples: [
+      [
+        'Explicitly specify multiple tests to run',
+        'datadog-ci synthetics import-tests --public-id pub-lic-id1 --public-id pub-lic-id2',
+      ],
+      ['Override the default glob pattern', 'datadog-ci synthetics import-tests -f ./component-1/**/*.synthetics.json'],
```
Alexis-Jacob-datadog, line 46: I think you need to update the -f example

src/commands/synthetics/import-tests-lib.ts
```diff
@@ -0,0 +1,143 @@
...
+          delete step.params.element.multiLocator
+        }
+        if ('bucketKey' in step.params.element) {
+          delete step.params.element['bucketKey']
+        }
+      }
+    }
+    for (const field of STEP_FIELDS_TRIM) {
+      if (field in step) {
+        delete step[field as keyof TestStep]
```
AntoineDona, line 137: Shouldn't this be `TestStepWithUnsupportedFields` ?

src/commands/synthetics/interfaces.ts
```diff
@@ -262,18 +277,24 @@ export interface LocalTestDefinition {
   }
   /** Can be used to link to an existing remote test. */
   public_id?: string
-  subtype: string
-  tags: string[]
+  subtype?: string
+  steps?: TestStepWithUnsupportedFields[] // From browser schema
```
AntoineDona, line 281: Here I think this should do the job, and you can remove `TestStep` which is redundant with `Step` and also remove `TestStepWithUnsupportedFields`.
You can then use `ServerTest["steps"]` to type `STEP_FIELDS_TRIM`
```suggestion
  steps?: {params: any, position: number, public_id: string}[] // From browser schema
```

src/commands/synthetics/reporters/default.ts
```diff
@@ -160,7 +160,7 @@ const renderResultOutcome = (
   }
 }
 
-const renderApiRequestDescription = (subType: string, config: Test['config']): string => {
+const renderApiRequestDescription = (subType = '<no subtype>', config: Test['config']): string => {
```
AntoineDona, line 163: I'm curious here, was this broken? Because this function is only called to render api tests and they have to have a subtype no? Or was there an instance whee we did not have any?

src/commands/synthetics/import-tests-command.ts
```diff
@@ -0,0 +1,140 @@
+import {Command, Option} from 'clipanion'
```
AntoineDona, line 1: I feel like we're starting to have a lot of files in the synthetics folder... We might want to think of a refactor to sort stuff better, all the more so since we're going to add a new command to upload the test definitions 🤔 Definitely out of scope for this pr though haha

src/commands/synthetics/__tests__/import-tests-lib.test.ts
```diff
@@ -0,0 +1,213 @@
...
+      jest.spyOn(tests, 'getTestConfigs').mockResolvedValue(mockTriggerConfig)
+
+      await importTests(mockReporter, config)
+
+      expect(apiHelper.getTest).toHaveBeenCalledTimes(2)
+      expect(tests.getTestConfigs).toHaveBeenCalledTimes(1)
+      expect(fsPromises.writeFile).toHaveBeenCalledTimes(1)
+      expect(fsPromises.writeFile).toHaveBeenCalledWith(filePath, expectedJsonData, 'utf8')
+    })
+  })
```
AntoineDona, line 212: It might be worth adding a few simple unit tests for `overwriteTestConfig` to make sur that we override the correct test and not delete another one! Might not be necessary as the `importTests` tests kinda act like integration tests 😁

AntoineDona (approved):
hey! I left some comments / nits / questions. Otherwise looks good to me, good job   🙌

AntoineDona:
Perfect, good job!!

Issue #1378, opened 2024-07-18
Deprecation Warning on Installation of packages(inflight@1.0.6, rimraf@3.0.2, glob@7.1.4)


oleksiyPetlyuk:
### Bug description

On installing I got the following warnings:
```
npm WARN deprecated inflight@1.0.6: This module is not supported, and leaks memory. Do not use it. Check out lru-cache if you want a good and tested way to coalesce async requests by a key value, which is much more comprehensive and powerful.
npm WARN deprecated rimraf@3.0.2: Rimraf versions prior to v4 are no longer supported
npm WARN deprecated glob@7.1.4: Glob versions prior to v9 are no longer supported
```

### Describe what you expected

I expect no warnings to be thrown.

### Steps to reproduce the issue

`npm install -g @datadog/datadog-ci`

### Additional context

OS Version: macOS 14.5
Node Version: v16.18.1
Package Manager: NPM 8.19.2

### Command

flutter-symbols

Drarig29:
Released in https://github.com/DataDog/datadog-ci/releases/tag/v2.40.2

Drarig29:
Just checked, and we now have a new warning...

```bash
npm warn deprecated google-p12-pem@4.0.1: Package is no longer maintained
```

Drarig29:
Will be fixed in the next release

arthurbarbero:
New error on installing dependencies with glob:11

`error glob@11.0.0: The engine "node" is incompatible with this module. Expected version "20 || >=22". Got "18.20.4"`

Will Node 18 still be supported on this new release?

Drarig29:
Hi! Here is another example of warnings from NPM:

```
npm WARN EBADENGINE Unsupported engine {
npm WARN EBADENGINE   package: 'glob@11.0.0',
npm WARN EBADENGINE   required: { node: '20 || >=22' },
npm WARN EBADENGINE   current: { node: 'v16.20.2', npm: '8.19.4' }
npm WARN EBADENGINE }
```

I'll revert the related PR because the original warning was totally OK to have:
```
npm WARN deprecated glob@7.1.4: Glob versions prior to v9 are no longer supported
```

Note that **we cannot use `glob` v9 either**, as it only supports Node.js >= 16, while we still currently support Node.js >= 14.

Drarig29:
[...]

PR #1560, opened 2025-02-24
[dep] Drop rimraf in production code

Drarig29:
### What and why?

> [!NOTE]  
> This PR is based on https://github.com/DataDog/datadog-ci/pull/1382, which had to be reverted (https://github.com/DataDog/datadog-ci/pull/1401) because it was bumping `rimraf` to its latest version (`6.0.1`), which has dropped support for Node 18: https://github.com/isaacs/rimraf#v5-to-v6

Related to https://github.com/DataDog/datadog-ci/issues/1378

This PR removes `rimraf` from production code (it's still used in `devDependencies`).

### How?

Replace `rimraf` with `rm` from `fs/promises`

### Review checklist


AntoineDona (approved):

src/commands/dsyms/utils.ts
```diff
@@ -25,8 +25,9 @@ export const createUniqueTmpDirectory = async (): Promise<string> => {
   return directoryPath
 }
 
-export const deleteDirectory = async (directoryPath: string): Promise<void> =>
-  new Promise((resolve, reject) => rimraf(directoryPath, () => resolve()))
+export const deleteDirectory = async (directoryPath: string): Promise<void> => {
+  await rimraf(directoryPath)
+}
```
BenoitZugmeyer, line 30: **💬 suggestion:** ‏since [we dropped Node 14 support](https://github.com/DataDog/datadog-ci/pull/1554), maybe just use `rm(directoryPath, { recursive: true })`?

https://nodejs.org/docs/v22.14.0/api/fs.html#fspromisesrmpath-options

BenoitZugmeyer (approved):

src/commands/dsyms/utils.ts
Drarig29: Really good point, thanks!

PR #1561, opened 2025-02-25
remove logic around unsupported fields

teodor2312:
### What and why?

### How?

### Review checklist

### DIFF
--- a/src/commands/synthetics/api.ts
+++ b/src/commands/synthetics/api.ts
@@ -23,6 +23,7 @@ import {
   Trigger,
   MobileAppUploadResult,
   MobileApplicationNewVersionParams,
+  LocalTestDefinition,
 } from './interfaces'
 import {MAX_TESTS_TO_TRIGGER} from './run-tests-command'
 import {ciTriggerApp, getDatadogHost, retry} from './utils/public'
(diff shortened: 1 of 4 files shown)


Drarig29 (approved):
Very nice! 🎉

src/commands/synthetics/api.ts
```diff
@@ -83,10 +84,13 @@ const triggerTests = (request: (args: AxiosRequestConfig) => AxiosPromise<Trigge
   return resp.data
 }
 
-const getTest = (request: (args: AxiosRequestConfig) => AxiosPromise<ServerTest>) => async (testId: string) => {
+const getTest = (request: (args: AxiosRequestConfig) => AxiosPromise<ServerTest>) => async (
```
teodor2312, line 87: So i'm not 100% set on changing getTest here. I definitely need it to be able to call a specific testType endpoint for the LTD further down, so updated it here for consistency and i also think it's nicer this way since it more closely mimics what you can actually do with the endpoint in dogweb, but at the same time this extra functionality is not really used anywhere in the code atm 🤔

Drarig29: It aligns better with our API spec, so let's keep your changes 👍

PR #1568, opened 2025-02-27
[SYNTH-18694] Do not prompt ML update confirmation in CI environments

AntoineDona:
### What and why?

Currently, running an LTD can result in pipelines stuck on hold waiting for a prompt from the user, when a ML update is detected. We want to skip this step if we detect that the `run-tests` command was used in a CI environment.

### How?

Only prompt the user for ml updates if in interactive mode.

- Add a new `isInteractive` function
- Reuse preexisting `requestConfirmation` function and improve its typing

### Review checklist

### DIFF
--- a/src/commands/synthetics/multilocator.ts
+++ b/src/commands/synthetics/multilocator.ts
@@ -1,6 +1,7 @@
 import {writeFile} from 'fs/promises'
 
-import inquirer from 'inquirer'
+import {isInteractive} from '../../helpers/ci'
+import {requestConfirmation} from '../../helpers/prompt'
 
 import {Result, MultiLocator, TestConfig, ImportTestsCommandConfig, MainReporter} from './interfaces'
 import {findUniqueLocalTestDefinition} from './local-test-definition'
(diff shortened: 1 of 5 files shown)


src/commands/synthetics/multilocator.ts
```diff
@@ -24,8 +25,13 @@ export const updateLTDMultiLocators = async (
     return reporter.log('No MultiLocator updates found. No changes will be made.\n')
   }
 
-  const userConfirmed = await promptUser(
-    'MultiLocator updates found. Do you want to apply them to your local test definition?'
+  if (!isInteractive()) {
+    return reporter.log('Cannot apply MultiLocator updates in non-interactive mode.\n')
```
teodor2312, line 29: You could add a quick test for this as well since it's a new outcome for the function 🤔

src/helpers/__tests__/ci.test.ts
```diff
@@ -363,3 +363,69 @@ describe('getCIEnv', () => {
...
+    expect(isInteractive()).toBe(true)
+  })
+
+  test('returns false when process.stdout is not a TTY and no stream is provided', () => {
+    delete process.env.CI
+    process.env.TERM = 'xterm'
+    process.stdout.isTTY = false
+
+    expect(isInteractive()).toBe(false)
+  })
```
teodor2312, line 430: sweet 👌

teodor2312 (approved):

src/helpers/prompt.ts
```diff
@@ -16,27 +19,28 @@ export const requestConfirmation = async (message: string, defaultValue = true)
...
-      {
-        type: 'input',
-        name: 'filePath',
-        message: 'Please enter a file path, or press Enter to finish:',
-      },
-    ])
+    const question: inquirer.InputQuestion<{filePath: string}> = {
+      type: 'input',
+      name: 'filePath',
+      message: 'Please enter a file path, or press Enter to finish:',
```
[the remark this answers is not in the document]
AntoineDona: Yep makes sense! Actually we don't use this one in synthetics so i.g we don't need a default param for now...

vitor-de-araujo (approved):

PR #1569, opened 2025-02-27
[K9VULN-3897] Filter large SARIF files

juli1:
### What and why?

We want to prevent large SARIF files from being uploaded. This leads to operational issues.

### How?

Add a check in the validation and do an early exit if validation fails.

## Testing

Tested by setting the max size to 1. There is the output below.

Screenshot 2025-02-27 at 2 39 57 PM

### DIFF
--- a/src/commands/sarif/validation.ts
+++ b/src/commands/sarif/validation.ts
@@ -7,12 +7,26 @@ import addFormats from 'ajv-formats'
 
 import sarifJsonSchema from './json-schema/sarif-schema-2.1.0.json'
 
+const maxSarifFileSize = 5 * 1024 * 1024 // 5MB in bytes
+
 /**
- * Validate the SARIF file against the SARIF schema.
+ * Validate the SARIF file and check if the file is too large or not valid
+ * against the SARIF schema.
  *
  * @param sarifReportPath - the path of the SARIF file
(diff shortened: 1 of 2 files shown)


tanishq-dubey (approved):

src/commands/sarif/validation.ts
```diff
@@ -7,12 +7,26 @@ import addFormats from 'ajv-formats'
...
  *
  * @param sarifReportPath - the path of the SARIF file
  */
 export const validateSarif = (sarifReportPath: string) => {
+  try {
+    const stats = fs.statSync(sarifReportPath) // Synchronously get file stats
+    const fileSize = stats.size
+
+    if (fileSize > maxSarifFileSize) {
+      return `file size too large (size: ${fileSize}, max size: ${maxSarifFileSize})`
```
amaanq, line 24: As it is, this will print something like:

```suggestion
    file size too large (size: 12247542, max size: 5242880)`
```

which is not really helpful for the user. At the very least, printing the sizes in MB would be quite helpful.

```suggestion
      return `file size too large (size: ${fileSize / 1024 / 1024} MB, max size: ${maxSarifFileSize / 1024 / 1024} MB)`
```

PR #1575, opened 2025-03-05
[lambda] [bugfix] Ensure ARM layers are used for python3.13 ARM functions

TalUsvyatsky:
### What and why?

In https://github.com/DataDog/datadog-ci/pull/1537, we added support for instrumenting Python3.13 lambdas using `datadog-ci lambda instrument` command. In that PR, we missed adding `python3.13` constant to the `ARM_LAYERS` array, which causes the command to apply non-ARM Python layers to ARM Python 3.13 lambdas.

### How?
This PR adds the constant and a unit test that catches the issue.

### Testing
- Ran yarn prepack to build CLI locally
- Created a Python 3.13 ARM lambda
  - Ran node `./dist/cli.js lambda instrument --function MY_FUNCTION_ARN --extensionVersion 68 --layer-version 105` 
  - Saw it successfully instrument with the `arn:aws:lambda:ca-central-1:464622532012:layer:Datadog-Python313-ARM:105` layer

### Review checklist

### DIFF
--- a/src/commands/lambda/__tests__/functions/instrument.part2.test.ts
+++ b/src/commands/lambda/__tests__/functions/instrument.part2.test.ts
@@ -108,25 +108,27 @@ describe('instrument', () => {
       [Runtime.python311, 'Datadog-Python311-ARM'],
       [Runtime.python312, 'Datadog-Python312-ARM'],
       [Runtime.python313, 'Datadog-Python313-ARM'],
-    ])('calculates an update request with just lambda library layers in arm architecture', async (runtime: Runtime, layer: string) => {
-      const config = {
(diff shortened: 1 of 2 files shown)


avangelillo (approved):
Will there always be both an ARM and other variant / does it make sense to have both generated from the same list so that we can't forget?

duncanista (approved):

TalUsvyatsky:
> Will there always be both an ARM and other variant / does it make sense to have both generated from the same list so that we can't forget?

My understanding is yes for .NET, Python, and Ruby, and no for the other runtimes, but I'd like to validate that. I agree that we should refactor these lists to make it harder to miss.

PR #1579, opened 2025-03-08
Fix upload of ELF dynamic symbols

nsavoire:
### What and why?

`objcopy --only-keep-debug` unexpectedly removes `.dynsym` and `.dynstr` sections. This leads to datadog-ci uplaoding ELF files with no symbol information when only dynamic symbols are available.

### How?

This PR workaround this by first dumping these sections to files with `objcopy --dump-section` and adding them back with `objcopy --add-section`.

Other changes (separate commits):
* fix typo in README
* fix `objcopy` supported targets listing
* when checking for a section existence also check that it is non empty (i.e. it has a type different from `NOBITS`): this what caused tests checking for dynamic symbol extraction to incorrectly succeed
* enable debug section compression (zstd if available otherwise zlib)
* correctly detect debug info compressed with `zlib-gnu` compression (`.zdebug_info`  section)
* allow duplicate buildIDs with different archs
* fix test case on node 23
* change codeowner of elf symbol upload to @DataDog/profiling-full-host 

### Review checklist


Drarig29 (approved):
LGTM

Gandem (approved):

PR #1580, opened 2025-03-10
[feature] Custom span command

CelianR:
### What and why?

This PR introduces a new command named `datadog-ci trace span` used to report custom spans similar to the `datadog-ci trace` command.
It is useful to report a custom span without tracing a command, multiple usages in the CI includes:
1. Monitoring tests / subtests
2. Monitoring build steps
3. Monitoring any function within a program and not the whole process

Like `trace`, it is possible to send tags and measures.

```
datadog-ci trace span [--name <name>] [--start-time <ms>] [--end-time <ms>] [--duration <ms>] [--tags] [--measures] [--dry-run]
```

See the command [README](https://github.com/DataDog/datadog-ci/blob/celian/innovation-week-25q1-custom-span/src/commands/span/README.md) for more details!

#### Example

Example usage within the datadog-agent to monitor unit test durations:

image

### How?

This is done in two steps:
- Refactoring `trace` to share some common code within helper.js (`CustomSpanCommand`) (+ same for its tests)
- Adding `span` command and its tests

### Extra discussions

- Should this use another endpoint / table (some fields like `command` or `exit_code` are not used)? Should we be able to provide these fields (especially `error_message` / `exist_code` which is useful to either obtain information or filter data)?
- Can we provide one payload for multiple spans to avoid overhead? e.g. if we want to monitor 500 unit tests, there is a lot of overhead to load and execute the process. This can be solved by providing a single payload within either a file or an argument.

### Review checklist

- [X] Feature or bugfix MUST have appropriate tests (unit, integration)


buraizu:
Created DOCS-10368 for documentation team review

CelianR:
@buraizu

> Created DOCS-10368 for documentation team review

I opened this [PR](https://github.com/DataDog/documentation/pull/28128) that updates the documentation 👍

buraizu:
@CelianR Thanks for following up, I see that the linked PR updates the `content/en/continuous_integration/pipelines/custom_commands.md` file. 

However, the documentation team was looped in to this PR because of the change in `src/commands/span/README.md`. Are the changes in your PR intended to replace the new `src/commands/span/README.md` file? Thanks in advance for any clarification.

CelianR:
> @CelianR Thanks for following up, I see that the linked PR updates the `content/en/continuous_integration/pipelines/custom_commands.md` file.
> 
> However, the documentation team was looped in to this PR because of the change in `src/commands/span/README.md`. Are the changes in your PR intended to replace the new `src/commands/span/README.md` file? Thanks in advance for any clarification.

Of course, the new README.md needs a review from your team. As it's my first PR in this repository I'm not used to the review process so I wasn't sure how it worked. Thanks for the clarification 👍

Drarig29 (approved):
LGTM structurally

src/commands/span/README.md
```diff
@@ -0,0 +1,56 @@
...
+    The resulting dictionary is merged with what is in the `DD_TAGS` environment variable. If a `key` appears both in `--tags` and `DD_TAGS`, the value in `DD_TAGS` takes precedence.
+- `--measures` is an array of key-value pairs with the format `key:value`. These measures are added to the custom span.
+    The `value` must be a number.
+- `--dry-run` (default: `false`) runs the command without sending the custom span. All other checks are performed.
+
+#### Environment variables
+
+Additionally you might configure the `span` command with environment variables:
+
+- `DD_API_KEY` (**required**): API key used to authenticate the requests.
```
buraizu, line 31: Is this only required if the user decides to configure the span command with environment variables?

buraizu (changes requested):
@CelianR  thanks for clarifying, and no worries. I've left some minor feedback here for accessibility, clarity, and conformity with our style guide. Let me know if you have any questions.

src/commands/span/README.md
CelianR: This is the same configuration as the trace command [here](https://github.com/DataDog/datadog-ci/blob/master/src/commands/trace/README.md#environment-variables) and it is the only way to pass the key so it is always required

src/commands/span/README.md
```diff
@@ -0,0 +1,58 @@
...
+```
+
+For example:
+
+```bash
+datadog-ci span --name "Say Hello" --duration 800 --tags responded-hello-too:true
+```
+
+- `--name` is a human-readable name for the reported span.
+- `--start-time` the span start time in milliseconds.
```
juan-fernandez, line 18: is this unix time in ms? Do we want to specify this? Or ms since the unix epoch or something of the sort. 

It might also be interested to support ISO 8601, but that can be done later

src/commands/span/span.ts
```diff
@@ -0,0 +1,85 @@
...
+      See README for details.
+    `,
+    examples: [
+      [
+        'Create span with name "Get Dependencies" and duration of 10s and report to Datadog',
+        'datadog-ci span --name "Get Dependencies" --duration 10000',
+      ],
+      [
+        'Create span with name "Get Dependencies" and duration of 10s and report to Datadog with tags and measures',
+        'datadog-ci span --name "Get Dependencies" --duration 10000 --tags "dependency-set:notify" --measures "n-dependencies:42"',
```
juan-fernandez, line 24: could we add examples of start and end time?

src/commands/trace/trace.ts
```diff
@@ -98,92 +72,28 @@ export class TraceCommand extends Command {
...
-
-    return apiConstructor(this.getBaseIntakeUrl(), this.config.apiKey)
-  }
-
-  private getBaseIntakeUrl() {
-    const site = process.env.DATADOG_SITE || process.env.DD_SITE || 'datadoghq.com'
-
-    return `https://api.${site}`
-  }
+    if (res !== 0) {
```
juan-fernandez, line 86: ⚠️ this behavior is different from the current `trace` implementation, right? It's kind of an edge case, but before this change, we'd still fail the command if the `exitCode` is !=0, even without the `--no-fail` flag. We'd only prevent the error if the `trace` command failed because the CI provider is not supported.

juan-fernandez: I think I'd keep the old behavior, as it's a breaking change, even if the behavior was surprising to the user

src/commands/span/span.ts
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    const endTime = this.endTimeInMs ? new Date(this.endTimeInMs) : new Date()
+    const startTime = new Date(endTime.getTime() - this.durationInMs)
+
+    return this.executeReportCustomSpan(this.generateSpanId(), startTime, endTime, {
+      name: this.name,
+      error_message: '',
+      exit_code: 0,
+      command: 'custom-span',
```
juan-fernandez, line 82: maybe this is easier to consume?
```suggestion
      command: 'datadog-ci span',
```

juan-fernandez (commented):
LGTM. Only a potentially important comment

src/commands/span/README.md
```diff
@@ -0,0 +1,58 @@
...
+- `--duration` is the duration of the span in milliseconds. If duration is provided instead of `--start-time` and `--end-time`, the span end time is the current time when executing this command. This method is less precise than using start and end times since launching the datadog-ci process requires some time.
+- `--tags` is an array of key-value pairs with the format `key:value`. These tags are added to the custom span.
+    The resulting dictionary is merged with the contents of the `DD_TAGS` environment variable. If a `key` appears both in `--tags` and `DD_TAGS`, the value in `DD_TAGS` takes precedence.
+- `--measures` is an array of key-value pairs with the format `key:value`. These measures are added to the custom span.
+    The `value` must be a number.
+- `--dry-run` (default: `false`) runs the command without sending the custom span. All other checks are performed.
+
+#### Environment variables
+
+You can also configure the `span` command with environment variables:
```
buraizu, line 29:
```suggestion
The `DD_API_KEY` environment variable is required for the `span` command. You can also set tags and the Datadog site:
```

src/commands/span/README.md
```diff
@@ -0,0 +1,56 @@
...
+For example:
+
+```bash
+datadog-ci span --name "Say Hello" --duration 800 --tags responded-hello-too:true
+```
+
+- `--name` is a human-friendly name for the reported span.
+- `--start-time` the span start time in milliseconds.
+- `--end-time` the span end time in milliseconds.
+- `--duration` is the duration of the span in milliseconds. If duration is provided instead of `--start-time` and `--end-time`, the end time will be the current time when executing the command (note that this method is less precise than using start / end times since launching the datadog-ci process takes some overhead).
```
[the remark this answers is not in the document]
buraizu: LGTM, thanks for following up and helping to clarify!

src/commands/span/README.md
buraizu: Thanks for confirming. Since the preceding sentence uses `might`, it sounds like this might not be required in all cases.

src/commands/span/README.md
CelianR: Done 👍

src/commands/trace/trace.ts
```diff
@@ -34,11 +21,11 @@ export class TraceCommand extends Command {
     examples: [
       [
         'Trace a command with name "Say Hello" and report to Datadog',
-        'datadog-ci trace --name "Say Hello" -- echo "Hello World"',
+        'datadog-ci trace span --name "Say Hello" -- echo "Hello World"',
```
tonyredondo, line 24: If I understand correctly we are changing the previous command by asking `span` here, is this a breaking change? how are you handling breaking changes with customers? is this going to be released in a major version?

tonyredondo (approved):
LGTM, just a question.

src/commands/trace/trace.ts
CelianR: Oops this is a typo, I'll remove this and check that I haven't done any extra unwanted changes. Thanks for noticing this 👍

buraizu (approved):

PR #1586, opened 2025-03-11
Implement logic to automatically discover code coverage and jUnit XML report files

nikita-tkachenko-datadog:
### What and why?

Updates `junit upload` and `coverage upload` commands with logic that automatically discovers jUnit XML and code coverage reports respectively.

### How?

The logic does recursive search in the provided folders, looking for files whose names match the common names for jUnit XML and Jacoco code coverage reports.

The logic is controlled with `--auto-discovery` flag: enabled for coverage upload by default (since this command is still in beta), but disabled for jUnit XML upload by default (to maintain backward compatibility).

`--ignored-paths` option can be used to exclude specific folders from auto-discovery search.

### Review checklist


maycmlee (approved):

src/commands/coverage/__tests__/upload.test.ts
```diff
@@ -128,12 +137,10 @@ describe('upload', () => {
       })
 
       const fileNames = Object.values(result).flatMap((paths) => paths)
-
-      expect(fileNames).toEqual([
-        './src/commands/coverage/__tests__/fixtures/jacoco-report.xml',
-        './src/commands/coverage/__tests__/fixtures/another-jacoco-report.xml',
-        './src/commands/coverage/__tests__/fixtures/subfolder/nested-jacoco-report.xml',
-      ])
+      expect(fileNames.length).toEqual(3)
```
juan-fernandez, line 140: **nitpick**: this is the same result as calling the function with `'should read all xml files and reject invalid ones'`. I guess the difference is that here you're not using `automaticReportsDiscovery` but a glob pattern. 

No biggie, but I think we could benefit from better test names. Something like `should fetch nested folders when using glob patterns`.

src/commands/coverage/__tests__/upload.test.ts
```diff
@@ -302,17 +307,6 @@ describe('execute', () => {
...
-  test('with git metadata without argument (default value is true)', async () => {
-    const {context, code} = await runCLI([
-      '--verbose',
-      process.cwd() + '/src/commands/coverage/__tests__/fixtures/single_file.xml',
-    ])
-    const output = context.stdout.toString().split(os.EOL)
-    expect(id).toHaveBeenCalled()
-    expect(code).toBe(0)
-    expect(output[4]).toContain('Syncing git metadata')
-  })
```
juan-fernandez, line 314: did we change the default for git metadata upload?

src/commands/coverage/__tests__/upload.test.ts
```diff
@@ -302,17 +307,6 @@ describe('execute', () => {
...
-
   test('without git metadata (with argument)', async () => {
```
juan-fernandez, line 310: have we tested with folders that include `.xml`? Like 
`src/commands/coverage/__tests__/fixtures/reports.xml/my-report.xml` or even an empty folder like 
`src/commands/coverage/__tests__/fixtures/empty-folder.xml/`. I know I had issues with this in the past

src/commands/coverage/upload.ts
```diff
@@ -38,14 +37,35 @@ import {
...
 
+const IS_COVERAGE_REPORT = (file: string): boolean => {
+  if (path.extname(file) !== '.xml') {
+    return false
+  }
+
+  const filename = path.basename(file)
+
+  return (
+    filename.startsWith('jacoco') || filename.includes('Jacoco') // jacoco*.xml, *Jacoco*.xml
```
juan-fernandez, line 56: is this case sensitive? If `jacoco` is not at the start of the filename, does it have to be capitalized? I wonder if we could simply do 
```suggestion
    filename.toLowerCase().includes('jacoco')
```
this wouldn't work if we _don't_ want to include filenames of the type `thing-jacoco-thing.xml` but we do want to include `thing-Jacoco-thing.xml`

src/commands/coverage/upload.ts
```diff
@@ -58,6 +78,10 @@ export class UploadCodeCoverageReportCommand extends Command {
     `,
     examples: [
       ['Upload all code coverage report files in current directory', 'datadog-ci coverage upload .'],
```
juan-fernandez, line 80: would it be current directory or current directory and all the subfolders?

src/commands/coverage/upload.ts
```diff
@@ -87,11 +111,17 @@ export class UploadCodeCoverageReportCommand extends Command {
   private measures = Option.Array('--measures')
   private tags = Option.Array('--tags')
   private format = Option.String('--format')
-  private skipGitMetadataUpload = Option.String('--skip-git-metadata-upload', 'false', {
+  private skipGitMetadataUpload = Option.String('--skip-git-metadata-upload', 'true', {
```
juan-fernandez, line 114: this change is because git metadata is not used right now, right? If that's the case, I would simply remove it now and we can add it later. Other commands have this `skip` weirdness because of breaking changes, but this command is beta

src/commands/coverage/upload.ts
```diff
@@ -38,14 +37,35 @@ import {
...
-import {detectFormat, isFile, validateCoverageReport} from './utils'
+import {detectFormat, validateCoverageReport} from './utils'
 
 const TRACE_ID_HTTP_HEADER = 'x-datadog-trace-id'
 const PARENT_ID_HTTP_HEADER = 'x-datadog-parent-id'
 const errorCodesStopUpload = [400, 403]
 
 const MAX_REPORTS_PER_REQUEST = 10
 
+const IS_COVERAGE_REPORT = (file: string): boolean => {
```
juan-fernandez, line 48: nitpick: it's a bit surprising that a function is in upper case and camel case. Could we do `function isCoverageReport`?

src/commands/coverage/upload.ts
```diff
@@ -38,14 +37,35 @@ import {
...
+  }
+
+  const filename = path.basename(file)
+
+  return (
+    filename.startsWith('jacoco') || filename.includes('Jacoco') // jacoco*.xml, *Jacoco*.xml
+  )
+}
+
+const VALIDATE_COVERAGE_REPORT = (explicitFormat: string | undefined, file: string): string | undefined => {
```
juan-fernandez, line 60: could we do `function validateCoverageReport`?

src/commands/junit/__tests__/upload.test.ts
```diff
@@ -34,9 +34,9 @@ describe('upload', () => {
     test('should read all xml files and reject invalid ones', async () => {
       const context = createMockContext()
       const command = new UploadJUnitXMLCommand()
-      const [firstFile, secondFile] = await command['getMatchingJUnitXMLFiles'].call(
+      const files = await command['getMatchingJUnitXMLFiles'].call(
         {
-          basePaths: ['./src/commands/junit/__tests__/fixtures'],
+          basePaths: ['src/commands/junit/__tests__/fixtures'],
```
juan-fernandez, line 39: will it work with `./src...` or that's no longer valid?

src/commands/junit/upload.ts
```diff
@@ -41,12 +41,25 @@ import {
   renderSuccessfulUpload,
   renderUpload,
 } from './renderer'
-import {isFile} from './utils'
 
 const TRACE_ID_HTTP_HEADER = 'x-datadog-trace-id'
 const PARENT_ID_HTTP_HEADER = 'x-datadog-parent-id'
 const errorCodesStopUpload = [400, 403]
 
+const IS_JUNIT_XML_REPORT = (file: string): boolean => {
```
juan-fernandez, line 49: same comment as above about function name

src/commands/junit/upload.ts
```diff
@@ -76,6 +89,14 @@ export class UploadJUnitXMLCommand extends Command {
     `,
     examples: [
       ['Upload all jUnit XML test report files in current directory', 'datadog-ci junit upload --service my-service .'],
+      [
+        'Discover and upload all jUnit XML test report files doing recursive search in current directory',
+        'datadog-ci junit upload --service my-service --auto-discovery .',
+      ],
+      [
+        'Discover and upload all jUnit XML test report files doing recursive search in current directory, ignoring ./src/ignored-module-a and ./src/ignored-module-b',
+        'datadog-ci junit upload --service my-service --ignored-paths ./src/ignored-module-a,./src/ignored-module-b --auto-discovery .',
```
juan-fernandez, line 98: do we say somewhere that `--ignored-paths` are comma separated?

juan-fernandez (approved):
left mostly nitpicks but otherwise LGTM

src/commands/coverage/__tests__/upload.test.ts
```diff
@@ -29,50 +28,56 @@ describe('upload', () => {
...
-      expect(output).toContain(
-        renderInvalidFile(
-          './src/commands/coverage/__tests__/fixtures/invalid-jacoco-report.xml',
-          "Unclosed tag 'report'."
-        )
-      )
-      expect(output).toContain(
-        renderInvalidFile(
-          './src/commands/coverage/__tests__/fixtures/random-file.xml',
-          'Could not detect format of ./src/commands/coverage/__tests__/fixtures/random-file.xml, please specify the format manually using the --format option'
```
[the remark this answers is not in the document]
nikita-tkachenko-datadog: the test case still verifies that invalid files are not included (as we're doing the auto-discovery in the fixtures folder where the invalid files are), I just thought it to be too much hassle to verify the exact log output (seemed to me a test of implementation details)

src/commands/coverage/__tests__/upload.test.ts
nikita-tkachenko-datadog: Makes sense, updated the test name

src/commands/coverage/__tests__/upload.test.ts
nikita-tkachenko-datadog: yes, changed the behaviour to skip git upload by default. Looking at your other comment, I think it makes sense to remove it altogether - it is true that we're not using it

src/commands/coverage/__tests__/upload.test.ts
nikita-tkachenko-datadog: renamed `subfolder` to `subfolder.xml` to cover this case

src/commands/coverage/upload.ts
nikita-tkachenko-datadog: renamed

src/commands/coverage/upload.ts
nikita-tkachenko-datadog: yup, that was case-sensitive and intentional

src/commands/coverage/upload.ts
nikita-tkachenko-datadog: renamed

src/commands/coverage/upload.ts
nikita-tkachenko-datadog: Subfolders are included. Added it to the description

src/commands/coverage/upload.ts
nikita-tkachenko-datadog: Makes sense, removed git upload altogether

src/commands/junit/upload.ts
nikita-tkachenko-datadog: yes, it is said in the readme

src/commands/junit/__tests__/upload.test.ts
nikita-tkachenko-datadog: `./src` is still valid, but then the outputs will also be prefixed with `./` (which only matters in tests, as far as I can tell)

juan-fernandez (approved):

Drarig29 (approved):
LGTM

PR #1588, opened 2025-03-13
Refactor the common config and setup into a base command

etnbrd:
### What and why?

This PR introduces a new `BaseCommand` that currently lives in the same file as `DeployTestCommand`, but is expected to be moved to a new file, and be extended from all of the Synthetics commands. Moving to a new file appears to fail in the CI (but not locally), and might be related to circular dependencies.

### How?

`abstract` and moving code around 📦 

### Review checklist


Drarig29 (approved):
LGTM

PR #1595, opened 2025-03-28
implement the build-and-test command

etnbrd:
### What and why?

This PR implements the `synthetics build-and-test` command.
Provided with a `buildCommand`, and an optional `buildPluginPort` (defaults to 4000) options, this command runs the build command, wait for a successful build, and a dev server spawned by the [Datadog build-plugin](https://github.com/DataDog/build-plugins/pull/153), to run Synthetics tests against this dev server using the tunnel.

Except the unit tests, this command hasn't been tested properly yet. It'll be done later.

### How?

The `build-and-test` command spawns the build command, and regularly request the expected dev server (e.g. `https://localhost:4000`) until the build is advertised as successful. Then it runs the tests.

### Review checklist

### DIFF
--- a/src/commands/synthetics/__tests__/build-and-test.test.ts
+++ b/src/commands/synthetics/__tests__/build-and-test.test.ts
@@ -1,0 +1,55 @@
+import {once} from 'events'
+import * as http from 'http'
+
+import {spawnBuildPluginDevServer} from '../build-and-test'
+
+import {mockReporter} from './fixtures'
+
+// if BUILD_PLUGINS_S8S_PORT is defined, this command does nothing forever, otherwise, it exits immediately
+const MOCKED_BUILD_COMMAND = (port: number) => `[ "$BUILD_PLUGINS_S8S_PORT" = "${port}" ] && tail -f /dev/null`
+// this command simulates a build command without the build plugin configured
+const MOCKED_BUILD_COMMAND_NOT_CONFIGURED = 'echo "build successful"'
+
+describe('build-and-test - spawnBuildPluginDevServer', () => {
+  test('wait for the dev server and return expected readiness and status', async () => {
+    // Given a dev server which listen on port 4000
+    const port = 4000
+    const requests: string[] = []
+    const server = http
+      .createServer((req, res) => {
+        requests.push(String(req.url))
+        res.writeHead(200, {'Content-Type': 'application/json'})
+        res.end(
+          JSON.stringify({
+            status: 'success',
+            publicPrefix: 'prefix/',
+          })
+        )
+      })
+      .listen(port)
+
+    // When calling spawnBuildPluginDevServer
+    const command = await spawnBuildPluginDevServer(MOCKED_BUILD_COMMAND(port), port, mockReporter)
+
+    // Then it should send requests to the dev server until it's ready to serve,
+    // and return the devServerUrl and the path prefix.
+    expect(command.devServerUrl).toBe('http://localhost:4000')
+    expect(command.publicPrefix).toBe('prefix/')
+    expect(requests.pop()).toBe('/_datadog-ci_/build-status')
+
+    // Close the server and the command at the end of the test.
+    server.close()
+    await Promise.all([command.stop(), once(server, 'close')])
+  })
+
+  test('alert when the build-plugin is not configured', async () => {
+    // Given a build command without the build plugin configured
+    // When calling spawnBuildPluginDevServer
+    const commandPromise = spawnBuildPluginDevServer(MOCKED_BUILD_COMMAND_NOT_CONFIGURED, 4000, mockReporter)
+
+    // Then it should throw when the command exits.
+    await expect(commandPromise).rejects.toThrow(
+      'Build command exited before the build plugin could be started. Is the build plugin configured?'
+    )
+  })
+})
--- a/src/commands/synthetics/__tests__/build-and-test.test.ts
+++ b/src/commands/synthetics/__tests__/build-and-test.test.ts
@@ -7,6 +7,17 @@ import {mockReporter} from './fixtures'
 const NODE_COMMAND = process.execPath
 
 describe('build-and-test - spawnBuildPluginDevServer', () => {
+  test('alert when the build-plugin is not configured', async () => {
+    // Given a build command without the build plugin configured
+    const MOCKED_BUILD_COMMAND_NOT_CONFIGURED = `${NODE_COMMAND} -e "console.log('build successful')"`
+
+    // When calling spawnBuildPluginDevServer
+    const commandPromise = spawnBuildPluginDevServer(MOCKED_BUILD_COMMAND_NOT_CONFIGURED, mockReporter)
+
+    // Then it should throw when the command exits.
+    await expect(commandPromise).rejects.toThrow(UnconfiguredBuildPluginError)
+  })
+
   test('wait for the dev server and return expected readiness and status', async () => {
     // Given a dev server which listen on the port provided in the environment variable BUILD_PLUGINS_S8S_PORT.
     // The implementation of this server is written as a function, and stringified into a single line,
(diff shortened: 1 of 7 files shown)


src/commands/synthetics/run-tests-command.ts
```diff
@@ -185,12 +202,42 @@ export class RunTestsCommand extends Command {
     let results: Result[]
     let summary: Summary
 
+    const [_, command] = this.path ?? []
+    if (command === 'build-and-test') {
```
yoannmoinet, line 206: Is this check really necessary?
I thought you were simply creating an alias to `run-test`.

So the user could either run `run-test` or `build-and-test` for the same command.
Simply adding `--buildCommand` as an option and do the process based on the presence of the flag.

src/commands/synthetics/run-tests-command.ts
```diff
@@ -185,12 +202,42 @@ export class RunTestsCommand extends Command {
...
+      if (!this.config.buildCommand) {
+        this.reporter.error('The `buildCommand` option is required for the `build-and-test` command.')
+
+        return 1
+      }
+      if (!this.config.buildPluginPort) {
+        this.reporter.error('The `buildPluginPort` option is required for the `build-and-test` command.')
+
+        return 1
+      }
```
yoannmoinet, line 216: You can declare the relationship of both `--buildCommand` and `--buildPluginPort` using Typanion's [`hasKeyRelationship`](https://mael.dev/typanion/docs/predicates/cascading#haskeyrelationship) directly in the command's [`schema`](https://mael.dev/clipanion/docs/validation#validating-commands).

```
static schema = [
  t.hasKeyRelationship('buildCommand', t.KeyRelationship.Requires, ['buildPluginPort'])
]
```

TBD the usefulness of `buildPluginPort` altogether 😄

yoannmoinet (commented):
Left some comments.

src/commands/synthetics/build-and-test.ts
```diff
@@ -0,0 +1,101 @@
+import {ChildProcess, spawn} from 'child_process'
+import {once} from 'events'
+
+import axios from 'axios'
```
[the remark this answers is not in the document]
Drarig29: We just dropped support for Node < 18, so we didn't have time to consider this yet. Since this is a repo-wide task, it's generally not prioritized.

src/commands/synthetics/run-tests-command.ts
```diff
@@ -139,6 +145,15 @@ export class RunTestsCommand extends Command {
...
   })
 
+  private buildCommand = Option.String('--buildCommand', {
+    description: 'The build command to generate the assets to run the tests against.',
+  })
+
+  private buildPluginPort = Option.String('--buildPluginPort', {
+    description: 'The port on which to listen for the build plugin. Default is 4000',
+    validator: validation.isInteger(),
+  })
```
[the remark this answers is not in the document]
Drarig29: Agreed: we are configuring both the server (build sever) and the client (tunnel with `resourceUrlSubstitutionRegexes`), so the port doesn't need to be configured.

src/commands/synthetics/__tests__/build-and-test.test.ts
```diff
@@ -0,0 +1,55 @@
...
+import * as http from 'http'
+
+import {spawnBuildPluginDevServer} from '../build-and-test'
+
+import {mockReporter} from './fixtures'
+
+// if BUILD_PLUGINS_S8S_PORT is defined, this command does nothing forever, otherwise, it exits immediately
+const MOCKED_BUILD_COMMAND = (port: number) => `[ "$BUILD_PLUGINS_S8S_PORT" = "${port}" ] && tail -f /dev/null`
+// this command simulates a build command without the build plugin configured
+const MOCKED_BUILD_COMMAND_NOT_CONFIGURED = 'echo "build successful"'
```
Drarig29, line 11: Can you avoid really running any bash command in unit tests? I'm working on running all our unit tests on Windows, where `tail` and `echo` won't be available.

src/commands/synthetics/build-and-test.ts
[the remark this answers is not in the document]
Drarig29: But I agree we could use `fetch` in this command because `axios` isn't bringing anything (`fetch` supports abort controllers)

Drarig29 (commented):
Mostly LTGM, left some comments

src/commands/synthetics/build-and-test.ts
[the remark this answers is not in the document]
yoannmoinet: > Since this is a repo-wide task

Yes for existing usage of `axios`, I just noted it here because it's a new usage which could simply not use `axios` to begin with, and not put more weight on the migration work of the rest of the codebase.

(Which you seem to also support based on your second comment 😄)

src/commands/synthetics/__tests__/build-and-test.test.ts
yoannmoinet: You could maybe mock `spawn` altogether?

src/commands/synthetics/build-and-test.ts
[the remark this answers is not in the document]
etnbrd: I started the implementation with `fetch`, and then moved to `axios` to get inline with the rest of the repo 🤦 
Let's keep it like this to avoid adding heterogeneity of several implementations. It's not adding much in term of migration work, and it's not entirely obvious yet if and when we want to migrate to `fetch`.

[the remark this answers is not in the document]
etnbrd: For the record, I believe that instead of this:
```
      const response = await axios.get<BuildStatus>(buildPluginServerUrl, {signal: abortSignal})
      const {status, publicPrefix = ''} = response.data
```
It would be this:
```
      const response = await fetch(buildPluginServerUrl, {signal: abortSignal})
      const {status, publicPrefix = ''} = await response.json as BuildStatus
```

src/commands/synthetics/build-and-test.ts
```diff
@@ -0,0 +1,101 @@
...
+    buildCommandExited,
+  ])
+  controller.abort()
+
+  if (buildCommandReturnValue === undefined) {
+    throw new Error('Unexpected state: buildCommandReturnValue is undefined')
+  }
+
+  if (buildCommandReturnValue.readiness === 'buildCommandExited') {
+    const errorMessage = `Build command exited before the build plugin could be started. Is the build plugin configured?`
```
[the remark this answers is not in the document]
etnbrd: I created a ticket to properly improve that later: https://datadoghq.atlassian.net/browse/SYNTH-19175
But in the mean time, I took inspiration from this message, yes. Thanks 🙇

src/commands/synthetics/__tests__/build-and-test.test.ts
etnbrd: I went with an agnostic implementation leveraging the node executable running the tests.
I'd rather avoid mocking `spawn` to make sure the env variable is propagated, and the child process is correctly executed and being killed, as it'll probably be the main source of issue in this implementation, I reckon.

src/commands/synthetics/run-tests-command.ts
etnbrd: I went ahead and removed the `buildPluginPort` altogether.

src/commands/synthetics/__tests__/build-and-test.test.ts
```diff
@@ -0,0 +1,55 @@
...
+        requests.push(String(req.url))
+        res.writeHead(200, {'Content-Type': 'application/json'})
+        res.end(
+          JSON.stringify({
+            status: 'success',
+            publicPrefix: 'prefix/',
+          })
+        )
+      })
+      .listen(port)
```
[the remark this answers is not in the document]
etnbrd: Are you concerned with the complexity reading the test implementation, or executing the test?
I understand it's bringing a bit of complexity in both, but given it allows to remain close to the expected implementation, and the test implementation remains quite straightforward, I thought it'd be acceptable.

src/commands/synthetics/interfaces.ts
```diff
@@ -542,6 +542,7 @@ export interface SyntheticsCIConfig extends APIHelperConfig {}
 export interface RunTestsCommandConfig extends SyntheticsCIConfig {
   batchTimeout?: number
   configPath: string
+  buildCommand?: string
```
Drarig29, line 545: I hate saying that, but can you alphabetically order it?

Drarig29 (approved):
LGTM, just one nit

src/commands/synthetics/__tests__/build-and-test.test.ts
yoannmoinet: > I'd rather avoid mocking spawn to make sure the env variable is propagated, and the child process is correctly executed and being killed

How do you make sure of this here?
I feel like these are not currently assessed in the tests.

src/commands/synthetics/interfaces.ts
etnbrd: Oh, good catch, thanks for raising that up 🙏

PR #1598, opened 2025-03-31
[K9VULN-3651] Automatically determine generation tool used from Sbom

piloulacdog:
### What and why?

As a developer, I want to automate the gathering of SBOM generator versions so that we can track the versions used in vulnerability scans. Users are able to manually specify the generation tool by using the command line: 
```
datadog-ci sbom upload --tags "tool.generator.name:$tool" --tags **"tool.generator.version:$version" .file
```

However, we would like also to get this value automatically if possible.

### How?
For example, `trivy`, `syft`, and `snyk add their name and version to their metadata. We will read it and pass this value as a datadog tag.

It can be found in the metadata field of the SBOM. example:
```
 "metadata": {
    "timestamp": "2025-02-17T12:32:14+00:00",
    "tools": {
      "components": [
        {
          "type": "application",
          "group": "aquasecurity",
          "name": "trivy", # here
          "version": "0.58.2" # here
        }
      ]
    },
    "component": {
      "bom-ref": "7ec72498-f24f-495f-bedc-7169be80ebff",
      "type": "application",
      "name": "../express",
      "properties": [
        {
          "name": "aquasecurity:trivy:SchemaVersion",
          "value": "2"
        }
      ]
    }
  },
```

### Note
you can ignore commit: 1b469a4546332d0eef4e0027aff0bdd11c03b610 when reviewing, this is only refactoring.


src/commands/sbom/__tests__/fixtures/cdxgen-cyclonedx1.6.json
```diff
@@ -1 +1,150 @@
...
\ No newline at end of file
+{
```
piloulacdog, line 1: I only formatted the JSON so that it's more usable in our tests

src/commands/sbom/__tests__/payload.test.ts
```diff
@@ -43,10 +48,12 @@ describe('generation of payload', () => {
     expect(payload?.dependencies[0].version).toBe('1.3.0')
     expect(payload?.dependencies[0].licenses.length).toBe(0)
     expect(payload?.dependencies[0].language).toBe(DependencyLanguage.PHP)
+    expect(payload?.tags[TOOL_GENERATOR_NAME]).toBe('cyclonedx-php-composer')
+    expect(payload?.tags[TOOL_GENERATOR_VERSION]).toBe('in-dev')
   })
 
-  test('should correctly work with a CycloneDX 1.6 file', async () => {
```
piloulacdog, line 48: I didn't modify this test, it's just that github is poorly resolving the code diff. I simply moved this test below. Review the single commit change [here](https://github.com/DataDog/datadog-ci/pull/1598/commits/2622607046b3c1d60fc858e4a8d820f3e6ffac39)

src/commands/sbom/payload.ts
```diff
@@ -240,3 +252,44 @@ const extractingRelations = (dependency: any): Relations => {
...
+  } else if (typeof tools === 'object') {
+    const preferredTool = findPreferredTool(tools)
+
+    return extractGenerationTool(preferredTool)
+  }
+
+  return undefined
+}
+
+const findPreferredTool = (tool: any): any => {
```
dastrong, line 274: Does CDX export a type for tool that we can use instead of `any` here?

dastrong (approved):
Left some non-blocking comments, but otherwise LGTM.

I wouldn't mind seeing us document in the sbom command readme what tags customers can use.

src/helpers/tags.ts
```diff
@@ -42,6 +42,10 @@ export const GIT_BASE_REF = 'git.commit.base_ref'
 export const GIT_PULL_REQUEST_BASE_BRANCH_SHA = 'git.pull_request.base_branch_sha'
 export const GIT_PULL_REQUEST_BASE_BRANCH = 'git.pull_request.base_branch'
 
+// Sbom
+export const TOOL_GENERATOR_NAME = 'tool.generator.name'
+export const TOOL_GENERATOR_VERSION = 'tool.generator.version'
```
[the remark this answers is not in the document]
piloulacdog: moved ✅

src/commands/sbom/payload.ts
piloulacdog: some might exist: https://cyclonedx-javascript-library.readthedocs.io/en/v6.6.1/typedoc/node/classes/Models.Metadata.html, but while the usage we have is very limited I might trade saving us the import of an extra library (but if we make further usage of it, I think we should definitely need).

Drarig29 (approved):

src/helpers/interfaces.ts
```diff
@@ -1,6 +1,8 @@
 import type {AxiosPromise, AxiosRequestConfig} from 'axios'
 import type {Writable} from 'stream'
 
+import {TOOL_GENERATOR_NAME, TOOL_GENERATOR_VERSION} from 'src/commands/sbom/constants'
```
Drarig29, line 4: This kind of import doesn't work in TS, I wonder why it even worked tbh 😁 And I didn't see it during review.

First, the import isn't relative here. And `src/` only works in the repo. Once packaged in NPM, we only have the `dist/` folder containing JS files compiled from TS.

Also, since we are defining global types in `src/helpers/interfaces` from these constants, they should be global too. Otherwise it's a circular dependency (command -> global -> command).

I'm creating a PR because it's breaking this build in another repo: https://github.com/DataDog/synthetics-ci-github-action/pull/301

PR #1603, opened 2025-04-02
[MLOB-2522] Add `--llmobs` flag for instrumenting Lambdas with LLM Observability

sabrenner:
### What and why?

This PR adds a `--llmobs` flag to the AWS Lambda instrument/uninstrument commands, which takes in a string. This sets
- `DD_LLMOBS_ENABLED="true"`
- `DD_LLMOBS_AGENTLESS_ENABLED="false"` (enabled as part of https://github.com/DataDog/datadog-lambda-extension/pull/628, our instructions will point users to make sure to install the extension layer)
- `DD_LLMOBS_ML_APP` to the string provided

Some additional changes/explanations/clarifications
1. Our documentation will highlight intended usage as:
```bash
datadog-ci lambda instrument -f <YOUR_LAMBDA_FUNCTION_NAME> -r <AWS_REGION> -v 106 -e 73 --llmobs <YOUR_ML_APP>
```
i.e, this should be used to instrument one layer, with both the language layer and extension layer specified (layer versions will be auto-populated).

2. With both layers, these three variables are the only ones needed to enable LLM Observability. `DD_LLMOBS_AGENTLESS_ENABLED="false"` will use the extension layer's agent as a proxy, enabled by https://github.com/DataDog/datadog-lambda-extension/pull/628

### How?

Adds an `--llmobs` option to read, and parses it onto the `llmobsMlApp` setting, which if it is set, sets `DD_LLMOBS_ENABLED`, `DD_LLMOBS_AGENTLESS_ENABLED`, and `DD_LLMOBS_ML_APP`. This is my first time contributing, so I did my best to add tests where I saw appropriate. Happy to add more where I should if I missed some spots!

### Review checklist


src/commands/lambda/README.md
```diff
@@ -106,6 +106,7 @@ You can pass the following arguments to `instrument` to specify its behavior. Th
...
 | `--no-upload-git-metadata`     |           | Disables Git metadata uploading, as a part of source code integration. Use this flag if you have the Datadog Github Integration installed, as it renders Git metadata uploading unnecessary.                                                                                                                                                  |         |
 | `--apm-flush-deadline`         |           | Used to determine when to submit spans before a timeout occurs, in milliseconds. When the remaining time in an AWS Lambda invocation is less than the value set, the tracer attempts to submit the current active spans and all finished spans. Supported for NodeJS and Python. Defaults to `100` milliseconds.                              |         |
+| `--llmobs`                     |           | If specified, enables LLM Observability for the instrumented function(s) with the provided ML application name.                                                                                                                                                                                                                               |         |
```
duncanista, line 109: Just curious, what happens if you don't set the app name but enable LLMObs? as in, just the two env vars

```json
      "DD_LLMOBS_ENABLED": "true",
      "DD_LLMOBS_AGENTLESS_ENABLED": "false"
```

What would happen to the `ML_APP` name?

sabrenner: yeah good question. when I tried doing this in the tests, ie
```javascript
const code = await cli.run(
  [
    'lambda',
    'instrument',
    // ...
    '--llmobs',
    // 'my-ml-app',
  ],
  context
)
```
i got a mismatched output of
```
Unknown Syntax Error: Command not found; did you mean one of:
...
While running lambda instrument -f arn:aws:lambda:us-east-1:123456789012:function:lambda-hello-world --dry-run --layerVersion 10 --logLevel debug --service middletier --env staging --version 0.2 --extra-tags layer:api,team:intake --no-source-code-integration --llmobs
```
and assumed that when using `Options.String('llmobs')` expects a value or otherwise causes a failure somewhere. But, i'm not too well-versed on this behavior here.

but, in general, if ml_app is not set, the LLMObs SDKs will throw/raise at runtime.

michaelcretzman (approved):

TalUsvyatsky (approved):

duncanista (approved):

PR #1604, opened 2025-04-02
chore: Improve error message for flare file upload

lym953:
### Problem

According to @SiddhithaBhoopathy , a few customers failed to send Lambda Flare due to potential API key error:
- https://datadog.zendesk.com/agent/tickets/2076284
- https://datadog.zendesk.com/agent/tickets/2077466
- https://datadoghq.atlassian.net/browse/SLES-1893

The current error message customers get is like:
> [Error] Failed to send flare file to Datadog Support: Request failed with status code 400. request does not match with Datadog or Zendesk information: VER_000
Is your Datadog API key correct?

where no clear action item is available, so the customer reached back to support engineers for help.

### What

Adds a doc link in the error message, and suggests the customer to follow it to set API key.

### Review checklist

### DIFF
--- a/src/helpers/flare.ts
+++ b/src/helpers/flare.ts
@@ -73,7 +73,8 @@ export const sendToDatadog = async (
           break
         case 400:
         case 403:
-          message += 'Is your Datadog API key correct?\n'
+          message += `Is your Datadog API key correct? Please follow this doc to set your API key: 
+https://docs.datadoghq.com/serverless/libraries_integrations/cli/#environment-variables\n`
           break
       }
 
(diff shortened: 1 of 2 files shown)


src/helpers/flare.ts
```diff
@@ -73,7 +73,8 @@ export const sendToDatadog = async (
           break
         case 400:
         case 403:
```
avangelillo, line 75: Mostly for my understanding, but I'm a bit surprised to see that 400 and 403 are handled the same here.  Does a 400 get thrown for access issues?

avangelillo (approved):
Are there any other places we check for the API key that should get similar messaging?

src/helpers/flare.ts
TalUsvyatsky: I'm also surprised API key issues result in a 403 and not a 401

lym953: I have no idea either. @nhulston do you know?

TalUsvyatsky (approved):

PR #1615, opened 2025-04-11
[Synthetics] Redesign build-and-test command to spawn the dev server

etnbrd:
### What and why?

The build command can trigger several build and neither `datadog-ci` nor the build-plugins have any control or insight over the number of builds triggered. These builds could be run in parallel or in sequence, and the only way to detect that all builds finished is when the build command exited successfully. (we'll handle the case when the build command exit with an error later, in [SYNTH-19418](https://datadoghq.atlassian.net/browse/SYNTH-19418)).

### How?

This PR redesign the `build-and-test` command to host the dev server in `datadog-ci` rather than in the build-plugin.
The build plugins will report the completed builds to the dev server through the `/_datadog-ci_/build` endpoint.

The unit tests for the static serving of the dev server will be implemented later in [SYNTH-19419](https://datadoghq.atlassian.net/browse/SYNTH-19419).

### Review checklist

[SYNTH-19418]: https://datadoghq.atlassian.net/browse/SYNTH-19418?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SYNTH-19419]: https://datadoghq.atlassian.net/browse/SYNTH-19419?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/commands/synthetics/build-and-test.ts
+++ b/src/commands/synthetics/build-and-test.ts
@@ -1,130 +1,235 @@
-import {ChildProcess, spawn} from 'child_process'
+import {spawn} from 'child_process'
 import {once} from 'events'
-
-import axios from 'axios'
+import {promises as fs} from 'fs'
+import * as http from 'http'
+import * as path from 'path'
 
 import {MainReporter} from './interfaces'
-import {poll} from './utils/internal'
 
-interface BuildStatus {
-  status: 'running' | 'fail' | 'success'
-  publicPrefix?: string
+interface ReportedBuild {
+  outputDirectory: string // the path to the assets directory, absolute or relative to the CWD
+  publicPath: string // the public prefix
 }
-
-type BuildCommandReturnValue =
-  | {
-      readiness: 'buildCommandReady'
-      publicPrefix?: string
-    }
-  | {
-      readiness: 'buildCommandExited'
-    }
-  | {
-      readiness: 'buildCommandErrored'
-      error: Error
-    }
-
-export const DEFAULT_BUILD_PLUGIN_PORT = 4000
 
 export const UnconfiguredBuildPluginError = new Error(`
 We couldn't detect the Datadog Build plugins within your build. Did you add it?
 If not, you can learn more about it here: https://github.com/DataDog/build-plugins#readme
 `)
 
-const watchBuildPluginServerReadiness = async (buildPluginServerUrl: string, abortSignal: AbortSignal) => {
-  return poll(async () => {
-    try {
-      const response = await axios.get<BuildStatus>(buildPluginServerUrl, {signal: abortSignal})
-      const {status, publicPrefix = ''} = response.data
+export const MalformedBuildError = new Error(
+  `Invalid payload. Expected payload is {\"outputDirectory\": string, \"publicPath\": string}`
+)
 
-      if (status === 'success') {
-        return {
-          readiness: 'buildCommandReady',
-          publicPrefix,
-        } as const
-      }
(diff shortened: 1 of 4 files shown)


src/commands/synthetics/build-and-test.ts
```diff
@@ -1,130 +1,254 @@
...
+  throw MalformedBuildError
+}
+
+const spawnDevServer = async (): Promise<{builds: ReportedBuild[]; server: http.Server; url: string}> => {
+  const builds: ReportedBuild[] = []
+  const requestHandler = getRequestHandler({
+    builds,
+    root: process.cwd(),
+    routes: {
+      '/_datadog-ci_/build': {
```
AntoineDona, line 169: nit (non-blocking): We could have a global variable for this path, as it is used 2 times in this file.

src/commands/synthetics/build-and-test.ts
```diff
@@ -1,130 +1,254 @@
...
+
+  for (const build of builds) {
+    const url = new URL(requestUrl, 'http://127.0.0.1')
+    const filePath = path.join(
+      path.resolve(staticPath, build.outputDirectory), // absolute path to the assets directory
+      path.relative(build.publicPath, url.pathname), // relative path to the file
+      url.pathname.endsWith('/') ? 'index.html' : '' // add index.html if the path ends with a slash
+    )
+    const found = await fileExists(filePath)
+
```
AntoineDona, line 89: nit (non-blocking): Could add this to prevent path traversal, but not 100% necessary as this feature will be used in a controlled dev environment.

```suggestion
// Verify path is within the intended directory
const outputDirPath = path.resolve(staticPath, build.outputDirectory)
if (!filePath.startsWith(outputDirPath)) {
  return { found: false }
}
```

src/commands/synthetics/build-and-test.ts
```diff
@@ -1,130 +1,254 @@
...
-        return {
-          readiness: 'buildCommandReady',
-          publicPrefix,
-        } as const
+const prepareFile = async (root = process.cwd(), builds: ReportedBuild[], requestUrl: string): Promise<File> => {
+  const staticPath = path.isAbsolute(root) ? root : path.resolve(process.cwd(), root)
+
+  for (const build of builds) {
+    const url = new URL(requestUrl, 'http://127.0.0.1')
+    const filePath = path.join(
```
AntoineDona, line 83: nit (non-blocking): Might be worth adding a TODO to add unit tests, otherwise I'll mostly forget to do it 😁

AntoineDona (approved):
LGTM 👍 Just left a few comment, only non-blocking nits.
Thanks for all of your work!!

src/commands/synthetics/build-and-test.ts
etnbrd: I added a ticket for you not to forget them 😁 
https://datadoghq.atlassian.net/browse/SYNTH-19419

src/commands/synthetics/build-and-test.ts
etnbrd: Sounds good. I added two checks:
- the request url is matching the `publicPath` of the current build in the iteration
- the file path is a direct descendant of the `outputDirectory` of the build.

PR #1617, opened 2025-04-14
[SYNTH-19349] bump the batch timeout safety margin to 60s

mtalec:
Need to deal with some smoke test triggers with the following error:

```
SMOKE TEST ERROR: The batch didn't timeout after the expected timeout period
```
The safety limit margin was set to 15s which is a bit to short of a margin, as any lag > 15s between the batch reducer and datadog-ci can cause this error.

Thus, went from 15s to 60s by having `12*POLLING_INTERVAL` instead of `3*`


AntoineDona (approved):
Looks good to me 👍

Issue #1577, opened 2025-03-06
Uploading source maps broken on Windows


derekcicerone:
### Bug description

We build a cross-platform Electron app which uploads source maps for its main and renderer processes. Before the latest major version upgrade (v3) these commands worked on Windows:
```
          datadog-ci sourcemaps upload .webpack/main \
            --minified-path-prefix=webpack://main \
            --release-version=${{ needs.setup.outputs.version }} \
            --service=roam-app
          datadog-ci sourcemaps upload .webpack/renderer \
            --minified-path-prefix=webpack://renderer \
            --release-version=${{ needs.setup.outputs.version }} \
            --service=roam-app
```

However now we get this output:
<img width="1326" alt="Image" src="https://github.com/user-attachments/assets/affc2b21-1380-4a73-b775-ec69ea27bfa6" />

This is the error we get about unminification not working when viewing a stack trace in Datadog:
<img width="1298" alt="Image" src="https://github.com/user-attachments/assets/1ae9f46e-52ea-4375-b8e7-76788ebbe3ce" />

### Command

sourcemaps

Drarig29:
Hi @derekcicerone! It's probably due to the `glob` package upgrade (#1559).

Can you show a screenshot of the logs before version `3.0.0`? It looks like there is a mix of `/` and `\` in your output, and I wonder if it's the cause.

derekcicerone:
Yeah, I think it is due to the glob upgrade. Here's what the output looked like from v2:
<img width="1325" alt="Image" src="https://github.com/user-attachments/assets/1c8790fc-dace-4310-a437-75e2d2f8cd59" />

I think there is a second bug as well. Initially I tried to fix this by running the upload command from the directory containing the source maps with this syntax: `datadog-ci sourcemaps .` but then the uploaded files lost their periods before the extension part of the file names:
<img width="1330" alt="Image" src="https://github.com/user-attachments/assets/04cfa3eb-9e63-4c9a-b097-80afc2bf77c2" />
I ended up working around that issue by switching to `./` for the upload base path.  

In the end, I worked around both issues by using these commands:
```
      - name: Upload main process source maps to Datadog
        run: |
          datadog-ci sourcemaps upload ./ \
            --minified-path-prefix=webpack://main \
            --release-version=${{ needs.setup.outputs.version }} \
            --service=roam-app
        env:
          DATADOG_API_KEY: ${{ secrets.DD_API_KEY }}
        shell: bash
[...]

PR #1622, opened 2025-04-18
[test] Simplify unit tests with new helpers

Drarig29:
### What and why?

- https://github.com/DataDog/datadog-ci/pull/1620
- https://github.com/DataDog/datadog-ci/pull/1622 (← **you are here**)
- https://github.com/DataDog/datadog-ci/pull/1623
- https://github.com/DataDog/datadog-ci/pull/1624

This PR introduces the `makeRunCLI()` helper, which returns a standardized `runCLI` function which can be used in all unit tests. Note that a few `runCLI` implementations were non-trivial, so I couldn't refactor them with `makeRunCLI()`.

It also:
- Removed all `createMockContext()` custom implementations
- Renamed `src/helpers/__tests__/fixtures.ts` to `src/helpers/__tests__/testing-tools.ts`
- Inlined all `makeCli()` helpers, which were superfluous and just adding indirection
- Used `createCommand()` when passing a mocked stdout `write` function was needed at least once in the file. ([example](https://github.com/DataDog/datadog-ci/blob/0944aea9bb1f61ad2a05faaf6978a367ba61d6e0/src/commands/sourcemaps/__tests__/upload.test.ts#L107))
- Removed many `any` by fixing typing: went **from 2551 to 1805 warnings** in total

As a result, a lot of boilerplate code was removed: image

### How?

Create helpers and make sure they are used everywhere.

### Review checklist

- [x] Feature or bugfix MUST have appropriate tests (unit, integration)

PR #1620, opened 2025-04-18
[chores] Fix lint and `.gitignore` file

Drarig29:
### What and why?

- https://github.com/DataDog/datadog-ci/pull/1620 (← **you are here**)
- https://github.com/DataDog/datadog-ci/pull/1622
- https://github.com/DataDog/datadog-ci/pull/1623
- https://github.com/DataDog/datadog-ci/pull/1624

This PR fixes minor things in various configurations.

### How?

Config changes

### Review checklist

- ~~[ ] Feature or bugfix MUST have appropriate tests (unit, integration)~~

PR #1624, opened 2025-04-18
[ci] Run unit tests on Windows

Drarig29:
### What and why?

- https://github.com/DataDog/datadog-ci/pull/1620
- https://github.com/DataDog/datadog-ci/pull/1622
- https://github.com/DataDog/datadog-ci/pull/1623
- https://github.com/DataDog/datadog-ci/pull/1624 (← **you are here**)

Closes https://github.com/DataDog/datadog-ci/issues/1577

This PR makes sure we run all our unit tests in the CI on Windows to catch regressions on that OS as well.

The goal being to have a consistent experience between Unix and Windows, where all `\` are `/`.

### How?

- Use [`upath`](https://www.npmjs.com/package/upath) to translate paths to a universal format
- Wrap `glob` into `globSync()` and `globAsync()` helpers which return `upath`-normalized paths.
  - Using `glob`'s `posix: true` option was not enough: it would return redundant `//?/` for absolute paths (https://github.com/isaacs/node-glob/issues/568), causing [this kind of failures](https://github.com/DataDog/datadog-ci/actions/runs/14520370304/job/40739604278).
  - Their behavior is tested in a `src/helpers/__tests__/glob.test.ts` file which purposely avoids mocking `fs` to test in a real world setup.
- Introduce linting rules to give guidelines about supporting Windows

### Review checklist


src/helpers/__tests__/plist.test.ts
```diff
@@ -44,7 +44,7 @@ describe('plist util', () => {
 
     it('throws an error if the file does not exist', () => {
       expect(() => parsePlist('non-existing-file')).toThrow(
-        "ENOENT: no such file or directory, open 'non-existing-file'"
+        /ENOENT: no such file or directory, open '.*non-existing-file'/
```
Drarig29, line 47: For some reason, only in Windows, Node 20 and 22 now return the full path instead of only the filename: `ENOENT: no such file or directory, open 'D:\\a\\datadog-ci\\datadog-ci\\non-existing-file'`

BenoitZugmeyer (approved):

teodor2312 (approved):

nsavoire (approved):

TalUsvyatsky (approved):

src/commands/dsyms/__tests__/upload.test.ts
```diff
@@ -279,7 +279,7 @@ describe('execute', () => {
   test('Should succeed with folder input', async () => {
     const {context, code} = await runCLI('src/commands/dsyms/__tests__/fixtures/')
     const outputString = context.stdout.toString()
-    const output = outputString.split(EOL)
+    const output = outputString.split('\n')
```
fuzzybinary, line 282: Wouldn't `\n` be less correct on Windows?

fuzzybinary (approved):

src/commands/dsyms/__tests__/upload.test.ts
Drarig29: I think it's only the case for **_file_** line-endings, and not for command line `stdout`/`stderr`. Or maybe Node.js changes them from `\r\n` to `\n`.

When running the following on Windows:
```sh
require('child_process').exec('git --help', (error, stdout, stderr) => console.log(JSON.stringify({stdout})))
```

You get (on both `cmd` and Powershell, so I don't think it's related to shell that is used): 

image

And now that _we do_ run the tests in the CI on Windows, we know that `.split('\n')` works, while `.split(os.EOL)` doesn't work ([example of failing CI](https://github.com/DataDog/datadog-ci/actions/runs/14222643753/job/39854313102#step:5:4521)).

rjcoulter22 (approved):

liashenko (approved):

PR #1626, opened 2025-04-24
[SVLS] Add a couple tests for the lambda uninstrument command

TalUsvyatsky:
### What and why?

PR https://github.com/DataDog/datadog-ci/pull/1622 refactored unit tests across the codebase. In that PR, we discovered a couple of misconfigured tests for the lambda uninstrument command. To ease the refactor, we removed those tests. 

This PR adds those tests back in, now set up correctly.

### How?

Tests now:
- Use the `uninstrument` command instead of `instrument`
- Set up functions to have layers so they don't get skipped over for uninstrumentation

### Review checklist


src/commands/lambda/__tests__/uninstrument.test.ts
```diff
@@ -565,6 +565,141 @@ describe('lambda', () => {
...
+        lambdaClientMock.on(UpdateFunctionConfigurationCommand).rejects('Unexpected error updating request')
+
+        const {code, context} = await runCLI([
+          '-f',
+          'arn:aws:lambda:us-east-1:123456789012:function:lambda-1-us-east-1',
+          '-f',
+          'arn:aws:lambda:us-east-2:123456789012:function:lambda-1-us-east-2',
+        ])
+        expect(code).toBe(1)
+        expect(context.stdout.toString()).toMatchSnapshot()
```
avangelillo, line 701: Do we want to have the entire output be snapshotted, or are there relevant sections that we can search for in the output?  Usually when I've done things that check the entire output it makes any update change a lot more tests than would strictly be needed, but that may or may not be something we'd run into here

TalUsvyatsky: Yeah good point, I think for this test checking for `✖ Failed updating every Lambda function.` is all that matters. On the other test though, I think we want to see that it successfully instrumented a few functions and failed on the rest, which is pretty much the whole output.

Drarig29 (approved):

Issue #1638, opened 2025-05-15
SBOM upload with service name fails due to duplication


atreptow:
### Bug description

I understand that the service flag was removed from the cli tool and replaced with "datadog-ci" by default.
https://github.com/DataDog/datadog-ci/pull/1532

Context:
I have a mono repo with a lot of services and want to send the sbom to datadog. The version for different services is often the same, when their images were regenerated at the same time. The images though are different and can contain libraries specific to that service.

Problem:
On such a duplication for different images datadog-ci complains that a duplicate SBOM was detected. Apparently based on version, env and service name.
> An analysis has already been processed for sha:v1-34a45e1332 env:prod service:datadog-ci

I'm not able to determine if this is because of the static service definition or a deep analysis of the SBOM. this logic seems to be server side https://github.com/DataDog/datadog-ci/blob/master/src/commands/sbom/upload.ts#L172

Maybe a better description of why the report was rejected as a duplicate would help. If it is really because of the conflicting version, env and service then my question is how to pass it correctly?

### Describe what you expected

A hint on how to avoid the conflict with the static service name given by the CLI

### Additional context

Note: i can not select the sbom command for this bug

### Command

None

dastrong:
Hey @atreptow,

I agree that we need to update the messaging here.

The issue stems from us accepting one upload per commit.

To understand how we can help you better, could you let me know which scanning tool (us, Synk, etc) is generating your SBOM and which languages/ecosystems it contains.

atreptow:
Hi @dastrong,

thanks for your answer.

I use syft and grype to generate the SBOM and convert to cyclonedx. It mostly contains Go, but also Dart/Flutter, C++ and Python as language. So there are a variety of different dependencies from go libraries/sdk over ClamAV to wasm/js from flutter transpilation.

My main issue is that Datadog seems to reject uploads (apparently responding with 409) if the key from version, env, service is not unique. The repo has also some common folders that are shared between most services. If one of them is changed and this is the last change to get deployed, nearly all the scanned images will have the same version.

I've got a working solution for now, which is reverting to datadog-ci 2.48.0, because there i can define the service and will not get the conflicts. Going forward i'd just be interested on how to achieve this with a 3.x.x version.

dastrong:
Hello again @atreptow, 

> My main issue is that Datadog seems to reject uploads

Discussing with our team how to best handle this moving forward. 

> It mostly contains Go, but also Dart/Flutter, C++ and Python as language

[...]

PR #1647, opened 2025-05-22
[K9VULN-5404] Update duplicate upload notice

dastrong:
### What and why?

> Reported in https://github.com/DataDog/datadog-ci/issues/1638

The notice displayed when a duplicate upload is detected is outdated and not very helpful.

### How?

We're updated the text that's rendered when this duplicate upload error occurs to be more accurate and forthcoming with what actions are needed.


rjcoulter22 (approved):
This seems fine, but would it be possible to test this manually before merging?

Drarig29 (approved):

PR #1660, opened 2025-06-04
[SYNTH-17738] Fix ERR_STREAM_PREMATURE_CLOSE tunnel error

AntoineDona:
### What and why?

This PR fixes errors that we had with the synthetics tunnel, mainly `ERR_STREAM_PREMATURE_CLOSE`, which turned out to be harmless.

### How?

- Hide `ERR_STREAM_PREMATURE_CLOSE` logs during tunnel execution
- Fix errors during Tunnel shutdown. 

### Review checklist

### DIFF
--- a/src/commands/synthetics/tunnel/tunnel.ts
+++ b/src/commands/synthetics/tunnel/tunnel.ts
@@ -120,10 +120,6 @@ export class Tunnel {
       }
     })
 
-    if (this.multiplexer) {
-      this.multiplexer.close()
-    }
-
     await this.ws.close()
   }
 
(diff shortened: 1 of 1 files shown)


teodor2312 (approved):

src/commands/synthetics/tunnel/tunnel.ts
```diff
@@ -120,10 +120,6 @@ export class Tunnel {
       }
     })
 
-    if (this.multiplexer) {
-      this.multiplexer.close()
-    }
```
Drarig29, line 125: Why don't we need to close the multiplexer here?

AntoineDona: The multiplexer is automatically closed when the websocket is closed, as it is built on top of it. Closing the multiplexer before the duplex ws seam to cause the error, although every stream of the multiplexed connection has been closed. I think a reason why is because the duplex is still opened in the duplex <-> multiplex pipeline, causing it to error out when the multiplexer closes first. The node streamed pipeline are just a bit too sensitive I think, and we don't really care at that point as we are done with the execution of tests and just want to close the tunnel.

PR #1022, opened 2023-08-06
[chore] Upgrade TypeScript and Jest

Drarig29:
- #1022 ←
- #1023 
---

### What and why?

There is no major upside of upgrading TypeScript. It's mostly to keep up-to-date, and benefit from newer completion and type-checking features.

The build time is roughly the same.

### How?

Most changes in this PR come from the fact that Jest has to be upgraded because of the new TS version.

This new Jest version has **better snapshots**, and it **fixes a long-standing issue about `jest.restoreAllMocks()`** (https://github.com/jestjs/jest/issues/13925).

### Review checklist

- [x] Feature or bugfix MUST have appropriate tests (unit, integration)

PR #1665, opened 2025-06-05
[SVLS-6923] create stub for aas instrumentation cli

ava-silver:
### What and why?

### How?

### Review checklist

### DIFF
--- a/src/commands/aas/README.md
+++ b/src/commands/aas/README.md
@@ -1,0 +1,40 @@
+You can use the CLI to instrument your Azure App Services with Datadog. The CLI enables instrumentation by modifying existing App Services' configuration and hence does *not* require redeployment. It is the quickest way to get started with Datadog serverless monitoring.
+
+You can also add the command to your CI/CD pipelines to enable instrumentation for *all* your serverless applications. Run the command *after* your normal serverless application deployment, so that changes made by the Datadog CLI command do not get overridden.
+
+## Installation
+
+To instrument your App Services using the `datadog-ci aas instrument` command, follow the instructions for a specific runtime listed below:
+
+| OS      | Runtime   | Documentation                                                                                                                |
+| ------- | --------- | ---------------------------------------------------------------------------------------------------------------------------- |
+| Windows | .NET      | [Windows .NET setup](https://docs.datadoghq.com/serverless/azure_app_services/azure_app_services_windows?tab=net#setup)      |
+| Windows | Java      | [Windows Java setup](https://docs.datadoghq.com/serverless/azure_app_services/azure_app_services_windows?tab=java#setup)     |
+| Windows | Node      | [Windows Node setup](https://docs.datadoghq.com/serverless/azure_app_services/azure_app_services_windows?tab=nodejs#setup)   |
+| Linux   | .NET      | [Linux .NET setup](https://docs.datadoghq.com/serverless/azure_app_services/azure_app_services_linux?tab=nodenetphppython)   |
+| Linux   | Node      | [Linux Node setup](https://docs.datadoghq.com/serverless/azure_app_services/azure_app_services_linux?tab=nodenetphppython)   |
+| Linux   | PHP       | [Linux PHP setup](https://docs.datadoghq.com/serverless/azure_app_services/azure_app_services_linux?tab=nodenetphppython)    |
+| Linux   | Java      | [Linux Java setup](https://docs.datadoghq.com/serverless/azure_app_services/azure_app_services_linux?tab=java)               |
+| Linux   | Python    | [Linux Python setup](https://docs.datadoghq.com/serverless/azure_app_services/azure_app_services_linux?tab=nodenetphppython) |
+| Linux   | Container | [Linux Container setup](https://docs.datadoghq.com/serverless/azure_app_services/azure_app_services_container)               |
+
+
+## Commands
+
+### `instrument`
+
+Run `datadog-ci aas instrument` to apply Datadog instrumentation to an App Service. This command adds the Datadog Lambda Library and/or the Datadog Lambda Extension as Lambda Layers to the instrumented App Services and modifies their configurations.
+
+```bash
+# Instrument a function by subscription ID, resource group, and app service name
+datadog-ci aas instrument -s <subscription-id> -r <resource-group-name> -w <app-service-name> -v 81 -e 49
+
+# Instrument multiple functions specified by full resource IDs
+datadog-ci aas instrument -f <lambda-arn> -f <another-lambda-arn> -f <a-third-lambda-arn> -v 81 -e 49
+
+# Instrument multiple functions that match a regex pattern
+datadog-ci aas instrument --functions-regex <valid-regex-pattern> -r us-east-1 -v 81 -e 49
+
+# Dry run of all updates
+datadog-ci aas instrument -f <function-name> -f <another-function-name> -r us-east-1 -v 81 -e 49 --dry-run
+```
--- a/src/commands/aas/common.ts
+++ b/src/commands/aas/common.ts
@@ -1,0 +1,46 @@
+import {Command, Option} from 'clipanion'
+import {AasConfigOptions} from './interfaces'
+import {DEFAULT_CONFIG_PATHS, resolveConfigFromFile} from '../../helpers/utils'
+import {dryRunTag} from '../../helpers/renderer'
+
+export abstract class AasCommand extends Command {
+  private dryRun = Option.Boolean('-d,--dry,--dry-run', false)
+  private subscriptionId = Option.String('-s,--subscription-id')
+  private resourceGroup = Option.String('-g,--resource-group')
+  private aasName = Option.String('-n,--name')
+  private configPath = Option.String('--config')
+
+  public get dryRunPrefix(): string {
+    return this.dryRun ? dryRunTag + ' ' : ''
+  }
+
+  async ensureConfig(): Promise<[AasConfigOptions, string[]]> {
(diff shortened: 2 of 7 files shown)


src/commands/aas/common.ts
```diff
@@ -0,0 +1,114 @@
...
+
+  return arr
+}
+
+export const equal = (x: unknown, y: unknown): boolean => {
+  return x && y && typeof x === 'object' && typeof x === typeof y
+    ? Object.keys(x).length === Object.keys(y).length &&
+        Object.keys(x).every((key) => equal((x as Record<string, unknown>)[key], (y as Record<string, unknown>)[key]))
+    : x === y
+}
```
avangelillo: Is there a deep equal function that already exists that we can use?  This probably handles all the cases we're likely to see, but especially in javascript things can get weird.  One instance that comes to mind is circular references

```
> a = {}
{}
> b = {}
{}
> a.someKey = b
{}
> b.someKey = a
<ref *1> { someKey: { someKey: [Circular *1] } }
> a
<ref *1> { someKey: { someKey: [Circular *1] } }
> b
<ref *1> { someKey: { someKey: [Circular *1] } }
> equal(a,b)
Uncaught RangeError: Maximum call stack size exceeded
    at REPL6:4:39
    at Array.every (<anonymous>)
    at equal (REPL6:4:24)
    at REPL6:4:39
    at Array.every (<anonymous>)
    at equal (REPL6:4:24)
    at REPL6:4:39
    at Array.every (<anonymous>)
    at equal (REPL6:4:24)

```

As well as arrays being objects
```
>  typeof []
'object'
```
Which causes
```
> equal([{}], [[]])
true
```

src/commands/aas/instrument.ts
```diff
@@ -0,0 +1,140 @@
...
+    this.enableFips()
+    const [config, errors] = await this.ensureConfig()
+    if (errors.length > 0) {
+      for (const error of errors) {
+        this.context.stdout.write(renderError(error))
+      }
+
+      return 1
+    }
+    const cred = new DefaultAzureCredential()
```
avangelillo, line 29: Does Azure have profiles for credentials?  I'm only familiar with AWS so if this isn't relevant feel free to say that, but using the default credentials for AWS is a fairly common cause of accidentally running things against environments that you don't intend to

src/commands/aas/instrument.ts
```diff
@@ -0,0 +1,140 @@
...
+
+      return 1
+    }
+    this.context.stdout.write(`${this.dryRunPrefix}🐶 Instrumenting Azure App Service\n`)
+    const client = new WebSiteManagementClient(cred, config.subscriptionId)
+
+    const siteConfig = await client.webApps.getConfiguration(config.resourceGroup, config.aasName)
+    if (siteConfig.kind && !siteConfig.kind.toLowerCase().includes('linux')) {
+      this.context.stdout.write(
+        renderSoftWarning(
```
avangelillo, line 51: What does "soft" warning mean here?

src/commands/aas/README.md
```diff
@@ -0,0 +1,34 @@
...
+| Linux   | Java      | [Linux Java setup](https://docs.datadoghq.com/serverless/azure_app_services/azure_app_services_linux?tab=java)               |
+| Linux   | Python    | [Linux Python setup](https://docs.datadoghq.com/serverless/azure_app_services/azure_app_services_linux?tab=nodenetphppython) |
+| Linux   | Container | [Linux Container setup](https://docs.datadoghq.com/serverless/azure_app_services/azure_app_services_container)               |
+
+
+## Commands
+
+### `instrument`
+
+Run `datadog-ci aas instrument` to apply Datadog instrumentation to an App Service. This command adds the Datadog Lambda Library and/or the Datadog Lambda Extension as Lambda Layers to the instrumented App Services and modifies their configurations.
```
TalUsvyatsky: nit: looks like this has some leftover lambda wording

src/commands/aas/instrument.ts
ava-silver: It renders it as a yellow warning, indicating that its an easily fixable problem -- at least that's what ive seen in other usages

src/commands/aas/common.ts
ava-silver: I was really hoping there was an existing one, but lodash isnt used here and we don't explicitly import [fast-deep-equal](https://www.npmjs.com/package/fast-deep-equal), I'm not sure what the etiquette here is but I'd be happy to update the dependencies to add a hard dependency on that instead

tsconfig.json
```diff
@@ -10,6 +10,7 @@
     "noImplicitAny": true,
     "moduleResolution": "node",
     "baseUrl": "./",
+    "skipLibCheck": true,
```
duncanista: Why is this change required?

ava-silver: Problems with the azure SDK that we can't fix:
https://github.com/DataDog/datadog-ci/actions/runs/15541322183/job/43752571133?pr=1665

ava-silver: If there's a preferred way to scope this to just `@azure/*` I'm happy to update it to that!

src/commands/aas/common.ts
duncanista: We already have `deep-object-diff`, would that work for you @ava-silver?

src/commands/aas/common.ts
```diff
@@ -0,0 +1,114 @@
...
+    }
+    if (!this.aasName) {
+      errors.push('App Service (--name) is required')
+    }
+
+    return [config as AasConfigOptions, errors]
+  }
+}
+
+export const collect = async <T>(it: PagedAsyncIterableIterator<T>): Promise<T[]> => {
```
duncanista: `collect` is not a very descriptive method, what does this do?

src/commands/aas/common.ts
```diff
@@ -0,0 +1,114 @@
...
+      errors.push('--subscription-id is required')
+    }
+    if (!this.resourceGroup) {
+      errors.push('--resource-group is required')
+    }
+    if (!this.aasName) {
+      errors.push('App Service (--name) is required')
+    }
+
+    return [config as AasConfigOptions, errors]
```
duncanista, line 96: Can't you type if in line 63?

src/commands/aas/common.ts
ava-silver: the signature pretty much says it all, but im happy to add a jsdoc if you think it would be useful, something like `collect collects a PagedAsyncIterableIterator into an array`

tsconfig.json
duncanista: Looks like that error expects something from node16, but this CLI should only work on node 18+, maybe worth checking if the CI is misconfigured?

src/commands/aas/common.ts
duncanista: Can't you just do an `Array.fromAsync(...)`?

src/commands/aas/common.ts
ava-silver: on line 63 its a `ValueOptional<AasConfigOptions>`, which is because there are multiple ways to populate the config, and the `ensureConfig` asserts that the values are no longer optional

this is maybe overkill since the only way to use the cli is via an explicit subscription, resource group, and app service name, but in the future we may want to have different options like a list of resource IDs, or a regex, or by tag or something else

given the size of this PR (and the command being in beta) I was going to defer that work for later, and take a closer look at config/validation later once i tackle that.

src/commands/aas/instrument.ts
ava-silver: Discussed offline -- this isn't a concern since they are explicitly specifying the subscription ID (equivalent to aws account) where the web app lives

additionally, web app names are actually globally unique so there can never be confusion

tsconfig.json
ava-silver: https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/7593#issuecomment-2784034284

hmm looks like someone is suggesting we could instead upgrade typescript, not sure which would be a bigger deal

edit: looks like [last time it was a pretty big change](https://github.com/DataDog/datadog-ci/pull/1022)

src/commands/aas/common.ts
ava-silver: hmm my IDE was complaining about that when I tried to use it:
```
Property 'fromAsync' does not exist on type 'ArrayConstructor'. 
Do you need to change your target library? 
Try changing the 'lib' compiler option to 'esnext' or later. ts(2550)
```

src/commands/aas/common.ts
ava-silver: Ahh [someone mentioned we can just use the dependency](https://dd.slack.com/archives/C0136D22VUZ/p1749240529504029?thread_ts=1749240357.885879&cid=C0136D22VUZ) i suggested

michaelcretzman (approved):

tsconfig.json
Drarig29: 👋 I created https://github.com/DataDog/datadog-ci/pull/1670 to unblock this PR

src/commands/aas/common.ts
Drarig29: `Array.fromAsync` was added in Node 22 so we can't use it in datadog-ci without a polyfill. Let's keep this function (and maybe rename it to `collectAsyncIterator`)

Drarig29 (approved):
LGTM once `skipLibCheck: true` is removed in favor of https://github.com/DataDog/datadog-ci/pull/1670

avangelillo (approved):

TalUsvyatsky (approved):

PR #1666, opened 2025-06-06
Implement getting PR info tags from CI providers

nikita-tkachenko-datadog:
### What and why?

Implements extracting PR info tags from CI providers and user-defined environment variables.
Updates CI spec tests fixtures with the latest changes.

### Review checklist


src/helpers/ci.ts
```diff
@@ -298,7 +328,6 @@ export const getCISpanTags = (): SpanTags | undefined => {
 
     if (GITHUB_BASE_REF) {
       // GITHUB_BASE_REF is defined if it's a pull_request or pull_request_target trigger
-      tags[GIT_BASE_REF] = GITHUB_BASE_REF
```
nikita-tkachenko-datadog, line 301: I have removed this tag altogether:
- it was not part of our CI spec (no other library was setting it)
- it was only set for GitHub
- it's value was the same as that of `git.pull_request.base_branch` (set below: `tags[GIT_PULL_REQUEST_BASE_BRANCH]`)

src/commands/gate/evaluate.ts
```diff
@@ -104,7 +104,7 @@ export class GateEvaluateCommand extends Command {
 
     const api = this.getApiHelper()
     const spanTags = await this.getSpanTags()
-    const headRef = spanTags[GIT_BASE_REF]
+    const headRef = spanTags[GIT_PULL_REQUEST_BASE_BRANCH]
```
nikita-tkachenko-datadog, line 107: Replaced with another tag with identical value

PR #1672, opened 2025-06-11
Add LCOV format support to coverage upload command

nikita-tkachenko-datadog:
### What and why?

Adds LCOV format support to `coverage upload` command.

### How?

The code that detects and validates coverage report files is updated with lcov-specific logic.
The logic that discovers code coverage report files is refactored to be simpler and avoid some redundant work.

### Review checklist

### DIFF
--- a/src/helpers/__tests__/testing-tools.ts
+++ b/src/helpers/__tests__/testing-tools.ts
@@ -125,7 +125,7 @@ export type ConstructorOf<T> = new (...args: unknown[]) => T
  */
 export const createCommand = <T extends Command>(
   commandClass: ConstructorOf<T>,
-  context: {stdout?: {write: () => void}; stderr?: {write: () => void}} = {},
+  context: {stdout?: {write: () => void}; stderr?: {write: () => void}} = {}
 ) => {
   // Create a new instance of commandClass and pass in the parameters
   const command = new commandClass()


src/commands/coverage/upload.ts
```diff
@@ -126,10 +108,6 @@ export class UploadCodeCoverageReportCommand extends Command {
   private skipGitMetadataUpload = Option.Boolean('--skip-git-metadata-upload', false)
   private gitRepositoryURL = Option.String('--git-repository-url')
 
-  private automaticReportsDiscovery = Option.String('--auto-discovery', 'true', {
-    validator: t.isBoolean(),
-    tolerateBoolean: true,
-  })
```
juan-fernandez, line 132: isn't this a breaking change? is this intended? The `coverage` command is no longer a beta command, is it?

src/helpers/__tests__/testing-tools.ts
```diff
@@ -120,19 +120,26 @@ export type ConstructorOf<T> = new (...args: unknown[]) => T
...
+): T => {
   // Create a new instance of commandClass and pass in the parameters
   const command = new commandClass()
   command.context = {...createMockContext(), ...context} as BaseContext
 
   resolveCommandOptionsDefaults(command)
 
+  if (patch) {
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-return,@typescript-eslint/no-unsafe-argument
+    return Object.assign(Object.create(Object.getPrototypeOf(command)), command, patch)
```
juan-fernandez: why is this needed?

juan-fernandez (commented):
mostly worried about the potential breaking change. Otherwise LGTM

src/commands/coverage/__tests__/upload.test.ts
```diff
@@ -23,15 +23,42 @@ describe('upload', () => {
...
-      })
+      const command = createCommand(
+        UploadCodeCoverageReportCommand,
+        {},
+        {
+          basePaths: ['src/commands/coverage/__tests__/fixtures'],
+          config: {},
+          context,
+        }
+      )
```
Drarig29: You should be able to simplify this with something like:
```js
const command = createCommand(UploadCodeCoverageReportCommand, context)
command['basePaths'] = ['src/commands/coverage/__tests__/fixtures']
command['config'] = {} // only if necessary
```

src/commands/coverage/upload.ts
nikita-tkachenko-datadog: This is only being used for dogfooding as of now (the backend filters the list of orgs allowed to post to coverage intake), so I think it's acceptable to drop this before we have a customer using the command.

src/commands/coverage/__tests__/upload.test.ts
nikita-tkachenko-datadog: Much nicer, thanks!

src/helpers/__tests__/testing-tools.ts
nikita-tkachenko-datadog: The previous way of calling a method inside the command in these tests was:
```
command['getMatchingCoverageReportFilesByFormat'].call({
        basePaths: ['src/commands/coverage/__tests__/fixtures'],
        config: {},
        context,
      })
```
The problem was with binding `this` to the object passed to `call` - the method called that way could not invoke any other methods in the command, as the object used as `this` did not have them.

I came up with this awkward workaround to change the `__proto__` of the object, but @Drarig29 suggested a much simpler solution below

PR #1673, opened 2025-06-11
[SVLS-6923] fix sidecar application settings

ava-silver:
### What and why?

Fixes an issue where settings would not be properly set in the sidecar, since the values are not actually the values, but rather keys to the application settings keys 

### How?

Fixes the check and creation to be in the correct format. Validated and was finally able to see traces in datadog

### Review checklist

### DIFF
--- a/src/commands/aas/instrument.ts
+++ b/src/commands/aas/instrument.ts
@@ -99,8 +99,8 @@ https://docs.datadoghq.com/serverless/azure_app_services/azure_app_services_wind
     const sidecarContainer = siteContainers.find((c) => c.name === SIDECAR_CONTAINER_NAME)
     const envVars = this.getEnvVars(config)
     // We need to ensure that the sidecar container is configured correctly, which means checking the image, target port,
-    // and environment variables. The environment variables are must have the matching names and values as the
-    // keys in the App Settings, since the value indicates the name of the environment variable.
+    // and environment variables. The sidecar environment variables must have matching names and values, as the sidecar
+    // env values point to env keys in the main App Settings. (essentially env var forwarding)
     if (
       sidecarContainer === undefined ||
       sidecarContainer.image !== SIDECAR_IMAGE ||
--- a/src/commands/aas/instrument.ts
+++ b/src/commands/aas/instrument.ts
@@ -1,7 +1,7 @@
-import {SiteContainer, StringDictionary, WebSiteManagementClient} from '@azure/arm-appservice'
+import {StringDictionary, WebSiteManagementClient} from '@azure/arm-appservice'
 import {DefaultAzureCredential} from '@azure/identity'
 import chalk from 'chalk'
-import {Command, Option} from 'clipanion'
+import {Command} from 'clipanion'
 import equal from 'fast-deep-equal/es6'
 
 import {renderError, renderSoftWarning} from '../../helpers/renderer'
--- a/src/commands/aas/instrument.ts
+++ b/src/commands/aas/instrument.ts
@@ -1,8 +1,8 @@
-import {StringDictionary, WebSiteManagementClient} from '@azure/arm-appservice'
+import {SiteContainer, StringDictionary, WebSiteManagementClient} from '@azure/arm-appservice'
 import {DefaultAzureCredential} from '@azure/identity'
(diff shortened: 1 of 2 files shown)


ava-silver:
* **#1673** <a href="https://app.graphite.dev/github/pr/DataDog/datadog-ci/1673?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/DataDog/datadog-ci/1673?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

src/commands/aas/instrument.ts
```diff
@@ -117,7 +118,7 @@ https://docs.datadoghq.com/serverless/azure_app_services/azure_app_services_wind
           image: SIDECAR_IMAGE,
           targetPort: SIDECAR_PORT,
           isMain: false,
-          environmentVariables: Object.entries(envVars).map(([name, value]) => ({name, value})),
+          environmentVariables: Object.keys(envVars).map((name) => ({name, value: name})),
```
avangelillo, line 121: Is this meant to be a mapping between sets of environment variables or something?  I'm not sure I understand what the name / value do on this

avangelillo (approved):

src/commands/aas/instrument.ts
ava-silver: yep, i added a [comment above](https://github.com/DataDog/datadog-ci/pull/1673/files#diff-9b49a1e4de0f131fd68960b72306312b227caebe6627c08689660f10bd889d7dR101-R103) to clarify, it confused me too 😅

src/commands/aas/instrument.ts
```diff
@@ -98,14 +98,15 @@ https://docs.datadoghq.com/serverless/azure_app_services/azure_app_services_wind
     const siteContainers = await collectAsyncIterator(client.webApps.listSiteContainers(resourceGroup, aasName))
     const sidecarContainer = siteContainers.find((c) => c.name === SIDECAR_CONTAINER_NAME)
     const envVars = this.getEnvVars(config)
+    // We need to ensure that the sidecar container is configured correctly, which means checking the image, target port,
+    // and environment variables. The environment variables are must have the matching names and values as the
```
Drarig29: **nit** I just rephrased a bit because it was difficult to understand (I hope I kept the meaning of it)
```suggestion
    // and environment variables. The sidecar environment variables must have matching names and values, and they point to keys in the App Settings.
```

src/commands/aas/instrument.ts
```diff
@@ -98,14 +98,15 @@ https://docs.datadoghq.com/serverless/azure_app_services/azure_app_services_wind
...
+    // keys in the App Settings, since the value indicates the name of the environment variable.
     if (
       sidecarContainer === undefined ||
       sidecarContainer.image !== SIDECAR_IMAGE ||
       sidecarContainer.targetPort !== SIDECAR_PORT ||
-      !equal(
-        Object.fromEntries(sidecarContainer.environmentVariables?.map(({name, value}) => [name, value]) ?? []),
-        envVars
-      )
+      !sidecarContainer.environmentVariables?.every(({name, value}) => name === value) ||
```
Drarig29, line 108: Is it really possible that `name !== value`, or is it just an assertion? I feel like it's assumed that `name === value` should always be true since we create the sidecar, but it feels weird that in the opposite case we fallback to the `else` case, saying "Sidecar already exists with correct configuration".

I guess it could be clearer to fail early when `!sidecarContainer.environmentVariables?.every(({name, value}) => name === value)`, WDYT?

src/commands/aas/instrument.ts
ava-silver: gonna rephrase slightly, but i do like the clarification

src/commands/aas/instrument.ts
ava-silver: the customer could change things manually, which would break instrumentation (we have no guarantees) -- if not every environment variable is set to forward the main app's env vars, we should fix it for the customer.

Drarig29 (approved):

PR #1678, opened 2025-06-12
[SVLS-7006] add api key validation

ava-silver:
### What and why?

Adds API key validation to the `aas instrument` CLI to check for a valid API key before adding it to the instrumentation.

### How?

Calls the `/api/v1/validate` endpoint to ensure the key is valid

### Review checklist

### DIFF
--- a/src/commands/aas/instrument.ts
+++ b/src/commands/aas/instrument.ts
@@ -4,12 +4,12 @@ import chalk from 'chalk'
 import {Command, Option} from 'clipanion'
 import equal from 'fast-deep-equal/es6'
 
+import {newApiKeyValidator} from '../../helpers/apikey'
 import {renderError, renderSoftWarning} from '../../helpers/renderer'
+import {maskString} from '../../helpers/utils'
 
 import {AasCommand, collectAsyncIterator, SIDECAR_CONTAINER_NAME, SIDECAR_IMAGE, SIDECAR_PORT} from './common'
 import {AasConfigOptions} from './interfaces'
-import {newApiKeyValidator} from '../../helpers/apikey'
-import {maskString} from '../../helpers/utils'
 
 export class InstrumentCommand extends AasCommand {
   public static paths = [['aas', 'instrument']]
@@ -45,6 +45,7 @@ export class InstrumentCommand extends AasCommand {
           )}\nEnsure you copied the value and not the Key ID.`
         )
       )
+
       return 1
     }
     const cred = new DefaultAzureCredential()
--- a/src/commands/aas/instrument.ts
+++ b/src/commands/aas/instrument.ts
@@ -124,28 +124,6 @@ https://docs.datadoghq.com/serverless/azure_app_services/azure_app_services_wind
     return envVars
   }
 
-  /**
-   * Validates the Datadog API key in the env by making a request to the Datadog API.
-   * @returns {boolean} - Returns true if the API key is valid, false otherwise.
-   */
-  public async validateAPIKey(): Promise<boolean> {
-    // Validate the Datadog API key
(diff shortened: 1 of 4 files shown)


src/commands/aas/instrument.ts
```diff
@@ -30,6 +30,26 @@ export class InstrumentCommand extends AasCommand {
...
+      method: 'GET',
+      headers: {
+        Accept: 'application/json',
+        'DD-API-KEY': apiKey,
+      },
+    })
+    // no-dd-sa:typescript-best-practices/no-explicit-any
+    const data: any = await response.json()
+    if (data?.valid !== true) {
+      const censoredKey =
```
duncanista: There is already a method for this in other commands, might want to reuse that

src/commands/aas/instrument.ts
```diff
@@ -30,6 +30,26 @@ export class InstrumentCommand extends AasCommand {
...
+        apiKey.length < 4 ? '(too short to display)' : '*'.repeat(apiKey.length - 4) + apiKey.slice(-4)
+      this.context.stdout.write(
+        renderSoftWarning(`Invalid API Key ${censoredKey}, ensure you copied the value and not the Key ID`)
+      )
+
```
duncanista: Looks like this whole thing could be a method

src/commands/aas/instrument.ts
```diff
@@ -30,6 +30,26 @@ export class InstrumentCommand extends AasCommand {
 
       return 1
     }
+    // Validate the Datadog API key
+    const apiKey = process.env.DD_API_KEY!
+    const response = await fetch(`https://api.${process.env.DD_SITE ?? 'datadoghq.com'}/api/v1/validate`, {
```
avangelillo: Similar comment to Jordan's, but is there already a validate API key function somewhere?  I assume other commands do the same thing

ava-silver: there is a validate helper, but its API design doesn't make it reusable, since it expects you already called some other DD api first 😢

ava-silver: https://github.com/DataDog/datadog-ci/blob/db96c742493257e5dddb80f61e6375974788c54a/src/helpers/apikey.ts#L44

ava-silver: gonna update the interface to also have the boolean method instead

Drarig29: > there is a validate helper, but its API design doesn't make it reusable, since it expects you already called some other DD api first 😢

The "other DD api" is an API endpoint to validate the keys, so it's pretty reusable 😁 

https://github.com/DataDog/datadog-ci/blob/db96c742493257e5dddb80f61e6375974788c54a/src/helpers/apikey.ts#L59

Were you concerned about the two API calls for performance?

ava-silver: yep, and I don't actually need to call a datadog api in this command at all -- this is validating for future use in the customer's environment

ava-silver: i've pushed some changes, let me know what y'all think!

src/commands/aas/instrument.ts
```diff
@@ -30,6 +32,22 @@ export class InstrumentCommand extends AasCommand {
 
       return 1
     }
+    if (
+      !(await newApiKeyValidator({
+        apiKey: process.env.DD_API_KEY,
+        datadogSite: process.env.DD_SITE ?? 'datadoghq.com',
+      }).validateApiKey())
```
duncanista: move this to a variable `is_valid_api_key` for readability, and maybe use the default side constant for `'datadoghq.com'`

duncanista (approved):

PR #1680, opened 2025-06-12
[K9VULN-5872] Send exclusion properties as a dependency field

piloulacdog:
### What and why?

Our current scans of Maven projects do it by analyzing `pom.xml` files and resolving transitive dependencies down the road. However, it does not currently support the Maven `<exclusions>` mechanism ([doc](https://maven.apache.org/guides/introduction/introduction-to-optional-and-excludes-dependencies.html)), which allows projects to intentionally omit specific transitive dependencies.
As a result, our SCA scanning includes dependencies that are explicitly excluded by the project, leading to inaccurate or misleading outputs. 

### How?

We now set as a property the detected exclusions for a given package, we should send it to the API as a property. 

We are reading:
```json
{
    "properties": [
        { "name": "datadog-sbom-generator:exclusion", "value": "com.github.jnr:jnr-ffi"}, // this
        { "name": "datadog-sbom-generator:exclusion", "value": "com.github.jnr:jnr-posix"}, // this
        { "name": "datadog-sbom-generator:exclusion", "value": "io.dropwizard.metrics:metrics-core"}, // this
        { "name": "datadog-sbom-generator:exclusion", "value": "io.netty:netty-handler"}, // this
        { "name": "osv-scanner:is-direct", "value": "true"},
        { "name": "osv-scanner:package-manager", "value": "Maven" }
    ]
}
```
and converting it to:
```json
{
    "exclusions": [
        "com.github.jnr:jnr-ffi",
        "com.github.jnr:jnr-posix",
        "io.dropwizard.metrics:metrics-core",
        "io.netty:netty-handler"
    ]
}
```
### Review checklist

### DIFF
--- a/src/commands/sbom/constants.ts
+++ b/src/commands/sbom/constants.ts
@@ -4,5 +4,6 @@ export const PACKAGE_MANAGER_PROPERTY_KEY = 'osv-scanner:package-manager'
 export const IS_DEPENDENCY_DIRECT_PROPERTY_KEY = 'osv-scanner:is-direct'
 export const IS_DEPENDENCY_DEV_ENVIRONMENT_PROPERTY_KEY = 'osv-scanner:is-dev'
 export const FILE_PACKAGE_PROPERTY_KEY = 'osv-scanner:package'
+export const EXCLUSION_KEY = 'datadog-sbom-generator:exclusion'
 
 export const REACHABLE_SYMBOL_LOCATION_KEY_PREFIX = 'datadog-sbom-generator:reachable-symbol-location'
--- a/src/commands/sbom/__tests__/fixtures/sbom-with-exclusion.json
+++ b/src/commands/sbom/__tests__/fixtures/sbom-with-exclusion.json
@@ -1,0 +1,44 @@
+{
+  "$schema": "http://cyclonedx.org/schema/bom-1.5.schema.json",
+  "bomFormat": "CycloneDX",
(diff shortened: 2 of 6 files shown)


src/commands/sbom/constants.ts
```diff
@@ -4,5 +4,6 @@ export const PACKAGE_MANAGER_PROPERTY_KEY = 'osv-scanner:package-manager'
 export const IS_DEPENDENCY_DIRECT_PROPERTY_KEY = 'osv-scanner:is-direct'
 export const IS_DEPENDENCY_DEV_ENVIRONMENT_PROPERTY_KEY = 'osv-scanner:is-dev'
 export const FILE_PACKAGE_PROPERTY_KEY = 'osv-scanner:package'
+export const EXCLUSION_KEY = 'datadog-sbom-generator:exclusion'
```
rjcoulter22, line 7: nit: should this be `exclusions` if we have multiple? If its annoying to change b/c its baked into the sbom generator its fine to keep just throwing it out there

rjcoulter22 (approved):
Approved with comment

Drarig29 (approved):
LGTM overall

src/commands/sbom/constants.ts
piloulacdog: actually each property will only contain on exclusion. If you have 2 exclusions, you will have 2 properties. that's why I keep it singular

PR #1682, opened 2025-06-12
RUM-8149 Fix fat dSYM processing

maxep:
### What and why?

dSYM bundle can include more than one DWARF file. The `dsyms upload` cmd processing was considering bundle with more than one DWARF as universal (multi-architecture), but that's not necessarily the case. When attempting to thin a single arch DWARF, the `lipo` cmd fails, the symbols are **not** uploaded and a warning msg is presented.

Reported in DataDog/dd-sdk-ios#2176

### How?

Never assume multi-architecture DWARF and instead fail gracefully when `lipo` fails. Bundles with multiple DWARF are split and multiple arch extracted to dedicated dSYM bundle.

This PR also include a refacto to use smaller functions and align naming.

### Review checklist

### DIFF
--- a/src/commands/dsyms/upload.ts
+++ b/src/commands/dsyms/upload.ts
@@ -268,8 +268,14 @@ export class UploadCommand extends Command {
           await promises.mkdir(upath.dirname(object), {recursive: true})
 
           try {
+            // Attempt to extract the single arch object file.
+            // We could use the `lipo -archs` command to get the list of archs,
+            // but it's more straightforward to just attempt to extract
+            // the single arch.
             await executeLipo(dwarf.object, dwarf.arch, object)
           } catch {
+            // The thinning using `lipo` failed, meaning the dSYM is already
+            // a single arch dSYM. Copy the object file to the new dSYM.
             await promises.copyFile(dwarf.object, object)
           }
 


src/commands/dsyms/renderer.ts
```diff
@@ -4,25 +4,25 @@ import upath from 'upath'
...
 
 export const renderConfigurationError = (error: Error) => chalk.red(`${ICONS.FAILED} Configuration error: ${error}.\n`)
 
 export const renderInvalidDsymWarning = (dSYMPath: string) =>
   chalk.yellow(`${ICONS.WARNING} Invalid dSYM file, will be skipped: ${dSYMPath}\n`)
 
-export const renderDSYMSlimmingFailure = (dSYM: Dsym, slice: ArchSlice) =>
-  chalk.yellow(`${ICONS.WARNING} Failed to export '${slice.arch}' arch (${slice.uuid}) from ${dSYM.bundlePath}\n`)
+export const renderDSYMSlimmingFailure = (dSYM: Dsym, dwarf: DWARF, error: Error) =>
+  chalk.yellow(`${ICONS.WARNING} Failed to export '${dwarf.arch}' arch (${dwarf.uuid}) from ${dSYM.bundle}: ${error}\n`)
```
mariedm, line 16: 👌

src/commands/dsyms/upload.ts
```diff
@@ -216,105 +218,143 @@ export class UploadCommand extends Command {
...
+        try {
+          const bundle = buildPath(output, `${dwarf.uuid}.dSYM`)
+          const object = buildPath(bundle, upath.relative(dsym.bundle, dwarf.object))
+
+          await promises.mkdir(upath.dirname(object), {recursive: true})
+
+          try {
+            await executeLipo(dwarf.object, dwarf.arch, object)
+          } catch {
+            await promises.copyFile(dwarf.object, object)
```
mariedm, line 279: What if the command fails for another reason than the file already being thin? Could there be a case where we don't want to copy the file?

mariedm (commented):
Just one question, otherwise LGMT! 👍

src/commands/dsyms/upload.ts
maxep: Good question, if the `dwarfdump` command succeed then we should upload the file. So if the `lipo` fails we just upload it as is. The `lipo` is our best effort to make the payload smaller, but should not prevent the upload.

ncreated: 💡 Maybe we can at least add a log message in this flow, so it's easier to troubleshoot?

ncreated (approved):
Looks good 👌. Well done with renaming, it fits the actual domain much better now 👍.

src/commands/dsyms/upload.ts
maxep: `lipo` failing is also a valid path, not an issue, it means that the dSYM includes multiple DWARF but is not a multi-arch. I should add a comment to clarify.

package.json
```diff
@@ -55,6 +55,7 @@
     "test:debug": "node --inspect-brk node_modules/jest/bin/jest --runInBand",
     "test:lambda": "jest src/commands/lambda --colors",
     "test:cloud-run": "jest src/commands/cloud-run --colors",
+    "test:dsyms": "jest src/commands/dsyms --colors",
```
AntoineDona, line 58: I think this is not really necessary, as you can run your tests like this:
```
yarn run test dsyms
```

AntoineDona (approved):
Looks good, just a small nit 👍

package.json
maxep: Ah good to know! 
I'm merging as-is to not go through the review loop again, also because other `test:*` exists.

PR #1683, opened 2025-06-12
Create Google Cloud Run `instrument` command skeleton; authenticate credentials

nhulston:
### What and why?

Create skeleton for `instrument` command and authenticate GCP credentials.

The `instrument` and `uninstrument` commands will have a really large diff, so I'm breaking up into many small PRs to make it easier to review.

I commented out `paths` so `datadog-ci cloud-run instrument` doesn't do anything right now.

### Review checklist


avangelillo (approved):

src/commands/cloud-run/instrument.ts
```diff
@@ -0,0 +1,60 @@
+import chalk from 'chalk'
+import {Command, Option} from 'clipanion'
+
+import {CloudRunConfigOptions} from './interfaces'
+import {renderAuthenticationInstructions, renderCloudRunInstrumentUninstrumentHeader} from './renderer'
+import {checkAuthentication} from './util'
+
+export class InstrumentCommand extends Command {
+  // TODO uncomment when commnand is ready
```
Drarig29: Can you add a TODO to update docs too? Especially https://github.com/DataDog/datadog-ci#cloud-run which could easily be forgotten

src/commands/cloud-run/cli.ts
```diff
@@ -1,3 +1,4 @@
 import {CloudRunFlareCommand} from './flare'
+import {InstrumentCommand} from './instrument'
 
-module.exports = [CloudRunFlareCommand]
+module.exports = [InstrumentCommand, CloudRunFlareCommand]
```
Drarig29: Since the command is a WIP, let's not add it to the list here. This way it won't be exposed to users

src/commands/cloud-run/util.ts
Drarig29, line 1: **nit** This file is named `utils.ts` across other commands, can you rename it? We all know it will grow and become plural anyway 😁

src/commands/cloud-run/instrument.ts
nhulston: Yes, great suggestion. Thanks

Drarig29 (approved):

PR #1691, opened 2025-06-13
Google Cloud Run Basic Instrument Command

nhulston:
### What and why?

### How?

### Review checklist

### DIFF
--- a/src/commands/cloud-run/instrument.ts
+++ b/src/commands/cloud-run/instrument.ts
@@ -7,7 +7,17 @@ import {google} from '@google-cloud/run/build/protos/protos'
 import chalk from 'chalk'
 import {Command, Option} from 'clipanion'
 
-import {DATADOG_SITE_US1} from '../../constants'
+import {
+  API_KEY_ENV_VAR,
+  DATADOG_SITE_US1,
+  ENVIRONMENT_ENV_VAR,
+  HEALTH_PORT_ENV_VAR,
+  LOGS_INJECTION_ENV_VAR,
+  LOGS_PATH_ENV_VAR,
+  SERVICE_ENV_VAR,
+  SITE_ENV_VAR,
+  VERSION_ENV_VAR,
+} from '../../constants'
 import {newApiKeyValidator} from '../../helpers/apikey'
 import {renderSoftWarning} from '../../helpers/renderer'
 import {maskString} from '../../helpers/utils'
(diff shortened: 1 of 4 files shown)


src/commands/cloud-run/__tests__/instrument.test.ts
```diff
@@ -0,0 +1,261 @@
...
+    let mockClient: Partial<ServicesClient>
+    const mockServicePath = 'projects/project/locations/region/services/service'
+
+    beforeEach(() => {
+      command = new InstrumentCommand()
+      // inject a fake context so we can spy on writes
+      command.context = {
+        stdout: {write: jest.fn()},
+        stderr: {write: jest.fn()},
+      } as any
```
Drarig29: I don't see those spies being used anywhere. Can you use `createCommand(InstrumentCommand)`? It does everything for you

src/commands/cloud-run/instrument.ts
```diff
@@ -55,6 +134,187 @@ export class InstrumentCommand extends Command {
...
+        {name: API_KEY_ENV_VAR, value: process.env.DD_API_KEY},
+        {name: HEALTH_PORT_ENV_VAR, value: '12345'},
+        {name: LOGS_INJECTION_ENV_VAR, value: 'true'},
+        {name: SERVICE_ENV_VAR, value: ddService},
+        ...(this.environment ? [{name: ENVIRONMENT_ENV_VAR, value: this.environment}] : []),
+        ...(this.version ? [{name: VERSION_ENV_VAR, value: this.version}] : []),
+      ],
+      startupProbe: {
+        tcpSocket: {
+          port: 12345,
```
avangelillo, line 242: Is 12345 a placeholder?  Also in the `HEALTH_PORT_ENV_VAR` just above

src/commands/cloud-run/instrument.ts
```diff
@@ -55,6 +134,187 @@ export class InstrumentCommand extends Command {
...
+
+    const service = await withSpinner(
+      `Fetching service configuration...`,
+      async () => {
+        try {
+          const [existingService] = await client.getService({name: servicePath})
+
+          return existingService
+        } catch (error) {
+          throw new Error(`Service ${serviceName} not found in project ${project}, region ${region}`)
```
avangelillo: Because this is in instrument service if someone passes a list of services that includes one that doesn't exist, some of the services that do exist could be instrumented while others aren't.  Is that fine or should we do the existence check before we start to instrument?

avangelillo (approved):

src/commands/cloud-run/instrument.ts
nhulston: Great catch. I think we should check the existence for all of them first, and exit without instrumenting any of them if there are invalid services

src/commands/cloud-run/instrument.ts
nhulston: Hmm, I was following [these docs](https://docs.datadoghq.com/serverless/google_cloud_run?tab=terraformdeploy) which use port 12345. But those docs are very outdated -- I'll reach out to the team and figure out if there's a better value or precedent we can use

avangelillo: It's entirely possible that 12345 is the right value for it, I don't actually know.  It just seems like the kind of thing that was a placeholder

src/commands/cloud-run/__tests__/instrument.test.ts
nhulston: Will do, thank you

src/commands/cloud-run/instrument.ts
nhulston: Yea I agree, it was probably just a placeholder/example port in the docs to indicate to users that they could use any value

nhulston: Okay, we're going to leave the default to `12345`, and then add a `--port` flag to configure it in a separate PR

PR #1697, opened 2025-06-17
Upload RN sourcemaps with debug-id

ImaneLargou:
### What and why?

Modifies the React Native Upload Command to extract the Debug ID from the sourcemap and add it to the multipart payload.

### Review checklist


src/commands/react-native/__tests__/upload.test.ts
```diff
@@ -19,6 +19,34 @@ describe('upload', () => {
...
+        'build'
+      )
+
+      // THEN
+      const event = payload.content.get('event')
+
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-unsafe-argument, @typescript-eslint/no-unsafe-member-access
+      const eventValue = JSON.parse((event as any)['value'])
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+      expect(eventValue['debug_id']).toBe('a422b269-0dba-4341-93c2-73e1bcf71fbb')
```
Drarig29: **nit** using the right types makes it more understandable IMO
```suggestion
const event = payload.content.get('event') as MultipartStringValue
const eventValue = JSON.parse(event['value']) as {debug_id: string}
expect(eventValue['debug_id']).toBe('a422b269-0dba-4341-93c2-73e1bcf71fbb')
```

src/commands/react-native/interfaces.ts
```diff
@@ -54,13 +57,33 @@ export class RNSourcemap {
     this.sourcemapPath = newSourcemapFilePath
   }
 
+  private extractDebugId = (): string | undefined => {
+    try {
+      const data = fs.readFileSync(this.sourcemapPath, 'utf8')
+      const sourcemap = JSON.parse(data)
+      const debugId = sourcemap['debugId']
+      if (debugId) {
+        console.info(`Extracted Debug ID from sourcemap: ${debugId}`)
```
Drarig29: Can you pass the `this.context` (the type is `CommandContext`) to `asMultipartPayload()`, and use `context.stdout.write`? We should avoid using `console` directly. (Note that `console.warn` writes to `stderr` so you have to use `context.stderr.write` to be equivalent)

I'm going to add an ESLint rule forbidding `console` soon

src/commands/react-native/__tests__/upload.test.ts
marco-saia-datadog: I totally agree!

src/commands/react-native/interfaces.ts
```diff
@@ -54,13 +67,34 @@ export class RNSourcemap {
...
+      if (debugId) {
+        context.stdout.write(`Extracted Debug ID from sourcemap: ${debugId}\n`)
+
+        return debugId
+      } else {
+        context.stderr.write(`Debug ID not found in sourcemap\n`)
+      }
+    } catch (err) {
+      const errorMsg = err instanceof Error ? err.message : String(err)
+      context.stderr.write(`Cannot extract Debug ID from sourcemap: ${errorMsg}`)
```
Drarig29:
```suggestion
      context.stderr.write(`Cannot extract Debug ID from sourcemap: ${errorMsg}\n`)
```

marco-saia-datadog: Good catch!

PR #1702, opened 2025-06-18
[React Native] Added inject-debug-id command for React Native

marco-saia-datadog:
### What and why?

Introduces a new `inject-debug-id` command, which adds a unique debug ID to JavaScript bundles and sourcemaps.

A debug ID allows Datadog to reliably match error stack traces from applications to the correct uploaded debug symbols (sourcemaps).

### How?

The command scans the given directory for `*.bundle` and `*.bundle.map` files.

If either the bundle or the sourcemap are found, the command will try to extract the Debug ID from it.

If the Debug ID is not present, the command will generate a new one.

After the Debug ID has been determined, the command will make sure that both the bundle and the sourcemap contain it.

Specifically:

- **In the sourcemap:** as a JSON top-level property `debugId`
- **In the JS Bundle:**
  - As a minified snippet of code, that allows our React Native SDK to retrieve it at runtime
  - As a comment at the end of the JS Bundle file, in the format: `//# debugId=${DEBUG_ID}` 

### Review checklist


src/commands/react-native/__tests__/inject-debug-id.test.ts
```diff
@@ -0,0 +1,183 @@
...
+
+const runCommand = async (directory: string) => {
+  const cli = makeCli()
+  const context = createMockContext() as any
+
+  const command = ['react-native', 'inject-debug-id', '--path', directory]
+  const code = await cli.run(command, context)
+
+  return {context, code}
+}
```
Drarig29: Can you use `makeRunCLI` from testing tools? This will replace your `runCommand`, `makeCli` and `createMockContext`.

src/commands/react-native/injectDebugId.ts
```diff
@@ -0,0 +1,258 @@
...
+        injecting a unique Debug ID into each file. These Debug IDs enable precise source map resolution in Datadog, ensuring accurate 
+        stack traces and error symbolication.
+    `,
+    examples: [
+      ['Inject Debug ID', 'datadog-ci react-native inject-debug-id --path ./dist'],
+      ['Inject Debug ID (dry run)', 'datadog-ci react-native inject-debug-id --path ./dist --dry-run'],
+    ],
+  })
+
+  private assetsPath = Option.String('--path', './dist')
```
Drarig29: Would it make sense to make this positional [like the `xcode` command](https://github.com/DataDog/datadog-ci/blob/1920e31f7b99bb2b0c584431d3ad2164395b6f5d/src/commands/react-native/xcode.ts#L81)?

src/commands/react-native/README.md
```diff
@@ -204,6 +204,22 @@ The following optional parameters are available:
...
+
+A debug ID allows Datadog to reliably match error stack traces from your application to the correct uploaded debug symbols (sourcemaps), ensuring accurate error resolution.
+
+To inject the debug ID, run the following command:
+
+```bash
+datadog-ci react-native inject-debug-id --path ./dist
+```
+
+The expected path is a folder containing both the JS bundle and the sourcemap for your build (e.g index.android.bundle and index.android.bundle.map).
```
buraizu:
```suggestion
The expected path is a folder containing both the JS bundle and the sourcemap for your build (for example, index.android.bundle and index.android.bundle.map).
```

src/commands/react-native/injectDebugId.ts
marco-saia-datadog: Yes :) seems cleaner that way. I have [modified it](https://github.com/DataDog/datadog-ci/compare/8ecec6958d3e2dfa9ee0baf9bc44151083354d91..ad7f1a97ee337193b9630830e26e818c4bb1f4f7#diff-0d132b2036e9693f9fd54d5fd946156294487fa5c3c4feef0f0b3b8a7f0ba841R32).

src/commands/react-native/__tests__/inject-debug-id.test.ts
```diff
@@ -0,0 +1,151 @@
...
+
+import {makeRunCLI} from '../../../helpers/__tests__/testing-tools'
+
+import {InjectDebugIdCommand} from '../injectDebugId'
+
+const tmpDir = upath.join(os.tmpdir(), 'inject-debug-id-tests')
+
+const runCommand = async (directory: string) => {
+  return makeRunCLI(InjectDebugIdCommand, ['react-native', 'inject-debug-id'])([directory])
+}
```
Drarig29: Sorry to be insistent on this, but I'd like unit tests to look similar in all commands so it's easier for repo-wide changes. In this case, the wrapper adds little benefit, so I suggest changing to:
```suggestion
const runCLI = makeRunCLI(InjectDebugIdCommand, ['react-native', 'inject-debug-id'])
```

src/commands/react-native/__tests__/inject-debug-id.test.ts
```diff
@@ -0,0 +1,151 @@
...
+    const fixturePath = './src/commands/react-native/__tests__/fixtures/sourcemap-with-no-files'
+
+    const tmpBundlePath = upath.join(tmpDir, 'test-1.js.bundle')
+    const tmpSourcemapPath = upath.join(tmpDir, 'test-1.js.bundle.map')
+
+    await promises.copyFile(upath.join(fixturePath, 'empty.min.js'), tmpBundlePath)
+    await promises.copyFile(upath.join(fixturePath, 'empty.min.js.map'), tmpSourcemapPath)
+
+    // WHEN
+    const {context, code} = await runCommand(tmpDir)
```
Drarig29:
```suggestion
    const {context, code} = await runCLI([tmpDir])
```

src/commands/react-native/__tests__/inject-debug-id.test.ts
marco-saia-datadog: No you are right, this is definitely better. I made a wrong assumption, and thought the context would be shared between different tests.

marco-saia-datadog: I have pushed the change now, thanks!

marco-saia-datadog:
@Drarig29 thank you very much for the reviews! :)

PR #1708, opened 2025-06-20
[chore] Bump `@google-cloud/run` to v2

nhulston:
### What and why?

Needed to get `instrument` working with Cloud Run functions.

I was getting Typescript build errors:
```
node_modules/@google-cloud/run/node_modules/google-auth-library/build/src/auth/jwtclient.d.ts:1:29 - error TS1479:
The current file is a CommonJS module whose imports will produce 'require' calls; however, the referenced file is an ECMAScript module and cannot be imported with 'require'.
Consider writing a dynamic 'import("gtoken")' call instead.
```

Since `gtoken` is [dual-shipped in ESM and CJS](https://github.com/googleapis/node-gtoken/blob/2c21cb7a8e9d4c8511000916af5a12cb268d83c0/package.json#L37-L44), but google-auth-library [is not](https://github.com/googleapis/google-auth-library-nodejs/blob/082c171cae021e184d68bf3c794fdf8f2eddbd94/package.json#L9-L10).
The workaround now is to temporarily use `require()` and lose all type annotations, until Google fixes this
(thanks for the help @Drarig29)

[More info here](https://dd.slack.com/archives/C0136D22VUZ/p1750439989678209)

### Review checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,7 +71,7 @@
     "@azure/arm-appservice": "^16.0.0",
     "@azure/identity": "^4.10.0",
     "@google-cloud/logging": "^11.1.0",
-    "@google-cloud/run": "^1.4.0",
+    "@google-cloud/run": "^2.1.0",
     "@smithy/property-provider": "^2.0.12",
     "@smithy/util-retry": "^2.0.4",
     "@types/datadog-metrics": "0.6.1",
--- a/src/commands/cloud-run/__tests__/instrument.test.ts
+++ b/src/commands/cloud-run/__tests__/instrument.test.ts
@@ -1,7 +1,5 @@
-import IContainer = google.cloud.run.v2.IContainer
-import IVolumeMount = google.cloud.run.v2.IVolumeMount
-
-import {google} from '@google-cloud/run/build/protos/protos'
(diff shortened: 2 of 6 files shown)


tsconfig.json
```diff
@@ -13,7 +13,8 @@
     "allowSyntheticDefaultImports": true,
     "useUnknownInCatchVariables": false,
     "esModuleInterop": true, // Allows default imports from modules with no default export
-    "resolveJsonModule": true // Used for SBOM json schemas
+    "resolveJsonModule": true, // Used for SBOM json schemas
+    "skipLibCheck": true
```
duncanista: Why is this change needed?

nhulston: I was getting a Typescript error from a file in node_modules

```
node_modules/@google-cloud/run/node_modules/google-auth-library/build/src/auth/jwtclient.d.ts:1:29 - error TS1479:
The current file is a CommonJS module whose imports will produce 'require' calls; however, the referenced file is an ECMAScript module and cannot be imported with 'require'.
Consider writing a dynamic 'import("gtoken")' call instead.
```

Let's see what they say in #datadog-ci

Drarig29 (approved):
LGTM as a temporary workaround 👍

PR #1709, opened 2025-06-20
[SVLS-6823] Support FIPS-compatible lambda functions

alexgallotta:
**What and why?**

Add FIPS support to deploy FIPS compatible lambda functions

**How?**

FIPS support requires different layers and configuration in a Lambda function, so the instrument/uninstrument functions have to be updated accordingly. When FIPS flag is set, change instrumentation:
- use the Lambda layer extension with the -FIPS suffix
- add the DD_LAMBDA_FIPS_MODE=true env variable

and clean up those changes on uninstrument
Since full FIPS compliance requires also a proper intake, at the moment only present in dd-gov, the readme states that dd-gov is not used


src/commands/lambda/functions/instrument.ts
```diff
@@ -333,7 +333,23 @@ export const calculateUpdateRequest = async (
     if (settings.interactive && !settings.extensionVersion) {
       extensionVersion = await findLatestLayerVersion(EXTENSION_LAYER_KEY as LayerKey, region)
     }
-    fullExtensionLayerARN = `${lambdaExtensionLayerArn}:${extensionVersion}`
+
+    let extensionLayerName = DD_LAMBDA_EXTENSION_LAYER_NAME
+    let extensionLayerArn = lambdaExtensionLayerArn
+    if (settings.fips) {
+      const currentSite = site ?? oldEnvVars[SITE_ENV_VAR] ?? 'datadoghq.com'
+      if (currentSite !== 'ddog-gov.com') {
```
ava-silver: nit: we should use the constants in `src/constants.ts` for us1 and fed

src/commands/lambda/interfaces.ts
```diff
@@ -69,6 +69,7 @@ export interface InstrumentationSettings extends InstrumentationTags {
   appsecEnabled?: boolean
   captureLambdaPayload?: boolean
   extensionVersion?: number
+  fips?: boolean
```
ava-silver: I've had the datadog linter complain about not starting boolean flags with `is` or `has`, should we call this `isFipsEnabled` maybe?

src/commands/lambda/functions/instrument.ts
```diff
@@ -333,7 +333,23 @@ export const calculateUpdateRequest = async (
...
+        console.warn('DD_SITE is not set to ddog-gov.com, FIPS will not be fully compliant')
+      }
+
+      if (oldEnvVars['DD_LAMBDA_FIPS_MODE'] !== 'true') {
+        needsUpdate = true
+        newEnvVars['DD_LAMBDA_FIPS_MODE'] = 'true'
+      }
+      extensionLayerName = `${DD_LAMBDA_EXTENSION_LAYER_NAME}-FIPS`
```
ava-silver:
```suggestion
      extensionLayerName += '-FIPS'
```

ava-silver (commented):
do we need any changes to `uninstrument`? or will that work properly with these changes?

alexgallotta:
> do we need any changes to `uninstrument`? or will that work properly with these changes?

oh right, forgot to push that, thanks!

src/commands/lambda/interfaces.ts
alexgallotta: There is a lot of other booleans without the is/has prefix, I would rather keep it consistent

src/commands/lambda/instrument.ts
```diff
@@ -510,6 +510,7 @@ export class InstrumentCommand extends Command {
       environment,
       extensionVersion,
       extraTags,
+      fips: this.fips || this.fipsConfig.fips,
```
apiarian-datadog: what is `fipsConfig`?

src/commands/lambda/functions/instrument.ts
```diff
@@ -333,7 +335,23 @@ export const calculateUpdateRequest = async (
...
+      const currentSite = site ?? oldEnvVars[SITE_ENV_VAR] ?? 'datadoghq.com'
+      if (currentSite !== DATADOG_SITE_GOV) {
+        console.warn('DD_SITE is not set to ddog-gov.com, FIPS will not be fully compliant')
+      }
+
+      if (oldEnvVars[DD_LAMBDA_FIPS_MODE_ENV_VAR] !== 'true') {
+        needsUpdate = true
+        newEnvVars[DD_LAMBDA_FIPS_MODE_ENV_VAR] = 'true'
+      }
+      extensionLayerName += '-FIPS'
```
apiarian-datadog: we have a function above which is called `getLayerArn`. can we move this logic in there? that way the `findLatestLayerVersion` function will find the latest layer version for the correct layer and not assume that we have correctly and consistently released the same version numbers to all environments (this has been "eventually consistent" in the past).

out of curiosity, where do we handle the amd/arm distinction?

src/commands/lambda/instrument.ts
alexgallotta: it's an object used to store FIPS config from env variables, initialized here
https://github.com/DataDog/datadog-ci/blob/master/src/commands/lambda/instrument.ts#L99-L102

src/commands/lambda/functions/instrument.ts
alexgallotta: good idea, I was trying to keep all the fips logic in the same place but it's better to avoid all this string manipulation

src/commands/lambda/instrument.ts
Drarig29: The `fips` option that was added to all commands is meant to enable FIPS mode in `datadog-ci` for traffic sent to Datadog: https://github.com/DataDog/datadog-ci#fips-support

And it seems like you are working on something else: something which instruments a lambda and sets `DD_LAMBDA_FIPS_MODE` in a layer.

So I feel like we shouldn't mix the two concepts and you should probably add a new option to your command, like `enableLambdaFipsMode`

src/commands/lambda/functions/instrument.ts
```diff
@@ -333,6 +335,18 @@ export const calculateUpdateRequest = async (
     if (settings.interactive && !settings.extensionVersion) {
       extensionVersion = await findLatestLayerVersion(EXTENSION_LAYER_KEY as LayerKey, region)
     }
+
+    if (settings.fips) {
+      const currentSite = site ?? oldEnvVars[SITE_ENV_VAR] ?? 'datadoghq.com'
+      if (currentSite !== DATADOG_SITE_GOV) {
+        console.warn('DD_SITE is not set to ddog-gov.com, FIPS will not be fully compliant')
```
Drarig29: > FIPS will not be fully compliant

I think we should rephrase this – it's not "FIPS" that would "not be fully compliant"

src/commands/lambda/instrument.ts
apiarian-datadog: i agree with @Drarig29 .

src/commands/lambda/functions/instrument.ts
apiarian-datadog: it's more like "the lambda instrumentation will be FIPS-compliant, but the data center is not"

src/commands/lambda/instrument.ts
ava-silver: I was potentially leaning the other way -- If a customer wants to ensure their infrastructure is fips compliant, forcing them to need two flags for this (one for the cli and one for the layer) feels like an easy way for customers to footgun themselves.

ava-silver: do either of @apiarian-datadog or @Drarig29 see a use case for using one option but not the other?

ava-silver (approved):
Looks good from my perspective, but we should resolve the discussion around the options (one vs two) before merging

src/commands/lambda/functions/commons.ts
```diff
@@ -343,11 +343,17 @@ export const getLayerArn = (
   }
   const account = settings?.layerAWSAccount ?? DEFAULT_LAYER_AWS_ACCOUNT
   const isGovCloud = region.startsWith('us-gov')
+  let arnBuilt
   if (isGovCloud) {
-    return `arn:aws-us-gov:lambda:${region}:${GOVCLOUD_LAYER_AWS_ACCOUNT}:layer:${layerName}`
+    arnBuilt = `arn:aws-us-gov:lambda:${region}:${GOVCLOUD_LAYER_AWS_ACCOUNT}:layer:${layerName}`
+  } else {
+    arnBuilt = `arn:aws:lambda:${region}:${account}:layer:${layerName}`
+  }
```
ava-silver: I might slightly favor a ternary here:
```suggestion
  let arnBuilt = isGovCloud
    ? `arn:aws-us-gov:lambda:${region}:${GOVCLOUD_LAYER_AWS_ACCOUNT}:layer:${layerName}`
    : `arn:aws:lambda:${region}:${account}:layer:${layerName}`
```

src/commands/lambda/functions/instrument.ts
Drarig29: Also, can you pass `context: CommandContext` to this function and replace this line with:
```suggestion
        context.stderr.write('FIPS mode was enabled on the extension layer, but DD_SITE is not set to ddog-gov.com\n')
```

src/commands/lambda/instrument.ts
Drarig29: Nevermind, I guess when a user is using GovCloud they are also using `ddog-gov.com` so let's keep one option. 

Can we add a message saying something like "Enabling FIPS mode on the extension layer"? Just to let the user know the `--fips` was taken into account for that as well

src/commands/lambda/functions/commons.ts
alexgallotta: I would rather keep it if/else if there is no strong opinion about it. The line is pretty long and do a bunch of concatenations, I think it's good to keep it more readable

src/commands/lambda/instrument.ts
alexgallotta: The team had the same discussion, it's a "separate options for clarity" vs "just one for simplicity" and we got to the same conclusion, not sure when FIPS would be ok in CI but not in lambda or vice versa.

Sounds good, I will add an info log for that, ty!

src/commands/lambda/__tests__/functions/instrument.part2.test.ts
```diff
@@ -728,6 +731,85 @@ describe('instrument', () => {
...
+      const settings = {
+        extensionVersion: 6,
+        fips: false,
+        flushMetricsToLogs: false,
+        layerAWSAccount: mockAwsAccount,
+        mergeXrayTraces: false,
+        tracingEnabled: false,
+        lambdaFips: true,
+      }
+      const region = 'sa-east-1'
```
apiarian-datadog: nit: maybe we should be using the proper gov regions and ddog-gov site in this test, to demonstrate the correctly supported combination?

src/commands/lambda/functions/commons.ts
```diff
@@ -343,11 +343,17 @@ export const getLayerArn = (
...
+  if (settings?.lambdaFips) {
+    arnBuilt += '-FIPS'
```
apiarian-datadog: nit: would this be cleaner as a layerName suffix instead?

apiarian-datadog (approved):
looks good. some nits, but otherwise seems fine. we chose not to include a warning when the fips mode, aws site, and datadog region are not in sync?

README.md
```diff
@@ -139,20 +160,28 @@ The option `fips-ignore-error` ignores this error.
...
 Ignore Node.js errors if FIPS cannot be enabled on the host system.
 
 **Note**: the absence of an error doesn't indicate that FIPS is enabled successfully.
 
 ENV variable: `DATADOG_FIPS_IGNORE_ERROR=true`
 CLI param: `--fips-ignore-error`
 
+#### `lambda-fips`
+
+This option is used to enable FIPS support in the lambda functions deployed using this tool. Note that for full FIPS compliance, a FIPS endpoint such as `ddog-gov.com` is required.
```
michaelcretzman:
```suggestion
This option is used to enable FIPS support in the Lambda functions deployed using this tool. Note that for full FIPS compliance, a FIPS endpoint such as `ddog-gov.com` is required.
```

src/commands/lambda/functions/uninstrument.ts
```diff
@@ -200,9 +198,7 @@ export const calculateUpdateRequest = (
     PROFILER_ENV_VAR,
     PROFILER_PATH_ENV_VAR,
     DOTNET_TRACER_HOME_ENV_VAR,
-    DD_LLMOBS_ENABLED_ENV_VAR,
-    DD_LLMOBS_ML_APP_ENV_VAR,
-    DD_LLMOBS_AGENTLESS_ENABLED_ENV_VAR,
```
TalUsvyatsky: I might have missed it - why are we removing these?

alexgallotta: oh thanks for catching this! I must have deleted it while rebasing conflicts

src/commands/lambda/__tests__/functions/instrument.part2.test.ts
alexgallotta: what is the region to be used?

alexgallotta:
> looks good. some nits, but otherwise seems fine. we chose not to include a warning when the fips mode, aws site, and datadog region are not in sync?

I decided to just add the warning to the readme. Logging with `context.stdout.write` is done only in the root class and the check is done a few levels deeper, where communication is done with throwing exception

src/commands/lambda/__tests__/functions/instrument.part2.test.ts
TalUsvyatsky: `us-gov-west-1` or `us-gov-east-1`

src/commands/lambda/functions/commons.ts
```diff
@@ -337,17 +337,23 @@ export const getLayerArn = (
...
+  let arnBuilt
   if (isGovCloud) {
-    return `arn:aws-us-gov:lambda:${region}:${GOVCLOUD_LAYER_AWS_ACCOUNT}:layer:${layerName}`
+    arnBuilt = `arn:aws-us-gov:lambda:${region}:${GOVCLOUD_LAYER_AWS_ACCOUNT}:layer:${layerNameSuffix}`
+  } else {
+    arnBuilt = `arn:aws:lambda:${region}:${account}:layer:${layerNameSuffix}`
+  }
+  if (settings?.lambdaFips) {
+    arnBuilt += '-FIPS'
   }
```
TalUsvyatsky: Suggestion: move this up to line 343 and have it be part of the layer name suffix

```ts
  let layerNameSuffix = LAYER_LOOKUP[layer]
  if (ARM_LAYERS.includes(layer) && config.Architectures?.includes(ARM64_ARCHITECTURE)) {
    layerNameSuffix += ARM_LAYER_SUFFIX
  }
  if (settings?.lambdaFips) {
    layerNameSuffix += '-FIPS'
  }

```

README.md
ava-silver, line 1: I see a lot of (i assume) unintended edits to the readme, can we revert the unrelated changes?

README.md
```diff
@@ -139,20 +160,28 @@ The option `fips-ignore-error` ignores this error.
...
 
 ENV variable: `DATADOG_FIPS_IGNORE_ERROR=true`
 CLI param: `--fips-ignore-error`
 
+#### `lambda-fips`
+
+This option is used to enable FIPS support in the Lambda functions deployed using this tool. Note that for full FIPS compliance, a FIPS endpoint such as `ddog-gov.com` is required.
+
+ENV variable: `DATADOG_LAMBDA_FIPS=true`
+CLI param: `--lambda-fips`
```
ava-silver: this should be in the `lambda instrument` specific documentation, it being in the main readme implies that it is available for all commands

src/commands/lambda/functions/commons.ts
```diff
@@ -337,17 +337,23 @@ export const getLayerArn = (
...
+  let arnBuilt
   if (isGovCloud) {
-    return `arn:aws-us-gov:lambda:${region}:${GOVCLOUD_LAYER_AWS_ACCOUNT}:layer:${layerName}`
+    arnBuilt = `arn:aws-us-gov:lambda:${region}:${GOVCLOUD_LAYER_AWS_ACCOUNT}:layer:${layerNameSuffix}`
+  } else {
+    arnBuilt = `arn:aws:lambda:${region}:${account}:layer:${layerNameSuffix}`
   }
 
-  return `arn:aws:lambda:${region}:${account}:layer:${layerName}`
+  return arnBuilt
```
ava-silver: is there a reason for the change in control flow? the previous structure seems simpler/more readable

also maybe im misunderstanding, but the shouldn't the variable be called `layerName` and now `layerNameSuffix`? it contains the whole layer name, right?

ava-silver (changes requested):

src/commands/lambda/instrument.ts
```diff
@@ -95,9 +96,11 @@ export class InstrumentCommand extends Command {
 
   private fips = Option.Boolean('--fips', false)
   private fipsIgnoreError = Option.Boolean('--fips-ignore-error', false)
+  private lambdaFips = Option.Boolean('--lambda-fips', false)
   private fipsConfig = {
     fips: toBoolean(process.env[FIPS_ENV_VAR]) ?? false,
     fipsIgnoreError: toBoolean(process.env[FIPS_IGNORE_ERROR_ENV_VAR]) ?? false,
+    lambdaFips: toBoolean(process.env[LAMBDA_FIPS_ENV_VAR]) ?? false,
   }
```
Drarig29: Can we keep `fipsConfig` totally separated from the command-specific config? We are planning to move that part to a base command which will be extended by all commands.
```suggestion
  private extraTags = ...
  private lambdaFips = Option.Boolean('--lambda-fips', false)
  private flushMetricsToLogs = ...
  
  private fips = Option.Boolean('--fips', false)
  private fipsIgnoreError = Option.Boolean('--fips-ignore-error', false)
  private fipsConfig = {
    fips: toBoolean(process.env[FIPS_ENV_VAR]) ?? false,
    fipsIgnoreError: toBoolean(process.env[FIPS_IGNORE_ERROR_ENV_VAR]) ?? false,
  }
```

ava-silver (commented):
i would appreciate if you make the changes for PR feedback before pinging me for review 🙏🏻

src/commands/lambda/instrument.ts
```diff
@@ -444,6 +447,8 @@ export class InstrumentCommand extends Command {
       environment,
       extensionVersion,
       extraTags,
+      fips: this.fips || this.fipsConfig.fips,
```
Drarig29: If possible, let's also remove `fips` from the `InstrumentationSettings`, since I believe it's used nowhere anyway.

README.md
alexgallotta: It is automatic formatting.
I can revert them, this file should not touched anyway

src/commands/lambda/interfaces.ts
```diff
@@ -69,6 +69,8 @@ export interface InstrumentationSettings extends InstrumentationTags {
...
+  lambdaFips?: boolean
```
Drarig29: Remove here too:
```suggestion
  lambdaFips?: boolean
```

src/commands/lambda/uninstrument.ts
```diff
@@ -71,9 +71,11 @@ export class UninstrumentCommand extends Command {
 
   private fips = Option.Boolean('--fips', false)
   private fipsIgnoreError = Option.Boolean('--fips-ignore-error', false)
+  private lambdaFips = Option.Boolean('--lambda-fips', false)
```
Drarig29: Same comment here, let's keep `lambdaFips` among command-specific options

src/commands/lambda/functions/commons.ts
alexgallotta: yes correct, bad AI for the naming.
For the control flow, do you mean using if-else rather than if with returns?
Usually it's better to avoid multiple exit points

src/commands/lambda/functions/uninstrument.ts
```diff
@@ -222,7 +224,10 @@ export const calculateUpdateRequest = (
   const originalLayerARNs = getLayers(config)
   const layerARNs = (config.Layers ?? [])
     .filter(
-      (layer) => !layer.Arn?.includes(lambdaLibraryLayerName) && !layer.Arn?.includes(DD_LAMBDA_EXTENSION_LAYER_NAME)
+      (layer) =>
+        !layer.Arn?.includes(lambdaLibraryLayerName) &&
+        !layer.Arn?.includes(DD_LAMBDA_EXTENSION_LAYER_NAME) &&
+        !layer.Arn?.includes(`${DD_LAMBDA_EXTENSION_LAYER_NAME}-FIPS`)
```
ava-silver: this line is redundant given the previous line

ava-silver (commented):
I agree with Corentin's comments, along with one last nit and then i'll approve

Drarig29 (approved):
LGTM

PR #1710, opened 2025-06-20
GCR Instrument improvements + additional instrument flags

nhulston:
### What and why?

This PR implements several missing features in the `datadog-ci cloud-run instrument` command:
- Move DD_TAGS, DD_LOG_LEVEL, DD_TRACE_ENABLED, and llm obs env vars from `/lambda/constants` to `/constants` file
- Implement logic to support each option: `--extra-tags`, `--log-level`, `--tracing`, and `--llmobs`
- Implement `--port` or `--health-check-port` to configure the sidecar health check port. Default to 5555, following the convention set by Datadog Kubernetes
- Make DD_SERVICE env var optional; log a warning and fallback to the GCR service name when missing
- Add default env vars to sidecar container when they're missing, but don't overwrite or delete any existing env vars in the sidecar container. This allows the user to manually configure their env vars and be able to run the `instrument` command without overwriting them.
- Extract some logic into `buildSidecarContainer` and `updateAppContainer` for better readability
- Mark `revision` as `undefined` in the config -- this allows CloudRun functions (gen 2) to be instrumented because it lets GCR autogenerate the new revision name. Otherwise, we were getting "revision already exists" errors

### How?

Manually tested with both container and function GCR apps.

Thorough unit tests

### Review checklist

https://datadoghq.atlassian.net/browse/SVLS-7072


src/commands/cloud-run/instrument.ts
```diff
@@ -406,6 +400,14 @@ export class InstrumentCommand extends Command {
...
     }
 
+    // LLMOBS
+    if (this.llmobs) {
+      updatedEnvVars.push({name: DD_LLMOBS_ENABLED_ENV_VAR, value: 'true'})
+      updatedEnvVars.push({name: DD_LLMOBS_ML_APP_ENV_VAR, value: this.llmobs})
+      // serverless-init is installed, so agentless mode should be false
+      updatedEnvVars.push({name: DD_LLMOBS_AGENTLESS_ENABLED_ENV_VAR, value: 'false'})
+    }
+
```
nhulston, line 410: these env vars should be added to the container running the tracer -- not the container running the sidecar

src/commands/cloud-run/instrument.ts
```diff
@@ -255,77 +364,59 @@ export class InstrumentCommand extends Command {
...
-    // Add sidecar if it doesn't exist
-    if (existingSidecarIndex < 0) {
-      updatedContainers.push(sidecarContainer)
+    // Replace LLMOBS env vars
+    if (this.llmobs) {
+      this.addOrReplaceEnvVar(updatedEnvVars, DD_LLMOBS_ENABLED_ENV_VAR, 'true')
+      this.addOrReplaceEnvVar(updatedEnvVars, DD_LLMOBS_ML_APP_ENV_VAR, this.llmobs)
+      // serverless-init is installed, so agentless mode should be false
+      this.addOrReplaceEnvVar(updatedEnvVars, DD_LLMOBS_AGENTLESS_ENABLED_ENV_VAR, 'false')
     }
```
nhulston: these LLMOBS env vars should live on the container with the tracer (the main app), not the sidecar container

src/commands/cloud-run/instrument.ts
```diff
@@ -216,32 +227,130 @@ export class InstrumentCommand extends Command {
...
+        // Let GCR generate the next revision name
+        revision: undefined,
+      },
+    }
+  }
+
+  public buildSidecarContainer(existingSidecarContainer: IContainer | undefined, ddService: string): IContainer {
+    // Add these env vars to the container if they don't already exist,
+    // but leave them unchanged if they already exist on the container.
+    const defaultEnvVars: IEnvVar[] = [
```
avangelillo: You end up having to do a lot of more awkward checks through this (find/some/addOrReplace) because this is a list.  I assume that the GCP call needs them in this format, but one thing to consider is if it ends up being easier to keep this as a an object with `{  envKey: envValue }` and convert it when you need to do that call instead of having to do a different check each time.

Up to you if you think it's worth making that change since I think they're functionally the same

avangelillo (approved):

src/commands/cloud-run/instrument.ts
nhulston: Yeah, that's actually much simpler. I've made the change and used a `Map` in both `buildSidecarContainer()` and `updateAppContainer()`

Thanks for the suggestion!

nhulston: actually, using `Record<String, String>` since `Map` is overkill

src/constants.ts
```diff
@@ -39,6 +39,17 @@ export const SITE_ENV_VAR = 'DD_SITE'
 export const LOGS_INJECTION_ENV_VAR = 'DD_LOGS_INJECTION'
 export const LOGS_PATH_ENV_VAR = 'DD_SERVERLESS_LOG_PATH'
 export const HEALTH_PORT_ENV_VAR = 'DD_HEALTH_PORT'
+export const LOG_LEVEL_ENV_VAR = 'DD_LOG_LEVEL'
+export const TRACE_ENABLED_ENV_VAR = 'DD_TRACE_ENABLED'
+export const DD_LLMOBS_ENABLED_ENV_VAR = 'DD_LLMOBS_ENABLED'
+export const DD_LLMOBS_ML_APP_ENV_VAR = 'DD_LLMOBS_ML_APP'
+export const DD_LLMOBS_AGENTLESS_ENABLED_ENV_VAR = 'DD_LLMOBS_AGENTLESS_ENABLED'
+export const EXTRA_TAGS_ENV_VAR = 'DD_TAGS'
```
duncanista: Can we rename this to `DD_TAGS_ENV_VAR`? At Datadog, idk why, there's the concept of `DD_EXTRA_TAGS` which is also used and at the end it merges back to `DD_TAGS`, but just wanna keep consistency here

src/constants.ts
```diff
@@ -39,6 +39,17 @@ export const SITE_ENV_VAR = 'DD_SITE'
 export const LOGS_INJECTION_ENV_VAR = 'DD_LOGS_INJECTION'
 export const LOGS_PATH_ENV_VAR = 'DD_SERVERLESS_LOG_PATH'
 export const HEALTH_PORT_ENV_VAR = 'DD_HEALTH_PORT'
+export const LOG_LEVEL_ENV_VAR = 'DD_LOG_LEVEL'
+export const TRACE_ENABLED_ENV_VAR = 'DD_TRACE_ENABLED'
```
duncanista: Should we rename this too? appending `DD`?

src/constants.ts
```diff
@@ -39,6 +39,17 @@ export const SITE_ENV_VAR = 'DD_SITE'
...
+export const DD_LLMOBS_ENABLED_ENV_VAR = 'DD_LLMOBS_ENABLED'
+export const DD_LLMOBS_ML_APP_ENV_VAR = 'DD_LLMOBS_ML_APP'
+export const DD_LLMOBS_AGENTLESS_ENABLED_ENV_VAR = 'DD_LLMOBS_AGENTLESS_ENABLED'
+export const EXTRA_TAGS_ENV_VAR = 'DD_TAGS'
+// DD_TAGS Regular Expression
+// This RegExp ensures that the --extra-tags string
+// matches a list of <key>:<value> separated by commas
+// such as: layer:api,team:intake
+export const EXTRA_TAGS_REG_EXP = /^(([a-zA-Z]+)[\w\-/\.]*:[^,]+)+((\,)([a-zA-Z]+)[\w\-/\.]*:[^,]+)*$/g
```
duncanista, line 54: Also, for the docs of this, this needs to be in the format of `// *` or else the IDE won't pick it up correctly, AFAIK

duncanista (approved):
left a couple nits

PR #1711, opened 2025-06-24
SDCD-1464: adding `--team` and `--custom-tags` as options in the DORA CLI

henrywalsh:
### What and why?

Adding two new options in the dora deployments cli command:
  `--team`: our documentation has stated for a long time that the [cli accepts a `--team` option](https://docs.datadoghq.com/dora_metrics/setup/deployments?tab=apiorcli#requirements) but that has not been this case. It has recently been brought to our attention that this has not, up until now, actually been an available option.
  `--custom-tags`: we have recently added the ability for customers to add custom tags to deployment events via our api, and we want them to be able to add tags in the cli as well. These tags take the form `key:value` (for example, `department:engineering`), and we accept a maximum of 100 tags per deployment event.

### How?

In order to add these optional parameters, we have updated the `DeploymentEvent` interface to include these fields, and we read any values provided for them like we do for the other optional fields (`version`, `git`, `env`). The tests have also been updated to include a test case where we provide these fields.

### Review checklist

### DIFF
--- a/src/commands/dora/README.md
+++ b/src/commands/dora/README.md
@@ -21,6 +21,8 @@ $ DD_BETA_COMMANDS_ENABLED=1 datadog-ci dora deployment [--service #0] [--env #0
   --finished-at #0           In Unix seconds or ISO8601. (Examples: 1699961048, 2023-11-14T11:24:08Z)
   --git-repository-url #0    Example: https://github.com/DataDog/datadog-ci.git
   --git-commit-sha #0        Example: 102836a25f5477e571c73d489b3f0f183687068e
+  --team #0                  Example: my-team
+  --custom-tags #0           Example: department:engineering,deployment-type:hotfix
   --skip-git                 Disables sending git URL and SHA. Change Lead Time will not be available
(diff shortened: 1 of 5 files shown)


src/commands/dora/README.md
```diff
@@ -21,6 +21,8 @@ $ DD_BETA_COMMANDS_ENABLED=1 datadog-ci dora deployment [--service #0] [--env #0
   --finished-at #0           In Unix seconds or ISO8601. (Examples: 1699961048, 2023-11-14T11:24:08Z)
   --git-repository-url #0    Example: https://github.com/DataDog/datadog-ci.git
   --git-commit-sha #0        Example: 102836a25f5477e571c73d489b3f0f183687068e
+  --team #0                  Example: my-team
+  --custom-tags #0           Example: department:engineering,deployment-type:hotfix
```
henrywalsh:
```suggestion
  --custom-tags #0           Example: department:engineering
```

brett0000FF (approved):

PR #1713, opened 2025-06-26
[elf-symbols] fix dynamic symbols extraction

nsavoire:
### What and why?

This change fixes and simplifies the dynamic symbols extraction process. 

### How?

Previous implementation was dumping `.dynsym/.dynstr` sections content in temporary files and then removing those sections before adding them back to the debug info file (because `--only-keep-debug` was setting their size to 0).

The new implementation simply sets the sections flags to `alloc,readonly,contents` (which is the default for those sections) and this seems to prevent `--only-keep-debug` from removing their content.

### Review checklist


peterg17 (approved):
nice fix 🎉

Drarig29 (approved):

PR #1715, opened 2025-06-27
[SYNTH-20553] Add link to Test Optimization

Drarig29:
### What and why?

This PR adds a link to Test Optimization at the end of the `datadog-ci synthetics run-tests` execution.

image

### How?

Add `metadata` to `Summary` and reuse existing helpers to create that link.

### Review checklist


src/commands/synthetics/run-tests-lib.ts
```diff
@@ -151,9 +153,15 @@ export const executeTests = async (
     }
   }
 
+  const metadata = getCIMetadata({
+    [GIT_COMMIT_MESSAGE]: 500,
```
Alexis-Jacob-datadog, line 157: why 500 ?

Drarig29: it's a length limit we already had where I copy/pasted this from, so I didn't change it

PR #1722, opened 2025-07-03
[Cloud Run Instrument] Dry run and interactive mode

nhulston:
### What and why?

In the Cloud Run instrument command:
- Implement `--dry` flag
- Implement `--interactive` flag
- And finally, display a git-style diff of the service configuration when running the instrument command:

Screenshot 2025-07-02 at 5 34 02 PM
<br/>
Screenshot 2025-07-02 at 5 33 54 PM
<br/>
Screenshot 2025-07-02 at 5 33 46 PM

### How?

### Review checklist

### DIFF
--- a/src/commands/cloud-run/instrument.ts
+++ b/src/commands/cloud-run/instrument.ts
@@ -235,7 +235,7 @@ export class InstrumentCommand extends Command {
     this.context.stdout.write(generateConfigDiff(existingService, updatedService))
     if (this.dryRun) {
       this.context.stdout.write(
-        `\n${dryRunPrefix(this.dryRun)}Would have updated service ${chalk.bold(serviceName)} with the above changes.\n`
+        `\n\n${dryRunPrefix(this.dryRun)}Would have updated service ${chalk.bold(serviceName)} with the above changes.\n`
       )
 
       return
--- a/src/commands/cloud-run/instrument.ts
+++ b/src/commands/cloud-run/instrument.ts
@@ -235,7 +235,9 @@ export class InstrumentCommand extends Command {
     this.context.stdout.write(generateConfigDiff(existingService, updatedService))
     if (this.dryRun) {
       this.context.stdout.write(
-        `\n\n${dryRunPrefix(this.dryRun)}Would have updated service ${chalk.bold(serviceName)} with the above changes.\n`
+        `\n\n${dryRunPrefix(this.dryRun)}Would have updated service ${chalk.bold(
+          serviceName
+        )} with the above changes.\n`
       )
 
       return
--- a/src/commands/cloud-run/instrument.ts
+++ b/src/commands/cloud-run/instrument.ts
@@ -27,7 +27,7 @@ import {maskString} from '../../helpers/utils'
 
 import {CloudRunConfigOptions} from './interfaces'
 import {renderAuthenticationInstructions, renderCloudRunInstrumentUninstrumentHeader, withSpinner} from './renderer'
-import {checkAuthentication} from './utils'
+import {checkAuthentication, generateConfigDiff} from './utils'
 
 // XXX temporary workaround for @google-cloud/run ESM/CJS module issues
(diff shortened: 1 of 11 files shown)


src/commands/cloud-run/utils.ts
```diff
@@ -14,3 +15,150 @@ export const checkAuthentication = async () => {
...
+ * TODO(@nhulston): update Lambda and AAS instrument to show this diff
+ * @param original The original configuration object
+ * @param updated The updated configuration object
+ * @returns A formatted diff string with colors
+ */
+export const generateConfigDiff = (original: any, updated: any): string => {
+  // Sort keys consistently before comparison
+  const sortedOriginal = sortObjectKeys(original)
+  const sortedUpdated = sortObjectKeys(updated)
+
```
nhulston, line 67: There is a package called `deep-object-diff` in package.json, but it doesn't display the unchanged lines; it is only able to show added/removed lines.

I plan on migrating this to a shared utils in the future and adding it to lambda/instrument and aas/instrument

avangelillo: What part of the unchanged JSON would be relevant for someone to see here?

src/commands/cloud-run/instrument.ts
```diff
@@ -108,25 +100,40 @@ export class InstrumentCommand extends Command {
       return 1
     }
 
+    if (this.interactive) {
```
avangelillo, line 103: Should only this subset of inputs be interactive or should everything be here?

nhulston: These inputs are required for the command to work; everything else is optional configuration

src/commands/cloud-run/utils.ts
nhulston: Mainly, we're just changing environment variables, adding or modifying the sidecar container, and the shared-volume.

If you see `instrument.test.ts.snap`, this useful information takes up most of the diff that's printed

avangelillo: You mentioned you did all of the diffing stuff because the package that is used in other places doesn't display unchanged lines so I was curious what part of the unchanged diff is important to display.  For example when you run something like a `cdk diff` command it shows you something to identify the resource, then only the parts that change underneath it, kind of similar to how `git diff` shows you the file, then which line numbers you're in, then just the diff.  If we're able to get those unchanged parts without needed to reimplement all the diffing I think it would be better if we didn't have to.

nhulston: Ah, well also the package doesn't really give me what I'm looking for. It gives something like this in a really weird format, and hides a lot of information:

```json
{
  "added": {},
  "deleted": {
    "template": {
      "containers": {
        "0": {
          "env": {
            "0": {},
            "1": {},
            "2": {}
          }
        },
        "1": {
          "env": {
            "0": {},
            "1": {},
            "2": {},
            "3": {},
            "4": {},
            "5": {},
            "6": {}
          },
          "resources": {},
          "startupProbe": {}
        }
      }
    }
  },
  "updated": {
    "template": {
      "containers": {
        "1": {
          "startupProbe": {
            "tcpSocket": {
              "port": "5555"
            }
          }
        }
      }
    }
  }
}
```

The current implementation also gives me the freedom to style it how I want, with the color coding and  +&- symbols

nhulston: Alternatively we could use a package like this? https://www.npmjs.com/package/print-diff

Drarig29: You can also try https://www.npmjs.com/package/jest-diff, which gives you a `contextLines` option. But I'm not against having custom code either

nhulston: Good idea, that's a much better/newer package. I will switch to `jest-diff`

Drarig29 (approved):
LGTM

PR #1724, opened 2025-07-07
[SVLS-7097] add additional env vars to aas commands

ava-silver:
### What and why?

Adds `DD_VERSION` as part of  [Unified Service Tagging](https://docs.datadoghq.com/getting_started/tagging/unified_service_tagging/). Also adds in `DD_PROFILING_ENABLED` for automatic profiling support where it is installed.

### How?

Adds the settings to both the main container and sidecar containers

### Review checklist

### DIFF
--- a/src/commands/aas/instrument.ts
+++ b/src/commands/aas/instrument.ts
@@ -29,13 +29,13 @@ export class InstrumentCommand extends AasCommand {
   })
 
   private service = Option.String('--service', {
-    description: 'How you want to tag your service. For example, `my-service`',
+    description: 'The value for the service tag. For example, `my-service`',
   })
   private environment = Option.String('--env,--environment', {
-    description: 'How you want to tag your env. For example, `prod`',
+    description: 'The value for the env tag. For example, `prod`',
   })
   private version = Option.String('--version', {
-    description: 'How you want to tag your AAS code version. For example, `1.0.0`',
+    description: 'The value for the version tag. For example, `1.0.0`',
   })
   private isInstanceLoggingEnabled = Option.Boolean('--instance-logging', false, {
     description:


aliciascott (approved):
looks good for docs

Drarig29 (approved):

src/commands/aas/instrument.ts
```diff
@@ -29,15 +29,22 @@ export class InstrumentCommand extends AasCommand {
...
+  private version = Option.String('--version', {
+    description: 'The value for the version tag. For example, `1.0.0`',
   })
   private isInstanceLoggingEnabled = Option.Boolean('--instance-logging', false, {
     description:
       'When enabled, log collection is automatically configured for an additional file path: /home/LogFiles/*$COMPUTERNAME*.log',
   })
+  private isProfilingEnabled = Option.Boolean('--profiling', true, {
+    description:
+      'Adds the `DD_PROFILING_ENABLED` env var for automatic profiling support. Defaults to enabled. Specify `--no-profiling` to disable.',
```
TalUsvyatsky, line 46: Why `--no-profiling` rather than `--profiling false`?

ava-silver: because of the way [clipanion boolean arguments](https://mael.dev/clipanion/docs/api/option/#optionboolean) work, they don't accept positional arguments

TalUsvyatsky: Gotcha, thanks for the link. What happens when both `--profiling` and `--no-profiling` are passed?

ava-silver: the latter one takes precedence

TalUsvyatsky: latter as in whichever one is passed last or latter as in `--no-profiling`? Either way, non-blocking, but I think a test for that would be nice to clarify the behavior

TalUsvyatsky (approved):

src/commands/aas/instrument.ts
ava-silver: it's:
> whichever one is passed last 

i could add a test, but we realistically never expect someone to use the `--profiling` flag since it is on by default, only the `--no-profiling` flag, I'm hesitant to add a test case to test library functionality, one option if we want would be to change this to something like what we have for `--no-restart`.

interestingly, that option doesn't allow `--no-no-restart` like id expect based on the behavior in clipanion for the profiling option.

ava-silver:
### Merge activity

* **Jul 8, 5:03 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/DataDog/datadog-ci/1724).

PR #1728, opened 2025-07-07
[Cloud Run Instrument] FIPs, source code integration, and customizable sidecar,shared-volume,logs-path

nhulston:
### What and why?
- Allow user to optionally customize sidecar image, sidecar name, log volume name, log volume path, and logs file path
- Implement `--fips` flag
- Implement source code integration in Cloud Run tags
- Implement `--sidecar-cpus` and `--sidecar-memory` flags to let users customize their sidecar allocation

This is the last PR required before we can release and document the `datadog-ci cloud-run instrument` command

If they pass invalid values for cpu/memory, the cloud-run API has a nice error message, so we don't need to validate anything:
Screenshot 2025-07-07 at 3 04 01 PM

There are cases where customizing the sidecar name, log volume name, log volume path, and logs file path might be useful. However, it's only recommended for users that know what they're doing. 

For example, if the user instruments their app with `--sidecar-name custom-name` and instruments it again without the `--sidecar-name` flag, they'll end up adding 2 sidecar containers. So we're marking them as "Not recommended"

### How?

Source code integration required some refactoring from `lambda/instrument.ts` to /helpers/git/instrument-helpers.ts`

### Review checklist

### DIFF
--- a/src/commands/cloud-run/instrument.ts
+++ b/src/commands/cloud-run/instrument.ts
@@ -85,16 +85,16 @@ export class InstrumentCommand extends Command {
     description: `The image to use for the sidecar container. Defaults to '${DEFAULT_SIDECAR_IMAGE}'`,
   })
   private sidecarName = Option.String('--sidecar-name', DEFAULT_SIDECAR_NAME, {
-    description: `The name to use for the sidecar container. Defaults to '${DEFAULT_SIDECAR_NAME}'`,
+    description: `(Not recommended) The name to use for the sidecar container. Defaults to '${DEFAULT_SIDECAR_NAME}'`,
   })
   private sharedVolumeName = Option.String('--shared-volume-name', DEFAULT_VOLUME_NAME, {
-    description: `The name to use for the shared volume. Defaults to '${DEFAULT_VOLUME_NAME}'`,
+    description: `(Not recommended) The name to use for the shared volume. Defaults to '${DEFAULT_VOLUME_NAME}'`,
   })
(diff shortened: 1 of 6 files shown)


ava-silver (approved):
lgtm! thanks for refactoring out the source code integration bits, I'll add a ticket to add source code integration for AAS instrument as well

src/commands/cloud-run/instrument.ts
```diff
@@ -69,15 +75,43 @@ export class InstrumentCommand extends Command {
...
+  private sourceCodeIntegration = Option.Boolean('--source-code-integration,--sourceCodeIntegration', true)
+  private uploadGitMetadata = Option.Boolean('-u,--upload-git-metadata,--uploadGitMetadata', true)
   private tracing = Option.String('--tracing')
   private version = Option.String('--version')
   private llmobs = Option.String('--llmobs')
   private healthCheckPort = Option.String('--port,--health-check-port,--healthCheckPort')
+  private sidecarImage = Option.String('--image,--sidecar-image', DEFAULT_SIDECAR_IMAGE, {
+    description: `The image to use for the sidecar container. Defaults to '${DEFAULT_SIDECAR_IMAGE}'`,
+  })
+  private sidecarName = Option.String('--sidecar-name', DEFAULT_SIDECAR_NAME, {
```
ava-silver, line 87: as discussed offline, lets indicate that sidecarName isnt generally recommended for ease of updating/un-instrumentation later

avangelillo (approved):

jminuscula (approved):

PR #1734, opened 2025-07-08
[Google Cloud Run] Implement Uninstrument Command

nhulston:
### What and why?

Implement the uninstrument command. After this, we can release and document the `instrument` and `uninstrument` commands.

Screenshot 2025-07-08 at 3 42 39 PM
<br/>
Screenshot 2025-07-08 at 3 42 17 PM
<br/>
Screenshot 2025-07-08 at 3 42 25 PM

### How?

- Remove the sidecar container
- Remove the shared volume mount
- Remove env vars beginning with `DD_` from the main app container

### Review checklist


src/commands/cloud-run/utils.ts
```diff
@@ -74,10 +78,42 @@ export const generateConfigDiff = (original: any, updated: any): string => {
   const configDiff = diff(obfuscatedOriginal, obfuscatedUpdated, {
     aColor: chalk.red,
     bColor: chalk.green,
+    omitAnnotationLines: true,
```
nhulston, line 81: This removes the lines
```
- Expected
+ Received
```

from the jest-diff string

src/commands/cloud-run/utils.ts
```diff
@@ -74,10 +78,42 @@ export const generateConfigDiff = (original: any, updated: any): string => {
...
+) => {
+  const existingServiceConfigs: IService[] = []
+  for (const serviceName of services) {
+    const servicePath = client.servicePath(project, region, serviceName)
+
+    const existingService = await withSpinner(
+      `Fetching configuration for ${chalk.bold(serviceName)}...`,
+      async () => {
+        try {
+          const [serv] = await client.getService({name: servicePath})
```
ava-silver, line 104: this is probably a follow up item, but it could be nice to make these calls in parallel

src/commands/cloud-run/uninstrument.ts
```diff
@@ -0,0 +1,222 @@
...
+
+import chalk from 'chalk'
+import {Command, Option} from 'clipanion'
+
+import {FIPS_ENV_VAR, FIPS_IGNORE_ERROR_ENV_VAR} from '../../constants'
+import {toBoolean} from '../../helpers/env'
+import {enableFips} from '../../helpers/fips'
+import {renderSoftWarning} from '../../helpers/renderer'
+
+import {DEFAULT_SIDECAR_NAME, DEFAULT_VOLUME_NAME} from './instrument'
```
ava-silver: part of me wonders if they should be in a utils/common file rather than imported from instrument

src/commands/cloud-run/uninstrument.ts
```diff
@@ -0,0 +1,222 @@
...
+  public static usage = Command.Usage({
+    category: 'Serverless',
+    description: 'Revert Datadog instrumentation in a Cloud Run app.',
+  })
+
+  private dryRun = Option.Boolean('-d,--dry,--dry-run', false)
+  private project = Option.String('-p,--project')
+  private services = Option.Array('-s,--service,--services', [])
+  private interactive = Option.Boolean('-i,--interactive', false)
+  private region = Option.String('-r,--region')
```
ava-silver: could these get descriptions? even something short would be useful

src/commands/cloud-run/uninstrument.ts
```diff
@@ -0,0 +1,222 @@
...
+
+      // Prompt for service if missing
+      if (this.services.length === 0) {
+        const serviceName = await requestServiceName()
+        this.services = [serviceName]
+      }
+    }
+
+    // Validate required variables
+    this.context.stdout.write(chalk.bold('\n🔍 Verifying command flags...\n'))
```
ava-silver: nit: not sure if this is logged in the instrument command, but it's probably a little verbose for it to happen every run

src/commands/cloud-run/uninstrument.ts
```diff
@@ -0,0 +1,222 @@
...
+      chalk.bold(`\n${dryRunPrefix(this.dryRun)}🚀 Uninstrumenting Cloud Run services with sidecar...\n`)
+    )
+    for (let i = 0; i < existingServiceConfigs.length; i++) {
+      const serviceConfig = existingServiceConfigs[i]
+      const serviceName = services[i]
+      try {
+        await this.uninstrumentService(client, serviceConfig, serviceName)
+      } catch (error) {
+        this.context.stderr.write(
+          chalk.red(`${dryRunPrefix(this.dryRun)}Failed to uninstrument service ${serviceName}: ${error}\n`)
```
ava-silver: the error logs should probably use the `renderError` helper for consistency across the CLI

src/commands/cloud-run/uninstrument.ts
```diff
@@ -0,0 +1,222 @@
...
+    }
+  }
+
+  // Remove volume mount and add required env vars
+  private updateAppContainer(appContainer: IContainer) {
+    const existingVolumeMounts = appContainer.volumeMounts || []
+    const updatedVolumeMounts = existingVolumeMounts.filter((v) => v.name !== this.sharedVolumeName)
+
+    const existingEnvVars = appContainer.env || []
+    // Remove env vars beginning with DD_
```
ava-silver, line 216: are there any we might add that don't begin with this? I know for AAS I added a couple that didnt: https://github.com/DataDog/datadog-ci/blob/decd2702e77340a813ee8b00f7c3779df5231e51/src/commands/aas/common.ts#L44-L46

src/commands/cloud-run/utils.ts
nhulston: Good suggestion! I'm worried about rate limiting, so maybe we can do 5 calls at a time?

nhulston: Ah it currently does not work well with the spinner. Apparently `ora` doesn't support multiple spinners at once. This could be implemented in the future if requested by customers

src/commands/cloud-run/uninstrument.ts
nhulston: No, in the instrument command we only add env vars beginning with `DD_`.

I think we're okay with skipping those specific env vars, since hardly anyone uses .NET on GCR

ava-silver: I think the only other risk here is that we could potentially nuke some of their variables since we have no guarantees that they won't prefix their variables with `DD_`.

I'd personally opt for a list of vars, (potentially even shared across commands), but I wont block this PR on it

src/commands/cloud-run/uninstrument.ts
```diff
@@ -0,0 +1,229 @@
...
+  private services = Option.Array('-s,--service,--services', [], {
+    description: 'Cloud Run service(s) to instrument',
+  })
+  private interactive = Option.Boolean('-i,--interactive', false, {
+    description: 'Prompt for flags one at a time',
+  })
+  private region = Option.String('-r,--region', {
+    description: 'GCP region your service(s) are deployed in',
+  })
+  // private regExPattern = Option.String('--services-regex,--servicesRegex') implement if requested by customers
```
ava-silver: nit: probably best for now to remove commented out code

src/commands/cloud-run/uninstrument.ts
```diff
@@ -0,0 +1,229 @@
...
+
+  public async execute(): Promise<0 | 1> {
+    enableFips(this.fips || this.fipsConfig.fips, this.fipsIgnoreError || this.fipsConfig.fipsIgnoreError)
+
+    this.context.stdout.write(
+      `\n${dryRunPrefix(this.dryRun)}🐶 ${chalk.bold('Uninstrumenting Cloud Run service(s)')}\n\n`
+    )
+
+    if (this.interactive) {
+      // Prompt for project if missing
```
ava-silver: nit: these comments are redundant, and just feels like ai

ava-silver (approved):
lgtm! small nits but nothing blocking 👍🏻

src/commands/cloud-run/uninstrument.ts
nhulston: very true, those comments were unnecessary

PR #1739, opened 2025-07-09
Fix coverage upload dir scan to handle symlinks

nikita-tkachenko-datadog:
### What and why?

Fixes the code that does directory scanning for code coverage reports upload.
The code mistakenly assumed that anything that is not `isFile()` is a directory, and this check failed for symlinks.

### How?

The fixed code uses `statSync` instead of `lstatSync` to follow symlinks. It also does not assume that anything that isn't a file is a directory, and does an explicit check as well.

There's a minor improvement to the `--ignored-paths` param handling as well: now it's possible to specify an ignored paths partially (no need to specify the full path relative to what the scan directory is).

### Review checklist

### DIFF
--- a/src/commands/coverage/upload.ts
+++ b/src/commands/coverage/upload.ts
@@ -138,7 +138,7 @@ export class UploadCodeCoverageReportCommand extends Command {
       return 1
     }
 
-    this.context.stderr.write(`Ignored paths: ${parsePathsList(this.ignoredPaths)}\n')
+    this.context.stderr.write(`Ignored paths: ${parsePathsList(this.ignoredPaths)}\n`)
 
     if (this.format && !isCoverageFormat(this.format)) {
       this.context.stderr.write(
--- a/src/commands/coverage/upload.ts
+++ b/src/commands/coverage/upload.ts
@@ -138,6 +138,9 @@ export class UploadCodeCoverageReportCommand extends Command {
       return 1
     }
 
+    // eslint-disable-next-line @typescript-eslint/restrict-plus-operands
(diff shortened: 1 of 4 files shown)


src/commands/coverage/upload.ts
```diff
@@ -138,6 +138,9 @@ export class UploadCodeCoverageReportCommand extends Command {
       return 1
     }
 
+    // eslint-disable-next-line @typescript-eslint/restrict-plus-operands
+    this.context.stderr.write('Ignored paths: ' + parsePathsList(this.ignoredPaths) + '\n')
```
Drarig29: **nit** to align with the rest of the file
```suggestion
    this.context.stderr.write(`Ignored paths: ${parsePathsList(this.ignoredPaths)}\n')
```

src/helpers/file-finder.ts
```diff
@@ -79,7 +80,14 @@ const partitionFilesRecursive = (
     return
   }
 
-  if (isFile(path)) {
+  let stats: Stats
+  try {
+    stats = statSync(path)
+  } catch (e) {
+    return
+  }
```
Drarig29: **nit** you can omit the `e` since it's not used
```suggestion
  try {
    stats = statSync(path)
  } catch {
    return
  }
```

Drarig29 (approved):
LGTM, just two nits

src/commands/coverage/upload.ts
Drarig29: oops, sorry for the syntax error 😅

nikita-tkachenko-datadog: No worries, should've spotted that myself 😁

PR #1751, opened 2025-07-15
Add 'deployment gate' command

GabrielAnca:
### What and why?

### How?

### Review checklist

### DIFF
--- a/src/commands/deployment/gate.ts
+++ b/src/commands/deployment/gate.ts
@@ -1,0 +1,289 @@
+import {isAxiosError} from 'axios'
+import chalk from 'chalk'
+import {Command, Option} from 'clipanion'
+import * as t from 'typanion'
+
+import {FIPS_ENV_VAR, FIPS_IGNORE_ERROR_ENV_VAR} from '../../constants'
+import {toBoolean} from '../../helpers/env'
+import {enableFips} from '../../helpers/fips'
+import {ICONS} from '../../helpers/formatting'
+import {Logger, LogLevel} from '../../helpers/logger'
+import {getApiHostForSite} from '../../helpers/utils'
+
+import {apiConstructor} from './api'
+import {APIHelper, GateEvaluationRequest} from './interfaces'
+
+/**
+ * This command evaluates deployment gates in Datadog.
+ * It handles the entire process of requesting a gate evaluation and polling for results
+ * with robust error handling and retry logic.
+ */
+export class DeploymentGateCommand extends Command {
+  public static paths = [['deployment', 'gate']]
+
+  public static usage = Command.Usage({
+    category: 'CI Visibility',
+    description: 'Evaluate deployment gates in Datadog.',
+    details: `
+      This command evaluates deployment gates in Datadog.\n
+      It handles the entire process of requesting a gate evaluation and polling for results
+      with robust error handling and retry logic.\n
+      The command will exit with status 0 when the gate passes and status 1 otherwise.
+    `,
+    examples: [
+      [
+        'Evaluate a deployment gate for payments-backend service in prod environment',
+        'datadog-ci deployment gate --service payments-backend --env prod',
+      ],
+      [
+        'Evaluate a deployment gate with custom timeout',
+        'datadog-ci deployment gate --service payments-backend --env prod --timeout 7200',
+      ],
+      [
+        'Evaluate a deployment gate that fails on errors',
+        'datadog-ci deployment gate --service payments-backend --env prod --fail-on-error',
+      ],
+      [
+        'Evaluate a deployment gate with version and APM primary tag',
+        'datadog-ci deployment gate --service payments-backend --env prod --version 1.2.3 --apm-primary-tag team:backend',
+      ],
+    ],
+  })
+
+  // Required parameters
+  private service = Option.String('--service', {
+    description: 'The service name (e.g., payments-backend)',
+    validator: t.isString(),
+  })
+  private env = Option.String('--env', {
+    description: 'The environment name (e.g., prod, staging)',
+    validator: t.isString(),
+  })
+
+  // Optional parameters
+  private identifier = Option.String('--identifier', 'default', {
+    description: 'The deployment identifier (defaults to "default")',
+  })
+  private version = Option.String('--version', {
+    description: 'The deployment version (required for gates with faulty deployment detection rules)',
+  })
+  private apmPrimaryTag = Option.String('--apm-primary-tag', {
+    description: 'The APM primary tag (only for gates with faulty deployment detection rules)',
+  })
+  private timeout = Option.String('--timeout', '10800', {
+    description: 'Maximum time to wait for the script execution in seconds (default: 10800 = 3 hours)',
+    validator: t.isString(),
+  })
+  private failOnError = Option.Boolean('--fail-on-error', false, {
+    description: 'When true, the script will consider the gate as failed when timeout is reached or errors occur',
+  })
+
+  // FIPS options
+  private fips = Option.Boolean('--fips', false)
+  private fipsIgnoreError = Option.Boolean('--fips-ignore-error', false)
+
+  private pollingInterval = 15000
+
+  private config = {
+    apiKey: process.env.DATADOG_API_KEY || process.env.DD_API_KEY,
+    appKey: process.env.DATADOG_APP_KEY || process.env.DD_APP_KEY,
+    fips: toBoolean(process.env[FIPS_ENV_VAR]) ?? false,
+    fipsIgnoreError: toBoolean(process.env[FIPS_IGNORE_ERROR_ENV_VAR]) ?? false,
+  }
+
+  private logger: Logger = new Logger((s: string) => this.context.stdout.write(s), LogLevel.INFO)
+  private startTime: number = Date.now()
+
+  public async execute() {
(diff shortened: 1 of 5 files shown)


src/commands/deployment/interfaces.ts
```diff
@@ -0,0 +1,46 @@
+import type {AxiosPromise} from 'axios'
+
+export interface GateEvaluationRequest {
+  service: string
+  env: string
+  identifier: string
```
andrea-mosk: This is optional as well right?

src/commands/deployment/gate.ts
```diff
@@ -0,0 +1,294 @@
...
+        'Evaluate a deployment gate with custom timeout',
+        'datadog-ci deployment gate --service payments-backend --env prod --timeout 7200',
+      ],
+      [
+        'Evaluate a deployment gate and fail if an error occurs',
+        'datadog-ci deployment gate --service payments-backend --env prod --fail-on-error',
+      ],
+      [
+        'Evaluate a deployment gate with version and APM primary tag',
+        'datadog-ci deployment gate --service payments-backend --env prod --version 1.2.3 --apm-primary-tag team:backend',
```
andrea-mosk:
```suggestion
        'datadog-ci deployment gate --service payments-backend --env prod --version 1.2.3 --apm-primary-tag region:us-central-1',
```

Something more realistic for the primary tag

src/commands/deployment/gate.ts
```diff
@@ -0,0 +1,294 @@
...
+    description: 'The deployment identifier (defaults to "default")',
+  })
+  private version = Option.String('--version', {
+    description: 'The deployment version (required for gates with faulty deployment detection rules)',
+  })
+  private apmPrimaryTag = Option.String('--apm-primary-tag', {
+    description: 'The APM primary tag (only for gates with faulty deployment detection rules)',
+  })
+  private timeout = Option.String('--timeout', '10800', {
+    description: 'Maximum time to wait for the script execution in seconds (default: 10800 = 3 hours)',
```
andrea-mosk:
```suggestion
    description: 'Maximum amount of seconds to wait for the script execution in seconds (default: 10800 = 3 hours)',
```

Let's just specify that it's in seconds

src/commands/deployment/gate.ts
```diff
@@ -0,0 +1,294 @@
...
+        this.logger.error(`Request failed with client error: ${error.response.status} ${error.response.statusText}`)
+      } else {
+        this.logger.error(
+          `Could not start gate evaluation with unknown error: ${
+            error instanceof Error ? error.message : String(error)
+          }`
+        )
+      }
+
+      return Promise.reject(error)
```
andrea-mosk: What does this do? Is it here that we should switch based on the --fail-on-error variable?

src/commands/deployment/gate.ts
```diff
@@ -0,0 +1,294 @@
...
+  ): Promise<number> {
+    this.logger.info('Waiting for gate evaluation results...')
+
+    const maxWaitTime = timeoutSeconds * 1000
+
+    while (true) {
+      const timePassed = Date.now() - this.startTime
+      const remainingTime = maxWaitTime - timePassed
+
+      if (remainingTime <= 0) {
```
andrea-mosk: This could be the condition of the while no?

src/commands/deployment/gate.ts
```diff
@@ -0,0 +1,294 @@
...
+            const rules = response.data.data.attributes.rules
+            const totalRules = rules.length
+            const completedRules = rules.filter((rule) => rule.status !== 'in_progress').length
+
+            this.logger.info(
+              `\tGate evaluation in progress (${completedRules}/${totalRules} rules completed). Retrying in ${waitTimeInSeconds}s...`
+            )
+
+            await new Promise((resolve) => setTimeout(resolve, waitTime))
+            continue
```
andrea-mosk: Do we need this `continue`?

src/commands/deployment/gate.ts
```diff
@@ -0,0 +1,294 @@
...
+          default:
+            throw new Error(`Unknown gate evaluation status: ${status}`)
```
andrea-mosk: As discussed, we should be as tolerable as we can when building the command. Introducing a new status in the API in future could cause all customers using this version of the command to fail, so we need to find a better default

src/commands/deployment/gate.ts
```diff
@@ -0,0 +1,294 @@
...
+        )
+
+        return this.getExitCodeForDatadogError()
+      }
+    }
+  }
+
+  private getExitCodeForDatadogError() {
+    if (this.failOnError) {
+      this.logger.info('Unexpected error happened, exiting with status 1')
```
andrea-mosk: Let's mention that it's because fail on error is `true` in the log?

src/commands/deployment/gate.ts
GabrielAnca: Great point! Updated so that we keep retrying when there is an unexpected status. I agree this helps if we ever decide to add cases like 'queued', 'pending' or 'paused'. It will just make us reach the timeout for other types of states such as 'cancelled', but I don't see any reason to implement a status like that in the near future

src/commands/deployment/gate.ts
GabrielAnca: Great point, updated!

src/commands/deployment/gate.ts
GabrielAnca: It's indeed redundant, removed!

src/commands/deployment/gate.ts
GabrielAnca: Oh true, I pushed an update that simplifies timeout handling a lot

src/commands/deployment/interfaces.ts
GabrielAnca: My bad, updated!

src/commands/deployment/gate.ts
GabrielAnca: This is the equivalent of throwing an error, but inside of an async function. Ultimately, this will cause an exception that will be caught by the exception handing in the caller function

https://github.com/DataDog/datadog-ci/blob/4c397ca980a6fc6707e2b56ba7cf45d5e13732f5/src/commands/deployment/gate.ts#L161-L174

Drarig29: You should just be able to `throw error`

GabrielAnca: Oh I was under the impression that it was discouraged - but it seems my memory is wrong 😅 Updated to a simple throw!

src/commands/deployment/gate.ts
```diff
@@ -255,21 +253,21 @@ export class DeploymentGateCommand extends Command {
...
 
-            await new Promise((resolve) => setTimeout(resolve, waitTime))
-            continue
           default:
-            throw new Error(`Unknown gate evaluation status: ${status}`)
+            this.logger.warn(`Unknown gate evaluation status: ${status}, retrying in ${waitTimeInSeconds}s...`)
         }
+
+        await new Promise((resolve) => setTimeout(resolve, waitTime))
+        timePassed = Date.now() - this.startTime
```
andrea-mosk: I guess the last retry might not be done based on this? if the max wait is 10 seconds, we will do a single poll, then wait for 10 seconds and then update this value which will be slightly higher than 10 seconds, so we will exit the while?

src/commands/deployment/gate.ts
```diff
@@ -278,11 +276,15 @@ export class DeploymentGateCommand extends Command {
         return this.getExitCodeForDatadogError()
       }
     }
+
+    this.logger.warn(`${ICONS.WARNING} Timeout reached (${timeoutSeconds} seconds). Gate evaluation did not complete.`)
```
andrea-mosk:
```suggestion
    this.logger.warn(`${ICONS.WARNING} Timeout reached (${timeoutSeconds} seconds). Gate evaluation did not complete in time.`)
```

andrea-mosk (approved):

src/commands/deployment/gate.ts
GabrielAnca: Updated!

src/commands/deployment/gate.ts
GabrielAnca: Great point, it was indeed possible so I've updated the logic to capture that case (sorry it's a big commit). Now we do the whole loop, I switched the `<=` with just `<` and added an extra retry after the loop. LMK what you think!

src/commands/deployment/gate.ts
```diff
@@ -229,68 +235,85 @@ export class DeploymentGateCommand extends Command {
...
-          `Error polling for gate evaluation results: ${error instanceof Error ? error.message : String(error)}`
-        )
+    while (timePassed < maxWaitTime) {
+      const remainingTime = maxWaitTime - timePassed
+      const waitTime = Math.min(this.pollingInterval, remainingTime)
+      const waitTimeInSeconds = Math.floor(waitTime / 1000)
 
-        return this.getExitCodeForDatadogError()
+      result = await this.getEvaluationResult(api, evaluationId)
+      if (result === 'PASS' || result === 'FAIL') {
```
andrea-mosk: Should we do `if (result)` also here, similarly to below?

andrea-mosk (approved):

src/commands/deployment/gate.ts
GabrielAnca: True! Updated!

Drarig29 (approved):

PR #1752, opened 2025-07-15
[SVLS-7166][AAS] remove explicit profiling option

ava-silver:
### What and why?

Removes the --profiling flag (which needs the `--no-profiling` option to disable it), since it is a paid product and adds a lot of overhead. 

Instead of manually adding env product options via unique CLI flags, we should instead have customers use the `--env-vars` options to add additional env vars for product enablement. Also adds a `-e` shorthand for it to make it simpler to do so.

Also Implements env var specification support for uninstrument, so customers can use the same config to both set and unset env vars.

### Review checklist

### DIFF
--- a/src/commands/aas/common.ts
+++ b/src/commands/aas/common.ts
@@ -68,6 +68,10 @@ export abstract class AasCommand extends Command {
     description:
       'Full Azure resource IDs to instrument, eg "/subscriptions/{subscriptionId}/resourceGroups/{resourceGroup}/providers/Microsoft.Web/sites/{aasName}"',
   })
+  private envVars = Option.Array('-e,--env-vars', {
+    description:
+      'Additional environment variables to set for the App Service. Can specify multiple in the form `--env-vars VAR1=VALUE1 --env-vars VAR2=VALUE2`.',
+  })
 
   private configPath = Option.String('--config', {
     description: 'Path to the configuration file',
@@ -100,6 +104,7 @@ export abstract class AasCommand extends Command {
             subscriptionId: this.subscriptionId,
             resourceGroup: this.resourceGroup,
             aasName: this.aasName,
+            envVars: this.envVars,
             ...this.additionalConfig,
           },
         },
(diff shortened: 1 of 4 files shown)


ava-silver:
* **#1752** <a href="https://app.graphite.dev/github/pr/DataDog/datadog-ci/1752?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/DataDog/datadog-ci/1752?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

src/commands/aas/common.ts
```diff
@@ -69,6 +68,10 @@ export abstract class AasCommand extends Command {
     description:
       'Full Azure resource IDs to instrument, eg "/subscriptions/{subscriptionId}/resourceGroups/{resourceGroup}/providers/Microsoft.Web/sites/{aasName}"',
   })
+  private envVars = Option.Array('-e,--env-vars', {
```
Drarig29, line 71: Each argument is singular so I'd suggest removing the "s" at the end of "--env-vars". And to align with other commands, can we use `-e,--env`?
```suggestion
  private envVars = Option.Array('-e,--env', {
```

ava-silver: hmm we already have --env as the DD_ENV. Additionally, this would be backwards incompatible, since I'm just moving this from the instrument command to both instrument and uninstrument. maybe i could add `--env-var` as an option?

Drarig29: Ok let's keep as is

Drarig29 (approved):

ava-silver:
> > removing an option is a breaking change... Should we deprecate it instead, undocument it and make it log a warning?
> 
> Normally I'd say we should keep `--profiling` working for some amount of time (how long?) as an alias for `-e DD_PROFILING_ENABLED=true`, adding the deprecation warning to give users time to switch to the new format. But if this feature was still in preview and it was only available for a short amount of time, then chances that anyone was already using it are very low and it's probably fine to remove outright?

I chatted with Corentin offline, we agreed that it makes the most sense to just remove the option outright. I've reverted the PR to that state (cc @lucaspimentel)

avangelillo (approved):

PR #1753, opened 2025-07-15
[SVLS-7166] [AAS .NET] use glibc by default for the profiler path, with the option for musl

ava-silver:
### What and why?

Previously, `musl` was the default backend for the profiler path, which should be glibc, since Azure is now using only ubuntu for dotnet code deployments. if customers are instrumenting a musl-based containerized web app running dotnet, they can now specify the `--musl` flag to specify, since we would have no way of knowing otherwise.

### Review checklist

### DIFF
--- a/src/commands/aas/instrument.ts
+++ b/src/commands/aas/instrument.ts
@@ -176,9 +176,18 @@ export class InstrumentCommand extends AasCommand {
         return false
       }
 
+      const isContainer = isLinuxContainer(site)
+      if (config.isMusl && !isContainer) {
+        this.context.stdout.write(
+          renderSoftWarning(
+            `The --musl flag is set, but the App Service ${chalk.bold(aasName)} is not a containerized app. \
+This flag is only applicable for containerized .NET apps (on musl-based distributions like Alpine Linux), and will be ignored.`
+          )
+        )
+      }
       await this.instrumentSidecar(
         aasClient,
-        {...config, isDotnet: config.isDotnet || isDotnet(site), isMusl: config.isMusl && isLinuxContainer(site)},
+        {...config, isDotnet: config.isDotnet || isDotnet(site), isMusl: config.isMusl && isContainer},
         resourceGroup,
         aasName
       )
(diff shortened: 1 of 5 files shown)


ava-silver:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/DataDog/datadog-ci/1753?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#1753** <a href="https://app.graphite.dev/github/pr/DataDog/datadog-ci/1753?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/DataDog/datadog-ci/1753?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1752** <a href="https://app.graphite.dev/github/pr/DataDog/datadog-ci/1752?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

src/commands/aas/instrument.ts
```diff
@@ -177,7 +183,7 @@ export class InstrumentCommand extends AasCommand {
 
       await this.instrumentSidecar(
         aasClient,
-        {...config, isDotnet: config.isDotnet || isDotnet(site)},
+        {...config, isDotnet: config.isDotnet || isDotnet(site), isMusl: config.isMusl && isLinuxContainer(site)},
```
Drarig29: Can we add a warning to let users know when their `config.isMusl` is forced to be false on non-linux containers? I don't know if it'd be obvious otherwise

lucaspimentel (approved):
LGTM, thanks!

src/commands/aas/instrument.ts
ava-silver: My main concern here is that they could be specifying it for some web apps, but not others, and logging a warning when it's disabled by default could get chatty. happy to make the change though

Drarig29: > logging a warning when it's disabled by default

I would only log the warning when they explicitly set it to `true` but it's still forced to be false for non-linux containers

avangelillo (approved):

PR #1757, opened 2025-07-17
[synthetics] Downgrade some tunnel warn logs to debug

Drarig29:
### What and why?

Use the `debug` package to support `DEBUG=synthetics:tunnel datadog-ci ...`

### How?

Replace `this.reporter.warn` calls with `debug` calls for some non-actionable warn logs.

All those warn logs were being printed even for successful runs, so they aren't interesting. In case the user suspects something is going wrong because of the tunnel, they can enable debug logs with `DEBUG=synthetics:tunnel`

### Review checklist

### DIFF
--- a/src/commands/synthetics/README.md
+++ b/src/commands/synthetics/README.md
@@ -454,7 +454,7 @@ Use the [Continuous Testing tunnel](https://docs.datadoghq.com/continuous_testin
 
 For more information, see [Using Local and Staging Environments](#using-local-and-staging-environments).
 
-To troubleshoot tests failing due to the tunnel, you may enable debug logs with `DEBUG=synthetics:tunnel`.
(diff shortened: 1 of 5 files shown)


src/commands/synthetics/README.md
```diff
@@ -454,6 +454,8 @@ Use the [Continuous Testing tunnel](https://docs.datadoghq.com/continuous_testin
 
 For more information, see [Using Local and Staging Environments](#using-local-and-staging-environments).
 
+To troubleshoot tests failing due to the tunnel, you may enable debug logs with `DEBUG=synthetics:tunnel`.
```
joepeeples:
```suggestion
To troubleshoot tests failing due to the tunnel, you can enable debug logs with `DEBUG=synthetics:tunnel`.
```

teodor2312 (approved):

PR #1762, opened 2025-07-21
Add retries to deployment gate command

GabrielAnca:
💡 [This PR is easier to review without whitespace changes](https://github.com/DataDog/datadog-ci/pull/1762/files?w=1)

### What and why?

This PR introduces retries in the deployment gate command so that we retry automatically up to the timeout when a request to either the gate evaluation or the polling fails.

### How?

Using the existing retry library

### Review checklist


src/commands/deployment/__tests__/__snapshots__/gate.test.ts.snap
```diff
@@ -1,14 +1,36 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`gate execute evaluation errors on gate evaluation request should fail when gate evaluation request fails with 500 and fail-on-error is true 1`] = `
+exports[`gate execute evaluation errors on gate evaluation request should fail when gate evaluation request fails with 400 1`] = `
```
GabrielAnca, line 3: This is a new snapshot (I missed this in my previous PR). The updated snapshot is just below

src/commands/deployment/__tests__/gate.test.ts
```diff
@@ -200,13 +239,17 @@ describe('gate', () => {
...
+          await jest.runAllTimersAsync()
+
+          const {context, code} = await runPromise
 
           expect(code).toBe(0)
           expect(context.stdout.toString()).toMatchSnapshot()
 
           expect(apiConstructorSpy).toHaveBeenCalledWith('https://api.datadoghq.com', 'test-api-key', 'test-app-key')
-          expect(mockApi.requestGateEvaluation).toHaveBeenCalledTimes(1)
+          expect(mockApi.requestGateEvaluation).toHaveBeenCalledTimes(6)
```
GabrielAnca, line 252: We retry 6 times in a minute

src/commands/deployment/gate.ts
```diff
@@ -91,7 +93,8 @@ export class DeploymentGateCommand extends Command {
   }
 
   private logger: Logger = new Logger((s: string) => this.context.stdout.write(s), LogLevel.INFO)
-  private pollingInterval = 15000
+  private evaluationRequestTimeout = 60000 // 1 minute
```
GabrielAnca, line 96: We want to have a smaller timeout to request an evaluation. If we can't request it in 1 minute, we stop immediately. If we can request it but it's running, we wait until the user-specified timeout is reached

src/commands/deployment/gate.ts
```diff
@@ -207,47 +211,61 @@ export class DeploymentGateCommand extends Command {
...
     }
+
+    const evaluationId = await retryRequest(doRequest, {
+      ...this.getRetryOptions(timeout),
+      onRetry: (e, attempt) => {
+        this.logger.info(`Retrying gate evaluation request (${attempt} attempts)...`)
+      },
+    })
+
+    return evaluationId
```
andrea-mosk, line 248: What happens if the API returns 5xx for the entire timeout? The `retryRequest` returns an error that is caught upstream?

src/commands/deployment/gate.ts
```diff
@@ -207,47 +211,61 @@ export class DeploymentGateCommand extends Command {
...
   }
 
   private async pollForEvaluationResults(
     api: APIHelper,
     evaluationId: string,
-    timeoutSeconds: number
+    timeout: number
```
andrea-mosk, line 254: So the overall behavior here is:
- We poll the endpoint every 15 seconds
- Each time we call the endpoint, we might do N calls if the endpoint is unavailable? 

Could we just wait the next call and threat it as a 404/in-progress perhaps?

src/commands/deployment/gate.ts
GabrielAnca: Exactly!

src/commands/deployment/gate.ts
GabrielAnca: You're right, we could definitely rely on the polling logic to retry on 404 & 500 errors. Updated!

Drarig29 (approved):

PR #1799, opened 2025-08-13
Add monitor variable option to deployment gate command

GabrielAnca:
### What and why?

Adding the `--monitor-variable` option to the `deployment gate` command.

I want to deliberately hide it from the command help and leave it undocumented because this is a trial we are running with a couple of customers and we are not sure if this will be the definitive approach or if this will change radically in the near future

### How?

When the option is present, pass it as-is to the API

### Review checklist

### DIFF
--- a/src/commands/deployment/gate.ts
+++ b/src/commands/deployment/gate.ts
@@ -65,12 +65,15 @@ export class DeploymentGateCommand extends Command {
   // Optional parameters
   private identifier = Option.String('--identifier', {
     description: 'The deployment identifier (defaults to "default")',
+    validator: t.isString(),
   })
   private version = Option.String('--version', {
     description: 'The deployment version (required for gates with faulty deployment detection rules)',
+    validator: t.isString(),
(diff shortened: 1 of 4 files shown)


src/commands/deployment/gate.ts
```diff
@@ -80,6 +83,11 @@ export class DeploymentGateCommand extends Command {
     description:
       'When true, the script will consider the gate as failed when timeout is reached or unexpected errors occur calling the Datadog APIs',
   })
+  // monitorVariable is hidden because it's not available yet
+  private monitorVariable = Option.String('--monitor-variable', '', {
+    validator: t.isString(),
+    hidden: true,
+  })
```
GabrielAnca, line 90: I want to deliberately hide it from the command help and leave it undocumented because this is a trial we are running with a couple of customers and we are not sure if this will be the definitive approach or if this will change radically in the near future

calvinbayer (approved):

PR #1804, opened 2025-08-19
[packages-migration] Create `@datadog/datadog-ci-base` package

Drarig29:
### What and why?

- https://github.com/DataDog/datadog-ci/pull/1804 (← **you are here**)
- https://github.com/DataDog/datadog-ci/pull/1817
- https://github.com/DataDog/datadog-ci/pull/1820
- https://github.com/DataDog/datadog-ci/pull/1821
- https://github.com/DataDog/datadog-ci/pull/1822

This PR moves `src/constants.ts` and `src/helpers/*` to a new `@datadog/datadog-ci-base` package, which will be imported by all other packages.

> [!NOTE]  
> **For an easier review:** you can **unselect** the first 3 commits because they are moving _a lot_ of files. They are only the result of calling `mv`, using search & replace to rewrite the imports and calling `yarn format` to reorder the imports.

### How?

- Move all `src/helpers` and their unit tests to a new package
- Update all imports to point to the new package
- Do not break the current API contract (e.g. `src/index.ts`)
- Fix the CI to support multiple packages

### Review checklist


.github/workflows/ci.yml
```diff
@@ -45,7 +45,7 @@ jobs:
           NODE_OPTIONS: -r dd-trace/ci/init
       - run: |
           mkdir artifacts
-          yarn workspace @datadog/datadog-ci pack --filename ../../artifacts/datadog-ci-${{ matrix.version }}.tgz
+          yarn loop-packages pack --out ../../artifacts/%s-${{ matrix.version }}.tgz
```
Drarig29, line 48: See https://yarnpkg.com/cli/pack#examples

`%s` refers to the package name, so that the `artifacts` directory contains one tarball file per package

nsavoire (approved):

packages/base/package.json
```diff
@@ -0,0 +1,44 @@
...
+  },
+  "files": [
+    "dist/**/*"
+  ],
+  "scripts": {
+    "build": "yarn package:clean; yarn package:build",
+    "prepack": "yarn package:clean-dist"
+  },
+  "dependencies": {
+    "@types/datadog-metrics": "0.6.1",
```
GabrielAnca, line 17: Small thing: any reason why we don't have this in devDependencies? Given that this is a CLI, I would expect types to matter internally only

.gitignore
```diff
@@ -87,3 +87,6 @@ packages/datadog-ci/src/commands/react-native/__tests__/fixtures/with-sources-co
 !.yarn/releases
 !.yarn/sdks
 !.yarn/versions
+
+# Jest cache directories
+.jest-cache/
```
AntoineDona: Missing trailing newline here 😁

GabrielAnca (approved):
LGTM! Only one small comment, but nothing serious

AntoineDona (approved):
LGTM! 👍 just one micro nit

packages/base/package.json
Drarig29: @GabrielAnca it's a CLI but it can also be used as a library, see https://github.com/DataDog/datadog-ci/pull/476.

And since https://github.com/DataDog/datadog-ci/pull/476, we had to have `@types/datadog-metrics` as a dependency because we are exporting the `BufferedMetricsLogger` type from `@types/datadog-metrics`: https://github.com/DataDog/datadog-ci/pull/488

That's a common practice: https://stackoverflow.com/a/46011417/3970387

The alternative is to define our own interface and export it, but we didn't choose to do this

GabrielAnca: Thanks for the context Corentin! I didn't know it was used as a library too. Makes perfect sense then!

marco-saia-datadog (approved):

dastrong (approved):

ava-silver (approved):
lgtm from serverless!

jminuscula (approved):
LGTM for source code integration 👍

juli1 (approved):

ManuelPalenzuelaDD (approved):
LGTM for junit

highfivedenis (approved):
LGTM for orchestrating

BenoitZugmeyer (approved):

PR #1805, opened 2025-08-20
Add --base-path parameter to coverage upload command

nikita-tkachenko-datadog:
### What and why?

This PR adds `--base-path` parameter to `coverage upload` command.

### How?

### Review checklist

### DIFF
--- a/src/commands/coverage/README.md
+++ b/src/commands/coverage/README.md
@@ -21,11 +21,8 @@ datadog-ci coverage upload --tags key1:value1 --tags key2:value2 unit-tests/cove
 ```
 
 - The positional arguments are directories, files, or glob patterns that will be used when looking for coverage report files. If you pass a folder, the CLI will do a recursive search looking for supported coverage reports.
-- `--ignored-paths` a comma-separated list of paths that should be excluded from automatic reports discovery (only applicable when `--auto-discovery` is set). Glob patterns are supported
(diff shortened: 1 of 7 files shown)


src/commands/coverage/README.md
```diff
@@ -21,11 +21,8 @@ datadog-ci coverage upload --tags key1:value1 --tags key2:value2 unit-tests/cove
 ```
 
 - The positional arguments are directories, files, or glob patterns that will be used when looking for coverage report files. If you pass a folder, the CLI will do a recursive search looking for supported coverage reports.
-- `--ignored-paths` a comma-separated list of paths that should be excluded from automatic reports discovery (only applicable when `--auto-discovery` is set). Glob patterns are supported
-- `--tags` is an array of key value pairs of the shape `key:value`. This will set global tags applied to all coverage reports.
-  - The resulting dictionary will be merged with whatever is in the `DD_TAGS` environment variable. If a `key` appears both in `--tags` and `DD_TAGS`, whatever value is in `DD_TAGS` will take precedence.
-- `--measures` is an array of key numerical value pairs of the shape `key:123`. This will set global measures applied to all coverage reports.
```
Drarig29, line 27: Why were `--tags` and `--measures` removed from the docs?

nikita-tkachenko-datadog: The values are currently being ignored by the backend. I left the code in there for the future, but decided to hide them from the docs for the time being

PR #1827, opened 2025-09-08
[CHORE] create migration script

ava-silver:
### What and why?

### How?

### Review checklist


ava-silver:
* **#1827** <a href="https://app.graphite.dev/github/pr/DataDog/datadog-ci/1827?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/DataDog/datadog-ci/1827?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

bin/migrate.sh
```diff
@@ -0,0 +1,178 @@
...
+
+set -euo pipefail
+
+if [ -z "$1" ]; then
+  echo "Usage: $0 <scope>"
+  exit 1
+fi
+
+if ! command -v gsed >/dev/null 2>&1; then
+  brew install gnu-sed
```
GabrielAnca: Given that we may have linux users, can we do this only for macOS? I understand linux sed is generally okay

bin/migrate.sh
```diff
@@ -0,0 +1,178 @@
...
+    "inquirer-checkbox-plus-prompt": "^1.4.2",
+    "jest-diff": "^30.0.4",
+    "js-yaml": "3.13.1",
+    "ora": "5.4.1",
+    "packageurl-js": "^2.0.1",
+    "semver": "^7.5.3",
+    "simple-git": "3.16.0",
+    "typanion": "^3.14.0",
+    "upath": "^2.0.1",
+    "uuid": "^9.0.0"
```
GabrielAnca, line 99: These deps will vary based on the scope. This may be hard to figure out for the script, so I'd add a callout at the end to manually remove unneeded dependencies

bin/migrate.sh
```diff
@@ -0,0 +1,178 @@
...
+echo 9. Check if we can automatically knip
+yarn install
+if yarn workspace @datadog/datadog-ci-base build && yarn workspace "$PLUGIN_PKG" build && yarn workspace "$PLUGIN_PKG" lint --fix; then
+  echo "Linting passed, running knip..."
+  yarn knip || true
+  echo Done
+else
+  echo "Linting failed. Please fix the issues manually."
+fi
+git add -A
```
GabrielAnca, line 174: This may generate a PR that's hard to review because of moving and updating files in the same commit. What if the script creates a branch at the beginning and creates a commit with each of the steps 1/9?

bin/migrate.sh
Drarig29: Agreed, I'd prefer to make the script assume the `sed` is the GNU one, on any OS.

We can provide instructions for the user to ensure that `sed --version` outputs `sed (GNU sed)` on their machine.
And if it doesn't, tell them to `brew install gnu-sed` and add the following to their `~/.zshrc`:
```sh
export PATH="/opt/homebrew/opt/gnu-sed/libexec/gnubin:$PATH"
```

This setup will likely serve them in the future anyway (and TBH it should part of the laptop setup)

bin/migrate.sh
Drarig29: This is handled by `knip` already! 🤩 So these deps are initialized to `@datadog/datadog-ci`'s current deps, then will be trimmed by `knip` automatically

GabrielAnca: Ohh TIL about knip. That's awesome!

bin/migrate.sh
ava-silver: interesting idea, although some of the steps may be a no-op, and from my experience, the diff isn't actually too bad, since it's mostly just moves + find/replace

maybe this can be a consideration as a follow up task?

GabrielAnca: Yeah, happy if we want to try this out as it is. If reviews get too complex we can revisit 🙌🏻

bin/migrate.sh
ava-silver: Given most of our contributors are datadog employees, I'd like to make the easy path to automatically use (and install) gsed here where applicable. I'll only do this for MacOS users, so linux users can still use this as normal.

PR #1833, opened 2025-09-11
packages-migration: move git-metadata to base

ava-silver:
### What and why?

Since other packages rely on it, we need to move git-metadata to the base package.

In the future, we could do a refactor of the command to split out the functionality into helpers which can be consumed by other commands, and keep the command as a separate package -- but that's outside the scope of my current work for serverless package migration.

https://datadoghq.atlassian.net/browse/SYNTH-21708

### How?

Moved git-metadata to the base package, along with moving the instrumentation git helpers to the base package as well for convenience for future package migrations

### Review checklist


packages/base/src/commands/git-metadata/renderer.ts
```diff
@@ -1,6 +1,7 @@
-import {ICONS} from '@datadog/datadog-ci-base/helpers/formatting'
 import chalk from 'chalk'
 
+import {ICONS} from '@datadog/datadog-ci-base/helpers/formatting'
```
Drarig29: We can change the `@datadog/datadog-ci-base/...` import with a relative import now that we are in `packages/base` already

packages/base/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@datadog/datadog-ci-base",
-  "version": "0.0.1",
+  "version": "3.20.0",
```
Drarig29, line 3: I wanted to keep this as `0.0.1` for the first release, just in case. Then we would synchronise all the versions later when we are sure it's stable, but I guess it's not necessary... and I agree with your move of the version command to the base package.

So let's also change it in `plugin-synthetics`

cswatt (approved):

packages/base/src/helpers/git-instrument-helpers.ts
Drarig29, line 1: In https://github.com/DataDog/datadog-ci/pull/1804/commits/40c8e5471b1805f853cc803860f92e02fb0513a0 I had renamed this file from `src/helpers/git/instrument-helpers.ts` to `src/git-instrument-helpers.ts`.

But I think we can now rename it to `src/helpers/git/source-code-integration.ts`.

The only function that is really meant to be exported from the module to be used by other commands is `handleSourceCodeIntegration()`, so can we move it at the top of the file?

The 2 other exported functions (`getCurrentGitStatus()` and `uploadGitData()`) are only exported for unit tests, so you can add a comment along the lines of "// Only exported to be mocked in unit tests"

Drarig29 (approved):
LGTM

packages/base/src/helpers/git/source-code-integration.ts
```diff
@@ -1,10 +1,36 @@
...
+      extraTags = `git.commit.sha:${gitData.commitSha},git.repository_url:${gitData.gitRemote}`
+    }
+  } catch (err) {
+    context.stdout.write(renderSoftWarning(`Couldn't add source code integration, continuing without it. ${err}`))
+  }
+
+  return extraTags
+}
+
+
```
Drarig29: **nit** this one is not exported 😬

ava-silver: oop good catch

nsavoire (approved):

samarcan (approved):

PR #1835, opened 2025-09-11
[chore] Use Node.js SEA for standalone binary

Drarig29:
### What and why?

This PR drops https://github.com/vercel/pkg in favor of https://github.com/yao-pkg/pkg.

We just use `@yao-pkg/pkg` with the `--sea` option instead of going through [those steps](https://nodejs.org/api/single-executable-applications.html) ourselves for now.

We now use `esbuild` to bundle `@datadog/datadog-ci` **and "all" its plugins together**, so that we have a standalone binary with all plugins pre-installed. (There is only 1 plugin for now.)

This PR also adds a `datadog-ci plugin check <args>` command to troubleshoot the loading of plugins.

The standalone binary size increase is due to moving from Node 18 to Node 22 as the base binary. Reason here: https://github.com/nodejs/single-executable/issues/107

### How?

We now explicitly list all commands instead of dynamically `require()`ing the `cli.ts` files.

### Review checklist


nsavoire (approved):

GabrielAnca (approved):

duncanista (approved):

.github/workflows/ci.yml
```diff
@@ -201,7 +201,7 @@ jobs:
       - run: yarn install --immutable
       - run: yarn build
       - name: Create standalone binary
-        run: yarn dist-standalone -t node18-linux-x64 -o datadog-ci_linux-x64
+        run: yarn dist-standalone -t node22-linux-x64 -o datadog-ci_linux-x64
```
GabrielAnca, line 204: Non blocking: any reason not to bump this to latest (node24)

GabrielAnca (approved):

.github/workflows/ci.yml
Drarig29: We [aren't testing against node24 yet](https://github.com/DataDog/datadog-ci/blob/24f2971fd3e820f935a7926f68ad2dddac487418/.github/workflows/ci.yml#L12) so I preferred to use the first version that we support and that works correctly with SEA

GabrielAnca: Oh true, makes sense! Thanks for clarifying it!

PR #1821, opened 2025-09-04
[chore] Use source files during development for better DevX

Drarig29:
### What and why?

- https://github.com/DataDog/datadog-ci/pull/1804
- https://github.com/DataDog/datadog-ci/pull/1817
- https://github.com/DataDog/datadog-ci/pull/1820
- https://github.com/DataDog/datadog-ci/pull/1821 (← **you are here**)
- https://github.com/DataDog/datadog-ci/pull/1822

This PR improves the DevX during the package migration, so that using `yarn build` is not needed in order to run `yarn launch`.

It also makes VSCode's TS Server map imports to package `src` files instead of `dist` files, so that rebuilding is not needed every time a package is changed.

### How?

Use https://nodejs.org/api/packages.html#resolving-user-conditions to introduce a `development` conditional export.

### Review checklist

- [x] Feature or bugfix MUST have appropriate tests (unit, integration)

PR #1837, opened 2025-09-12
packages-migration: migrate aas package

ava-silver:
### What and why?

Migrate the AAS package: https://datadoghq.atlassian.net/browse/SYNTH-21700

### How?

First, I ran `./bin/migrate.sh aas`, then i split the command as noted [here](https://datadoghq.atlassian.net/wiki/spaces/dtdci/pages/5472846600/How+to+Split+a+command+scope+into+a+plugin+package#Split-command-classes), and then I refactored the base class to have the functional helpers in `package-aas`, where the options are all in the base package, along with some lightweight helpers needed throughout the command.

### Review checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -122,7 +122,7 @@ jobs:
               "@datadog/datadog-ci-base": "file:./artifacts/@datadog-datadog-ci-base-${{ matrix.version }}.tgz",
               "@datadog/datadog-ci-plugin-sarif": "file:./artifacts/@datadog-datadog-ci-plugin-sarif-${{ matrix.version }}.tgz",
               "@datadog/datadog-ci-plugin-sbom": "file:./artifacts/@datadog-datadog-ci-plugin-sbom-${{ matrix.version }}.tgz",
-              "@datadog/datadog-ci-plugin-aas": "file:./artifacts/@datadog/datadog-ci-plugin-aas-${{ matrix.version }}.tgz",
+              "@datadog/datadog-ci-plugin-aas": "file:./artifacts/@datadog-datadog-ci-plugin-aas-${{ matrix.version }}.tgz",
               "@datadog/datadog-ci-plugin-synthetics": "file:./artifacts/@datadog-datadog-ci-plugin-synthetics-${{ matrix.version }}.tgz"
             },
             "dependencies": {


packages/datadog-ci/tsconfig.json
```diff
@@ -11,6 +11,9 @@
       "path": "../base"
     },
     // List every plugin here, so they keep being built even when we drop them from `@datadog/datadog-ci` in the future.
```
Drarig29: Let's remove this line, I didn't catch it in an earlier PR, but it's wrong to have that comment here.

When we drop a plugin from `@datadog/datadog-ci` in the future, _we will_ remove it here, but keep it in the root `tsconfig.json`

package.json
```diff
@@ -62,7 +62,6 @@
     "prettier": "^3.6.2",
     "rimraf": "^5.0.0",
     "ts-jest": "29.1.1",
-    "ts-json-schema-generator": "^2.4.0",
```
Drarig29: I think I removed a script by mistake when resolving conflicts... can you bring back [this script](https://github.com/DataDog/datadog-ci/pull/1821/files#r2344631035)?

```json
"generate-json-schemas:synthetics": "ts-json-schema-generator --markdown-description --no-type-check --no-top-ref --additional-properties --path packages/plugin-synthetics/src/interfaces.ts --type TestConfig > packages/plugin-synthetics/src/test-config.schema.json && prettier --write packages/plugin-synthetics/src/test-config.schema.json",
```

.github/workflows/ci.yml
```diff
@@ -120,6 +120,7 @@ jobs:
             "name": "datadog-ci-e2e-tests",
             "resolutions": {
               "@datadog/datadog-ci-base": "file:./artifacts/@datadog-datadog-ci-base-${{ matrix.version }}.tgz",
+              "@datadog/datadog-ci-plugin-aas": "file:./artifacts/@datadog/datadog-ci-plugin-aas-${{ matrix.version }}.tgz",
```
Drarig29:
```suggestion
              "@datadog/datadog-ci-plugin-aas": "file:./artifacts/@datadog-datadog-ci-plugin-aas-${{ matrix.version }}.tgz",
```

Drarig29: gotta love those tiny characters 😁

ava-silver: ahh! gotta update the script :)

packages/datadog-ci/tsconfig.json
ava-silver: wait -- do we keep any of the plugins here? or not

the comment i was using as an anchor to add the additional plugins in, so if we still want to add plugins here i'll need to use a different comment

Drarig29: > wait -- do we keep any of the plugins here? or not

yes, until the major release where we will drop the largest plugins only

ava-silver: okay, so i'll just change the comment to
```
    // List every plugin here.
```

if that's ok?

Drarig29: Maybe `// Add plugins here.`, just to avoid "every"

avangelillo (approved):

PR #1852, opened 2025-09-18
[PACKAGES-MIGRATION] Migrate tag command to base package

GabrielAnca:
### What and why?

This command is simple and used by the `deployment mark` command, so it needs to be moved to the base package. In the future we should refactor the command so that there is a common `tag` helper used by both the `tag` command and the `deployment mark` command.

This PR is required for me to migrate the `deployment` scope to a plugin

### How?

Following the guide.

### Review checklist


packages/base/src/commands/tag/__tests__/tag.test.ts
```diff
@@ -69,7 +71,6 @@ describe('execute', () => {
       {
         BUILDKITE: 'true',
         BUILDKITE_BUILD_ID: 'id',
-        BUILDKITE_JOB_ID: 'id',
```
GabrielAnca, line 72: Something changed and this test started timing out because it was hanging trying to call the API (which is not mocked). I suspect that timeout handled differently before and because this test only ensures that there is no output, it just passed. I removed this line to make the job invalid, so that the test is still useful: it verifies that there is no output when `--silent` is present.

Aaron-9900 (approved):

PR #1854, opened 2025-09-19
[PACKAGES-MIGRATION] Migrate dora scope into a plugin

GabrielAnca:
### What and why?

Migrates `dora` scope into a plugin 

### How?

Using the `bin/migrate` script and [this guide](https://datadoghq.atlassian.net/wiki/spaces/dtdci/pages/5472846600/How+to+Split+a+command+scope+into+a+plugin+package)

### Review checklist


bin/migrate.sh
```diff
@@ -167,6 +167,7 @@ yarn install
 if yarn workspace @datadog/datadog-ci-base build && yarn workspace "$PLUGIN_PKG" build && yarn workspace "$PLUGIN_PKG" lint --fix; then
   echo "Linting passed, running knip..."
   yarn knip || true
+  yarn
```
GabrielAnca, line 170: Tiny fix: `yarn.lock` is not updated after running knip, so we need to run `yarn` now

packages/datadog-ci/src/commands/dora/__tests__/deployment.test.ts
```diff
@@ -31,7 +31,6 @@ describe('execute', () => {
     })
 
     test('with all parameters provided', async () => {
-      /* eslint-disable prettier/prettier */
```
GabrielAnca, line 34: The next line already disabled prettier for the block, so this is not needed. I think disabling prettier for readability in the next block is good

michaelcretzman (approved):

PR #1855, opened 2025-09-22
Fix failed Windows tests and timeout

GabrielAnca:
### What and why?

This test started failing with timeouts trying to make API calls. I have no idea how this test ever worked because the API calls weren't mocked.

### How?

There are already tests for all the api functionality in `gitdb.test.ts`, so I'm mocking the gitdb library for this test

### Review checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,7 +65,7 @@ jobs:
         version: [18, 20, 22]
 
     name: Unit tests (Windows)
-    runs-on: windows-latest
+    runs-on: windows-2022
     env:
       FORCE_COLOR: 1
 


.github/workflows/ci.yml
```diff
@@ -65,7 +65,7 @@ jobs:
         version: [18, 20, 22]
 
     name: Unit tests (Windows)
-    runs-on: windows-latest
+    runs-on: windows-2022
```
GabrielAnca, line 68: The folks in github are doing a slow rollout of Windows Server 2025 this month (https://github.com/actions/runner-images/issues/12677). They are basically changing the alias `windows-latest` from 2022 to 2025. Version 2025 causes several issues in our CI. We use `binutils` and we get this error:

```
Command failed: objcopy -I elf64-little --only-keep-debug --compress-debug-sections=zstd --remove-section=.gdb_index ./src/commands/elf-symbols/__tests__/fixtures/dyn_aarch64 C:\Users\RUNNER~1\AppData\Local\Temp/2141447330357995.8/dyn_aarch64.debug
    objcopy: --compress-debug-sections=zstd: binutils is not built with zstd support
```

The end goal should be for us to support Windows 2025 but at this point I think we should first unlock the CI to allow new commits, and we can fix this in parallel

PR #1859, opened 2025-09-23
[chore] Symlink README and duplicate LICENSE files

Drarig29:
### What and why?

This PR prepares the release of all packages by duplicating the LICENSE file in every package, and it moves the root `README.md` to the `packages/datadog-ci` folder then symlinks the root README to the package README.

We can't symlink the other way around, because `npm pack` ignores symlinks and would not include the `packages/datadog-ci` README if it was a symlink. (see https://github.com/npm/cli/issues/6746)

I found another repo doing it: https://github.com/SBoudrias/Inquirer.js/blob/main/README.md

In GitHub, it will look like this: 

image

### How?

Copy licenses, update migration script, move the root readme, then run `ln -s packages/datadog-ci/README.md README.md`

### Review checklist


joepeeples (approved):

GabrielAnca (approved):

packages/base/LICENSE
```diff
@@ -0,0 +1,201 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright [yyyy] [name of copyright owner]
```
AntoineDona: Shouldn't we have `Copyright 2020 Datadog, Inc.` in every package ?

AntoineDona (approved):
Looks good! just left one small comment 😉

packages/base/LICENSE
GabrielAnca: Based on feedback received from the open source team we should have something like `Copyright 2020-Present Datadog, Inc.` 🤔

Drarig29: Good catch, thanks! We never did it apparently

PR #1876, opened 2025-09-26
[chore] Fix plugin `peerDependencies` for PNPM

### LINKED ISSUE #1870: PNPM add @datadog/datadog-ci fails to install (v3.21.1+)
### Bug description

Hi team,

The following two commands hang across a wide variety of environments and appear to be unable to download or install globally based on NPM records:

Image

* `pnpm add @datadog/datadog-ci@v3.21.1`
* `pnpm add @datadog/datadog-ci@v3.21.2`

The result is the same:

Image

Switching back to `v3.20.0` works fine.  This broke our deploy process and entire CI/CD pipeline (we install datadog dependencies only during deployed environments).

### Describe what you expected

To be able to install the library.

### Steps to reproduce the issue

_No response_

### Additional context

_No response_

### Command

None

Drarig29:
### What and why?

Closes #1870 

This PR changes the peer dependency in all plugins to be `@datadog/datadog-ci-base` instead of `@datadog/datadog-ci`.

It looks like the problem was that we had **both** a circular reference and a **mix between dep and peer dep**:
```mermaid
graph LR
    A["@datadog/datadog-ci-base"] -->|peerDependency| B["@datadog/datadog-ci-plugin-*"]
    B -->|dependency| A
```

A similar issue was fixed in PNPM, but with circular references involving **peer deps only**: https://github.com/pnpm/pnpm/pull/8755.

We can use this fix by sticking to peer deps only in our circular reference:
```mermaid
graph LR
    A["@datadog/datadog-ci-base"] -->|peerDependency| B["@datadog/datadog-ci-plugin-*"]
    B -->|peerDependency| A
    
    linkStyle 1 stroke:#333,stroke-width:1px,color:#ff6b6b
```

Let's keep this circular reference for now for a quick fix without a big refactor. We can try to solve it in the future.

Tested this PR with NPM, Yarn (PnP and `node_modules`) and PNPM.

### How?

- Move `@datadog/datadog-ci-base` from normal dep to peer dep in all plugins
- Remove `@datadog/datadog-ci` from peer deps in all plugins

### Review checklist


AntoineDona (approved):
Discussed in person, LGTM ! 👍

packages/base/src/cli.ts
```diff
@@ -27,6 +29,7 @@ import {TagCommand} from './commands/tag/tag-command'
 
 // prettier-ignore
 export const commands = {
+  'aas': [AasInstrumentCommand, AasUninstrumentCommand],
```
teodor2312, line 32: what is AAS?

packages/base/src/helpers/plugin.ts
```diff
@@ -11,11 +12,17 @@ import {messageBox} from './message-box'
 export type PluginPackageJson = {name: string; version: string}
 export type PluginSubModule = {PluginCommand: CommandClass}
 
+// Use `DEBUG=plugins` to enable debug logs
+const debug = createDebug('plugins')
```
teodor2312, line 16: Why do we need debug level logs?

packages/plugin-aas/src/cli.ts
```diff
@@ -1,4 +0,0 @@
-import {PluginCommand as InstrumentCommand} from './commands/instrument'
-import {PluginCommand as UninstrumentCommand} from './commands/uninstrument'
-
-module.exports = [InstrumentCommand, UninstrumentCommand]
```
teodor2312, line 4: Why are we deleting all of these files?

teodor2312 (approved):
LGTM with some questions for my understanding :)

packages/base/src/cli.ts
Drarig29: It's one of the command scopes: Azure App Service

packages/base/src/helpers/plugin.ts
Drarig29: I think it's a good practice and it helped me debug what was happening. This should be helpful for users reporting issues too

packages/plugin-aas/src/cli.ts
Drarig29: Sorry, those changes are unrelated. I found out `aas` was not listed anymore (probably due to a past conflict resolution), then I noticed this file was extraneous for AAS, then found out it was still present in all plugins. So I diverged a bit

PR #1886, opened 2025-10-03
Add `plugin list` command

Drarig29:
### What and why?

**Related PRs:**
- https://github.com/DataDog/datadog-ci/pull/1887
- https://github.com/DataDog/datadog-ci/pull/1886
- https://github.com/DataDog/datadog-ci/pull/1890
- https://github.com/DataDog/datadog-ci/pull/1891

This PR adds a `plugin list` command that lists the plugins that are **NOT built-in**, and thus installable separately by the user. For now, there is none. But after next major version, there will be `aas`, `cloud-run`, `lambda`, `stepfunctions` and `synthetics`.

### How?

Set a custom `CommandContext` for all commands thanks to a `BaseCommand`. This is required to have the right typings.

Thanks to this, the `@datadog/datadog-ci` package can pass any kind of information globally to all commands of the CLI – including plugin commands – through `this.context`.

For now, I only added `this.context.builtinPlugins`, but we may add more in the future.

### Review checklist

### DIFF
--- a/packages/base/src/commands/plugin/__tests__/list.test.ts
+++ b/packages/base/src/commands/plugin/__tests__/list.test.ts
@@ -1,0 +1,48 @@
+import {Cli} from 'clipanion'
+
+import {createMockContext} from '../../../helpers/__tests__/testing-tools'
+import {listAllPlugins} from '../../../helpers/plugin'
+
+import {CommandContext} from '../../..'
+import {PluginListCommand} from '../list'
+
+describe('PluginListCommand', () => {
+  test.each([
+    ['', 'All plugins are currently built-in. We will start splitting them in next major release.\n'],
+    [' (--json)', '[]\n'],
+  ])('all plugins are built-in%s', async (jsonArg, expectedOutput) => {
+    const cli = new Cli<CommandContext>()
+    cli.register(PluginListCommand)
+
+    const context = createMockContext() as CommandContext
+    context.builtinPlugins = listAllPlugins()
+
+    const code = await cli.run(['plugin', 'list', ...(jsonArg ? ['--json'] : [])], context)
+
+    expect(context.stdout.toString()).toStrictEqual(expectedOutput)
+    expect(code).toBe(0)
+  })
+
+  test.each([
+    [
+      '',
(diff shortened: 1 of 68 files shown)


packages/base/src/commands/plugin/__tests__/list.test.ts
```diff
@@ -0,0 +1,48 @@
...
+import {listAllPlugins} from '../../../helpers/plugin'
+
+import {CommandContext} from '../../..'
+import {PluginListCommand} from '../list'
+
+describe('PluginListCommand', () => {
+  test.each([
+    ['', 'All plugins are currently built-in. We will start splitting them in next major release.\n'],
+    [' (--json)', '[]\n'],
+  ])('all plugins are built-in%s', async (jsonArg, expectedOutput) => {
```
GabrielAnca: nit: These tests may be a bit easier to follow if we use an object rather than an array for the test cases

packages/base/src/commands/plugin/list.ts
```diff
@@ -0,0 +1,68 @@
...
+export class PluginListCommand extends BaseCommand {
+  public static paths = [['plugin', 'list']]
+
+  public static usage = Command.Usage({
+    category: 'Plugins',
+    description: 'List the available plugins.',
+    details: `
+      This command lists the plugins that can be installed with the \`datadog-ci plugin install\` command.
+
+      All other plugins are considered **built-in** and are not listed.
```
GabrielAnca: Nit: Using `considered` makes me thing that they really aren't built in but we somehow make them built-in. I'd say they are built-in

```suggestion
      All other plugins are **built-in** and are not listed.
```

GabrielAnca (approved):
Looks great! Two minor comments

packages/base/src/commands/plugin/__tests__/list.test.ts
Drarig29: true, I was lazy 🙇

packages/base/src/commands/plugin/list.ts
Drarig29: good point, let's make it simple

nsavoire (approved):

PR #1902, opened 2025-10-07
[FIX] cloud run env var secret support

ava-silver:
### What and why?

### How?

### Review checklist


packages/plugin-cloud-run/src/types.ts
```diff
@@ -2,8 +2,20 @@
...
+        secretKeyRef?:
+          | {
+              secret?: string | null | undefined
+              version?: string | null | undefined
+            }
+          | null
+          | undefined
+      }
+    | null
+    | undefined
```
ava-silver, line 18: this was pulled from the library types rather than just a rough approximation

packages/plugin-cloud-run/src/commands/instrument.ts
```diff
@@ -273,59 +273,40 @@ export class PluginCommand extends CloudRunInstrumentCommand {
...
+    const setEnv = (name: string, value: string | undefined) => {
+      if (name && value) {
+        newEnvVars[name] = {name, value}
       }
     }
 
-    const newEnv: IEnvVar[] = Object.entries(newEnvVars).map(([name, value]) => ({name, value}))
+    // Overwrite existing env vars with these if they already exist
+    // and add them to the container if they don't exist yet.
+    setEnv(API_KEY_ENV_VAR, process.env[API_KEY_ENV_VAR])
```
TalUsvyatsky, line 296: Is this a behavior change (if `API_KEY_ENV_VAR` is unset)?

TalUsvyatsky: Similar question for `LOGS_PATH_ENV_VAR` and potentially `SERVICE_ENV_VAR`

ava-silver: the API key is required -- the command will fail to validate without it

ava-silver: for logs path, that's passed in as a CLI flag and has a default value

for service, we automatically populate it if unset

ava-silver: so no -- those will all remain the same, setting the variable will always happen

TalUsvyatsky: 👌 perfect, that's what I was looking for

PR #1904, opened 2025-10-07
[UPDATE] type definitions to use sdk

ava-silver:
### What and why?

### How?

### Review checklist


packages/plugin-cloud-run/tsconfig.json
```diff
@@ -1,6 +1,7 @@
 {
   "extends": "../../tsconfig.base.json",
   "compilerOptions": {
+    "skipLibCheck": true,
```
Drarig29, line 4: Now that plugins are separate packages with their own build settings, I'm OK with setting `skipLibCheck: true` for this plugin only, since it doesn't expose any `index.ts` file and thus has no public library API, so it shouldn't be consumed as a library.

For `@datadog/datadog-ci`, `@datadog/datadog-ci-base` and `@datadog/datadog-ci-plugin-synthetics` however, we'll stick to the rationale here: https://github.com/DataDog/datadog-ci/pull/1249#discussion_r1543119046

PR #1911, opened 2025-10-13
Extract Github job name automatically for `tag`, `measure`, `trace`, `span` and `deployment mark`

rodrigo-roca:
### What and why?

We want to remove the requirement of exposing the `DD_GITHUB_JOB_NAME` environment variable manually when correlating data with Github jobs in order to improve the UX. Instead, the command will try to determine the Github job name automatically. Customers can still pass that variable which takes more priority (we will not do this logic in that case).

This affects the `tag`, `measure`, `trace`, `span` and `deployment mark` commands.

### How?

We do workaround where we search the diagnostic logs in the Github Actions worker and apply a regex to find the job display name. It is not ideal but we have not found any other way to achieve this since Github does not expose this in the environment variables.

### Review checklist

### DIFF
--- a/packages/base/src/helpers/__tests__/ci.test.ts
+++ b/packages/base/src/helpers/__tests__/ci.test.ts
@@ -4,7 +4,6 @@ import {BaseContext} from 'clipanion'
 import upath from 'upath'
 
 import {
-  envDDGithubJobName,
   getCIEnv,
   getCIMetadata,
   getCISpanTags,


packages/base/src/helpers/ci.ts
```diff
@@ -56,13 +62,18 @@ export const CI_ENGINES = {
...
 
 // DD_GITHUB_JOB_NAME is an override that is required for adding custom tags and metrics
 // to GHA jobs if the 'name' property is used. It's ok for it to be missing in case the name property is not used.
-const envAllowedToBeMissing = ['DD_GITHUB_JOB_NAME']
+const envAllowedToBeMissing = [envDDGithubJobName]
+
+export const githubWellKnownDiagnosticDirs = [
+  '/home/runner/actions-runner/cached/_diag', // for SaaS
+  '/home/runner/actions-runner/_diag', // for self-hosted
+]
```
Drarig29, line 74: Can you try on Windows too? Maybe in a separate PR

packages/plugin-deployment/src/__tests__/mark.test.ts
```diff
@@ -1,15 +1,18 @@
-import {
-  createCommand,
-  getEnvVarPlaceholders,
-  makeRunCLI,
-} from '@datadog/datadog-ci-base/helpers/__tests__/testing-tools'
+import fs from 'fs'
+import {afterEach} from 'node:test'
```
Drarig29: We don't need this import – `afterEach` is part of Jest globals

packages/base/src/helpers/ci.ts
```diff
@@ -1020,3 +1031,81 @@ const filterEnv = (values: string[]): Record<string, string> => {
...
+/**
+ * Extracts the job display name from the GitHub Actions diagnostic log files.
+ *
+ * @returns The job display name, or undefined if not found
+ */
+export const getGithubJobNameFromLogs = (context: BaseContext): string | undefined => {
+  if (!shouldGetGithubJobDisplayName()) {
+    return
+  }
+  context.stdout.write('Determining github job name\n')
```
Drarig29:
```suggestion
  context.stdout.write('Determining GitHub job name\n')
```

packages/base/src/helpers/ci.ts
```diff
@@ -1020,3 +1031,81 @@ const filterEnv = (values: string[]): Record<string, string> => {
...
+        foundDiagDir = currentDir
+        workerLogFiles = potentialLogs
+        break
+      }
+    } catch (error) {
+      // If the directory does not exist, just try the next one
+      if (error instanceof Error && 'code' in error && error.code === 'ENOENT') {
+        continue
+      }
+      let errMessage = 'error reading Github diagnostic log files'
```
Drarig29:
```suggestion
      let errMessage = 'error reading GitHub diagnostic log files'
```

packages/base/src/helpers/ci.ts
```diff
@@ -1020,3 +1031,81 @@ const filterEnv = (values: string[]): Record<string, string> => {
...
+      } else {
+        errMessage += `: ${String(error)}`
+      }
+      context.stderr.write(`${chalk.yellow.bold('[WARNING]')} ${errMessage}`)
+
+      return
+    }
+  }
+  if (workerLogFiles.length === 0 || foundDiagDir === '') {
+    context.stderr.write(`${chalk.yellow.bold('[WARNING]')} could not find Github diagnostic log files`)
```
Drarig29:
```suggestion
    context.stderr.write(`${chalk.yellow.bold('[WARNING]')} could not find GitHub diagnostic log files`)
```

packages/base/src/helpers/ci.ts
```diff
@@ -1020,3 +1031,81 @@ const filterEnv = (values: string[]): Record<string, string> => {
...
+  // 2. Get the job display name via regex
+  for (const logFile of workerLogFiles) {
+    const filePath = upath.join(foundDiagDir, logFile)
+    const content = fs.readFileSync(filePath, 'utf-8')
+
+    const match = content.match(githubJobDisplayNameRegex)
+
+    if (match && match[1]) {
+      // match[1] is the captured group with the display name
+      context.stdout.write(`Succesfully extracted job name: ${match[1]}\n`)
```
Drarig29:
```suggestion
      context.stdout.write(`Successfully extracted job name: ${match[1]}\n`)
```

packages/base/src/helpers/ci.ts
```diff
@@ -1020,3 +1031,81 @@ const filterEnv = (values: string[]): Record<string, string> => {
...
+
+      return match[1]
+    }
+  }
+
+  context.stderr.write(
+    `${chalk.yellow.bold('[WARNING]')} could not find "jobDisplayName" attribute in Github diagnostic logs`
```
Drarig29:
```suggestion
    `${chalk.yellow.bold('[WARNING]')} could not find "jobDisplayName" attribute in GitHub diagnostic logs`
```

packages/base/src/helpers/ci.ts
Drarig29: According to [this line](https://github.com/github-aws-runners/terraform-aws-github-runner/blob/67fadae8e978fd5e5ec981a44f46e19cedae6702/modules/runners/logging.tf#L21) in a different project, I think you can add `/opt/actions-runner/_diag` to the list, and `C:/actions-runner/_diag` for Windows.

This directory does show up in [some GitHub issues](https://github.com/search?q=+%22C%3A%2Factions-runner%2F_diag%22&type=issues).

packages/base/src/commands/tag/__tests__/tag.test.ts
```diff
@@ -92,4 +98,59 @@ describe('execute', () => {
...
+        GITHUB_RUN_ID: '123',
+        GITHUB_RUN_ATTEMPT: '1',
+        GITHUB_JOB: 'fake job name',
+      },
+      ['--dry-run']
+    )
+    expect(result.code).toBe(0)
+    const out = result.context.stdout.toString()
+    expect(out).not.toContain('Determining github job name')
+    expect(out).not.toContain('"DD_GITHUB_JOB_NAME": "real job name"')
```
juan-fernandez: nitpick: a slightly better test is
```suggestion
    expect(out).not.toContain('DD_GITHUB_JOB_NAME')
```
as the `DD_GITHUB_JOB_NAME` could be a wrongly calculated value (different from `real job name`)  which would not be caught by this test

packages/datadog-ci/src/commands/trace/helper.ts
```diff
@@ -53,7 +53,9 @@ export abstract class CustomSpanCommand extends BaseCommand {
 
       return 1
     }
-    const ciSpanTags = getCISpanTags()
+    const realGithubJobName = getGithubJobNameFromLogs(this.context)
```
juan-fernandez, line 56: this means that this is executed regardless of whether the user has set `DD_GITHUB_JOB_NAME`, right? This seems like a wasted effort, doesn't it? As the value is not used at all if the env var is there: https://github.com/DataDog/datadog-ci/pull/1911/files#diff-dc1fcf04816c2d334efaa97fa563c1431218db229a8e620d4023b5925304bafeR345

packages/base/src/helpers/ci.ts
```diff
@@ -1020,3 +1031,92 @@ const filterEnv = (values: string[]): Record<string, string> => {
...
+      } else {
+        errMessage += `: ${String(error)}`
+      }
+      context.stderr.write(`${chalk.yellow.bold('[WARNING]')} ${errMessage}`)
+
+      return
+    }
+  }
+  if (workerLogFiles.length === 0 || foundDiagDir === '') {
+    context.stderr.write(`${chalk.yellow.bold('[WARNING]')} could not find GitHub diagnostic log files`)
```
juan-fernandez, line 1094: should we return here? Otherwise we'll print both this and `could not find "jobDisplayName" attribute in GitHub diagnostic logs` which seems confusing

packages/base/src/helpers/ci.ts
```diff
@@ -1020,3 +1031,92 @@ const filterEnv = (values: string[]): Record<string, string> => {
 export const isInteractive = ({stream = process.stdout}: {stream?: NodeJS.WriteStream} = {}) => {
   return Boolean(!('CI' in process.env) && process.env.TERM !== 'dumb' && stream && stream.isTTY)
 }
+
+export const shouldGetGithubJobDisplayName = (): boolean => {
+  return getCIProvider() === CI_ENGINES.GITHUB && process.env.DD_GITHUB_JOB_NAME === undefined
+}
+
+export const getGithubJobNameFromLogsAndUpdateEnv = (context: BaseContext, ciEnv: Record<string, string>) => {
```
juan-fernandez: do we need this function as it just calls `getGithubJobNameFromLogs` and updates a dictionary? It does not seem useful, as we could simply return the value from `getGithubJobNameFromLogs` and update the dictionary from outside?

packages/datadog-ci/src/commands/trace/helper.ts
rodrigo-roca: We do a check for that env variable there too here. 
https://github.com/DataDog/datadog-ci/blob/61135e09857a91fe226608667d40cbf810785f03/packages/base/src/helpers/ci.ts#L1063

I understand why it can be confusing tho, maybe I shovelled too much logic with the purpose of not repeating that if statement everywhere

packages/base/src/helpers/ci.ts
rodrigo-roca: makes sense, good catch!

packages/base/src/helpers/ci.ts
rodrigo-roca: mmh you are right maybe I tried to reuse too much, will refactor this

packages/base/src/helpers/ci.ts
rodrigo-roca: thank you! I think I will do a follow up PR for this

PR #1924, opened 2025-10-20
[Cloud Run Instrument] Set label `service` to DD_SERVICE value

nhulston:
### What and why?

Add a label to Cloud Run services `service: <DD_SERVICE>`

Used by the GCP Integration and frontend

Screenshot 2025-10-20 at 1 23 23 PM

### Review checklist


ava-silver (approved):
lgtm, we should also add `env` and `version` for the same reason, either in this PR or as a follow up

PR #1943, opened 2025-10-31
[SVLS] refactor container apps and cloud run to use the same logic

ava-silver:
### What and why?

We have a handful of subtle bugs caused by trying to implement the same thing multiple times (primarily across cloud run and container apps), this consolidates that logic and ensures that env vars are similarly deduplicated for AAS as well.

Each service now pulls the base env vars, and then adds any product specific ones.

Container apps now use a diff based comparison like cloud run, and both share the same instrumentation logic since they both use K8s under the hood.

Also mocked an unmocked AWS function which now reduces flakiness/latency in lambda tests

### How?

Consolidates shared logic under `packages/base/src/helpers/serverless/common.ts`, and refactors the three gcp/azure commands to use the helpers.

Consolidates shared serverless code under `packages/base/src/helpers/serverless`

Remains backwards compatible for cloud run (only adds additional env vars which will have no effect, this is intentional to let us add custom env vars like we do for aas/container apps)

### Follow up work

In future PRs:
- add startup probe to container apps like we do for container apps
- add custom env var support for cloud run (eg for specifying additional agent config)
- add a flag on all commands to fully disable logging support

### Review checklist

### DIFF
--- a/packages/base/src/helpers/serverless/common.ts
+++ b/packages/base/src/helpers/serverless/common.ts
@@ -1,3 +1,5 @@
+import {isDeepStrictEqual} from 'node:util'
+
 import chalk from 'chalk'
 import {diff} from 'jest-diff'
 
@@ -116,7 +118,7 @@ export const sortedEqual = (a: any, b: any): boolean => {
   const sortedA = sortObject(a)
   const sortedB = sortObject(b)
 
-  return JSON.stringify(sortedA) === JSON.stringify(sortedB)
+  return isDeepStrictEqual(sortedA, sortedB)
 }
 
 /**


packages/plugin-cloud-run/src/__tests__/__snapshots__/instrument.test.ts.snap
ava-silver, line 1: diffs are now sorted (stable) and env vars are only added in places they won't matter (ie sidecar env vars on the main container or vice versa), so it's not a breaking change

packages/base/src/helpers/serverless/common.ts
```diff
@@ -71,11 +74,213 @@ export const getBaseEnvVars = (config: ServerlessConfigOptions): Record<string,
...
+        sorted[key] = sortObject(obj[key])
+      })
+
+    return sorted
+  }
+
+  return obj
+}
+
+export const sortedEqual = (a: any, b: any): boolean => {
```
avangelillo, line 117: Is there a reason to do a sorted equal here instead of something like a deep equal?

packages/base/src/helpers/serverless/common.ts
```diff
@@ -71,11 +74,213 @@ export const getBaseEnvVars = (config: ServerlessConfigOptions): Record<string,
...
+export const generateConfigDiff = (original: any, updated: any): string => {
+  // Sort keys consistently before comparison
+  const sortedOriginal = sortObject(original)
+  const sortedUpdated = sortObject(updated)
+
+  const originalJson = JSON.stringify(sortedOriginal, undefined, 2)
+  const updatedJson = JSON.stringify(sortedUpdated, undefined, 2)
+
+  const obfuscatedOriginal = originalJson.split('\n').map(obfuscateSensitiveValues).join('\n')
+  const obfuscatedUpdated = updatedJson.split('\n').map(obfuscateSensitiveValues).join('\n')
```
avangelillo, line 149: This obfuscates the values then does a diff.  Does this show any changes for an API key swap?

packages/base/src/helpers/serverless/common.ts
nhulston: I think we can only compare equality by reference but not deep equal. Unless we add another package

ava-silver: to clarify, this is a sorted deep equal, meaning all keys and lists are sorted -- necessary to check if the resource schema would be different from what we want

if the name is confusing, i could rename to "sortedDeepEqual" or something similar

packages/base/src/helpers/serverless/common.ts
ava-silver: ooh i guess not, this is moved from the cloud run helpers previously, so i didn't make any changes here

but that's a good edge case that is worth addressing in a follow up PR!

packages/base/src/helpers/serverless/common.ts
```diff
@@ -71,11 +74,213 @@ export const getBaseEnvVars = (config: ServerlessConfigOptions): Record<string,
...
+  }
+
+  return configDiff
+}
+
+export const byName = <T extends FullyOptional<{name: string}>>(xs: T[]): Record<string, T> => {
+  return Object.fromEntries(xs.filter((x) => x.name).map((x) => [x.name, x]))
+}
+
+const DEFAULT_HEALTH_CHECK_PORT = 5555
```
nhulston: Micro nit: maybe this should be in the new serverless constants file?

ava-silver: ooh yep!

packages/base/src/helpers/serverless/common.ts
avangelillo: I see that it does do deep equality, but because this compares stringified versions of the object it can be surprising sometimes.  Some extra things get pulled to null, like:

```
> JSON.stringify(null) === JSON.stringify(NaN)
true
```

There might be other things I'm not thinking of, but I'd just import `isDeepStrictEqual` from `node:util` since its probably already there and we don't need to reimplement and think of all the edge cases for this

ava-silver: how am i just now finding out about `isDeepStrictEqual` 😭 

i'll absolutely switch to that, but I'd like to keep the sorting since we wanna make sure any lists are also sorted for stability, since the APIs this is being used against assume all lists are unordered.

PR #1947, opened 2025-11-03
[npm] Use OIDC to publish

yoannmoinet:
### What and why?

Publish to NPM using OIDC instead of a token.

### How?

A few things were necessary to algin everything:

- Update yarn to 4.10.* so it supports OIDC.
- Use Node 24+ to publish, so it supports OIDC.
- Small tweaks to the publish scripts.
- Delete pre-release process as it never ran, and NPM only supports 1 action.

Configuration was changed on NPM's side for [these packages](https://www.npmjs.com/search?q=author%3Adatadog%20keywords%3Adatadog-ci).

Once merged, I'll remove the token from the repo too.


package.json
```diff
@@ -23,7 +23,7 @@
     "package:clean": "cd $INIT_CWD && rimraf dist tsconfig.tsbuildinfo",
     "package:clean-dist": "cd $INIT_CWD && rimraf --glob 'dist/**/__tests__'",
     "package:lint": "eslint --cache --quiet $INIT_CWD/src/**/*.ts",
-    "publish:all": "yarn loop-packages --topological --verbose npm publish --provenance",
+    "publish:all": "yarn loop-packages --topological --verbose npm publish",
```
ava-silver, line 26: for my own understanding, why are we getting rid of `--provenance`? is this related to:
> Configuration was changed on NPM's side for [these packages](https://www.npmjs.com/search?q=author%3Adatadog%20keywords%3Adatadog-ci).

Also, what's the path going forward for new packages, will any additional steps be needed?

yoannmoinet: The `--provenance` flag is not necessary anymore, the link is done automatically by NPM now that we've identified the workflow file where the publish happens.

For new packages, they will need to be configured on NPM's side too.

ava-silver: can we make sure that the process for new packages is documented properly then? (Since I'm working on a new package currently)

ava-silver (approved):

package.json
yoannmoinet: Totally yes.
I'll send you a link via DM for now, I'm not sure exactly how to configure a "new" package though, so it will be TDB once you're there.

Drarig29 (approved):
LGTM

PR #1952, opened 2025-11-06
[synthetics] Export type of JUnit report files

hegrec:
### What and why?

In v3.x.x, there was no package.json exports field, meaning that consumers of datadog-ci could dig into dist files arbitrarily and use exported values.

As of 4.x.x, this is no longer the case, with the splitting out of the plugins, their new respective package.json files now use the more modern "exports" format instead of solely relying on a "main" entrypoint.

With the usage of "exports" comes the encapsulation of code not explicitly exported from package.json or transitively through the already present exported files.

My team has code relying on some types in the junit reporter, and by adding an export for this file, we are able to once again consume the exported types there.

This will enable us to upgrade from 3.x.x to 4.x.x.

### How?

I modified the package.json and tested locally using yarn portals to make my other repository depend on my local working copy of datadog-ci. I verified that the types, such as XMLJSON were able to be imported via the path: `@datadog/datadog-ci-plugin-synthetics/reporters/junit`

### Review checklist

### DIFF
--- a/packages/plugin-synthetics/src/index.ts
+++ b/packages/plugin-synthetics/src/index.ts
@@ -4,6 +4,7 @@ export {CiError, CriticalError} from './errors'
 export * from './interfaces'
 export {DefaultReporter} from './reporters/default'
 export {JUnitReporter} from './reporters/junit'
+export type {XMLJSON} from './reporters/junit'
 export {executeTests, execute} from './run-tests-lib'
 export * as utils from './utils/public'
 


hegrec:
* **#1952** <a href="https://app.graphite.com/github/pr/DataDog/datadog-ci/1952?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/datadog-ci/1952?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

packages/plugin-synthetics/package.json
```diff
@@ -24,6 +24,10 @@
     "./commands/*": {
       "development": "./src/commands/*.ts",
       "default": "./dist/commands/*.js"
+    },
+    "./reporters/junit": {
+      "development": "./src/reporters/junit.ts",
+      "default": "./dist/reporters/junit.js"
     }
```
Drarig29: I'd like to keep the same API for all datadog-ci plugins.

Can you instead add `type XMLJSON` to this line:
https://github.com/DataDog/datadog-ci/blob/f51dfce243973bfc88a372e57e4c265b75355058/packages/plugin-synthetics/src/index.ts#L6

hegrec: Sure thing! Swapped it to export from index.ts

Drarig29 (approved):
LGTM

PR #1959, opened 2025-11-11
[SVLS-7764] add container app uninstrument command

ava-silver:
### What and why?

Adds container app uninstrument command for removing instrumentation from container apps

### Testing

Added unit tests and ensured the same state before and after running container-app instrument followed by container-app uninstrument

```sh
az containerapp show -g tf-container-app-datadog-avasilver -n cli-test-ava > ~/before.json
yarn launch container-app instrument -s $(az account show --query id -o tsv) -g tf-container-app-datadog-avasilver -n cli-test-ava
yarn launch container-app uninstrument -s $(az account show --query id -o tsv) -g tf-container-app-datadog-avasilver -n cli-test-ava
az containerapp show -g tf-container-app-datadog-avasilver -n cli-test-ava > ~/after.json
diff ~/before.json ~/after.json
```

### Review checklist

### DIFF
--- a/packages/plugin-container-app/src/__tests__/instrument.test.ts
+++ b/packages/plugin-container-app/src/__tests__/instrument.test.ts
@@ -1350,6 +1350,18 @@ Updating tags for my-container-app
                 {name: 'DD_AZURE_RESOURCE_GROUP', value: 'my-resource-group'},
                 {name: 'DD_SERVERLESS_LOG_PATH', value: '/shared-volume/logs/*.log'},
               ]),
+              probes: [
+                {
+                  failureThreshold: 3,
+                  initialDelaySeconds: 0,
+                  periodSeconds: 10,
+                  tcpSocket: {
+                    port: 5555,
+                  },
+                  timeoutSeconds: 1,
+                  type: 'Startup',
+                },
+              ],
               resources: {
                 cpu: 0.25,
                 memory: '0.5Gi',


packages/plugin-container-app/src/commands/uninstrument.ts
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    return true
+  }
+
+  public async removeTags(subscriptionId: string, resourceGroup: string, containerApp: ContainerApp): Promise<void> {
+    const updatedTags = {...containerApp.tags}
+    delete updatedTags.service
+    delete updatedTags.env
+    delete updatedTags.version
```
TalUsvyatsky, line 97: Are these always added by us (the instrument command) or do users sometimes set these themselves?

ava-silver: it's possible for users to add it themselves, but there isn't a great way to determine if they set them manually or through the CLI

TalUsvyatsky: Chatted offline - I think there's a risk of removing tags that users have set themselves which could be unexpected or frustrating, but I don't think we have a great alternative either.

Maybe if people run into it enough we could add some kind of option to leave certain tags alone

nhulston (approved):

PR #1980, opened 2025-11-17
[SVLS-7759] add windows web app support

ava-silver:
### What and why?

Add windows web app support to the cli through the extension. Previously we assumed customers would use the legacy powershell scripts but this should be a better cross platform solution, unified with the linux and container web app types.

### How?

Automatically detect if an app is on windows or not, and automatically detect the windows runtime to determine which extension to install. 

For uninstrument, remove all datadog extensions currently installed, regardless of detected runtime.

Also migrated the tests to use snapshots for the stdout rather than relying on raw strings which were much more tedious to update.

### Review checklist

### DIFF
--- a/packages/plugin-aas/src/__tests__/__snapshots__/instrument.test.ts.snap
+++ b/packages/plugin-aas/src/__tests__/__snapshots__/instrument.test.ts.snap
@@ -154,7 +154,7 @@ exports[`aas instrument execute Validates windows runtime parameter 1`] = `
 
 exports[`aas instrument execute Warns and exits if App Service is Windows but runtime cannot be detected 1`] = `
 "🐶 Beginning instrumentation of Azure App Service(s)
-[!] Unable to detect runtime for Windows App Service my-web-app. Skipping instrumentation.
+[!] Unable to detect runtime for Windows App Service my-web-app. Skipping instrumentation. Try manually specifying your runtime with \`--windows-runtime\`
 🐶 Instrumentation completed with errors, see above for details.
 "
 `;
--- a/packages/plugin-aas/src/__tests__/__snapshots__/uninstrument.test.ts.snap
+++ b/packages/plugin-aas/src/__tests__/__snapshots__/uninstrument.test.ts.snap
@@ -134,10 +134,3 @@ Updating tags for my-web-app
 🐶 Uninstrumentation completed successfully!
 "
 `;
-
-exports[`aas instrument execute Warns and exits if App Service is Windows but runtime cannot be detected 1`] = `
-"🐶 Beginning uninstrumentation of Azure App Service(s)
-[!] Unable to detect runtime for Windows App Service my-web-app. Skipping uninstrumentation.
-🐶 Uninstrumentation completed with errors, see above for details.
-"
-`;
--- a/packages/plugin-aas/src/__tests__/uninstrument.test.ts
+++ b/packages/plugin-aas/src/__tests__/uninstrument.test.ts
@@ -208,19 +208,6 @@ describe('aas instrument', () => {
       expect(updateTags).toHaveBeenCalledWith(WEB_APP_ID, {properties: {tags: {}}})
     })
 
-    test('Warns and exits if App Service is Windows but runtime cannot be detected', async () => {
-      webAppsOperations.get.mockClear().mockResolvedValue({...CONTAINER_WEB_APP, kind: 'app,windows'})
-      const {code, context} = await runCLI(DEFAULT_ARGS)
-      expect(code).toEqual(1)
-      expect(context.stdout.toString()).toMatchSnapshot()
-      expect(getToken).toHaveBeenCalled()
-      expect(webAppsOperations.get).toHaveBeenCalledWith('my-resource-group', 'my-web-app')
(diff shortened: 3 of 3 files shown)


packages/plugin-aas/README.md
```diff
@@ -93,11 +93,12 @@ You can pass the following arguments to `instrument` to specify its behavior. Th
...
+| `--dotnet` or `--dotnet-container` |  | Add in required .NET-specific configuration options, is automatically inferred for code runtimes. This should be specified if you are using a containerized .NET app. | `false` |
 | `--musl` |  | Add in required .NET-specific configuration options for musl-based .NET apps. This should be specified if you are using a containerized .NET app on a musl-based distribution like Alpine Linux. | `false` |
 | `--source-code-integration` or `--sourceCodeIntegration` |  | Enable source code integration to add git metadata as tags. Defaults to enabled. Specify `--no-source-code-integration` to disable. | `true` |
 | `--upload-git-metadata` or `--uploadGitMetadata` |  | Upload git metadata to Datadog. Defaults to enabled. Specify `--no-upload-git-metadata` to disable. | `true` |
 | `--extra-tags` or `--extraTags` |  | Additional tags to add to the service in the format "key1:value1,key2:value2" |  |
+| `--windows-runtime` |  | Manually specify Windows runtime (node, dotnet, or java) to override automatic detection for the extension to use. |  |
```
domalessi:
```suggestion
| `--windows-runtime` |  | Manually specify the Windows runtime (`node`, `dotnet`, or `java`) used by the extension to override automatic detection. |  |
```

packages/base/src/commands/aas/common.ts
```diff
@@ -127,6 +148,9 @@ export abstract class AasCommand extends BaseCommand {
         '--musl can only be set if --dotnet is also set, as it is only relevant for containerized .NET applications.'
       )
     }
+    if (config.windowsRuntime && !(config.windowsRuntime in WINDOWS_RUNTIME_EXTENSIONS)) {
+      errors.push('--windows-runtime must be one of: node, dotnet, java')
```
avangelillo: Are these really the only 3 that are supported, or are we planning on adding support for others later?

If we think this would change we should make it generated based on the list so there's one less place for us to make changes to, like
```
`--windows-runtime must be one of: ${Object.keys(WINDOWS_RUNTIME_EXTENSIONS).join(', ')}`
```

packages/plugin-aas/src/commands/uninstrument.ts
```diff
@@ -74,18 +74,43 @@ export class PluginCommand extends AasUninstrumentCommand {
...
 
-      await this.uninstrumentSidecar(
-        client,
-        {...config, isDotnet: config.isDotnet || isDotnet(site), service: config.service ?? aasName},
-        resourceGroup,
-        aasName
-      )
-      await this.removeTags(client.subscriptionId!, resourceGroup, aasName, site.tags ?? {})
+          return false
+        }
```
avangelillo: Why does this check matter if we don't need the runtime to uninstrument?

packages/base/src/commands/aas/common.ts
ava-silver: yep those are the only supported runtimes, Python hasnt been supported on windows in a while, [PHP windows web app EOL was 2022](https://learn.microsoft.com/en-us/azure/app-service/quickstart-php?tabs=cli&pivots=platform-windows), and as far as we're aware there aren't any additional runtimes expected for windows.

packages/plugin-aas/src/commands/uninstrument.ts
ava-silver: good catch, this was from a previous iteration

packages/base/src/commands/aas/common.ts
```diff
@@ -16,6 +15,27 @@ import {BaseCommand} from '../..'
  */
 export type AasBySubscriptionAndGroup = Record<string, Record<string, string[]>>
 
+export const WINDOWS_RUNTIME_EXTENSIONS = {
+  node: 'Datadog.AzureAppServices.Node.Apm',
+  dotnet: 'Datadog.AzureAppServices.DotNet',
```
TalUsvyatsky, line 20: Just checking that the dotnet one is expected to be different?

ava-silver: yep, this is taken directly from self monitoring, you can see it in [the csproj files in the extension repo as well](https://github.com/DataDog/datadog-aas-extension)

packages/plugin-aas/src/commands/instrument.ts
```diff
@@ -125,12 +149,8 @@ This flag is only applicable for containerized .NET apps (on musl-based distribu
           )
         )
       }
-      config = {
-        ...config,
-        isDotnet: config.isDotnet || isDotnet(site),
-        isMusl: config.isMusl && config.isDotnet && isContainer,
-        service: config.service ?? aasName,
```
nhulston, line 132: is this `config.service` intentionally removed?

nhulston (approved):

packages/plugin-aas/src/commands/instrument.ts
ava-silver: i just consolidated it to the top of the function

PR #1981, opened 2025-11-17
[SVLS-8009] add custom env var support to cloud run

ava-silver:
### What and why?

For cloud run, adds the `-e` and `--service-tag` flags which allow the customer to specify custom env vars, as well as specify the DD_SERVICE variables through options rather than just the env var (which wasn't documented in the readme).

Also updates the readme script to fix some issues when parsing the commands, and migrates cloud run to use the autogenerated usage docs.

Allows us to have the more general use case instead of implementing every possible option as a flag, as an alternative to https://github.com/DataDog/datadog-ci/pull/1972

### Review checklist

### DIFF
--- a/bin/lint-readme-usage.ts
+++ b/bin/lint-readme-usage.ts
@@ -38,7 +38,7 @@ const parseCommandFile = (filePath: string): OptionDefinition[] => {
   const constants = new Map<string, string>()
 
   const collectConstants = (file: ts.SourceFile) => {
-    const visit = (node: ts.Node) => {
+    const visitNode = (node: ts.Node) => {
       if (ts.isVariableStatement(node)) {
         for (const declaration of node.declarationList.declarations) {
           if (ts.isIdentifier(declaration.name) && declaration.initializer) {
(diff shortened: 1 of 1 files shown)


packages/plugin-cloud-run/src/commands/instrument.ts
```diff
@@ -105,7 +105,7 @@ export class PluginCommand extends CloudRunInstrumentCommand {
       this.context.stdout.write(chalk.yellow('Invalid or missing region. Please use the --region flag.\n'))
     }
 
-    const ddService = process.env[SERVICE_ENV_VAR]
+    const ddService = this.serviceTag ?? process.env[SERVICE_ENV_VAR]
```
avangelillo, line 108: When parsing this, is a customer ever going to intend to set both?  Should we validate that only one is set?

ava-silver: the flag is new, so customers who have been setting it through the env var won't be affected. and at least to me this seems like the intuitive precedence, and we also already use the service name as a default value if the env var isn't set, so adding another precedence feels like a reasonable approach rather than erroring. 

happy to chat more if you have a strong preference

avangelillo (approved):

PR #1992, opened 2025-11-20
Add a tip and a known issue about standalone binary

Drarig29:
### What and why?

We recently had an issue where any datadog-ci command would be terminated with a segfault:

> Segmentation fault      (core dumped) datadog-ci tag --level "$level" --no-fail --tags

As a first step we tried to determine the version of `datadog-ci`, but it was built into a Docker image by downloading the latest standalone binary without printing the downloaded version in build logs, so it was unclear what version was built into the image.

Running `datadog-ci --version` was either returning `Segmentation fault (core dumped)` or `cannot execute binary file: Exec format error` depending on the context. We determined the version with `strings` but the version wasn't easy to find in the big output, even with `grep`.

With this PR, it's now possible to do `strings path/to/datadog-ci | grep STANDALONE_BINARY_VERSION` or `grep --text STANDALONE_BINARY_VERSION path/to/datadog-ci`.

Then we found out that the segfault was not caused by a regression in datadog-ci, but by us calling `strip --strip-unneeded path/to/datadog-ci` to reduce the size of the layers in our Docker image.

### How?

- Add warning about segfaults when using `strip` (see https://github.com/nodejs/postject/issues/90)
- Add note about grepping `STANDALONE_BINARY_VERSION` to determine the version of the binary without running it
- Improve standalone binary docs

### Review checklist

### DIFF
--- a/packages/datadog-ci/README.md
+++ b/packages/datadog-ci/README.md
@@ -259,7 +259,7 @@ Ignore Node.js errors if FIPS cannot be enabled on the host system.
 
 If installing NodeJS in the CI is an issue, standalone binaries are attached to each [GitHub release](https://github.com/DataDog/datadog-ci/releases).
 
-Supported architectures are:
+Supported architectures:
 - `linux-x64`
 - `linux-arm64`
 - `darwin-x64` (MacOS)
@@ -270,7 +270,7 @@ Supported architectures are:
 > Using `strip` to remove debugging symbols from the standalone binary as an attempt to make it smaller may cause segmentation faults when running `datadog-ci`. See https://github.com/nodejs/postject/issues/90.
 
 > [!NOTE]  
-> To determine the version when unable to run `datadog-ci --version` (for example because of segmentation faults or architecture mismatch), run `grep --text STANDALONE_BINARY_VERSION path/to/datadog-ci`.
+> To determine the version when unable to run `datadog-ci --version` (for example, because of segmentation faults or an architecture mismatch), run `grep --text STANDALONE_BINARY_VERSION path/to/datadog-ci`.
 
 To install the standalone binary:
 


.yarn/patches/@yao-pkg-pkg-fetch-npm-3.5.24-3b96980136.patch
```diff
@@ -1,3 +1,17 @@
...
+index 595ffced210b76d21bb9348cb4f201b1cf7052ca..12a55e61055909e34e12527339ab25f092b33dc6 100644
+--- a/lib-es5/build.js
++++ b/lib-es5/build.js
+@@ -130,7 +130,8 @@ function tarFetch(nodeVersion) {
+         return __generator(this, function (_c) {
+             switch (_c.label) {
+                 case 0:
+-                    log_1.log.info('Fetching Node.js source archive from nodejs.org...');
++                    log_1.log.error('Aborting before fetching Node.js source archive from nodejs.org. The above error should be transient, you can retry the CI job.');
++                    process.exit(1)
```
Drarig29, line 11: This is quite unrelated, but a transient error happened on a few commits in this PR ([here](https://github.com/DataDog/datadog-ci/actions/runs/19548036477/job/55971875317?pr=1992) and [here](https://github.com/DataDog/datadog-ci/actions/runs/19548231160/job/55972544154?pr=1992)), so I figured I would make the job fail earlier and add a message suggesting to retry the job.

This was caused by a `> Error! 503: Service Unavailable` when fetching binaries from `https://github.com/Drarig29/pkg-fetch/releases/download/...`.

packages/datadog-ci/README.md
```diff
@@ -241,41 +241,63 @@ The released `datadog-ci` binary now uses Node.js version 18 to be compatible wi
...
 
 ## More ways to install the CLI
 
 ### Standalone binary
 
-If installing NodeJS in the CI is an issue, standalone binaries are provided with [releases](https://github.com/DataDog/datadog-ci/releases). _linux-x64_, _linux-arm64_, _darwin-x64_, _darwin-arm64_ (MacOS) and _win-x64_ (Windows) are supported.
+If installing NodeJS in the CI is an issue, standalone binaries are attached to each [GitHub release](https://github.com/DataDog/datadog-ci/releases).
 
-To install:
+Supported architectures are:
```
domalessi:
```suggestion
Supported architectures:
```

packages/datadog-ci/README.md
```diff
@@ -241,41 +241,63 @@ The released `datadog-ci` binary now uses Node.js version 18 to be compatible wi
...
+- `linux-arm64`
+- `darwin-x64` (MacOS)
+- `darwin-arm64` (MacOS)
+- `win-x64` (Windows)
+
+> [!WARNING]  
+> Using `strip` to remove debugging symbols from the standalone binary as an attempt to make it smaller may cause segmentation faults when running `datadog-ci`. See https://github.com/nodejs/postject/issues/90.
+
+> [!NOTE]  
+> To determine the version when unable to run `datadog-ci --version` (for example because of segmentation faults or architecture mismatch), run `grep --text STANDALONE_BINARY_VERSION path/to/datadog-ci`.
```
domalessi:
```suggestion
> To determine the version when unable to run `datadog-ci --version` (for example, because of segmentation faults or an architecture mismatch), run `grep --text STANDALONE_BINARY_VERSION path/to/datadog-ci`.
```

PR #1993, opened 2025-11-20
[SVLS-8008] add custom memory and cpu options for container app sidecar

ava-silver:
### What and why?

Adds support for specifying custom memory and cpu for the sidecar, also updates the default memory cpu to be in line with the terraform module defaults. (.5 cpu/1gi mem)

### Review checklist

### DIFF
--- a/packages/plugin-container-app/src/__tests__/instrument.test.ts
+++ b/packages/plugin-container-app/src/__tests__/instrument.test.ts
@@ -742,13 +742,15 @@ Ensure you copied the value and not the Key ID.
     test('Validates sidecar CPU is a number', async () => {
       const {code, context} = await runCLI([...DEFAULT_INSTRUMENT_ARGS, '--sidecar-cpu', 'invalid'])
       expect(code).toEqual(1)
-      expect(context.stdout.toString()).toContain('[Error] sidecarCpu must be a number\n')
+      expect(context.stdout.toString()).toContain('Invalid value for --sidecar-cpu: expected a number (got "invalid")')
     })
 
     test('Validates sidecar memory is a number', async () => {
       const {code, context} = await runCLI([...DEFAULT_INSTRUMENT_ARGS, '--sidecar-memory', 'invalid'])
       expect(code).toEqual(1)
-      expect(context.stdout.toString()).toContain('[Error] sidecarMemory must be a number\n')
+      expect(context.stdout.toString()).toContain(
+        'Invalid value for --sidecar-memory: expected a number (got "invalid")'
+      )
     })
   })
 


packages/base/src/commands/container-app/common.ts
```diff
@@ -124,6 +126,17 @@ export abstract class ContainerAppCommand extends BaseCommand {
...
     }
+    // Validate that sidecarCpu is a valid number
+    if (config.sidecarCpu) {
+      const cpu = parseFloat(config.sidecarCpu)
+      if (isNaN(cpu) || cpu <= 0) {
+        errors.push('sidecarCpu must be a valid positive number')
+      }
+    }
+    // Validate that sidecarMemory has valid format
+    if (config.sidecarMemory && !/^\d+(\.\d+)?(Gi|Mi|G|M)$/.test(config.sidecarMemory)) {
```
nhulston: maybe this Regex should be in a constant at the top of file

nhulston (approved):
nice

packages/base/src/commands/container-app/common.ts
ava-silver: actually for azure you can only specify the amount in `Gi` anyways, so I'll just have them put in a float to simplify

packages/base/src/commands/container-app/instrument.ts
```diff
@@ -41,6 +44,12 @@ export class ContainerAppInstrumentCommand extends ContainerAppCommand {
   private logsPath = Option.String('--logs-path', DEFAULT_LOGS_PATH, {
     description: `(Not recommended) Specify a custom log file path. Must begin with the shared volume path. Defaults to '${DEFAULT_LOGS_PATH}'`,
   })
+  private sidecarCpu = Option.String('--sidecar-cpu', {
+    description: `The number of CPUs to allocate to the sidecar container. Defaults to '${DEFAULT_SIDECAR_CPU}'.`,
+  })
+  private sidecarMemory = Option.String('--sidecar-memory', {
+    description: `The amount of memory (in Gibibytes) to allocate to the sidecar container. Defaults to '${DEFAULT_SIDECAR_MEMORY}'.`,
+  })
```
avangelillo, line 55: If we're going to parse this as a number, is there a number type we can use instead of string?

ava-silver: one would think that this would be a feature, but I dont see a number/float option type: https://mael.dev/clipanion/docs/options/

avangelillo: 😞

Drarig29: @ava-silver you need to use typanion for that: https://mael.dev/typanion/docs/predicates/types#isnumber

For example, we created an [`isInteger()`](https://github.com/DataDog/datadog-ci/blob/c48bb73ad8f4af51d7bef1b156b38a9392d5958b/packages/base/src/helpers/validation.ts#L57) helper.

For floats, I believe `isNumber()` without `isInteger()` coercion should be  enough

Drarig29: And can we abbreviate "in Gibibytes" into `GiB`? I feel like it's odd to see it in all letters

ava-silver: Yeah we could do that @Drarig29, but we also need the validation for the config, so I'd prefer to keep the validation as is.

ava-silver: oh I didnt realize that also changes the type of the value. I'll make that change as well

Drarig29: Yes it coerces the TypeScript type as well 😄

PR #2004, opened 2025-12-05
[SVLS-8092] use layer versions from a cached file

ava-silver:
### What and why?

Uses layer versions from a file rather than searching for the layer version in AWS, which improves performance and enables version pinning via pinning datadog-ci.

This sets us up to improve the UX by changing the default behavior of leaving out the layer version from 'none' to 'latest' by just changing the default value

### How?

adds a generated json file with the runtime latest versions, which we can automatically create PRs to update over time

### Review checklist


ava-silver:
* **#2004** <a href="https://app.graphite.com/github/pr/DataDog/datadog-ci/2004?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/datadog-ci/2004?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

packages/plugin-lambda/src/layer-versions.json
duncanista, line 1: I'm worried you'd have to be releasing a new patch version every time this is updated – I don't think our customers update their CI/CLI tool that often – I feel like although I agree this is faster, I think it's a step back from the current interactive automatic mode to pull latest layer.

I suggest we could use this versions as a starting point and then we could keep checking which would be the newest latest layer? That should definitely reduce instrumentation time

Happy to hear your thoughts on this

ava-silver: currently for terraform, we don't do a new release whenever there is a new layer version -- we typically do it in bulk. I'd imagine we'd follow the same practice here, where the default version is pinned to the version in the specific datadog-ci release the customer is using.

This was the conclusion we came to from our discussion during the agent roundtable sync, which was pushed for by @lym953. Personally, I'm not attached to the method we do, I just wanted to get the code ready for us to be able to change the default from `none` to `latest`, and figured I'd include Yiming's suggestion.

duncanista: I do like that change – definitely agree with you there!

And sounds good – will talk with Yiming and also gather your input ofc!

packages/plugin-lambda/src/interfaces.ts
```diff
@@ -60,6 +60,8 @@ export interface InstrumentedConfigurationGroup {
   region: string
 }
 
+export type Version = number | 'latest' | 'none'
```
duncanista, line 63: So this would be breaking because now the type is required, do we want to explicitly set the 'none' to act as our `null`?

packages/plugin-lambda/src/commands/instrument.ts
```diff
@@ -319,30 +320,34 @@ export class PluginCommand extends LambdaInstrumentCommand {
   }
 
   private getSettings(): InstrumentationSettings | undefined {
-    const layerVersionStr = this.layerVersion ?? this.config.layerVersion
-    const extensionVersionStr = this.extensionVersion ?? this.config.extensionVersion
+    const layerVersionStr = this.layerVersion ?? this.config.layerVersion ?? 'none'
```
duncanista, line 323: Ah I see, if not set, you just set it to `none`, so essentially, following this `null` pattern,

I'm just confused, can't we just expect it to be `null` instead of using an explicit `none` string?

What's the advantage of this?

duncanista (approved):
LGTM – the only thing I'm not sure about is the explicit `none` string, as opposed to having a nullable string which can be either `null/undefined`

packages/plugin-lambda/src/commands/instrument.ts
ava-silver: `'none'` is much more explicit rather than using `null`, which gets coalesced the same way as `undefined` like when the user doesnt specify. this will allow us to more easily swap for `'latest'` in a future PR before the breaking change release

packages/plugin-lambda/src/interfaces.ts
ava-silver: not yet -- right now in the code the default value for `null`/`undefined` is set to none, so we keep behavior the same while changing the internal types

bin/lint-readme-usage.ts
ava-silver, line 1: I had claude convert this from parsing the source files to just importing them directly, which ends up in a much cleaner approach

cswatt (approved):

Issue #1912, opened 2025-10-14
Automatic plugin installation does not work in version 4.0.0


florianmutter:
### Bug description

When running we get the following output:

```bash
$ npx @datadog/datadog-ci synthetics run-tests
Could not find @datadog/datadog-ci-plugin-synthetics. Installing...

┏━ [Installed plugin 🔌]
┃ Successfully installed @datadog/datadog-ci-plugin-synthetics@4.0.0
┗━


┏━ [Plugin not installed 🔌]
┃ The datadog-ci synthetics run-tests command could not be found.
┃ To use this command, please install @datadog/datadog-ci-plugin-synthetics alongside datadog-ci.
┗━

You can install the plugin using:
  datadog-ci plugin install synthetics
or
  datadog-ci plugin install @datadog/datadog-ci-plugin-synthetics

To troubleshoot, run:
  datadog-ci plugin check synthetics
or
  datadog-ci plugin check synthetics run-tests
```

Next run does not fail anymore

### Describe what you expected

We expect the command to work on the first try.

### Steps to reproduce the issue

Run `npx @datadog/datadog-ci synthetics run-tests`

### Additional context

- macOS 26
- node v20.19.5
- npx v10.8.2

### Command

synthetics

tgentry300:
Hello @Drarig29! We're just using `npm`.

PR #2005, opened 2025-12-07
[SYNTH-22415] Temporary auto-install with NPX

Drarig29:
### What and why?

Fixes https://github.com/DataDog/datadog-ci/issues/1912

This PR replaces the auto-install mechanism with a temporary installation using NPX.

Benefits:
- Auto-installing a plugin doesn't have side effects on the current folder anymore.
  -  If side effects are expected, use `datadog-ci plugin install <name>`
- This new technique works when NPX is initially used to launch datadog-ci.

### How?

- Use `npx` (short for `npm exec`) to install plugins in a temporary environment
- Find that folder by executing `printenv PATH` with a shell in that temporary environment
- Use `require('module').Module._initPaths()` to add the path to module resolution paths

Inspired from https://www.npmjs.com/package/npx-import

### Review checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -235,8 +235,69 @@ jobs:
             exit 1
           fi
 
-  test-plugin-auto-install-in-project:
-    name: Test plugin auto-install in existing project
+  # Using `npx datadog-ci` only runs the binary without fetching any package.
+  test-plugin-auto-install-in-project-npm:
+    name: Test plugin auto-install in existing project (NPM)
+    runs-on: ubuntu-latest
+    needs: build-and-test
+    env:
+      FORCE_COLOR: 1
+
+    steps:
+      - name: Install node
+        uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
+        with:
+          node-version: 20
+      - uses: actions/download-artifact@d3f86a106a0bac45b974a628896c90dbdf5c8093 # v4.3.0
+        with:
+          name: artifacts-20
+          path: artifacts/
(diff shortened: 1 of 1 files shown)


packages/base/src/helpers/plugin.ts
```diff
@@ -130,6 +166,65 @@ export const installPlugin = async (packageOrScope: string): Promise<boolean> =>
...
+      if (error) {
+        reject(error)
+      } else {
+        resolve(stdout)
+      }
+    })
+  })
+  debug('Output:', output)
+
+  const isWindows = process.platform === 'win32'
```
GabrielAnca: Nit: you check this condition on line 130 too, maybe it's cleaner to move this to the beginning of the method and use it there

.github/workflows/ci.yml
```diff
@@ -191,6 +191,113 @@ jobs:
...
+            "name": "datadog-ci-plugin-auto-install",
+            "resolutions": {
+              "@datadog/datadog-ci-base": "file:./artifacts/@datadog-datadog-ci-base-20.tgz"
+            },
+            "dependencies": {
+              "@datadog/datadog-ci": "./artifacts/@datadog-datadog-ci-20.tgz"
+            }
+          }' > package.json
+
+          yarn set version stable
```
GabrielAnca, line 327: Will this also work with other package managers like npm? I assume it will, but it may be good to add a test at least for npm

GabrielAnca (approved):

.github/workflows/ci.yml
Drarig29: Just added an e2e test for NPM

PR #1873, opened 2025-09-26
Fix git metadata import

GabrielAnca:
### What and why?

Towards https://github.com/DataDog/datadog-ci/issues/1871

This import was outdated

### How?

Update it with the correct path

### Review checklist

- [ ] Feature or bugfix MUST have appropriate tests (unit, integration)

PR #2010, opened 2025-12-10
[chores] Remove `git-metadata` exports entry

Drarig29:
### What and why?

Follow-up of https://github.com/DataDog/datadog-ci/pull/1873

This PR removes a redundant exports entry for `git-metadata` to have more consistent entries.

### How?

Remove an exports entry that was too specific in favor of importing the right module.

### Review checklist

### DIFF
--- a/packages/base/package.json
+++ b/packages/base/package.json
@@ -27,10 +27,6 @@
       "development": "./src/commands/*.ts",
       "default": "./dist/commands/*.js"
     },
-    "./commands/git-metadata": {
-      "development": "./src/commands/git-metadata/index.ts",
-      "default": "./dist/commands/git-metadata/index.js"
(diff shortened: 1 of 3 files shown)


packages/base/package.json
```diff
@@ -27,10 +27,6 @@
       "development": "./src/commands/*.ts",
       "default": "./dist/commands/*.js"
     },
-    "./commands/git-metadata": {
-      "development": "./src/commands/git-metadata/index.ts",
-      "default": "./dist/commands/git-metadata/index.js"
-    },
```
Drarig29, line 35: Technically a breaking change (so we'll release in v5), but what matters is for `packages/datadog-ci/src/index.ts`'s API not to change.

PR #2012, opened 2025-12-11
[SVLS-8092][BREAKING CHANGE] use latest layer as default instead of none

ava-silver:
### What and why?

The default behavior of `datadog-ci lambda instrument -f <FUNCTION_ARN>` is very unintuitive, by not adding the layer, unless a forwarder is set up, it will not fully instrument the function, requiring users to look up the latest layer versions and specify them.
This PR makes the default behavior for the `--layer-version` and `--extension-version` pick the latest version when not specified, rather than not adding the layer.

⚠️**BREAKING CHANGE**⚠️: This changes the default behavior of `lambda instrument` from not adding the layers to adding the latest layer version.

### How?

As of https://github.com/DataDog/datadog-ci/pull/2004, users can specify `latest` to use the (generally) latest version. This PR just makes that the default behavior, so users don't need to manually specify the layer version.

### Review checklist


packages/plugin-lambda/src/commands/instrument.ts
```diff
@@ -320,8 +320,8 @@ export class PluginCommand extends LambdaInstrumentCommand {
   }
 
   private getSettings(): InstrumentationSettings | undefined {
-    const layerVersionStr = this.layerVersion ?? this.config.layerVersion ?? 'none'
-    const extensionVersionStr = this.extensionVersion ?? this.config.extensionVersion ?? 'none'
+    const layerVersionStr = this.layerVersion ?? this.config.layerVersion ?? 'latest'
+    const extensionVersionStr = this.extensionVersion ?? this.config.extensionVersion ?? 'latest'
```
ava-silver, line 324: this is the breaking change

Drarig29 (approved):

GabrielAnca (approved):

TalUsvyatsky (approved):

PR #2017, opened 2025-12-12
[CHORE] remove unnecessary env var

ava-silver:
### What and why?

Historically this env var was required, but [it is now optional](https://github.com/DataDog/dd-trace-dotnet/blob/3bc005f36e1cf62601a2aa0a18a7713726828878/tracer/src/Datadog.Trace/Logging/Internal/DatadogLoggingFactory.cs#L204-L208), so we don't need to set it.

### How?

No longer add it, and update the env var removal to remove `DD_*` like we do for container apps and cloud run, to keep consistent and allow us to not check for the env var explicitly

### Review checklist


ava-silver:
* **#2017** <a href="https://app.graphite.com/github/pr/DataDog/datadog-ci/2017?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/datadog-ci/2017?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

PR #2035, opened 2026-01-06
[sourcemaps] sourceMappingURL support

buranmert:
### What and why?

`sourcemaps upload` now looks at the last non-empty lines of minified JS files in the given directory and reads their `sourceMappingURL` values in order to find their corresponding sourcemap files.

Before this mapping logic was hardcoded as the following: `foo.min.js` -> `foo.map.js`

### How?

### Review checklist

### DIFF
--- a/packages/datadog-ci/src/commands/sourcemaps/upload.ts
+++ b/packages/datadog-ci/src/commands/sourcemaps/upload.ts
@@ -191,7 +191,7 @@ export class SourcemapsUploadCommand extends BaseCommand {
     const sourcemaps = (
       await doWithMaxConcurrency(this.maxConcurrency, jsFiles, async (minifiedFilePath) => {
         try {
-          const lastLine = readLastLine(minifiedFilePath)
+          const lastLine = await readLastLine(minifiedFilePath)
 
           // Look for sourceMappingURL comment
           const sourceMappingMatch = lastLine.match(/\/\/# sourceMappingURL=(.+\.map)/)
--- a/packages/datadog-ci/src/commands/sourcemaps/utils.ts
+++ b/packages/datadog-ci/src/commands/sourcemaps/utils.ts
@@ -1,15 +1,15 @@
-import {openSync, fstatSync, readSync, closeSync} from 'fs'
+import {open, FileHandle} from 'fs/promises'
 
 import upath from 'upath'
 
 // Reads the last non-empty line from a file using a buffer from the end
-export const readLastLine = (filePath: string): string => {
-  let fd: number | undefined
+export const readLastLine = async (filePath: string): Promise<string> => {
+  let fh: FileHandle | undefined
   let lastLine = ''
 
   try {
-    fd = openSync(filePath, 'r')
-    const stats = fstatSync(fd)
+    fh = await open(filePath, 'r')
+    const stats = await fh.stat()
     const fileSize = stats.size
 
     // Read up to 1KB from the end (should be enough for sourceMappingURL comment)
@@ -17,7 +17,7 @@ export const readLastLine = (filePath: string): string => {
     const buffer = Buffer.alloc(bufferSize)
     const position = Math.max(0, fileSize - bufferSize)
 
-    readSync(fd, buffer, 0, bufferSize, position)
+    await fh.read(buffer, 0, bufferSize, position)
     const tailContent = buffer.toString('utf-8')
 
     // Get the last non-empty line (handle multiple trailing newlines)
@@ -29,8 +29,8 @@ export const readLastLine = (filePath: string): string => {
       }
     }
   } finally {
-    if (fd !== undefined) {
-      closeSync(fd)
+    if (fh !== undefined) {
+      await fh.close()
     }
   }
 
--- a/packages/datadog-ci/src/commands/sourcemaps/__tests__/fixtures/with-encoded-sourcemap-url/[some-bundler]_runtime.js.map
+++ b/packages/datadog-ci/src/commands/sourcemaps/__tests__/fixtures/with-encoded-sourcemap-url/[some-bundler]_runtime.js.map
@@ -1,0 +1,4 @@
+{
+    "sources": ["webpack:///./src/commands/sourcemaps/__tests__/upload.test.ts"]
+}
+
--- a/packages/datadog-ci/src/commands/sourcemaps/__tests__/fixtures/with-encoded-sourcemap-url/runtime.min.js
+++ b/packages/datadog-ci/src/commands/sourcemaps/__tests__/fixtures/with-encoded-sourcemap-url/runtime.min.js
@@ -1,0 +1,4 @@
+/* minified runtime */
+var x=10,y=20;console.log(x*y);
+//# sourceMappingURL=%5Bsome-bundler%5D_runtime.js.map
+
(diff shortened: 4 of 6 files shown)


packages/datadog-ci/src/commands/sourcemaps/upload.ts
```diff
@@ -186,6 +187,75 @@ export class SourcemapsUploadCommand extends BaseCommand {
...
+          // Read up to 1KB from the end (should be enough for sourceMappingURL comment)
+          const bufferSize = Math.min(1024, fileSize)
+          const buffer = Buffer.alloc(bufferSize)
+          const position = Math.max(0, fileSize - bufferSize)
+
+          readSync(fd, buffer, 0, bufferSize, position)
+          const tailContent = buffer.toString('utf-8')
+
+          // Get the last non-empty line (handle multiple trailing newlines)
+          const lines = tailContent.split('\n')
```
Miz85: Is there a way for us to log something to dd here in case we don't find a `\n`? Just to be aware if it happens?

packages/datadog-ci/src/commands/sourcemaps/upload.ts
```diff
@@ -186,6 +187,75 @@ export class SourcemapsUploadCommand extends BaseCommand {
...
+        // Read only the last line from the file using a buffer from the end
+        let fd: number | undefined
+        let lastLine = ''
+
+        try {
+          fd = openSync(minifiedFilePath, 'r')
+          const stats = fstatSync(fd)
+          const fileSize = stats.size
+
+          // Read up to 1KB from the end (should be enough for sourceMappingURL comment)
```
Miz85: might be worth to create a `readLastLine` utility function in order to get that logic out and test it.

packages/datadog-ci/src/commands/sourcemaps/upload.ts
```diff
@@ -186,6 +187,75 @@ export class SourcemapsUploadCommand extends BaseCommand {
...
+            new Sourcemap(minifiedFilePath, minifiedURL, sourcemapPath, relativePath, this.minifiedPathPrefix)
+          )
+        }
+      } catch (error) {
+        // Skip files that can't be read
+        continue
+      }
+    }
+
+    // Fall back to legacy method if no sourcemaps were found
```
Miz85, line 219: 💯

packages/datadog-ci/src/commands/sourcemaps/upload.ts
Drarig29: No, the only thing we can do is log a warning to the user and assume they'll come to us if there is an issue

packages/datadog-ci/src/commands/sourcemaps/upload.ts
```diff
@@ -186,6 +187,76 @@ export class SourcemapsUploadCommand extends BaseCommand {
   // Looks for the sourcemaps and minified files on disk and returns
   // the associated payloads.
   private getMatchingSourcemapFiles = async (): Promise<Sourcemap[]> => {
+    const jsFiles = globSync(buildPath(this.basePath, '**/*.js'))
```
Drarig29, line 189: Since we are in an async context, could we change all IO calls to be async and use `Promise.all()`?

Drarig29: We could use `doWithMaxConcurrency()` (with the same `this.maxConcurrency` as the uploading part?) to avoid exhausting file descriptors

buranmert: good idea 👍  done here https://github.com/DataDog/datadog-ci/pull/2035/changes/e65b3e3ce93a7eea3a9ae3dd7872cba715b1a503#diff-bfd7b106993d41526420dafda6e971770a9d5a88e1af544f96a0224e3ee995a5R191-R192

packages/datadog-ci/src/commands/sourcemaps/upload.ts
```diff
@@ -186,6 +186,47 @@ export class SourcemapsUploadCommand extends BaseCommand {
...
+
+    const sourcemaps = (
+      await doWithMaxConcurrency(this.maxConcurrency, jsFiles, async (minifiedFilePath) => {
+        try {
+          const lastLine = readLastLine(minifiedFilePath)
```
Drarig29: I think `readLastLine()` should be async

packages/datadog-ci/src/commands/sourcemaps/upload.ts
```diff
@@ -186,6 +186,47 @@ export class SourcemapsUploadCommand extends BaseCommand {
...
+
+          // Look for sourceMappingURL comment
+          const sourceMappingMatch = lastLine.match(/\/\/# sourceMappingURL=(.+\.map)/)
+
+          if (sourceMappingMatch) {
+            // mert: nextjs/turbopack uses url-percent encoding
```
Drarig29, line 200: Can you add a fixture with URL-percent encoding?

packages/datadog-ci/src/commands/sourcemaps/utils.ts
```diff
@@ -1,5 +1,42 @@
+import {openSync, fstatSync, readSync, closeSync} from 'fs'
```
Drarig29: Let's use the async FS API:
```suggestion
import {open} from 'node:fs/promises'

import type {FileHandle} from 'node:fs/promises'
```

packages/datadog-ci/src/commands/sourcemaps/utils.ts
```diff
@@ -1,5 +1,42 @@
...
+
 import upath from 'upath'
 
+// Reads the last non-empty line from a file using a buffer from the end
+export const readLastLine = (filePath: string): string => {
+  let fd: number | undefined
```
Drarig29:
```suggestion
  let fd: FileHandle | undefined
```

packages/datadog-ci/src/commands/sourcemaps/utils.ts
```diff
@@ -1,5 +1,42 @@
...
+  let lastLine = ''
+
+  try {
+    fd = openSync(filePath, 'r')
+    const stats = fstatSync(fd)
+    const fileSize = stats.size
```
Drarig29:
```suggestion
    fd = await open(filePath, 'r')
    const stats = await fd.stat()
    const fileSize = stats.size
```

packages/datadog-ci/src/commands/sourcemaps/utils.ts
```diff
@@ -1,5 +1,42 @@
...
+
+    // Read up to 1KB from the end (should be enough for sourceMappingURL comment)
+    const bufferSize = Math.min(1024, fileSize)
+    const buffer = Buffer.alloc(bufferSize)
+    const position = Math.max(0, fileSize - bufferSize)
+
+    readSync(fd, buffer, 0, bufferSize, position)
```
Drarig29:
```suggestion
    await fd.read(buffer, 0, bufferSize, position)
```

packages/datadog-ci/src/commands/sourcemaps/utils.ts
```diff
@@ -1,5 +1,42 @@
...
+    const lines = tailContent.split('\n')
+    for (const line of lines.reverse()) {
+      if (line.trim().length !== 0) {
+        lastLine = line
+        break
+      }
+    }
+  } finally {
+    if (fd !== undefined) {
+      closeSync(fd)
```
Drarig29:
```suggestion
      await fd.close()
```

Drarig29 (commented):
Let's use the async FS API to unblock the event loop

Drarig29 (approved):
LGTM, thanks for the changes!

PR #2047, opened 2026-01-07
[SVLS-8267] fix error on unspecified runtime

ava-silver:
### What and why?

Fixes an error where the default arguments error for custom runtime functions. Now, we check for a specified layer number rather than specifically "none" so the default value or "none" or "latest" don't error.

### How?

### Review checklist

### DIFF
--- a/packages/plugin-lambda/src/__tests__/instrument.test.ts
+++ b/packages/plugin-lambda/src/__tests__/instrument.test.ts
@@ -1197,6 +1197,31 @@ describe('lambda', () => {
         expect(code).toBe(0)
       })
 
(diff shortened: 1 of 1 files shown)


packages/plugin-lambda/src/functions/instrument.ts
```diff
@@ -183,7 +183,7 @@ export const calculateUpdateRequest = async (
   const runtimeType = RUNTIME_LOOKUP[runtime]
 
   if (runtimeType === RuntimeType.CUSTOM) {
-    if (settings.layerVersion !== 'none') {
+    if (typeof settings.layerVersion === 'number') {
```
TalUsvyatsky, line 186: is it possible to add a test for this?

PR #2063, opened 2026-01-19
[chores] Add label guardrails in release process

Drarig29:
### What and why?

This PR changes the job order during the release process to have quick access to the draft release.

After the draft release is created, `pre-approval-checks` and all `build-*` jobs run in parallel, and are required before starting `npm-publish`.

And the `pre-approval-checks` job can add a `oidc-setup-required ⚠️` label, which needs to be cleared with a `oidc-setup-done ✅` label. Adding `oidc-setup-done ✅` is now part of the first-time publishing process.

### How?

- Move the `pre-approval-checks` job after the `create-draft-release` job
- Change the single line YAML array into a multiline YAML array for better readability
- Fix the ARM build jobs not being listed as requirements
- Add `pre-approval-checks` as a direct requirement for `npm-publish`
- Add label management

### Review checklist

Tested with `GITHUB_TOKEN=$(gh auth token) GITHUB_SHA=04b0d1cc47ab5f3ad479d0603df5acc1c6e01ba0 bin/check-npm-packages.sh`  on https://github.com/DataDog/datadog-ci/pull/2062

### DIFF
--- a/bin/check-npm-packages.sh
+++ b/bin/check-npm-packages.sh
@@ -110,6 +110,7 @@ echo -e "${RED}The following packages are not published to NPM yet:${NC}"
 for pkg in "${missing_packages[@]}"; do
 	echo "  - $pkg"
 done
+echo
 
 # In CI environment, post a comment on the PR
 if [ -n "${GITHUB_TOKEN:-}" ] && [ -n "${GITHUB_SHA:-}" ]; then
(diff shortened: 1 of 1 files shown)


bin/check-npm-packages.sh
```diff
@@ -181,3 +220,13 @@ if [ "$DRY_RUN" = true ]; then
...
+
+echo
+echo -e "${RED}${BOLD}⚠️  You are not done!${NC} Please setup OIDC on each package at the links below."
+echo
+
+for pkg in "${missing_packages[@]}"; do
+	echo -e "  - Opening ${BLUE}https://www.npmjs.com/package/$pkg/access${NC}"
+	open "https://www.npmjs.com/package/$pkg/access"
+done
+echo
```
Drarig29, line 232: Image

bin/check-npm-packages.sh
```diff
@@ -63,7 +64,41 @@ while IFS= read -r pkg; do
...
+	PR_RESPONSE=$(curl -s -H "Authorization: token $GITHUB_TOKEN" \
+		"https://api.github.com/repos/$GITHUB_REPOSITORY/commits/$GITHUB_SHA/pulls")
+	PR_LABELS=$(echo "$PR_RESPONSE" | jq '[.[0].labels[].name]' 2>/dev/null || true)
+
+	echo -e "${BLUE}PR labels:${NC} $PR_LABELS"
+	echo
+fi
+
+# In CI, check the labels on the PR
+if [ -n "${GITHUB_TOKEN:-}" ] && [ -n "${GITHUB_SHA:-}" ]; then
```
GabrielAnca: This is the same if as above, and we don't use either of these variables within the block. Do we maybe want to check if `PR_LABELS` is not empty?

Drarig29: Good idea, done!

PR #2068, opened 2026-01-21
[K9VULN-9613] Add target framework scope properties

piloulacdog:
## 🚀 Motivation 
Modern .NET projects commonly use multi-targeting, where different versions of dependencies are required for different target frameworks (e.g., net6.0, net7.0, net8.0). Currently, there is no way to represent or filter this framework-specific dependency information in SBOMs, making it difficult to accurately track which dependencies apply to which runtime environments.

## 📝 Summary
This PR adds support for storing target framework information in SBOM component properties using the `datadog:target-framework` property key. The implementation:

## 🧪 Testing
- [x] New tests were added for new logic.
- [x] Existing tests were updated for new logic, and not only so that they pass!
- [ ] Benchmark results prove that performance is the same or better.

## 🆘 Recovery
Notes for on-call - **select only one**:
- [x] The change can be rolled back.
- [ ] Do not roll back. Why?:

### DIFF
--- a/packages/plugin-sbom/src/__tests__/validation.test.ts
+++ b/packages/plugin-sbom/src/__tests__/validation.test.ts
@@ -85,6 +85,7 @@ describe('validation of sbom file', () => {
         is_dev: undefined,
         reachable_symbol_properties: undefined,
         exclusions: undefined,
+        target_frameworks: undefined,
       })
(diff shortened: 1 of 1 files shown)


Drarig29 (approved):
LGTM (I re-triggered the mergegate)

PR #2072, opened 2026-01-26
fix(flutter-symbols): Validate version parameter is not empty before upload

jhssilva:
### What and why?

- Adds comprehensive validation for the `--version` parameter in the `flutter-symbols upload` command to prevent uploads with empty or invalid version strings.
- The deobfuscation backend requires a valid version to correctly associate symbol files. 
- [Jira](https://datadoghq.atlassian.net/browse/RUM-12254)

### How?

- **Version validation**: If `--version` is not provided, empty, or whitespace-only, the command now attempts to extract the version from `pubspec.yaml`
- **Final validation**: After all resolution attempts, if the version is still empty/whitespace, the command fails with a clear error message: `Error: parameter "version" is required`
- **Tests**: Added 4 new test cases covering:
  - Empty string version without valid pubspec (should fail)
  - Whitespace-only version without valid pubspec (should fail)
  - Empty string version with valid pubspec (should succeed using pubspec version)
  - Whitespace-only version with valid pubspec (should succeed using pubspec version)

### Review checklist

### Test check
```bash
yarn test packages/datadog-ci/src/commands/flutter-symbols
```

### DIFF
--- a/packages/datadog-ci/src/commands/flutter-symbols/README.md
+++ b/packages/datadog-ci/src/commands/flutter-symbols/README.md
@@ -59,7 +59,6 @@ datadog-ci flutter-symbols upload --dart-symbols-location ./debug-info --service
 | `--disable-git`              | Optional  | Prevents the command from invoking git in the current working directory and sending repository-related data to Datadog (such as the hash, remote URL, and paths within the repository of sources referenced in the source map).                                                                                  |
 | `--repository-url`           | Optional  | Overrides the remote repository with a custom URL. For example, `https://github.com/my-company/my-project`.                                                                                                                                                                                                      |
 
-**Note:** A version is required for the upload to succeed. If `--version` is not provided, the version will be extracted from your `pubspec.yaml`. The upload will fail if no version can be determined from either source.
 
 ## Further reading
 


packages/datadog-ci/src/commands/flutter-symbols/upload.ts
```diff
@@ -540,7 +540,16 @@ export class FlutterSymbolsUploadCommand extends BaseCommand {
...
+    // If version is not provided or is empty/whitespace, try to get it from pubspec
+    if (!this.version || this.version.trim().length === 0) {
+      if (await this.parsePubspecVersion(this.pubspecLocation)) {
+        parametersOkay = false
+      }
+    }
+
+    // Final validation: ensure we have a non-empty version after all resolution attempts
+    if (!this.version || this.version.trim().length === 0) {
+      this.context.stderr.write(renderArgumentMissingError('version'))
```
buranmert, line 552: you also need to update README in this directory. it says `version` is optional, we should state it's required

jhssilva: Done. Updated the README in the latest commit.
**Context**: 
- This PR is primarily a bug fix. The --version parameters was never truly "optional" - a version was always required for the upload to succeeded (either from the flag ``version`` or from the ``pubspec.yaml``).
- The bug allowed to pass whitespace-only values like ``--version "  "``, and this would bypass the pubspec fallback ``(!this.version && (await this.parsePubspecVersion(this.pubspecLocation))`` (since non-empty strings are truthy in JS).
- The fix ensures whitespace-only versions are treated the same as empty/missing versions, and adds an explicit client-side error message when no valid version can be determined.

Added a note in the ``README.MD``, clarifying the need for the ``version`` having a fallback for ``pubspec.yaml``.

PR #2084, opened 2026-02-02
[SVLS-8477] ensure consistency for WEBSITES_ENABLE_APP_SERVICE_STORAGE

ava-silver:
### What and why?

After some discussions with support and other folks, we should make sure we're consistent about ensuring the `WEBSITES_ENABLE_APP_SERVICE_STORAGE` env var is set to true, since the default value is not always set properly.

### How?

Always add WEBSITES_ENABLE_APP_SERVICE_STORAGE, since it can fix logging issues but doesn't hurt to do.

### Review checklist

### DIFF
--- a/packages/plugin-aas/src/__tests__/common.test.ts
+++ b/packages/plugin-aas/src/__tests__/common.test.ts
@@ -41,6 +41,7 @@ describe('aas common', () => {
         DD_API_KEY: 'test-api-key',
         DD_SITE: 'datadoghq.com',
         DD_AAS_INSTANCE_LOGGING_ENABLED: 'false',
+        WEBSITES_ENABLE_APP_SERVICE_STORAGE: 'true',
       })
     })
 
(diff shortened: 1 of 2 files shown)


packages/plugin-aas/src/common.ts
```diff
@@ -59,6 +60,7 @@ export const getEnvVars = (config: AasConfigOptions, isContainer: boolean): Reco
   // Get base environment variables
   let envVars = getBaseEnvVars(config)
   envVars = {
+    WEBSITES_ENABLE_APP_SERVICE_STORAGE: 'true',
```
avangelillo, line 63: I'm not super familiar with this specifically, but is there any risk associated changing this without customers knowing?

AI overview said this about turning it to true:
> Files written to /home (including subdirectories like /home/site/wwwroot, LogFiles, etc.) will persist across app restarts and scale instances.

Which makes it sound like this could cause stored files to build up or state to act differently

ava-silver: Discussed offline -- the TLDR is logging won't work without this environment variable so regardless of how the customer has configured their app, this is necessary.

avangelillo (approved):

PR #2089, opened 2026-02-03
Fix job display name extraction: handle escaped quotes and identify correct Worker log on non-ephemeral runners

LudovicTOURMAN:
## Problems

This PR fixes two related issues (cf [Issue](https://github.com/DataDog/datadog-ci/issues/2092)) with GitHub Actions job display name extraction from Runner diagnostic logs:

### Problem 1: Escaped Quotes and Special Characters ✅

The current regex for extracting job display names fails when job names contain escaped quotes or other special characters.

**Failing Example:**
- Job name: `End-to-End Tests (@org/backend, "features/a*", apps/backend)`
- Current result: `End-to-End Tests (@org/backend, \` (truncated at first quote)

**Root Cause:**
In Worker diagnostic logs, job names are stored as JSON strings with proper escaping. The current regex `/"jobDisplayName":\s*"([^"]+)"/` uses `[^"]+` which stops at the first quote character, even when it's escaped as `\"`.

### Problem 2: Wrong Worker Log on Non-Ephemeral Runners 🆕

On non-ephemeral (reusable) runners, multiple `Worker_*.log` files accumulate from different job executions. The current implementation iterates through logs and returns the **first** jobDisplayName found, which may belong to a **previous job run** rather than the current one.

**Real-World Failing Scenario:**
We encountered this issue in production where:
- Runner had multiple logs from different job runs
- Current job was `build-amd64-base-rails-assets`
- But datadog-ci extracted `Docker build / Docker build` from an older log file
- This caused incorrect job tagging in Datadog CI Visibility

**Root Cause:**
No filtering mechanism to identify which Worker log corresponds to the currently executing job.

---

### Solution 1: Proper Escape Handling ✅

1. **Updated regex**: `/"jobDisplayName":\s*"((?:[^"\\]|\\.)*)"/`
   - Uses `(?:[^"\\]|\\.)*` to correctly match JSON-escaped strings
   - Handles all escape sequences: `\"`, `\\`, `\n`, `\t`, `\uXXXX`, etc.

2. **Proper unescaping**: Added `unescapeJsonString()` function using `JSON.parse()`
   - Correctly unescapes all JSON escape sequences
   - Handles edge cases like Unicode, backslashes, control characters

### Solution 2: Worker Log Identification 🆕

Use the `ACTIONS_ORCHESTRATION_ID` environment variable to identify the correct Worker log file:

1. Read `process.env.ACTIONS_ORCHESTRATION_ID` (format: `<guid>.<job-id>.__default`)
2. Extract the planId GUID (everything before first dot)
3. Search Worker logs for one containing this GUID as `"planId": "<guid>"`
4. Only extract jobDisplayName from the matched log
5. Fall back to checking all logs if ACTIONS_ORCHESTRATION_ID unavailable or planId not found

**Why this works:**
- `ACTIONS_ORCHESTRATION_ID` is available in GitHub Actions environment (Runner v2.331.0+)
- The planId GUID appears in Worker log JSON as `"planId": "<guid>"`
- The planId is globally unique per job execution (type: `Guid` in Runner source code)
- This ensures we always extract the job name from the correct log, even on non-ephemeral runners

---

### Existing Tests (Escape Sequences) ✅
Added comprehensive tests covering:
- Job names with quotes (single and multiple)
- Backslashes (Windows paths, regex patterns)
- Emojis and Unicode (🚀, Chinese, Japanese, Arabic, Cyrillic)
- Matrix jobs with various formats
- Reusable workflows with slashes
- Special characters and complex combinations

### New Tests (Worker Log Identification) 🆕
Added 4 new test cases:
1. **Single log with ACTIONS_ORCHESTRATION_ID**: Verifies correct planId matching
2. **Multiple logs on non-ephemeral runner**: Ensures correct log selected among multiple Worker logs from different jobs
3. **Fallback without ACTIONS_ORCHESTRATION_ID**: Falls back to checking all logs (backward compatibility)
4. **Fallback when planId not found**: Falls back to checking all logs if planId doesn't match any log

### Bonus Fix 🎁
Fixed test helpers to properly JSON-escape job names using `JSON.stringify()`, which resolved 8 pre-existing test failures.

**All 1012 tests now pass** ✅

---

## Issues We Faced

During development and testing, we encountered several challenges:

1. **Test Failures Due to Invalid JSON**: The test helper functions were directly interpolating job names into JSON templates without escaping, causing invalid JSON when job names contained quotes or special characters. Fixed by using `JSON.stringify()` in test helpers.

2. **Multiple File Reads**: Initial test expectations were incorrect - the implementation reads logs multiple times (once while searching for planId, then again to extract job name). Adjusted test assertions to match actual behavior (3 reads instead of 2).

3. **Real Production Issue**: On non-ephemeral runners with multiple Worker logs, datadog-ci was consistently extracting job names from old logs instead of the current job. This PR completely resolves this by using the unique planId from `ACTIONS_ORCHESTRATION_ID`.

---

## Verification

Tested against real-world failing cases:
1. ✅ Job names with escaped quotes now parse correctly
2. ✅ Non-ephemeral runners now select the correct Worker log using planId matching
3. ✅ Backward compatibility maintained: works without ACTIONS_ORCHESTRATION_ID
4. ✅ All existing functionality preserved with fallback behavior

## Compatibility

- **Requires**: Runner v2.331.0+ for `ACTIONS_ORCHESTRATION_ID` environment variable
- **Fallback**: Works with older Runner versions (checks all logs, as before)
- **No breaking changes**: Fully backward compatible

---

This PR fixes critical issues that were causing incorrect job identification in Datadog CI Visibility, especially on non-ephemeral runners commonly used in production environments.


packages/base/src/helpers/ci.ts
```diff
@@ -1144,17 +1161,51 @@ export const getGithubJobNameFromLogs = (context: BaseContext): string | undefin
...
+      if (content.includes(`"planId": "${planId}"`)) {
+        targetLogFile = logFile
+        context.stdout.write(`Found Worker log for planId ${planId}: ${logFile}\n`)
+        break
+      }
+    }
+
+    if (!targetLogFile) {
+      context.stdout.write(
+        `${chalk.yellow.bold('[INFO]')} Could not find Worker log for planId ${planId}, checking all logs\n`
```
rodrigo-roca:
```suggestion
        `${chalk.yellow.bold('[WARNING]')} Could not find Worker log for planId ${planId}, checking all logs\n`
```

packages/base/src/helpers/ci.ts
```diff
@@ -1144,17 +1161,51 @@ export const getGithubJobNameFromLogs = (context: BaseContext): string | undefin
...
-  for (const logFile of workerLogFiles) {
+  // For non-ephemeral runners, multiple Worker logs may exist from different jobs
+  // Use ACTIONS_ORCHESTRATION_ID to identify the correct log if available
+  const orchestrationId = process.env.ACTIONS_ORCHESTRATION_ID
+  let targetLogFile: string | undefined
+
+  if (orchestrationId) {
+    // Extract planId GUID (everything before first dot)
+    const planId = orchestrationId.split('.')[0]
+    context.stdout.write(`Searching for Worker log with planId: ${planId}\n`)
```
rodrigo-roca: I would remove this print since below we already have either: `Found Worker log for planId` or `Could not find Worker log for planId`

LudovicTOURMAN:
Thanks for the feedback @rodrigo-roca
I've updated the code following so.

PR #2097, opened 2026-02-10
[SVLS-8314] add `lambda cloudwatch` command

ava-silver:
### What and why?

This PR adds a new `lambda cloudwatch` command to the Lambda plugin that accepts `enable` or `disable` as an action. This allows users to disable CloudWatch Logs for Lambda functions by attaching a deny IAM policy, and re-enable them by removing that policy. This provides a way to control CloudWatch logging costs while maintaining Lambda function operation.

### How?

- Added a unified `LambdaCloudwatchCommand` that takes an action parameter (`enable` or `disable`)
- Implemented the corresponding plugin command that handles AWS interactions
- Created CloudWatch-specific functions for managing IAM policies
- Added a renderer for CloudWatch-related output
- Added comprehensive tests for the command
- Updated the CLI command registry to include the new command

### Testing

Note, for all commands below, alias:
```sh
alias aws-sso='aws-vault exec sso-serverless-sandbox-account-admin -- '
```

Dry run:
```sh
aws-sso yarn launch lambda cloudwatch disable --dry-run -f <function-arn>
aws-sso yarn launch lambda cloudwatch enable --dry-run -f <function-arn>
```

Live test:
```sh
aws-sso yarn launch lambda cloudwatch disable -f <function-arn>
# Verify it gets added:
aws-sso aws iam get-role-policy --role-name <role-from-output> --policy-name DenyCloudWatchLogs

aws-sso yarn launch lambda cloudwatch enable -f <function-arn>
# Verify: above command should now return NoSuchEntity

# Multi-function & regex
aws-sso yarn launch lambda cloudwatch disable --dry-run --functions-regex 'my-prefix' -r us-east-1
aws-sso yarn launch lambda cloudwatch disable --dry-run -f <arn1> -f <arn2>
```

[Example notebook](https://ddserverless.datadoghq.com/notebook/13888424/cloudwatch-enable-disablement-via-datadog-ci?refresh_mode=paused&from_ts=1770906778089&to_ts=1770910523312)

### Review checklist

### DIFF
--- a/packages/plugin-lambda/src/commands/cloudwatch.ts
+++ b/packages/plugin-lambda/src/commands/cloudwatch.ts
@@ -180,7 +180,10 @@ export class PluginCommand extends LambdaCloudwatchCommand {
     const roleMap = new Map<string, {logGroups: string[]; functionARNs: string[]}>()
     for (const fn of functionARNs) {
       try {
-        const {roleName, logGroup} = await getFunctionDetails(lambdaClient, fn)
+        const {roleName, logGroup, hasExtensionLayer} = await getFunctionDetails(lambdaClient, fn)
+        if (this.action === 'disable' && !hasExtensionLayer) {
+          stdout.write(cloudwatchRenderer.renderNoExtensionWarning(fn))
+        }
         const entry = roleMap.get(roleName) ?? {logGroups: [], functionARNs: []}
         entry.logGroups.push(logGroup)
         entry.functionARNs.push(fn)
--- a/packages/plugin-lambda/src/functions/cloudwatch.ts
+++ b/packages/plugin-lambda/src/functions/cloudwatch.ts
@@ -1,6 +1,8 @@
 import {DeleteRolePolicyCommand, GetRolePolicyCommand, IAMClient, PutRolePolicyCommand} from '@aws-sdk/client-iam'
 import {GetFunctionCommand, LambdaClient} from '@aws-sdk/client-lambda'
 
+import {DD_LAMBDA_EXTENSION_LAYER_NAME} from '../constants'
+
 export const DENY_POLICY_NAME = 'DenyCloudWatchLogs'
 
 export const getDenyPolicyDocument = (logGroups: string[]) =>
@@ -18,7 +20,7 @@ export const getDenyPolicyDocument = (logGroups: string[]) =>
 export const getFunctionDetails = async (
   lambdaClient: LambdaClient,
   functionIdentifier: string
-): Promise<{roleName: string; functionName: string; logGroup: string}> => {
+): Promise<{roleName: string; functionName: string; logGroup: string; hasExtensionLayer: boolean}> => {
   const resp = await lambdaClient.send(new GetFunctionCommand({FunctionName: functionIdentifier}))
   const roleArn = resp.Configuration?.Role
   if (!roleArn) {
@@ -33,8 +35,10 @@ export const getFunctionDetails = async (
   // Role ARN format: arn:aws:iam::ACCOUNT:role/ROLE_NAME or arn:aws:iam::ACCOUNT:role/path/ROLE_NAME
   const roleName = roleArn.split('/').pop()!
   const logGroup = resp.Configuration?.LoggingConfig?.LogGroup ?? `/aws/lambda/${functionName}`
+  const hasExtensionLayer =
+    resp.Configuration?.Layers?.some((l) => l.Arn?.includes(DD_LAMBDA_EXTENSION_LAYER_NAME)) ?? false
 
-  return {roleName, functionName, logGroup}
+  return {roleName, functionName, logGroup, hasExtensionLayer}
 }
 
 const LOG_GROUP_RESOURCE_PATTERN = /^arn:aws:logs:\*:\*:log-group:(.+):\*$/
--- a/packages/plugin-lambda/src/__tests__/cloudwatch.test.ts
+++ b/packages/plugin-lambda/src/__tests__/cloudwatch.test.ts
@@ -28,6 +28,8 @@ import {DENY_POLICY_NAME, getDenyPolicyDocument} from '../functions/cloudwatch'
 
 import {mockAwsCredentials} from './fixtures'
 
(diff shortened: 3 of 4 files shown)


duncanista (approved):
This looks great, wondering if it would make sense to not make it Lambda specific? I know right now we are just disabling for Lambda functions, but is there a world where we add more services where we disable CWL?

packages/plugin-lambda/src/commands/cloudwatch.ts
```diff
@@ -0,0 +1,200 @@
...
+      if (!region) {
+        stdout.write(commonRenderer.renderNoDefaultRegionSpecifiedError())
+
+        return 1
+      }
+
+      const lambdaClient = new LambdaClient({
+        region,
+        credentials,
+        retryStrategy: EXPONENTIAL_BACKOFF_RETRY_STRATEGY,
```
avangelillo: I would recommend using [adaptive retries](https://docs.aws.amazon.com/sdkref/latest/guide/feature-retry-behavior.html) here, it will tradeoff latency for a higher likelihood of succeeding.  We [do this in the remote instrumenter](https://github.com/DataDog/serverless-remote-instrumentation/blob/prod/src/aws-resources.ts#L7) for the same reasons.  You should use one client per region though, so the pattern there of getting it with a function call that checks a cache is pretty common.  If you can avoid remaking them on lines 126-136 anyway that's better too.

packages/plugin-lambda/src/functions/cloudwatch.ts
```diff
@@ -0,0 +1,71 @@
...
+  return {roleName, functionName}
+}
+
+export const disableCloudwatchLogs = async (
+  iamClient: IAMClient,
+  roleName: string,
+  functionName: string
+): Promise<void> => {
+  await iamClient.send(
+    new PutRolePolicyCommand({
```
avangelillo, line 78: What will happen if someone calls the disable command twice for a function, will there be two role policies or will the second call fail?

packages/plugin-lambda/src/functions/cloudwatch.ts
```diff
@@ -0,0 +1,71 @@
...
+export const getDenyPolicyName = (functionName: string) => `DenyCloudWatchLogs-${functionName}`
+
+export const getDenyPolicyDocument = (functionName: string) =>
+  JSON.stringify({
+    Version: '2012-10-17',
+    Statement: [
+      {
+        Effect: 'Deny',
+        Action: ['logs:CreateLogGroup', 'logs:CreateLogStream', 'logs:PutLogEvents'],
+        Resource: `arn:aws:logs:*:*:log-group:/aws/lambda/${functionName}:*`,
```
avangelillo: This is most likely the log group that is being logged to, but on the [GetFunctionConfig](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/client/lambda/command/GetFunctionCommand/) command that we're already doing a LoggingConfig comes back with the response.  We should pull the log group from there in case someone has configured it differently

packages/plugin-lambda/src/commands/cloudwatch.ts
```diff
@@ -0,0 +1,200 @@
...
+        } catch (err) {
+          failures++
+          stdout.write(cloudwatchRenderer.renderFunctionError(fn, err))
+        }
+      }
+
+      return {successes, failures}
+    }
+
+    for (const fn of functionARNs) {
```
avangelillo: I would add one more step here to take all of the role arns from the function and dedupe them, then operate on role arns instead of function arns for the rest of the process.  If someone has multiple lambdas using a single role you'd end up with multiple policies disallowing logging, then if they re enable logging it won't work

packages/plugin-lambda/src/commands/cloudwatch.ts
```diff
@@ -0,0 +1,200 @@
...
+  getAWSProfileCredentials,
+  getLambdaFunctionConfigsFromRegex,
+} from '../functions/commons'
+import {LambdaConfigOptions} from '../interfaces'
+import * as cloudwatchRenderer from '../renderers/cloudwatch-renderer'
+import * as commonRenderer from '../renderers/common-renderer'
+
+type CloudwatchAction = (iamClient: IAMClient, roleName: string, functionName: string) => Promise<void>
+
+export class PluginCommand extends LambdaCloudwatchCommand {
```
avangelillo, line 24: What kinds of validations do we want to do for this command?  Should we allow someone to disable logging on a function that doesn't have the extension?  Is it possible to check if other functions not matched by the regex are using the role, or if it is used elsewhere?

packages/plugin-lambda/src/functions/cloudwatch.ts
ava-silver: good callout!

packages/plugin-lambda/src/functions/cloudwatch.ts
ava-silver: PutRolePolicyCommand is an upsert: if an inline policy with the same PolicyName already exists on the role, it replaces the document, so we don't need to worry about duplication/errors

ava-silver: Although I'm realizing my current implementation based on your comments is at risk of clobbering previous runs -- let me fix that

packages/plugin-lambda/src/commands/cloudwatch.ts
ava-silver: Good question, what are your thoughts on either adding a warning or a confirmation that they want to disable logs on a function without the agent? 

Assuming we check for the agent by looking for the extension, are there any issues if a customer is using a custom docker/container setup on their lambda?

> Is it possible to check if other functions not matched by the regex are using the role, or if it is used elsewhere?

We could in theory check for other usages of the role but I'd assume that would be a pretty expensive API call -- I wonder if we should just document this caveat, since we can't really prevent against the case of a customer using the same log group with the same IAM role

Maybe we can chat more about this separately?

ava-silver: TLDR from our discussion:

I'll add a warning if the customer is missing the extension -- for the issue of lambdas sharing both a role and a log group, I'll add that caveat to the docs since there's no way to fully prevent the issue and checking would be too costly in general.

PR #2125, opened 2026-02-24
Add file fixes generation for code coverage uploads

CarlosNietoP:
Identifies non-executable lines (comments, blank lines, brackets, etc.) in tracked source files and uploads them alongside coverage reports so the backend can exclude them from coverage calculations.

### What and why?

Code coverage reports sometimes flag non-executable lines (comments, blank lines, closing brackets, etc.) as "uncovered", creating false negatives. This PR adds **file fixes** — a bitmap per source file marking which lines are non-executable — so the backend can exclude them from coverage calculations.

### How?

During `datadog-ci coverage upload`, the CLI:
1. Lists all git-tracked files (`git ls-files`) and filters to supported languages (Go, Kotlin, C/C++, Swift, ObjC, PHP)
2. For each file, scans lines against language-specific regex patterns (empty lines, comments, block comments, brackets, etc.)
3. Encodes matches as a **1-bit-per-line bitmap** (base64-encoded) per file
4. Uploads the result as a gzipped JSON payload (`file_fixes.json.gz`) alongside the coverage report

**File discovery** uses two strategies:
- **Default (`git ls-files`)**: fast, respects `.gitignore`, only tracked files. Used when running inside a git repo without `--file-fixes-search-path`.
- **Filesystem walk fallback**: recursive directory traversal, skipping common non-source folders (node_modules, .git, .venv, etc.). Used when git is not available or `--file-fixes-search-path` is provided. This is useful in monorepos to scope the scan to a specific subdirectory.

Files are processed concurrently (8 async workers).

New CLI flags:
- `--disable-file-fixes` — skip file fixes generation entirely
- `--file-fixes-search-path` — override the root directory used to scan source files (defaults to repo root). Triggers the filesystem walk instead of `git ls-files`.

### Review checklist

### DIFF
--- a/packages/plugin-coverage/src/file-fixes.ts
+++ b/packages/plugin-coverage/src/file-fixes.ts
@@ -42,8 +42,7 @@ const PARENTHESIS_LINE = /^\s*[()]\s*(\/\/.*)?$/
 
 const BLOCK_COMMENT_OPEN = /^\s*\/\*/
 
-const GO_FUNC_LINE =
-  /^\s*func\s+(?:\([^)]*\)\s*)?[A-Za-z_][A-Za-z0-9_]*(?:\s*\[[^\]]+\])?\s*\(.*\)\s*.*\{\s*(\/\/.*)?$/
+const GO_FUNC_LINE = /^\s*func\s+(?:\([^)]*\)\s*)?[A-Za-z_][A-Za-z0-9_]*(?:\s*\[[^\]]+\])?\s*\(.*\)\s*.*\{\s*(\/\/.*)?$/
 const PHP_END_BRACKET = /^\s*\);\s*(\/\/.*)?$/
 const PHP_HEREDOC_START = /<<<\s*(['"]?)([A-Za-z_][A-Za-z0-9_]*)\1/
 
@@ -124,6 +123,7 @@ const escapeRegExp = (value: string): string => {
 
 const isPhpHeredocTerminator = (line: string, label: string): boolean => {
   const escapedLabel = escapeRegExp(label)
+
   return new RegExp(`^\\s*${escapedLabel};?\\s*$`).test(line)
 }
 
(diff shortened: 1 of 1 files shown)


CarlosNietoP:
@codex review

packages/plugin-coverage/src/file-fixes.ts
```diff
@@ -0,0 +1,262 @@
...
+    () => undefined, // no validation
+    () => undefined // no error rendering
+  )
+
+  return absolutePaths.map((absPath) => upath.relative(rootPath, absPath))
+}
+
+const collectResults = (results: FileFixResult[]): FileFixes => {
+  const fileFixes: FileFixes = {}
+  let estimatedSize = 2 // {}
```
Drarig29: not sure what this means, and why it's useful

packages/plugin-coverage/src/file-fixes.ts
```diff
@@ -0,0 +1,262 @@
...
+    repoRoot = upath.resolve((await git.revparse(['--show-toplevel'])).trim())
+  } else {
+    repoRoot = upath.resolve(process.cwd())
+  }
+
+  if (git && !searchPath) {
+    // Use git ls-files for tracked files (respects .gitignore)
+    const output = await git.raw(['ls-files'])
+    const allFiles = output
+      .split('\n')
```
Drarig29: will `.split('\n')` work on Windows?

Drarig29: you may use https://github.com/DataDog/datadog-ci/blob/18498aaebdda0e831f67e4574c2b21639a9e9ddf/packages/base/src/helpers/git/get-git-data.ts#L50-L55

packages/plugin-coverage/src/file-fixes.ts
CarlosNietoP: It's 2 bytes for the estimated size of the string "{}". Comment is not great, I'll improve that!

packages/plugin-coverage/src/file-fixes.ts
CarlosNietoP: Good catch 👍

packages/plugin-coverage/src/file-fixes.ts
Drarig29: oh I read it as a Python euclidian division 🤦‍♂️

PR #2130, opened 2026-02-25
[SVLS-8622] add web app slot support

ava-silver:
### What and why?

This PR adds support for Azure Web App deployment slots in the AAS (Azure App Service) instrumentation commands. Previously, the CLI only supported instrumenting main Web Apps, but Azure allows creating deployment slots for staging, testing, and blue-green deployments. This enhancement enables users to instrument and uninstrument specific deployment slots.

Also, to be more clear on language, opts for 'Web App' as the resource type instead of the more generic 'App Service'. I'll make a follow up PR to also update the command to have `web-app` as an alias for the scope

I also discovered a bug in the windows runtime detection which I've fixed, where sometimes a nodejs app will not be properly identified since it uses the weird convention of requiring a specific env var to be set. Sometimes, the `netFrameworkVersion` is specified even for other runtimes, so we should check the other runtimes first, and then check the `netFrameworkVersion` as a last resort.

### How?

- Added a new `--slot` parameter to specify the slot name for instrumentation
- Updated the `WebApp` interface to include an optional `slot` property
- Modified Azure resource ID parsing to handle slot-specific resource paths (e.g., `/sites/{name}/slots/{slotName}`)
- Updated all Azure API calls to use slot-specific methods when a slot is specified (e.g., `getSlot`, `restartSlot`, `updateApplicationSettingsSlot`)
- Added utility functions `renderWebApp` and `resourceIdSegment` for consistent display and resource path generation
- Updated command descriptions and documentation to reflect "Web App (or slot)" terminology
- Updated the env var fetching to happen at the beginning which allows them to be properly used for the windows node case.

### Validation

I've done some thorough manual testing with web apps and validating that telemetry comes through, and I've also added/updated the unit tests in the plugin.

### Review checklist

### DIFF
--- a/packages/plugin-aas/src/commands/instrument.ts
+++ b/packages/plugin-aas/src/commands/instrument.ts
@@ -77,7 +77,7 @@ export class PluginCommand extends AasInstrumentCommand {
       )
     }
 
-    this.context.stdout.write(`${this.dryRunPrefix}🐶 Beginning instrumentation of Azure App Service(s)\n`)
+    this.context.stdout.write(`${this.dryRunPrefix}🐶 Beginning instrumentation of Web App(s)\n`)
     const results = await Promise.all(
       Object.entries(appServicesToInstrument).map(([subscriptionId, resourceGroupToWebApps]) =>
         this.processSubscription(subscriptionId, resourceGroupToWebApps, config)
@@ -109,7 +109,7 @@ export class PluginCommand extends AasInstrumentCommand {
   }
 
   /**
-   * Process an Azure App Service for instrumentation.
+   * Process an Web App or slot for instrumentation.
    * @returns A promise that resolves to a boolean indicating success or failure.
    */
   public async processWebApp(
@@ -181,9 +181,7 @@ This flag is only applicable for containerized .NET apps (on musl-based distribu
             ? aasClient.webApps.restartSlot(resourceGroup, webApp.name, webApp.slot)
             : aasClient.webApps.restart(resourceGroup, webApp.name))
         } catch (error) {
-          this.context.stdout.write(
-            renderError(`Failed to restart Web App ${renderWebApp(webApp)}: ${error}`)
-          )
+          this.context.stdout.write(renderError(`Failed to restart Web App ${renderWebApp(webApp)}: ${error}`))
 
           return false
         }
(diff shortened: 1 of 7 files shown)


packages/plugin-aas/src/commands/instrument.ts
```diff
@@ -144,28 +158,32 @@ export class PluginCommand extends AasInstrumentCommand {
...
       if (!this.dryRun) {
         try {
-          await aasClient.webApps.restart(resourceGroup, aasName)
+          await (webApp.slot
+            ? aasClient.webApps.restartSlot(resourceGroup, webApp.name, webApp.slot)
+            : aasClient.webApps.restart(resourceGroup, webApp.name))
         } catch (error) {
-          this.context.stdout.write(renderError(`Failed to restart Azure App Service ${chalk.bold(aasName)}: ${error}`))
+          this.context.stdout.write(
+            renderError(`Failed to restart Azure App Service ${renderWebApp(webApp)}: ${error}`)
```
TalUsvyatsky: Do we want web app here too?
```suggestion
            renderError(`Failed to restart Web App ${renderWebApp(webApp)}: ${error}`)
```

packages/plugin-aas/src/commands/instrument.ts
```diff
@@ -214,61 +232,75 @@ This flag is only applicable for containerized .NET apps (on musl-based distribu
...
+      await this.updateEnvVars(client, resourceGroup, webApp, existingEnvVars, envVars)
 
       return
     }
 
-    const envVarsPromise = this.updateEnvVars(client, resourceGroup, aasName, envVars)
+    const envVarsPromise = this.updateEnvVars(client, resourceGroup, webApp, existingEnvVars, envVars)
 
-    this.context.stdout.write(`${this.dryRunPrefix}Stopping Azure App Service ${chalk.bold(aasName)}\n`)
+    this.context.stdout.write(`${this.dryRunPrefix}Stopping Azure App Service ${renderWebApp(webApp)}\n`)
```
TalUsvyatsky:
```suggestion
    this.context.stdout.write(`${this.dryRunPrefix}Stopping Web App ${renderWebApp(webApp)}\n`)
```

PR #2137, opened 2026-03-02
chore: consolidate api helpers

ava-silver:
### What and why?

This PR centralizes Datadog site configuration handling by introducing new helper functions `getDatadogSite()`, `getDatadogSiteFromEnv()`, `getIntakeUrl()`, and `getApiUrl()` in the API helpers module. Previously, site configuration was handled inconsistently across the codebase with repeated patterns of `process.env.DATADOG_SITE || process.env.DD_SITE || 'datadoghq.com'`.

### How?

- Added `getDatadogSite()` and `getDatadogSiteFromEnv()` functions to provide a single source of truth for site resolution
- Added `getIntakeUrl()` function to build intake URLs with optional environment variable overrides
- Added `getApiUrl()` function to construct API URLs
- Replaced all manual site configuration patterns across commands (dsyms, elf-symbols, flutter-symbols, sourcemaps, synthetics, etc.) with calls to the new helper functions
- Updated intake URL construction in various modules to use the centralized `getIntakeUrl()` function

### Review checklist


packages/base/src/commands/measure/measure.ts
```diff
@@ -141,8 +143,7 @@ export class MeasureCommand extends BaseCommand {
       throw new Error('API key is missing')
     }
 
-    const site = process.env.DATADOG_SITE || process.env.DD_SITE || 'datadoghq.com'
-    const baseAPIURL = `https://${getApiHostForSite(site)}`
```
ava-silver, line 145: this helper has a weird remapping of staging to point at `app.datad0g.com` instead of `api.datad0g.com`, so I'm leaving these as is to not break anything

packages/base/src/commands/measure/measure.ts
```diff
@@ -5,12 +5,14 @@ import {Command, Option} from 'clipanion'
...
 import {FIPS_ENV_VAR, FIPS_IGNORE_ERROR_ENV_VAR} from '@datadog/datadog-ci-base/constants'
+import {getDatadogSite} from '@datadog/datadog-ci-base/helpers/api'
 import {envDDGithubJobName, getGithubJobNameFromLogs, getCIEnv} from '@datadog/datadog-ci-base/helpers/ci'
 import {toBoolean} from '@datadog/datadog-ci-base/helpers/env'
 import {enableFips} from '@datadog/datadog-ci-base/helpers/fips'
 import {retryRequest} from '@datadog/datadog-ci-base/helpers/retry'
 import {parseMeasuresFile} from '@datadog/datadog-ci-base/helpers/tags'
-import {getApiHostForSite, getRequestBuilder} from '@datadog/datadog-ci-base/helpers/utils'
+import {getApiHostForSite} from '@datadog/datadog-ci-base/helpers/utils'
+import {getRequestBuilder} from '@datadog/datadog-ci-base/helpers/utils'
```
Drarig29:
```suggestion
import {getApiHostForSite, getRequestBuilder} from '@datadog/datadog-ci-base/helpers/utils'
```

Drarig29 (approved):
LGTM

packages/base/src/commands/measure/measure.ts
ava-silver: interesting, im surprised `yarn format` doesnt catch this -- i'll see if i can fix that in a different PR

ava-silver: https://github.com/DataDog/datadog-ci/pull/2140

Drarig29 (approved):

seungleelee (approved):

jasonforal (approved):

avangelillo (approved):

packages/base/src/helpers/api.ts
```diff
@@ -1,13 +1,46 @@
...
  */
-export const getBaseIntakeUrl = (intake: string) => {
-  if (process.env.DATADOG_SITE || process.env.DD_SITE) {
-    return `https://${intake}.${process.env.DATADOG_SITE || process.env.DD_SITE}`
+export const getDatadogSite = (site?: string): string => site || getDatadogSiteFromEnv() || DATADOG_SITE_US1
+
+/**
+ * Build `https://{subdomain}.{site}` with optional env var override.
+ */
+export const getIntakeUrl = (
```
GabrielAnca, line 18: Now that we have a common helper it would be handy to have some unit tests to cover these functions 😄

packages/base/src/helpers/base-intake-url.ts
```diff
@@ -1,9 +1,4 @@
...
-    return process.env.DATADOG_SOURCEMAP_INTAKE_URL
-  } else if (datadogSite) {
-    return 'https://sourcemap-intake.' + datadogSite
-  }
+import {getIntakeUrl} from './api'
 
-  return 'https://sourcemap-intake.datadoghq.com'
-}
+export const getBaseSourcemapIntakeUrl = (datadogSite?: string) =>
+  getIntakeUrl('sourcemap-intake', {overrideEnvVar: 'DATADOG_SOURCEMAP_INTAKE_URL', site: datadogSite})
```
GabrielAnca, line 4: I find it odd that we have methods for the intake url here and in api.ts. Could we maybe move this method to api.ts? Or maybe group all intake url functions in a single file? Or is it maybe a breaking change and it's safer to leave things where they are?

packages/base/src/helpers/metrics.ts
```diff
@@ -14,7 +16,7 @@ export interface MetricsLoggerOptions {
 }
 
 export const getMetricsLogger = (opts: MetricsLoggerOptions): MetricsLogger => {
-  const apiUrl = getBaseAPIUrl(opts.datadogSite)
+  const apiUrl = 'api.' + getDatadogSite(opts.datadogSite)
```
GabrielAnca, line 19: Can we use `getApiHostForSite` here?


```suggestion
  const apiUrl = getApiHostForSite(opts.datadogSite)
```

GabrielAnca (approved):
Super happy to see this standardized ❤️ Just a couple of minor feedback

packages/base/src/helpers/base-intake-url.ts
ava-silver: if any customers are relying on imports then it would be a breaking change unless we re-export it. for now, I'm inclined to move things as little as possible and leave that up to individuals for what makes sense

packages/base/src/helpers/api.ts
ava-silver: good callout! I'll do that in a separate PR to minimize the diff here

packages/base/src/helpers/metrics.ts
ava-silver: unfortunately no, that function has subtly wrong behavior for staging: https://github.com/DataDog/datadog-ci/blob/89048e446ab5f4c1ac1d394f46e4b426576c1851/packages/base/src/helpers/utils.ts#L231-L234

I'm honestly not sure why that got introduced 5 years ago, if we can confirm that its current usages would be fine to migrate to the proper behavior (using `api.datad0g.com`), then we can make it a helper and use it across the board. But I'm calling that out of scope for now.

packages/base/src/helpers/api.ts
ava-silver: https://github.com/DataDog/datadog-ci/pull/2143

PR #1688, opened 2025-06-13
[SYNTH-20482] Mark `/w` as safe in docker image entrypoint

Drarig29:
### What and why?

Closes #1658 

Same as https://github.com/DataDog/datadog-sca-github-action/blob/e373cb84ac57d35fbefd4c7c7ee5b4592e976fbc/entrypoint.sh#L71

### How?

Git fails when the current user inside the `docker/ci` container is not the user owning the files:

```sh
fatal: detected dubious ownership in repository at '/w'
```

And `/w` is the directory where we expect users to mount their host's repository folder: [see docs](https://github.com/DataDog/datadog-ci/blob/42ce08a40d1905ffb9846c1d292db09a916edec2/README.md?plain=1#L197).

### Review checklist

- [x] Feature or bugfix MUST have appropriate tests (unit, integration)

PR #1661, opened 2025-06-04
Add current folder as a safe.directory in git config

nikita-tkachenko-datadog:
### What and why?

When creating a git client, adds current folder as a `safe.directory` in git config.

Some CI providers check out repository as one user, then run the CI jobs as a different user.
When trying to use git CLI in a folder that belongs to a different user, the `detected dubious ownership in repository at ...` error occurs.
To avoid this, the repo folder needs to be added a safe directory.

### Review checklist

- [ ] Feature or bugfix MUST have appropriate tests (unit, integration)

Issue #1658, opened 2025-06-04
docker image can get git metatdata


staffanf:
### Bug description

When trying to upload git metadata you are hit with this

```
docker run --rm -v "$(pwd):/w" \
                    -e DD_API_KEY \
                    -e DD_APP_KEY \
                    -e DD_SITE \
                    -e DD_TAGS \
                    datadog/ci:latest git-metadata upload --dry-run
⚠️ DRY-RUN MODE ENABLED. WILL NOT UPLOAD
Uploading list of tracked files...
❌ Failed upload: fatal: detected dubious ownership in repository at '/w'
To add an exception for this directory, call:

	git config --global --add safe.directory /w

Failed upload of tracked files: Error: fatal: detected dubious ownership in repository at '/w'
To add an exception for this directory, call:

	git config --global --add safe.directory /w

Syncing GitDB...
Could not write to GitDB: Error: fatal: detected dubious ownership in repository at '/w'
To add an exception for this directory, call:

	git config --global --add safe.directory /w

Command failed. See messages above for more details.
```

since my current directory is not owned by root

### Describe what you expected

I expect the git-metadata upload to work from the dirs I mount into the docker image.

This can be fixed by adding 

`RUN git config --global --add safe.directory /w`

to the Dockerfile but that only works if the workdir is not changed, so maybe there is a better fix.

This also fail for junit upload but with no fail message

### Steps to reproduce the issue

```
docker run --rm -v "$(pwd):/w" \
                    -e DD_API_KEY \
                    -e DD_APP_KEY \
                    -e DD_SITE \
                    -e DD_TAGS \
                    datadog/ci:latest git-metadata upload --dry-run
```
from a git dir

### Additional context

Run from latest docker-ce in Ubuntu 24.04 using the latest datadog/ci image

### Command

git-metadata

Drarig29:
Hi @staffanf! This should be fixed in next release. Feel free to reopen the issue if it's not.

PR #2152, opened 2026-03-06
Only add `safe.directory` to git global config in CI and Docker contexts

Drarig29:
### What and why?

Personal laptops were having their global git config modified on every `yarn test` and normal runs. The `safe.directory` workaround is only needed when the repo is checked out by a different user (Docker volume mounts, CI envs).

Related:
- https://github.com/DataDog/datadog-ci/issues/1658
- https://github.com/DataDog/datadog-ci/pull/1661
- https://github.com/DataDog/datadog-ci/pull/1688

### How?

Only add cwd to safe directories in Docker and CI contexts.

### Review checklist

### DIFF
--- a/packages/base/src/commands/git-metadata/git.ts
+++ b/packages/base/src/commands/git-metadata/git.ts
@@ -3,6 +3,7 @@ import {URL} from 'url'
 import * as simpleGit from 'simple-git'
 import {GitConfigScope} from 'simple-git'
 
+import {isCI} from '../../helpers/ci'
 import {gitRemote} from '../../helpers/git/get-git-data'
 
 import {CommitInfo} from './interfaces'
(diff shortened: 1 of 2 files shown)


packages/base/src/commands/git-metadata/git.ts
```diff
@@ -18,12 +18,17 @@ export const newSimpleGit = async (): Promise<simpleGit.SimpleGit> => {
 
   const git = simpleGit.simpleGit(options)
 
-  try {
-    // In some CI envs repo may be checked out as a different user than the one running the command.
-    // To be able to run git commands, we need to add the current directory as a safe directory.
-    await git.addConfig('safe.directory', currentDir, true, GitConfigScope.global)
-  } catch (e) {
-    // Ignore the error
+  const isCI = 'CI' in process.env
```
nikita-tkachenko-datadog: I don't think every provider that we support sets the `CI` variable (for instance, Claude suggests that Jenkins, Azure Pipelines, AWS CodePipeline, TeamCity, and Codefresh don't).
To err on the side of caution, could we add a fallback that relies on our CI provider detection?
```suggestion
  const isCI = 'CI' in process.env || getCIProvider() !== 'unknown'
```
I think `import {getCIProvider} from '../../helpers/ci'` would have to be add for that

nikita-tkachenko-datadog (approved):

PR #2165, opened 2026-03-13
[SYNTH-24187] Add `planDryRun` public API and `--dry-run` flag to `run-tests`

Drarig29:
### What and why?

Exposes test planning as a public library API so callers can inspect what would be triggered before (or instead of) running.

- **`planDryRun`**: New public function that resolves the test plan (tests, overrides, execution rules) without running anything. Exported from `index.ts`.
- **`--dry-run` flag**: When passed to `run-tests`, calls `planDryRun` and prints a human-readable summary via the new `dryRunEnd` reporter method.
- **`TestPlan` / `TestPlanItem`**: New exported types representing the resolved plan. Each item carries `test`, `testOverrides`, and `executionRule`.
- **`filterTestPlan`**: New public utility to filter a plan while keeping `initialSummary` in sync — useful for callers who want to customize what gets run before calling `executeTests`.
- **`executeTests` / `executeWithDetails`**: Accept an optional pre-resolved `testPlan` and `initialSummary` to skip re-fetching when planning and executing separately.

### How?

`planDryRun` reuses the existing `getTriggerConfigs` + `getTestsToTrigger` pipeline and returns `{testPlan, initialSummary}`. `executeTests` was updated to accept an optional `testPlan`; when provided it skips the fetch phase and goes straight to triggering. `TestSkipped` and `TestWithOverride` were extended with `executionRule` so callers have full context.

### Review checklist

### DIFF
--- a/packages/plugin-synthetics/src/utils/internal.ts
+++ b/packages/plugin-synthetics/src/utils/internal.ts
@@ -111,11 +111,11 @@ export const isTimedOutRetry = (
 }
 
 export const isLocalTriggerConfig = (triggerConfig?: TriggerConfig): triggerConfig is LocalTriggerConfig => {
-  return triggerConfig ? 'localTestDefinition' in triggerConfig : false
+  return !!triggerConfig && 'localTestDefinition' in triggerConfig
 }
 
 export const isLocalTestPayload = (testPayload: TestPayload): testPayload is LocalTestPayload => {
-  return testPayload ? 'local_test_definition' in testPayload : false
+  return !!testPayload && 'local_test_definition' in testPayload
 }
 
 export const isBrowserServerResult = (serverResult: ServerResult): serverResult is BrowserServerResult => {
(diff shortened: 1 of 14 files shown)


packages/plugin-synthetics/src/utils/internal.ts
```diff
@@ -113,6 +114,10 @@ export const isLocalTriggerConfig = (triggerConfig?: TriggerConfig): triggerConf
   return triggerConfig ? 'localTestDefinition' in triggerConfig : false
 }
 
+export const isLocalTestPayload = (testPayload: TestPayload): testPayload is LocalTestPayload => {
+  return testPayload ? 'local_test_definition' in testPayload : false
```
ava-silver: nit:

```suggestion
  return !!testPayload && 'local_test_definition' in testPayload
```

ava-silver (approved):

Matt-Cam (approved):

PR #2170, opened 2026-03-18
[SVLS-7940] add container-app integration tests

ava-silver:
### What and why?

Adds end-to-end integration tests for the `container-app instrument/uninstrument` commands against real Azure infrastructure. The existing unit tests mock the Azure SDK, which can miss integration issues (auth flows, API contract changes, SDK incompatibilities). This is the first cloud-provider e2e test in the repo.

### How?

**Ephemeral container apps**: each CI run creates a fresh container app (`dd-ci-{run_id}-n{version}`), runs instrument → verify → uninstrument → verify, then deletes it. This avoids race conditions between concurrent CI runs and ensures a clean baseline without needing Bicep resets.

**Azure OIDC auth**: uses `azure/login` with federated identity credentials (no client secrets stored in GitHub). Flexible federated credentials (preview) allow all branches to authenticate.

**Path-based gating**: a `check-container-app-changes` job uses `dorny/paths-filter` to skip the Azure steps when no relevant files changed (`packages/plugin-container-app/**`, `packages/base/package.json`, `packages/base/src/helpers/serverless/**`, `.github/workflows/e2e/container-app/**`, `.github/workflows/ci.yml`).

**Verification script** (`verify-container-app.js`): validates the container app state after each command by checking sidecar container, shared volume, DD_* env vars, secrets, volume mounts, and tags.

**New files**:
- `.github/workflows/e2e/container-app/main.bicep` — Bicep template for one-time environment setup
- `.github/workflows/e2e/container-app/README.md` — setup instructions
- `.github/workflows/e2e/container-app/verify-container-app.js` — post-command verification

**New GitHub secrets**: `AZURE_CLIENT_ID_E2E`, `AZURE_TENANT_ID_E2E`, `AZURE_SUBSCRIPTION_ID_E2E`, `AZURE_RESOURCE_GROUP_E2E`, `AZURE_CONTAINER_APP_ENV_E2E`

### Future considerations

- The e2e test steps currently run sequentially within the existing `e2e-test` job. As more cloud-provider integration tests are added (lambda, cloud-run, aas), we should consider splitting them into parallel jobs to avoid the e2e suite becoming a bottleneck — individual integration tests can take multiple minutes each.

### Review checklist

- [x] Azure infrastructure provisioned and secrets configured
- [x] OIDC federated credentials configured per SECENG guide
- [x] Verified full cycle locally: create → instrument → verify → uninstrument → verify → delete


.github/workflows/ci.yml
```diff
@@ -200,11 +203,84 @@ jobs:
...
           DD_API_KEY: ${{ secrets.DD_API_KEY_CI_VISIBILITY }}
           DD_APP_KEY: ${{ secrets.DD_APP_KEY_CI_VISIBILITY }}
+      - name: Azure login (OIDC)
+        if: needs.check-container-app-changes.outputs.should_run == 'true'
+        uses: azure/login@a457da9ea143d694b1b9c7c869ebb04ebe844ef5 # v2
+        with:
+          client-id: ${{ secrets.AZURE_CLIENT_ID_E2E }}
+          tenant-id: ${{ secrets.AZURE_TENANT_ID_E2E }}
+          subscription-id: ${{ secrets.AZURE_SUBSCRIPTION_ID_E2E }}
+      - name: Create ephemeral container app
```
avangelillo, line 213: What are your thoughts on doing this in the github scripts like this vs using a testing framework like jest / vitest?

.github/workflows/ci.yml
```diff
@@ -200,11 +203,84 @@ jobs:
...
+          AZURE_RESOURCE_GROUP: ${{ secrets.AZURE_RESOURCE_GROUP_E2E }}
+      - name: Verify container app is uninstrumented
+        if: needs.check-container-app-changes.outputs.should_run == 'true'
+        run: node artifacts/e2e/container-app/verify-container-app.js uninstrumented
+        env:
+          AZURE_SUBSCRIPTION_ID: ${{ secrets.AZURE_SUBSCRIPTION_ID_E2E }}
+          AZURE_RESOURCE_GROUP: ${{ secrets.AZURE_RESOURCE_GROUP_E2E }}
+          AZURE_CONTAINER_APP_NAME: dd-ci-${{ github.run_id }}-n${{ matrix.version }}
+      - name: Delete ephemeral container app
+        if: always() && needs.check-container-app-changes.outputs.should_run == 'true'
```
avangelillo, line 257: Is there a way to do this by deploying an arm template and destroying the template at the end?  If this fails to run we could end up with resources that stay afterwards or if we want to add resources that get deployed we need to remember to clean them up since they're not grouped in the template

ava-silver: you can't destroy an arm template like you can with cloudformation -- arm templates are also significantly slower than direct API calls like we're doing here

I think it could be worth adding something to the e2e environment that auto-deletes apps that are older than an hour or something, which would fix that

.github/workflows/ci.yml
ava-silver: This follows what other integration tests have done, but especially as we add more integration tests, itll probably get way more unwieldy -- i'll check in with Corentin if there's a reason we dont use jest for integration tests

ava-silver: I'll do a migration after this to use Jest for the integration tests, which should help with maintainability

PR #2175, opened 2026-03-20
[SVLS-8731] add aas integration tests

ava-silver:
### What and why?

This PR adds end-to-end testing for Azure App Service (AAS) instrumentation and uninstrumentation functionality. The tests verify that the `datadog-ci aas` commands correctly configure both Linux and Windows web apps with the necessary Datadog monitoring settings, extensions, and sidecar containers.

### How?

The implementation adds a new CI job `check-aas-changes` that triggers AAS tests when relevant files are modified. The end-to-end tests create ephemeral Azure web apps (one Linux, one Windows) during test execution, run the instrumentation commands, verify the expected configuration changes, then clean up the resources. The verification logic checks for proper app settings, sidecar containers on Linux, site extensions on Windows, and resource tags.

### Review checklist

### DIFF
--- a/e2e/helpers/exec.ts
+++ b/e2e/helpers/exec.ts
@@ -32,7 +32,16 @@ export const execPromise = async (command: string, env?: Record<string, string |
 }
 
 // Transient Azure errors that are safe to retry
-const RETRYABLE_PATTERNS = ['GatewayTimeout', 'RestError', 'Operation was canceled', 'ETIMEDOUT', 'ECONNRESET']
+const RETRYABLE_PATTERNS = [
+  'GatewayTimeout',
+  'RestError',
+  'Operation was canceled',
+  'ETIMEDOUT',
+  'ECONNRESET',
+  "doesn't exist",
+  'Conflict',
+  'TooManyRequests',
+]
 
 const isRetryable = (result: ExecResult): boolean => {
   const output = `${result.stdout} ${result.stderr}`


OliviaShoup (approved):
thanks for the PR! 🚀

e2e/aas-linux.test.ts
```diff
@@ -0,0 +1,62 @@
...
+
+import {verifyLinuxInstrumented, verifyLinuxUninstrumented} from './helpers/aas-verifier'
+import {DATADOG_CI_COMMAND, execPromise, execPromiseWithRetries} from './helpers/exec'
+
+const describeOrSkip = process.env.SKIP_AAS_TESTS === 'true' ? describe.skip : describe
+
+describeOrSkip('aas (Linux)', () => {
+  const subscriptionId = process.env.AZURE_SUBSCRIPTION_ID!
+  const resourceGroup = process.env.AZURE_RESOURCE_GROUP!
+  const linuxAppName = `dd-ci-aas-linux-${crypto.randomBytes(4).toString('hex')}`
```
avangelillo, line 11: Is it useful to have the app name be based on the PR title or user running the tests?  We do something like that for remote instrumentation so if resources get abandoned you can tell if they can be deleted or not by the name

e2e/aas-linux.test.ts
```diff
@@ -0,0 +1,62 @@
...
+  const linuxPlan = process.env.AZURE_AAS_LINUX_PLAN!
+
+  beforeAll(async () => {
+    const result = await execPromise(
+      `az webapp create` +
+        ` --name "${linuxAppName}"` +
+        ` --resource-group "${resourceGroup}"` +
+        ` --plan "${linuxPlan}"` +
+        ` --runtime "NODE:20-lts"` +
```
avangelillo: Do we want node20 specifically or can we upgrade to something that isn't [at its EOL](https://nodejs.org/en/about/previous-releases)

e2e/aas-linux.test.ts
ava-silver: instead i'll set up a job to clean things up by age -- that way we dont have to worry about it at all

e2e/aas-linux.test.ts
ava-silver: good callout!

ava-silver: I updated them to 22lts since 24 isnt supported on windows

PR #2180, opened 2026-03-23
Support stages and steps for custom tags / measures

### LINKED ISSUE #1260: Send tags and measures on step level

When enriching our CI Visibility with tags and measure, we often need to add them on a step, in addition to pipeline and job.

### Solution

Allow `tag` and `measure` command to take `step` as a valid value for `--level`

### Additional context

_No response_

### Command

tag

rodrigo-roca:
### What and why?
Closes: https://github.com/DataDog/datadog-ci/issues/1260

Supports stages for: Azure, Gitlab and Jenkins.
Support steps for: Github Actions.

### How?

We read new environment variables which will be used in the backend to perform the correlation. Also, the Jenkins env variable requires the new release of the Datadog plugin in order to work.

### Review checklist


packages/base/src/helpers/ci.ts
```diff
@@ -1259,6 +1313,156 @@ export const getGithubJobNameFromLogs = (context: BaseContext): string | undefin
...
+ */
+export const getGithubStepInfoFromLogs = (context: BaseContext): {jobDisplayName: string; stepIndex: number} => {
+  if (getCIProvider() !== CI_ENGINES.GITHUB) {
+    throw new Error('Step level is only supported for GitHub Actions')
+  }
+
+  const githubAction = process.env.GITHUB_ACTION
+  if (!githubAction) {
+    throw new Error('GITHUB_ACTION environment variable is not set, cannot determine step index')
+  }
```
juan-fernandez, line 1364: why is `getCIProvider() !== CI_ENGINES.GITHUB` not enough? Are there cases where `process.env.GITHUB_ACTIONS` is defined but `process.env.GITHUB_ACTION` isn't?

juan-fernandez (approved):

Drarig29 (approved):
Same question as Juan

packages/base/src/helpers/ci.ts
rodrigo-roca: `GITHUB_ACTION` tells the CLI which step to look for. E.g: `__run2` means it is the second step that does not use an action. This is  correlated with what we find in the worker log

packages/base/src/commands/measure/README.md
```diff
@@ -14,9 +14,10 @@ For example:
 datadog-ci measure --level job --measures binary.size:1024
 ```
 
-- `--level` Has to be one of `[pipeline, job]`. It will determine in what span the measures will be added. If pipeline
-  is selected then the measures will be added to the pipeline trace span. If job is selected it will be added to the
-  span for the currently running job.
+- `--level` Has to be one of `[pipeline, job, stage, step]`. It will determine in what span the measures will be added.
+  If pipeline is selected then the measures will be added to the pipeline trace span. If job is selected it will be added
```
jeff-morgan-dd: You might consider putting the values in backticks in these descriptions, i.e. "If `pipeline` is selected", "If `stage` is selected", etc. Non-blocking, up to you!

jeff-morgan-dd (approved):
Just one suggestion for your consideration. Thank you!

packages/base/src/helpers/ci.ts
rodrigo-roca: This is indeed for cases  where process.env.GITHUB_ACTIONS is defined but process.env.GITHUB_ACTION isn't, which should not happen, but it is preferable to have that covered in the code

PR #2195, opened 2026-03-31
[security] Pin all direct dependencies to exact versions

### LINKED ISSUE #2194: Please consider locking all dependencies version to mitigate Supply Chain attacks

There might be transitive malicious dependency that may be resolved during installation if the package is not locked. Since this tool is heavily downloaded, please consider hardening the dependency version.

https://www.stepsecurity.io/blog/axios-compromised-on-npm-malicious-versions-drop-remote-access-trojan

### Solution

_No response_

### Additional context

_No response_

### Command

None

Drarig29:
### What and why?

Closes #2194 

The goal of this PR is to protect ourselves and consumers from supply chain attacks due to usage of `@datadog/datadog-ci`.

### How?

- Pin exact versions of all our direct dependencies
- Audit of automatic plugin installations

### Validation

You can run this command to prove we aren't doing any accidental dependency bumps (the goal is only to pin the versions we already have in `yarn.lock`):
```sh
diff <(git show master:yarn.lock | grep '^  resolution:' | sort) <(git show HEAD:yarn.lock | grep '^  resolution:' | sort)
```


packages/base/src/helpers/plugin.ts
```diff
@@ -191,7 +191,7 @@ const temporarilyInstallPluginWithNpx = async (scope: string) => {
   const {basePackage, pluginPackage} = getPackagesToInstall(scope)
 
   const emitPath = isWindows ? 'set PATH' : 'printenv PATH'
-  const cmd = `npx -y -p ${basePackage} -p ${pluginPackage} ${emitPath}`
+  const cmd = `npx --ignore-scripts -y -p ${basePackage} -p ${pluginPackage} ${emitPath}`
```
marco-saia-datadog, line 194: I have wrote a script to check which packages rely on post-install scripts, and did a quick analysis.

It seems like only some `devDependencies` rely on `postinstall`, so this change should be fine.

These are all the dependencies that use postinstall scripts, if you want to double-check or have some more context:

#### `esbuild@0.25.9` -> `node install.js`
**Necessary**: downloads and installs the correct platform-specific binary for esbuild (e.g. `@esbuild/darwin-arm64`). Without it, the `esbuild` executable is missing and the package is non-functional.
**Type:** Transitive `devDependency` (pulled in by `tsx` and `@datadog/datadog-ci`)

---

#### `oxc-resolver@11.8.4` -> `napi-postinstall oxc-resolver 11.8.4 check`
**Necessary** -> uses `napi-postinstall` to detect the current OS/arch and ensure the correct platform-specific native `.node` binary is present (downloading it from npm if missing). Without it, the Rust-compiled resolver cannot load.
**Type:** Transitive devDependency (pulled in by `knip`)

---

#### `unrs-resolver@1.11.1` -> `napi-postinstall unrs-resolver 1.11.1 check`
**Necessary** -> same as `oxc-resolver`: ensures the correct platform-specific native binary is present. Without it, the resolver fails to load.
**Type:** Transitive devDependency (pulled in by `eslint-import-resolver-typescript` and `eslint-plugin-import-x`)

---

#### `protobufjs@7.3.2` (×2, under `google-gax` and `proto3-json-serializer`)
**Not strictly necessary** -> checks whether the parent package declares the dependency using the correct `~` (compatible) version scheme (e.g. `~7.3.2` instead of `^7.3.2`), and prints a warning to stderr if not. It's a developer-advisory script only and has zero impact on runtime functionality.
**Type:** Transitive devDependency (pulled in by `@datadog/datadog-ci-plugin-cloud-run` → `@google-cloud/logging` / `@google-cloud/run`)

---

#### `protobufjs@7.5.3` (top-level `node_modules/protobufjs`)
Same as above -> version-scheme advisory warning, not necessary for the package to function.
**Type:** Transitive devDependency (same chain as above)

marco-saia-datadog (approved):

vksrinivasan1 (approved):

ava-silver (approved):

PR #2201, opened 2026-03-31
chore: migrate base package from axios to fetch/undici

ava-silver:
### What and why?

Replace `axios` (and its transitive dependency `follow-redirects`, which has had multiple CVEs) with `undici.fetch` (what powers native fetch in node) in `packages/base`. This is **Stage 1** of a multi-stage migration — it covers the base package only; downstream plugins continue to compile via backward-compat type aliases.

This is something we have been discussing internally and was also brought up in #2199

### How?

- New `request.ts` module wraps `undici.fetch` with an axios-like API surface (`httpRequest`, `RequestError`, `RequestConfig`, `RequestResponse`)
- Proxy support via `undici.ProxyAgent` (explicit proxy URL) and `EnvHttpProxyAgent` (env-var auto-detection)
- All `packages/base` consumers (`apikey`, `flare`, `trace`, `tag`, `measure`, `gitdb`) switched to new request module
- `interfaces.ts` exports backward-compat type aliases (`AxiosRequestConfig`, `AxiosPromise`) so downstream plugins compile unchanged
- Tests rewired from `jest.spyOn(axios)` to `jest.mock('../request')` pattern

**Dependency changes:**

- Removed: `axios` (+ transitive `follow-redirects`), `proxy` (test devDep)
- Added: `undici` 7.24.6 (zero transitive deps)

### Known Stage 1 workarounds (cleaned up in Stage 2)

- `RequestConfig` fields typed as `any` for compat with axios-typed plugin callers
- `RequestBuilder = (args: any) => Promise<any>` — permissive so both old and new callers compile
- `AxiosRequestConfig` / `AxiosPromise` type aliases exported from `interfaces.ts`
- `getAxiosError` test helper typed as `any` for synthetics test compat
- `proxy-agent` remains as a dependency (used by synthetics WebSocket tunnel and `datadog-metrics`)

### Behavioral note

`undici.ProxyAgent` only tunnels HTTPS traffic (via CONNECT); plain HTTP bypasses the proxy. The proxy routing tests were removed accordingly. This matches real-world usage since all Datadog API endpoints are HTTPS.

### Remaining stages

- **Stage 2**: Migrate 9 plugin packages — mechanical type swaps (`AxiosRequestConfig` → `RequestConfig`, etc.) across ~23 files, remove compat aliases
- **Stage 3**: Migrate `plugin-synthetics` separately — has custom retry logic, `paramsSerializer`, and heavier test mocking
- **Stage 4**: Remove `proxy-agent` dep, clean up lockfile, verify `follow-redirects` is fully gone from the tree

### Review checklist

- [x] All existing tests pass
- [x] CI still passes

### DIFF
--- a/packages/base/src/commands/trace/api.ts
+++ b/packages/base/src/commands/trace/api.ts
@@ -1,26 +1,20 @@
 import type {Payload} from './interfaces'
-import type {AxiosPromise, AxiosRequestConfig, AxiosResponse} from 'axios'
+import type {RequestBuilder} from '@datadog/datadog-ci-base/helpers/interfaces'
 
 import {getRequestBuilder} from '@datadog/datadog-ci-base/helpers/utils'
 
-// Dependency follows-redirects sets a default maxBodyLength of 10 MB https://github.com/follow-redirects/follow-redirects/blob/b774a77e582b97174813b3eaeb86931becba69db/index.js#L391
-// We don't want any hard limit enforced by the CLI, the backend will enforce a max size by returning 413 errors.
-const maxBodyLength = Infinity
-
-export const reportCustomSpan =
-  (request: (args: AxiosRequestConfig) => AxiosPromise<AxiosResponse>) => async (customSpan: Payload) => {
-    return request({
+export const reportCustomSpan = (request: RequestBuilder) => async (customSpan: Payload) => {
+  return request({
+    data: {
       data: {
-        data: {
-          type: 'ci_app_custom_span',
-          attributes: customSpan,
-        },
+        type: 'ci_app_custom_span',
+        attributes: customSpan,
       },
-      maxBodyLength,
-      method: 'POST',
-      url: '/api/intake/ci/custom_spans',
-    })
-  }
+    },
+    method: 'POST',
+    url: '/api/intake/ci/custom_spans',
+  })
+}
 
 export const apiConstructor = (baseUrl: string, apiKey: string) => {
   const requestIntake = getRequestBuilder({baseUrl, apiKey})
--- a/packages/base/src/helpers/__tests__/testing-tools.ts
+++ b/packages/base/src/helpers/__tests__/testing-tools.ts
@@ -1,12 +1,12 @@
 import type {MockCommandContext} from '../interfaces'
-import type {AxiosResponse, InternalAxiosRequestConfig} from 'axios'
 import type {BaseContext, CommandClass} from 'clipanion'
 import type {CommandOption} from 'clipanion/lib/advanced/options'
 import type {Writable} from 'stream'
 
-import {AxiosError} from 'axios'
 import {Cli, Command} from 'clipanion'
 import upath from 'upath'
+
+import {RequestError} from '../request'
 
 export const MOCK_BASE_URL = 'https://app.datadoghq.com/'
 export const MOCK_DATADOG_API_KEY = '02aeb762fff59ac0d5ad1536cd9633bd'
@@ -137,10 +137,12 @@ export const createCommand = <T extends Command>(
   return command
 }
 
-export const getAxiosError = (status: number, {errors, message}: {errors?: string[]; message?: string}) => {
-  const serverError = new AxiosError(message) as AxiosError<any> & {config: InternalAxiosRequestConfig}
-  serverError.config = {baseURL: MOCK_BASE_URL, url: 'example'} as InternalAxiosRequestConfig
-  serverError.response = {data: {errors}, status} as AxiosResponse
+export const getRequestError = (status: number, {errors, message}: {errors?: string[]; message?: string}) =>
+  new RequestError(
+    message ?? 'Request failed',
+    {baseURL: MOCK_BASE_URL, url: 'example'},
+    {data: {errors}, status, statusText: ''}
+  )
 
-  return serverError
-}
+// Backward-compat alias for plugin tests (removed in Stage 2)
+export const getAxiosError: any = getRequestError
--- a/packages/base/src/helpers/__tests__/testing-tools.ts
+++ b/packages/base/src/helpers/__tests__/testing-tools.ts
@@ -145,4 +145,4 @@ export const getRequestError = (status: number, {errors, message}: {errors?: str
   )
 
 // Backward-compat alias for plugin tests (removed in Stage 2)
-export const getAxiosError: any = getRequestError
+export const getAxiosError = getRequestError
--- a/packages/base/src/helpers/__tests__/utils.test.ts
+++ b/packages/base/src/helpers/__tests__/utils.test.ts
@@ -1,12 +1,14 @@
 import http from 'http'
 
-import type {AxiosPromise, AxiosRequestConfig} from 'axios'
+import type {RequestConfig} from '../request'
 import type {AddressInfo} from 'net'
 
-import axios from 'axios'
-import {createProxy} from 'proxy'
-import {ProxyAgent} from 'proxy-agent'
+jest.mock('../request', () => ({
+  ...jest.requireActual('../request'),
+  httpRequest: jest.fn(),
+}))
 
+import * as requestModule from '../request'
 import * as ciUtils from '../utils'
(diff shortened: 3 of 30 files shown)


ava-silver:
* **#2249** <a href="https://app.graphite.com/github/pr/DataDog/datadog-ci/2249?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2248** <a href="https://app.graphite.com/github/pr/DataDog/datadog-ci/2248?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2247** <a href="https://app.graphite.com/github/pr/DataDog/datadog-ci/2247?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2201** <a href="https://app.graphite.com/github/pr/DataDog/datadog-ci/2201?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/datadog-ci/2201?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

nina9753 (approved):

rodrigo-roca (approved):

packages/base/src/helpers/__tests__/testing-tools.ts
```diff
@@ -137,10 +137,12 @@ export const createCommand = <T extends Command>(
...
+  new RequestError(
+    message ?? 'Request failed',
+    {baseURL: MOCK_BASE_URL, url: 'example'},
+    {data: {errors}, status, statusText: ''}
+  )
 
-  return serverError
-}
+// Backward-compat alias for plugin tests (removed in Stage 2)
+export const getAxiosError: any = getRequestError
```
GabrielAnca: Can we use something more specific than `any` to ensure we don't break anything with this change? I'm just worried that setting `any` will make almost anything to compile after the change

packages/base/src/helpers/request.ts
```diff
@@ -0,0 +1,189 @@
...
+import type {Readable} from 'stream'
+
+import {EnvHttpProxyAgent, ProxyAgent, fetch} from 'undici'
+
+export interface RequestConfig {
+  baseURL?: string
+  data?: any
+  dispatcher?: any // undici Dispatcher — typed as any to avoid @types/node version conflict
+  headers?: any // permissive to allow AxiosHeaders during Stage 1 migration
+  maxBodyLength?: number // accepted for compat, ignored (no limit with fetch)
```
GabrielAnca: Should we move this to the bottom together with the other compat fields?

packages/plugin-synthetics/src/__tests__/api.test.ts
```diff
@@ -285,7 +305,7 @@ describe('dd-api', () => {
         baseV2Url: `http://127.0.0.1:${port}`,
       })
 
-      await expect(localApi.getSyntheticsOrgSettings).rejects.toThrow('socket hang up')
+      await expect(localApi.getSyntheticsOrgSettings).rejects.toThrow('other side closed')
```
GabrielAnca, line 308: Wondering if the change in error message could be a breaking change. May users rely on this exact error message? This looks like an edge case, but I'm just asking 😄

GabrielAnca (approved):
Looks great! Just some minor comments

rpelliard (approved):

packages/plugin-synthetics/src/__tests__/api.test.ts
ava-silver: This string comes from the underlying transport layer and is never documented or surfaced in our CLI output as a meaningful user-visible message -- it only appears in raw error stacks, which to me feels more like an implementation detail change than a breaking change. There are potentially others that could arise depending on network conditions, and unfortunately there isnt an existing mapping we can use afaik

That being said, we can definitely document it in the release that this comes in, mentioning that users who are relying on axios specific errors or other axios specific implementation details should be aware before updating

packages/base/src/commands/trace/api.ts
```diff
@@ -1,26 +1,20 @@
 import type {Payload} from './interfaces'
-import type {AxiosPromise, AxiosRequestConfig, AxiosResponse} from 'axios'
+import type {RequestBuilder} from '@datadog/datadog-ci-base/helpers/interfaces'
 
 import {getRequestBuilder} from '@datadog/datadog-ci-base/helpers/utils'
 
-// Dependency follows-redirects sets a default maxBodyLength of 10 MB https://github.com/follow-redirects/follow-redirects/blob/b774a77e582b97174813b3eaeb86931becba69db/index.js#L391
-// We don't want any hard limit enforced by the CLI, the backend will enforce a max size by returning 413 errors.
-const maxBodyLength = Infinity
```
Drarig29, line 8: Did you confirm fetch/undici doesn't enforce a max body length either?

packages/base/src/helpers/__tests__/utils.test.ts
```diff
@@ -247,132 +255,38 @@ describe('utils', () => {
...
-          url: 'test-from-proxy',
-        })
-        expect(targetServer.spy.mock.calls.length).toBe(1)
-        expect(proxyServer.spy.mock.calls.length).toBe(1)
-      } finally {
-        await targetServer.close()
-        await proxyServer.close()
-      }
-    })
+    // Proxy routing for plain HTTP is not testable here without TLS certs — undici
```
Drarig29: that comment is weird with the em-dash 😁

packages/base/src/helpers/__tests__/utils.test.ts
```diff
@@ -247,132 +255,38 @@ describe('utils', () => {
...
-        })
-        expect(targetServer.spy.mock.calls.length).toBe(1)
-        expect(proxyServer.spy.mock.calls.length).toBe(1)
-      } finally {
-        await targetServer.close()
-        await proxyServer.close()
-      }
-    })
-
-    test('Proxy configured through env var', async () => {
```
Drarig29, line 332: Do we still test somewhere that the proxy can be configured through env vars?

packages/plugin-synthetics/src/__tests__/api.test.ts
Drarig29: sounds good! we can have a small note about it

Drarig29 (approved):

packages/base/src/helpers/__tests__/utils.test.ts
ava-silver: i'll un-ai-ify it lol

packages/base/src/commands/trace/api.ts
ava-silver: Yep, the docs don't mention anything about it and everything I've read says that undici uses automatic streaming to handle it so theres no need for the maxBodyLength there. if we were setting this setting to a specific limit, we couldn't set that the same way in undici to make it error (we'd have to do a check after the fact). The TLDR is that we're all good here

packages/base/src/helpers/__tests__/utils.test.ts
ava-silver: The old test (`'Proxy configured through env var'`) worked by routing plain HTTP through a local proxy server. With undici, `ProxyAgent` only tunnels HTTPS (via CONNECT), so that approach no longer works.

It is technically testable -- you'd spin up a TCP server, set `HTTPS_PROXY=http://localhost:PORT`, make a request to an `https://` target, and assert the server received a `CONNECT` request. But it felt like a lot of plumbing just to test that undici's `EnvHttpProxyAgent` reads env vars correctly, which is really undici's own test coverage responsibility.

I've updated the existing test to assert the dispatcher is an `EnvHttpProxyAgent` instance (rather than just truthy), which at least verifies we're wiring up the right dispatcher type. Happy to add the full integration test if you think it's worth it, maybe in a follow-up PR if thats okay?

Drarig29: nah, no need it for it. indeed, it's undici's responsibility

Issue #1907, opened 2025-10-09
React Native react-native inject-debug-id bundle name limitation


AZagatti:
### Feature request description

The `inject-debug-id ./dir` command fails on standard iOS builds because it strictly looks for a `*.bundle` file, but the default iOS output is `main.jsbundle`. This requires a manual file renaming step in our CI pipeline.

### Solution

We suggest two possible solutions:

1. Update the file search pattern to be more flexible, like `*bundle`, to match `main.jsbundle`.
2. Allow specifying file paths directly via flags, for example:
```bash
npx @datadog/datadog-ci react-native inject-debug-id \
  --bundle ./path/to/main.jsbundle \
  --sourcemap ./path/to/main.jsbundle.map
```

### Command

react-native

PR #2211, opened 2026-04-02
[RUMS-5750] Fix iOS bundle lookup path for `inject-debug-id` command

marco-saia-datadog:
### What and why?

The `react-native inject-debug-id` command was only scanning for `*.bundle` files, which works for Android bundles but not for iOS. iOS bundles use the `*.jsbundle` extension (e.g. `main.jsbundle` and `main.jsbundle.map`), causing the command to fail with a "bundle not found" error when run on iOS build output.

Additionally, the command only supported scanning a directory, with no way to target specific bundle/sourcemap files directly. This is inconvenient for non-standard project layouts where explicitly specifying paths is preferable.

### How?

- Extended the bundle file filter in `injectDebugIds` to match both `*.bundle` (Android) and `*.jsbundle` (iOS) extensions.
- Added `--bundle` and `--sourcemap` flags to allow targeting specific files directly, bypassing directory scanning. When `--bundle` is provided without `--sourcemap`, the sourcemap path is derived automatically as `<bundle>.map`.
- Extracted shared per-file injection logic into a new `injectDebugIdIntoFiles` method, used by both the directory scan loop and the explicit-path code path.
- Updated error messages and usage description to reflect both naming conventions and the new flags.
- Added test cases for `*.jsbundle` files, `--bundle`/`--sourcemap` flag usage, auto-derived sourcemap path, and validation errors.

### Additional Notes

Solves issue https://github.com/DataDog/datadog-ci/issues/1907

### Review checklist

### DIFF
--- a/packages/base/src/commands/react-native/README.md
+++ b/packages/base/src/commands/react-native/README.md
@@ -210,15 +210,37 @@ The inject-debug-id command adds a debug ID to your JavaScript bundle and its so
 
 A debug ID allows Datadog to reliably match error stack traces from your application to the correct uploaded debug symbols (sourcemaps), ensuring accurate error resolution.
 
-To inject the debug ID, run the following command:
+#### Scan a directory
+
+To scan a directory and inject debug IDs into all bundles found:
 
 ```bash
 datadog-ci react-native inject-debug-id ./dist
 ```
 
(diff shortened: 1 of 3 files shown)


Drarig29 (commented):
Anything to update in the README?

Drarig29 (approved):

Drarig29 (approved):

iadjivon (approved):

PR #2212, opened 2026-04-02
[security] Vendor CLI and plugins as zero-dependency bundles

Drarig29:
### What and why?

Vendor the CLI and all plugins as zero-dependency CJS/ESM bundles before publishing to npm.

This closes a supply chain attack vector: previously, installing `@datadog/datadog-ci` or any plugin would trigger transitive `postinstall` scripts from npm dependencies. With vendored bundles, the versions of all dependencies and transitive dependencies are set in stone and baked into published `@datadog/datadog-ci` and `@datadog/datadog-ci-plugin-*` packages.

### How?

Replaced esbuild with [tsdown](https://tsdown.dev) to produce self-contained CJS bundles. Three new scripts handle the different build targets: `tsdown-cli-npm.mjs` (NPM bundle), `tsdown-cli-sea.mjs` (standalone binary), and `tsdown-plugin.mjs` (individual plugins). Shared utilities live in `tsdown-shared.mjs`. A new `datadog-ci.meta.json` at the repo root drives per-plugin bundling config (e.g. extra bundles for `@datadog/datadog-ci-plugin-lambda/functions/*`).

Plugin loading now has a three-tier resolution:
1. **Bundled injection** — built-in plugins injected at build time (used by both SEA and the npm bundle)
2. **Self-contained bundle** — installed plugins resolved via their package root export (vendored)
3. **Dev/source mode** — submodule path resolution (unchanged, for local development)

Because plugin bundles are now self-contained, `@datadog/datadog-ci-base` no longer needs to be installed alongside a plugin. The plugin installation and npx temporary-install paths are updated accordingly, and CI e2e tests no longer package or resolve the base package separately.

### Validation

```sh
# Clean up
yarn install
yarn clean

# Build the codebase
yarn build

# Create the datadog-ci bundle
yarn workspace @datadog/datadog-ci prepack

# packages/datadog-ci/dist/bundle.d.ts should be a self-contained .d.ts file with zero deps
# packages/datadog-ci/dist/bundle.js should be a self-contained .js file with zero deps

# This should print the version thanks to the shebang
./packages/datadog-ci/dist/bundle.js --version

# This should print that file:///.../packages/plugin-synthetics/dist/commands/run-tests.js was resolved
DEBUG=plugins node ./packages/datadog-ci/dist/bundle.js synthetics run-tests

# Build a plugin bundle
yarn workspace @datadog/datadog-ci-plugin-synthetics prepack

# The same command should now print "Loading self-contained plugin bundle"
DEBUG=plugins node ./packages/datadog-ci/dist/bundle.js synthetics run-tests
```

### Review checklist


scripts/esbuild-shared.mjs
```diff
@@ -0,0 +1,147 @@
...
+
+const REPO_ROOT = path.resolve(import.meta.dirname, '..')
+
+/**
+ * esbuild plugin that appends license attributions for bundled packages that have a LICENSE
+ * file but no inline license comment (which esbuild's `legalComments: 'linked'` picks up
+ * automatically). Runs in the `onEnd` hook so the base .LEGAL.txt already exists.
+ *
+ * @returns {import('esbuild').Plugin}
+ */
```
ava-silver: This JSDoc block looks orphaned — it describes `appendMissingLicensesPlugin` but sits above `resolvePluginCommandsPlugin`. Looks like the function order was swapped and the comment wasn't moved with it.

packages/datadog-ci/package.json
```diff
@@ -16,18 +16,18 @@
   "engines": {
     "node": ">=20"
   },
-  "bin": "dist/cli.js",
-  "main": "dist/index.js",
-  "types": "dist/index.d.ts",
+  "bin": "dist/bundle.mjs",
+  "main": "dist/bundle.mjs",
+  "types": "dist/cli.d.ts",
```
ava-silver: The `dist/cli.d.ts` generated by `tsc` will still contain bare module specifiers referencing `@datadog/datadog-ci-base` (e.g. `export * as gitMetadata from '@datadog/datadog-ci-base/commands/git-metadata/library'`). Since `@datadog/datadog-ci-base` is now a devDependency, it won't be installed for consumers — so TypeScript will fail to resolve these types.

The same issue applies to all plugin packages that ship `dist/**/*.d.ts` — notably `plugin-synthetics`, whose `.d.ts` files reference types like `Metadata`, `ProxyConfiguration`, `SpanTags` from the base package. Also `plugin-junit`, `plugin-coverage`, etc.

The bundled JS works fine at runtime, but any TypeScript consumer importing from these packages as libraries will get:
```
Cannot find module '@datadog/datadog-ci-base/helpers/interfaces'
```


Could we potentially generate a bundled/standalone `.d.ts` (e.g. `dts-bundle-generator`, `@microsoft/api-extractor`)? I'd love to validate that this works properly for consumers of this library - I can help test this with a repo like https://github.com/DataDog/serverless-remote-instrumentation

scripts/esbuild-plugin.mjs
```diff
@@ -0,0 +1,64 @@
...
+let hasIndex
+try {
+  await access(path.join(packageDir, 'src', 'index.ts'))
+  hasIndex = true
+} catch {
+  hasIndex = false
+}
+
+const virtualEntryLines = [
+  ...(hasIndex ? [`Object.assign(exports, require('./index'));`] : []),
```
ava-silver: Nit: `Object.assign(exports, require('./index'))` merges all index exports into the same namespace as the command entries. If a plugin ever exports a name from `index.ts` that matches a command filename, one would silently overwrite the other.

Currently safe — command names are kebab-case (`run-tests`) while index exports are camelCase/PascalCase (`executeTests`). But it might be worth adding a check in `esbuild-plugin.mjs` (or `bin/lint-packages.ts`) that errors if there's a collision, to catch this proactively.

packages/datadog-ci/package.json
ava-silver: I tested this locally against [serverless-remote-instrumentation](https://github.com/DataDog/serverless-remote-instrumentation), which imports from `@datadog/datadog-ci-plugin-lambda/functions/*` subpaths.

Beyond the `.d.ts` issue, there's a **runtime break** here: `plugin-lambda` keeps the `./functions/*` export map pointing to `dist/functions/*.js`, but the `files` array only ships `dist/bundle.js` and `dist/**/*.d.ts`. The individual `.js` files are not included in the package.

After packing (`yarn pack`) and inspecting the tarball:
- `dist/functions/instrument.d.ts` ✅ present
- `dist/functions/instrument.js` ❌ missing

So `require("@datadog/datadog-ci-plugin-lambda/functions/instrument")` fails with `MODULE_NOT_FOUND` at runtime — not just a types issue.

This would break any consumer using the `./functions/*` subpath (serverless-remote-instrumentation uses `instrument`, `uninstrument`, and `commons`).

That being said, only `plugin-lambda` has this issue — it's the only plugin with extra subpath exports (`./functions/*`). All other plugins just expose `.` -> `dist/bundle.js`.

One way to fix this without breaking consumers: produce separate self-contained bundles per function subpath (e.g. `dist/functions/instrument.js`, `dist/functions/commons.js`, etc.) as a second esbuild pass in the plugin build, then add `dist/functions/*.js` to the `files` array. The export map stays the same and consumers don't need to change their imports. (Ideally I'd love if this could be generalized in case other packages end up doing this as well)

README.md
```diff
@@ -18,20 +23,25 @@ npm install --save-dev @datadog/datadog-ci
...
+# npx
+npx @datadog/datadog-ci@v5 [scope]
+
+# yarn dlx
+yarn dlx @datadog/datadog-ci@v5 [scope]
 ```
 
-For more ways to install the CLI, see [this section](#more-ways-to-install-the-cli).
+> [!WARNING]
+> Datadog recommends pinning an exact version of `datadog-ci` to avoid being impacted by breaking changes released in major versions of datadog-ci,
```
rtrieu:
```suggestion
> Datadog recommends pinning an exact version of `datadog-ci` to avoid being impacted by breaking changes introduced in major versions of datadog-ci,
```

README.md
```diff
@@ -49,7 +59,9 @@ Use `datadog-ci plugin install` to install a plugin:
 datadog-ci plugin install <scope>
 ```
 
-By default, running a command that requires a plugin will automatically install the plugin if it is not already installed. You can disable this behavior with `DISABLE_PLUGIN_AUTO_INSTALL=1`.
+For backward compatibility, running a command that requires a plugin automatically installs the plugin if it is not already installed by default.
+
+**Datadog recommends explicitly installing plugins with `datadog-ci plugin install <scope>` and disabling auto-installation with `DISABLE_PLUGIN_AUTO_INSTALL=1`.**
```
rtrieu:
```suggestion
**Datadog recommends explicitly installing plugins with `datadog-ci plugin install <scope>`. To disable auto-installation, set `DISABLE_PLUGIN_AUTO_INSTALL=1`.**
```

rtrieu (approved):
minor suggestions

datadog-ci.meta.json
Drarig29, line 1: I want to use this new file for other things in future PRs, like listing all endpoints used per command, and to contain source of truth for various things like [this](https://github.com/DataDog/datadog-ci/blob/387b539136ad68f5228d7134ffe518b5de386ed3/bin/lint-packages.ts#L9-L58) and [this](https://github.com/DataDog/datadog-ci/blob/387b539136ad68f5228d7134ffe518b5de386ed3/.github/advanced-issue-labeler.yml#L1)

packages/datadog-ci/package.json
Drarig29: @ava-silver the presence of axios is breaking stuff in this PR so I'll wait for your PRs removing axios to be merged 😁

hongkai (approved):

juli1 (approved):

rodrigo-roca (approved):

vishal-joshi-datadog (approved):

KevinYeramian (approved):

ava-silver (approved):

Drarig29:
I fixed the conflicts and tested the tarballs in https://github.com/DataDog/synthetics-ci-github-action to see if the project still builds.

In datadog-ci repo:
```sh
yarn workspace @datadog/datadog-ci pack
yarn workspace @datadog/datadog-ci-plugin-synthetics pack
```

In https://github.com/DataDog/synthetics-ci-github-action repo:
```diff
diff --git a/package.json b/package.json
index 9acace6..98fbbdb 100644
--- a/package.json
+++ b/package.json
@@ -34,8 +34,8 @@
   "packageManager": "yarn@3.4.1",
   "dependencies": {
     "@actions/core": "^1.11.1",
-    "@datadog/datadog-ci": "^5.9.1",
-    "@datadog/datadog-ci-plugin-synthetics": "^5.9.1",
+    "@datadog/datadog-ci": "/path/to/datadog-ci/packages/datadog-ci/package.tgz",
+    "@datadog/datadog-ci-plugin-synthetics": "/path/to/datadog-ci/packages/plugin-synthetics/package.tgz",
     "deep-extend": "0.6.0"
   },
   "devDependencies": {
diff --git a/src/fixtures.ts b/src/fixtures.ts
index d898378..c563f07 100644
--- a/src/fixtures.ts
+++ b/src/fixtures.ts
@@ -13,6 +13,7 @@ export const mockReporter: synthetics.MainReporter = {
   testTrigger: jest.fn(),
   testWait: jest.fn(),
   testsWait: jest.fn(),
+  dryRunEnd: jest.fn(),
 }
 
 export const EMPTY_SUMMARY: synthetics.Summary = {
```

And there was a build issue with `import * as synthetics from '@datadog/datadog-ci-plugin-synthetics'`:
```
src/fixtures.ts:5:28 - error TS2503: Cannot find namespace 'synthetics'.

5 export const mockReporter: synthetics.MainReporter = {
                             ~~~~~~~~~~
```

That's what https://github.com/DataDog/datadog-ci/pull/2212/changes/6a022ace6c00a69466df75a096e25312e25f2dd7 is fixing.

hongkai (approved):

ManuelPalenzuelaDD (approved):

gregoirechauvet (approved):

PR #2213, opened 2026-04-03
[chore] Print datadog-ci and plugin versions

Drarig29:
### What and why?

This PR prints the datadog-ci version (and plugin version if applicable) at the beginning of every command for visibility. This is useful to know in CI logs what version was downloaded if the version is unpinned.

### How?

- Print datadog-ci version
- Print plugin version
- CTA about version skew: if the installed plugin version is different than the installed datadog-ci version

image

Also improves `datadog-ci plugin install --help`:

image

### Review checklist

### DIFF
--- a/packages/base/src/helpers/plugin.ts
+++ b/packages/base/src/helpers/plugin.ts
@@ -380,11 +380,10 @@ export const VERSION_OVERRIDE_REGEX = /^(\d+\.\d+\.\d+|file:\.\/[a-zA-Z0-9.\-/@]
 
 const getPackageToInstall = (scope: string): PackageInfo => {
   const pluginName = scopeToPackageName(scope)
-
-  const pluginVersionOverride = process.env['PLUGIN_AUTO_INSTALL_PLUGIN_VERSION_OVERRIDE']
+  const pluginVersionOverride = process.env['PLUGIN_INSTALL_VERSION_OVERRIDE']
 
   if (pluginVersionOverride && !VERSION_OVERRIDE_REGEX.test(pluginVersionOverride)) {
-    throw new Error(`Invalid PLUGIN_AUTO_INSTALL_PLUGIN_VERSION_OVERRIDE value: ${pluginVersionOverride}`)
+    throw new Error(`Invalid PLUGIN_INSTALL_VERSION_OVERRIDE value: ${pluginVersionOverride}`)
   }
 
   const pluginVersion = pluginVersionOverride ?? cliVersion
--- a/packages/base/src/helpers/plugin.ts
+++ b/packages/base/src/helpers/plugin.ts
@@ -16,7 +16,7 @@ import {cliVersion} from '../version'
 import {isStandaloneBinary} from './is-standalone-binary'
 import {messageBox} from './message-box'
 
-export type PackageInfo = {name: string; version: string; descriptor: string}
+export type PackageInfo = {name: string; version: string}
 export type PluginSubModule = {PluginCommand: CommandClass<CommandContext>}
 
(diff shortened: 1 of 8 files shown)


packages/base/src/helpers/plugin.ts
```diff
@@ -256,6 +259,17 @@ const handlePluginAutoInstall = async (scope: string) => {
   }
 }
 
+const printPluginVersion = (plugin: PackageInfo) => {
+  console.log(chalk.dim(`${plugin.name} v${plugin.version}`))
+  if (plugin.version !== cliVersion) {
+    console.log(
+      chalk.yellow(
+        `The plugin is not the same version as datadog-ci, which could lead to unexpected behavior. Consider syncing the plugin version with datadog-ci.`
```
ava-silver: maybe giving them a CTA (like a command to run) could be good here - would `datadog-ci plugin install` do the trick?

packages/base/src/helpers/plugin.ts
```diff
@@ -349,26 +363,28 @@ const getPackageToInstall = (scope: string) => {
     throw new Error(`Invalid PLUGIN_AUTO_INSTALL_PLUGIN_VERSION_OVERRIDE value: ${pluginVersionOverride}`)
   }
 
-  return `${pluginName}@${pluginVersionOverride ?? cliVersion}`
+  const pluginVersion = pluginVersionOverride ?? cliVersion
+
+  return {
+    name: pluginName,
+    version: pluginVersion,
+    descriptor: `${pluginName}@${pluginVersion}`,
```
ava-silver: nit: i usually prefer having a helper function to compute this rather than storing duplicate data (that could potentially get skewed)

packages/base/src/helpers/plugin.ts
```diff
@@ -383,7 +399,7 @@ const extractPackageJson = (content: unknown): PluginPackageJson => {
 
   const {name, version} = content
 
-  return {name, version}
+  return {name, version, descriptor: `${name}@${version}`}
```
ava-silver: a helper (see previous comment) could also help consolidate logic like this

ava-silver (approved):
a few comments but approving to not block

packages/base/src/helpers/plugin.ts
Drarig29: good idea, note that `datadog-ci@vX.Y.Z plugin install` always installs the version `vX.Y.Z` of the plugin.

so it only supports `datadog-ci → plugin` sync, and not `plugin → datadog-ci` sync

also, I'll squash both on the same line and avoid yellow because it could become annoying

packages/base/src/helpers/plugin.ts
```diff
@@ -364,48 +381,41 @@ const isValidScope = (scope: string): boolean => {
...
+    throw new Error(`Invalid PLUGIN_INSTALL_VERSION_OVERRIDE value: ${pluginVersionOverride}`)
   }
 
-  return `${pluginName}@${pluginVersionOverride ?? cliVersion}`
+  const pluginVersion = pluginVersionOverride ?? cliVersion
+
+  return {
+    name: pluginName,
+    version: pluginVersion,
+    descriptor: packageDescriptor(pluginName, pluginVersion),
```
ava-silver: for my previous comment, I was thinking this would be used at the callsite (maybe with the signature `packageDescriptor({name, version}: PackageInfo): string`) to remove duplicate stored state, and then call it in the places you currently use `.descriptor` -- that being said its not a big deal either way so i'll let you decide what you prefer

ava-silver (approved):

PR #2251, opened 2026-04-09
[security] Set `User-Agent` for requests done by datadog-ci

Drarig29:
### What and why?

This PR replaces the default `unidici` user-agent by a value with [same format as `datadog-api-client-typescript`](https://github.com/DataDog/datadog-api-client-typescript/blob/507c098afb5224efe8bf4cbcd0e3d84fd2bc4525/userAgent.ts#L5):

- Node.js, with plugin: `datadog-ci/1.0.0 (node 18.17.0; os darwin; arch arm64) datadog-ci-plugin-synthetics/1.0.0`
- Node.js, without plugin: `datadog-ci/1.0.0 (node 18.17.0; os darwin; arch arm64)`
- Unknown runtime, with plugin: `datadog-ci/1.0.0 (runtime unknown) datadog-ci-plugin-synthetics/1.0.0`
- Unknown runtime, without plugin: `datadog-ci/1.0.0 (runtime unknown)`

### How?

- Use `AsyncLocalStorage` to store plugin version as async context when a plugin is loaded
- Override user-agent header

### Review checklist

### DIFF
--- a/packages/base/src/helpers/plugin.ts
+++ b/packages/base/src/helpers/plugin.ts
@@ -15,6 +15,7 @@ import {cliVersion} from '../version'
 
 import {isStandaloneBinary} from './is-standalone-binary'
 import {messageBox} from './message-box'
+import {withPluginUserAgent} from './user-agent'
 
 export type PackageInfo = {name: string; version: string}
 export type PluginSubModule = {PluginCommand: CommandClass<CommandContext>}
@@ -30,9 +31,12 @@ export const executePluginCommand = async <T extends Command>(instance: T): Prom
     const submodule = await importPluginSubmodule(scope, command)
     debug(`Done importing plugin command`)
 
+    const plugin = await importPlugin(scope)
+    printPluginVersion(plugin)
+
     const pluginCommand = Object.assign(new submodule.PluginCommand(), instance)
 
-    return pluginCommand.execute()
+    return withPluginUserAgent(plugin.name, plugin.version, () => pluginCommand.execute())
   } catch (error) {
     debug('Error in executePluginCommand:', error)
 
@@ -92,7 +96,10 @@ export const checkPlugin = async (scope: string, command?: string): Promise<bool
   }
 
   try {
-    const module = command ? await importPluginSubmodule(scope, command) : await importPlugin(scope)
+    const plugin = await importPlugin(scope)
+    printPluginVersion(plugin)
+
+    const module = command ? await importPluginSubmodule(scope, command) : plugin
 
     console.log(
       [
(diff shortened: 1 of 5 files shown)


packages/base/src/helpers/plugin.ts
```diff
@@ -30,9 +31,12 @@ export const executePluginCommand = async <T extends Command>(instance: T): Prom
     const submodule = await importPluginSubmodule(scope, command)
     debug(`Done importing plugin command`)
 
+    const plugin = await importPlugin(scope)
+    printPluginVersion(plugin)
```
Drarig29, line 35: Moved `printPluginVersion()` outside of `importPlugin()` because `importPluginSubmodule()` already calls it under-the-hood, so it would result in a double-write of the plugin version

packages/base/src/helpers/request.ts
```diff
@@ -132,10 +134,11 @@ export const httpRequest = async <T = any>(config: RequestConfig): Promise<Reque
   const resolvedUrl = resolveUrl(config)
   const method = (config.method ?? 'GET').toUpperCase()
   const {body, headers} = serializeBody(config.data, config.headers ?? {})
+  const headersWithUserAgent = {...headers, 'User-Agent': getUserAgent()}
```
ava-silver, line 137: Worth considering whether callers should be able to override `User-Agent`. Right now the spread order means `getUserAgent()` always wins. If that's intentional its fine but wanted to mention.

packages/base/src/helpers/user-agent.ts
```diff
@@ -0,0 +1,41 @@
...
+const formatPluginName = (pluginName: string) => pluginName.replace(/^@datadog\//, '')
+const formatPluginUserAgent = (pluginName: string, pluginVersion: string) =>
+  `${formatPluginName(pluginName)}/${pluginVersion}`
+
+// The user agent has the same format as the one used in the Datadog API client for TypeScript.
+// https://github.com/DataDog/datadog-api-client-typescript/blob/507c098afb5224efe8bf4cbcd0e3d84fd2bc4525/userAgent.ts#L5
+
+let baseUserAgent: string
+if (typeof process !== 'undefined' && process.release && process.release.name === 'node') {
+  baseUserAgent = `datadog-ci/${cliVersion} (node ${process.versions.node}; os ${process.platform}; arch ${process.arch})`
```
ava-silver, line 16: **user-agent.ts:16** — Minor: `process.versions.node` returns e.g. `20.19.0` (no `v` prefix), so the actual output is `node 20.19.0` rather than `node v18.17.0` as shown in the PR description. Might want to either add the `v` or update the description.

packages/base/src/helpers/user-agent.ts
```diff
@@ -0,0 +1,41 @@
+import {AsyncLocalStorage} from 'node:async_hooks'
+
+import {cliVersion} from '../version'
+
+const pluginUserAgentStorage = new AsyncLocalStorage<string | undefined>()
```
ava-silver: nit: `AsyncLocalStorage<string | undefined>` -- the `| undefined` is redundant since `getStore()` already returns `T | undefined`. `AsyncLocalStorage<string>` would be slightly cleaner.

ava-silver (approved):
a couple small things but overall lgtm!

packages/base/src/helpers/request.ts
Drarig29: That's intentional, and we don't override it anywhere currently 👍

packages/base/src/helpers/user-agent.ts
Drarig29: updated the PR description to align with datadog-api-client-typescript

GabrielAnca (approved):

PR #2255, opened 2026-04-10
[serverless] Migrate to `@inquirer/prompts` and remove `inquirer-checkbox-plus-prompt`

Drarig29:
### What and why?

This PR migrates our interactive prompt handling from the legacy `inquirer` package to `@inquirer/prompts` and removes `inquirer-checkbox-plus-prompt`. The rewritten prompt stack is ESM-only, so this also adds a small runtime loader that keeps prompt usage compatible with the repo's current CommonJS build. As a side effect, removing the old prompt packages drops one direct `lodash` path from the runtime dependency graph.

### How?

- Replace legacy question-object prompt calls in `packages/base`, `packages/plugin-lambda`, and `packages/plugin-cloud-run` with `@inquirer/prompts` helpers.
- Reimplement Lambda function selection with a local filter + checkbox flow instead of `inquirer-checkbox-plus-prompt`.
- Add a shared prompt loader for the ESM package, and teach Knip to ignore `@inquirer/prompts`, `glob`, and `tsdown` because they are used through dynamic/runtime tooling paths that Knip does not statically resolve.
- Update prompt tests, dependency metadata, `LICENSE-3rdparty.csv`, and `yarn.lock` to match the migration.
- Validate locally with `yarn lint:packages`, `yarn lint:all`, `yarn tsc:build`, `JEST_USE_WATCHMAN=0 yarn test --watchman=false packages/base/src/helpers/__tests__/prompt.test.ts packages/plugin-lambda/src/__tests__/prompt.test.ts`, and `node bin/check-licenses.js`.

### Review checklist

### DIFF
--- a/packages/base/src/helpers/inquirer.ts
+++ b/packages/base/src/helpers/inquirer.ts
@@ -51,10 +51,60 @@ export type InquirerPrompts = {
   select: <Value = string>(config: SelectConfig<Value>) => Promise<Value>
 }
 
+type InquirerPromptStatus = 'idle' | 'loading' | 'done'
+
+type KeypressEvent = {
+  ctrl: boolean
+  name: string
+  shift: boolean
+}
+
+type Readline = {
+  clearLine: (dir: number) => void
+  line: string
+  write: (input: string) => void
+}
+
+type InquirerTheme = {
+  style: {
+    answer: (text: string) => string
+    error: (text: string) => string
+    highlight: (text: string) => string
+    message: (text: string, status: InquirerPromptStatus) => string
+  }
+}
+
+type PaginationConfig<Item> = {
+  active: number
+  items: readonly Item[]
+  loop?: boolean
+  pageSize: number
+  renderItem: (options: {index: number; isActive: boolean; item: Item}) => string
+}
+
+type PromptView<Value, Config> = (config: Config, done: (value: Value) => void) => string | [string, string | undefined]
+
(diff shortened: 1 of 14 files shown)


packages/base/src/helpers/inquirer.ts
ava-silver, line 1: any way we can avoid handwriting the types here? this feels like itll quickly become a problem to update in the future

packages/base/src/helpers/inquirer.ts
```diff
@@ -0,0 +1,110 @@
...
+
+// eslint-disable-next-line @typescript-eslint/no-implied-eval -- TypeScript rewrites plain `import()` to `require()` in our CommonJS emit.
+const importInquirerModule = new Function('specifier', 'return import(specifier)') as (
+  specifier: string
+) => Promise<unknown>
+
+// Preserve a real runtime dynamic import so Node can load the ESM-only prompt package from CommonJS output.
+export const loadPrompts = () => importInquirerModule('@inquirer/prompts') as Promise<InquirerPrompts>
+
+export const loadCore = () => importInquirerModule('@inquirer/core') as Promise<InquirerCore>
```
ava-silver, line 110: is this needed because we emit commonjs and the new inquirer package is esm only? if so, I'd love to discuss with the guild what it would take to migrate our package from commonjs to esm since it seems like that's the future of the ecosystem

PR #2256, opened 2026-04-10
[chore] Upgrade `datadog-metrics` with patch to use `undici`

Drarig29:
### What and why?

Upgrades `datadog-metrics` from `0.9.3` to `0.13.0-pre.1` and patches it to use `undici` instead of `cross-fetch` as its HTTP transport. This lets metrics honor proxy settings (via undici's dispatcher) without relying on the removed `agent` option or `proxy-agent`.

### How?

- upgrade `datadog-metrics` to `0.13.0-pre.1` and remove the legacy `@types/datadog-metrics` dependency
- apply a yarn patch to swap `cross-fetch` → `undici` in `datadog-metrics`'s reporter
- remove the `ProxyAgent` instantiation and `agent` option from `metrics.ts`
- update the `flush` call to use `async/await` to match the new promise-based API
- update the Jest mock to match the new async reporter and class-based logger

### Review checklist

### DIFF
--- a/.yarn/patches/datadog-metrics-npm-0.13.0-pre.1-d40793195b.patch
+++ b/.yarn/patches/datadog-metrics-npm-0.13.0-pre.1-d40793195b.patch
@@ -1,0 +1,41 @@
+diff --git a/lib/reporters.js b/lib/reporters.js
+index ebde40f910026af9503179a89ca29057310d88f7..0f093f78651d86532945915cfe799c1dc1c730f6 100644
+--- a/lib/reporters.js
++++ b/lib/reporters.js
+@@ -1,5 +1,5 @@
+ 'use strict';
+-const { fetch } = require('cross-fetch');
++const { fetch, EnvHttpProxyAgent } = require('undici');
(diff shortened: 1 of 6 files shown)


standalone-e2e/standalone-binary.test.ts
```diff
@@ -2,20 +2,29 @@ import child_process from 'node:child_process'
...
 import {version} from '../packages/datadog-ci/package.json'
 
+const sanitizeStderr = (stderr: string): string => {
+  return stderr
+    .replace(
+      /\(node:\d+\) \[DEP0040\] DeprecationWarning: The `punycode` module is deprecated\. Please use a userland alternative instead\.\n\(Use `.+ --trace-deprecation \.\.\.` to show where the warning was created\)\n?/g,
+      ''
+    )
+    .trim()
+}
```
Drarig29: Related to https://github.com/DataDog/datadog-api-client-typescript/issues/3605

ava-silver (approved):
Is there a plan to support this without the patch longer term?

PR #2269, opened 2026-04-14
fix(ci): use `jobId` to disambiguate Worker logs on multi-runners sharing same `planId`

### LINKED ISSUE #2271: bug(ci): wrong job name extracted on multi-runners when sequential jobs from the same workflow run share the same planId
## Summary

On non-ephemeral (multi) runners, `datadog-ci tag --level job` and `datadog-ci measure --level job` extract the wrong job display name from the GitHub Actions diagnostic logs when the runner has already processed a previous job from the **same workflow run**. All runner tags and measures sent for the current job are effectively discarded — the current job's span receives no custom metadata.

## Root cause

`getTargetWorkerLogFiles()` (introduced in #2089) identifies which `Worker_*.log` file belongs to the current job by scanning for the `planId` from `ACTIONS_ORCHESTRATION_ID`. However, all jobs in the same GitHub Actions workflow run share the same `planId`. On a multi-runner that has already processed an earlier job from the same run, multiple Worker log files will match the planId check. Since log files are iterated in alphabetical order (which is chronological), the **first** (oldest) matching file is always returned — which belongs to the earlier job.

```
# ACTIONS_ORCHESTRATION_ID = <planId>.<jobId>.__default
# All jobs in the same workflow run have the same planId
#
# Runner state after processing job-a, while running job-b:
#   _diag/Worker_20251014-083000-utc.log  ← job-a — contains "planId": "9f25551c-..."
#   _diag/Worker_20251014-090000-utc.log  ← job-b — contains "planId": "9f25551c-..."
#
# getTargetWorkerLogFiles() iterates in order and returns job-a's log
```

The wrong `jobDisplayName` is then written into `ciEnv['DD_GITHUB_JOB_NAME']`. The Datadog backend cannot match this to the current job's span, so all tags and measures are lost.

## Reproduction

1. Set up a **non-ephemeral** (multi) self-hosted GitHub Actions runner
2. Create a workflow with two sequential jobs targeting the same runner:
   ```yaml
   jobs:
     job-a:
       runs-on: [self-hosted]
       steps:
         - run: sleep 5
     job-b:
       needs: job-a
       runs-on: [self-hosted]
       steps:
         - run: datadog-ci tag --level job --tags foo:bar
   ```
3. Observe the output of `datadog-ci tag` in job-b:
   ```
   Determining GitHub job name
   Found Worker log for planId 9f25551c-...: Worker_20251014-083000-utc.log  ← job-a's log
   Successfully extracted job name: job-a                                     ← WRONG
   Tags sent
   ```
4. In Datadog CI Visibility, job-b's span has no custom tags; job-a's span is unaffected.

The bug is **intermittent** in practice: it only triggers when both jobs run on the same runner instance within the same workflow run. If the runner picks up job-b as its first job, or from a different workflow run (different planId), the planId match is unambiguous.

## Expected behavior

`datadog-ci tag --level job` should extract `job-b` and apply tags to job-b's span.

## Affected versions

Any version that includes the `ACTIONS_ORCHESTRATION_ID`-based filtering introduced in #2089, running against GitHub Actions runner v2.331.0+ (which is when `ACTIONS_ORCHESTRATION_ID` became available).

## Fix

`ACTIONS_ORCHESTRATION_ID` has the format `<planId>.<jobId>.__default`. The `jobId` (second segment) is serialized as `"jobId"` in the Worker log's job message JSON (`AgentJobRequestMessage.JobId` in the runner source) and is **unique per job** within a workflow run.

`getTargetWorkerLogFiles()` should require both `planId` and `jobId` to match. The `jobId` check should be skipped gracefully when absent (older runner versions), falling back to planId-only matching.

A fix is proposed in #2269.

LudovicTOURMAN:
Fixes #2271.

## Changes

- `packages/base/src/helpers/ci.ts` — `getTargetWorkerLogFiles()` now requires both `planId` **and** `jobId` (second segment of `ACTIONS_ORCHESTRATION_ID`) to match, since `jobId` is unique per job within a workflow run. The `jobId` check is skipped gracefully when absent, preserving backward compatibility with older runner versions.
- `packages/base/src/helpers/__tests__/ci.test.ts` — update existing tests to use matching jobIds in `ACTIONS_ORCHESTRATION_ID`; add regression test for the same-planId multi-runner scenario.

## Testing

```sh
yarn test packages/base/src/helpers/__tests__/ci.test.ts
# 1026 tests pass, including the new regression test:
# ✓ should disambiguate logs sharing the same planId using jobId on multi-runners
```

Follows up on #2089 (original fix for Worker log detection on non-ephemeral runners).

### DIFF
--- a/packages/base/src/helpers/__tests__/ci.test.ts
+++ b/packages/base/src/helpers/__tests__/ci.test.ts
@@ -572,6 +572,19 @@ describe('getGithubJobDisplayNameFromLogs', () => {
       "jobName": "__default"
     }`
 
+  // Variant with explicit jobId for testing same-planId disambiguation on multi-runners
+  const sampleLogContentWithPlanIdAndJobId = (jobDisplayName: string, planId: string, jobId: string): string => `
+    [2025-09-15 10:14:00Z INFO Worker] Waiting to receive the job message from the channel.
+    [2025-09-15 10:14:00Z INFO ProcessChannel] Receiving message of length 22985, with hash 'abcdef'
+    [2025-09-15 10:14:00Z INFO Worker] Message received.
+    [2025-09-15 10:14:00Z INFO Worker] Job message:
+    {
+      "planId": "${planId}",
+      "jobId": "${jobId}",
+      "jobDisplayName": ${JSON.stringify(jobDisplayName)},
+      "jobName": "__default"
+    }`
+
   const sampleLogFileName = 'Worker_20251014-083000.log'
   const sampleJobDisplayName = 'build-and-test'
 
@@ -1025,8 +1038,10 @@ describe('getGithubJobDisplayNameFromLogs', () => {
   describe('Worker log identification with ACTIONS_ORCHESTRATION_ID', () => {
     test('should use ACTIONS_ORCHESTRATION_ID to find correct Worker log', () => {
       const planId = '9f25551c-ce16-4f8f-a662-8575df3d1354'
+      // jobId must match what sampleLogContentWithPlanId embeds in the log
+      const jobId = '95a4619c-e316-542f-8a21-74cd5a8ac9ca'
       const originalEnv = process.env
-      process.env = {...originalEnv, ACTIONS_ORCHESTRATION_ID: `${planId}.test-job.__default`}
+      process.env = {...originalEnv, ACTIONS_ORCHESTRATION_ID: `${planId}.${jobId}.__default`}
 
       const targetDir = githubWellKnownDiagnosticDirsUnix[0]
       const logContent = sampleLogContentWithPlanId('correct-job-name', planId)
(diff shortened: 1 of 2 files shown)


Drarig29 (approved):

LudovicTOURMAN:
Thanks for the review @Drarig29.
Notice that as en external contributor, I'm not able to merge this PR.

Drarig29:
@LudovicTOURMAN I need to ask the relevant teams for review, then we'll merge the PR! Thanks for the 2nd contribution 🥂

rodrigo-roca (approved):
Good catch!

LudovicTOURMAN:
Thanks for being such reactive guys! :heart:

PR #2270, opened 2026-04-14
Parse datadog:opaque property from CycloneDX SBOM

anderruiz:
## 🚀 Motivation 

\`datadog-sbom-generator\` marks certain JAR packages as **opaque** — JARs whose contents cannot be fully analyzed (e.g., nested JARs in fat JARs) — using a custom \`datadog:opaque\` CycloneDX property. This information was being lost during ingestion in \`datadog-ci\`, preventing the SCA backend from applying the appropriate analysis strategy for those packages.

This PR propagates the \`datadog:opaque\` flag from the CycloneDX SBOM through \`datadog-ci\` all the way to the SCA API payload.

## Documentation
**Document** | **Link or Detail**
-------------|------------------
RFC | N/A
Incident | N/A
Jira Ticket | N/A

## 📝 Summary

- Added \`OPAQUE_KEY = 'datadog:opaque'\` constant in \`constants.ts\`
- Added \`opaque?: boolean\` field to the \`Dependency\` interface in \`types.ts\`
- Extended \`extractingDependency()\` in \`payload.ts\` to parse the \`datadog:opaque\` property from CycloneDX SBOM components and map it to the \`opaque\` field (\`\"true\"\` → \`true\`, anything else → \`undefined\`)
- Added fixture \`sbom-with-opaque.json\` with one opaque and one non-opaque Maven component
- Added unit tests covering correct propagation of the \`opaque\` flag and absence of the field for non-opaque dependencies

## 🧪 Testing
- [X] New tests were added for new logic.
- [X] Existing tests were updated for new logic, and not only so that they pass!
- [ ] Benchmark results prove that performance is the same or better.

## 🚧 Staging validation
- [ ] Deployed and monitored using Datadog dashboards.
- [ ] Proof that it works as expected, including profiling or UX screenshots.

## 🆘 Recovery
Notes for on-call - **select only one**:
- [X] The change can be rolled back.
- [ ] Do not roll back. Why?:

### DIFF
--- a/packages/plugin-sbom/src/__tests__/fixtures/sbom-with-opaque.json
+++ b/packages/plugin-sbom/src/__tests__/fixtures/sbom-with-opaque.json
@@ -1,0 +1,39 @@
+{
+    "$schema": "http://cyclonedx.org/schema/bom-1.5.schema.json",
+    "bomFormat": "CycloneDX",
+    "specVersion": "1.5",
+    "version": 1,
+    "components": [
+        {
+            "bom-ref": "pkg:maven/org.apache.logging.log4j/log4j-core@2.14.1",
+            "type": "library",
+            "name": "log4j-core",
+            "version": "2.14.1",
+            "purl": "pkg:maven/org.apache.logging.log4j/log4j-core@2.14.1",
+            "properties": [
+                {
+                    "name": "datadog:opaque",
+                    "value": "true"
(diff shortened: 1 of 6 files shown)


piloulacdog (approved):
LGTM

Drarig29 (approved):

PR #2273, opened 2026-04-15
fix(request): Fix JSON response parsing for v2 endpoints

Drarig29:
### What and why?

The axios-to-undici migration (#2201) introduced a regression: `httpRequest` only auto-parsed JSON responses when the `content-type` header was exactly `application/json`. Some Datadog API endpoints (e.g. `/api/v2/git/repository/search_commits`) return `application/vnd.api+json`, which was not handled — leaving `response.data` as a raw string instead of a parsed object.

This caused `git-metadata upload` to fail with `Invalid API response: [object Object]` during GitDB sync.

### How?

- Extract the media type from the content-type header (stripping parameters like `; charset=utf-8`) and compare with strict equality
- Add `application/vnd.api+json` as an accepted JSON content-type alongside `application/json`
- Added unit tests covering both content-types with and without charset parameters, plus a negative case for `text/plain`

### Review checklist

### DIFF
--- a/e2e/git-metadata.test.ts
+++ b/e2e/git-metadata.test.ts
@@ -1,0 +1,15 @@
+import {DATADOG_CI_COMMAND, execPromise} from './helpers/exec'
+
+describe('git-metadata', () => {
+  it('uploads without errors', async () => {
+    const result = await execPromise(`${DATADOG_CI_COMMAND} git-metadata upload`, {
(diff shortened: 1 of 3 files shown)


Drarig29:
Added a non-regression test which fails without our fix:

image

ava-silver (approved):

PR #2276, opened 2026-04-16
fix(ci): use `system.orchestrationId` to disambiguate Worker logs on multi-runners

LudovicTOURMAN:
### What and why?

On non-ephemeral (multi) runners, sequential jobs from the same workflow run accumulate Worker diagnostic log files that all share the same `planId`. The previous implementation matched only on `planId` and iterated alphabetically (oldest-first), so it returned the wrong log when a previous job in the same run had already run on the same runner. This caused `getGithubJobNameFromLogs()` to return an incorrect display name, which in turn caused `datadog-ci tag --level job` to tag the wrong CI span.

### How?

Runner v2.331.0 (January 2026, controlled by feature flag `actions_set_orchestration_id_env_for_actions`) introduced `ACTIONS_ORCHESTRATION_ID`, sourced directly from the `system.orchestrationId` variable in the job request message. The runner unconditionally serializes the full job message JSON — including all variables — into the Worker diagnostic log on startup (`Runner.Worker/Worker.cs`), so `system.orchestrationId` is always present in the log when the env var is available.

The orchestration ID is unique per job within a workflow run:
- Non-matrix job: `<planId>.<yaml-job-key>.__default`
- Single-dimension matrix: `<planId>.<yaml-job-key>.<matrix-value>`
- Multi-dimension matrix: `<planId>.<yaml-job-key>.<val1>.<val2>[…]`

The new `getTargetWorkerLogFiles` implementation:
1. **Primary path** (runner ≥ v2.331.0, flag enabled): scans Worker logs newest-first, returns the single log whose `system.orchestrationId` value matches `ACTIONS_ORCHESTRATION_ID`. This uniquely identifies the current job's log regardless of how many previous logs exist.
2. **Fallback** (older runner or flag disabled): returns all logs sorted newest-first. Because a non-ephemeral runner processes jobs strictly sequentially, the current Worker process always produces the most recently timestamped log, so newest-first reliably identifies the correct log in all practical cases.

### Review checklist

### DIFF
--- a/packages/base/src/helpers/__tests__/ci.test.ts
+++ b/packages/base/src/helpers/__tests__/ci.test.ts
@@ -560,29 +560,23 @@ describe('getGithubJobDisplayNameFromLogs', () => {
       "jobName": "__default"
     }`
 
-  const sampleLogContentWithPlanId = (jobDisplayName: string, planId: string): string => `
+  // Variant with system.orchestrationId variable — reflects real runner log format.
+  // ACTIONS_ORCHESTRATION_ID is sourced from this variable, so they always share the same value.
+  const sampleLogContentWithOrchestrationId = (jobDisplayName: string, orchestrationId: string): string => `
     [2025-09-15 10:14:00Z INFO Worker] Waiting to receive the job message from the channel.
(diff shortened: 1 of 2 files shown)


Drarig29 (approved):

afontan (approved):
LGTM

PR #2300, opened 2026-05-04
[chore] Replace `@yao-pkg/pkg` with `node --build-sea`

Drarig29:
### What and why?

The `datadog-ci` standalone binaries have been segfaulting on AMD64 systems (exit 139, SIGSEGV). Investigation showed that `@yao-pkg/pkg` + postject produced ELF binaries with a non-standard program header offset (~38MB instead of 64 bytes), which caused QEMU to oscillate between x32 and 64-bit execution mode and crash. This was introduced when we migrated to Node.js SEA in v3.21.1.

### How?

Replace `@yao-pkg/pkg` (and both yarn patches) with a minimal `scripts/build-sea.mjs` that uses Node.js 25's built-in `--build-sea` command:

1. Downloads the stripped Node.js 26.1.0 binary from [`Drarig29/pkg-fetch`](https://github.com/Drarig29/pkg-fetch/releases/tag/v1.0). See [pkg-fetch#5](https://github.com/Drarig29/pkg-fetch/pull/5).
2. Verifies the SHA256 checksum from the accompanying `.sha256sum` file
3. Writes a `sea-config.json` pointing to the downloaded binary as `executable`
4. Runs `node --build-sea sea-config.json` — no external injection tools needed
5. Signs the output on macOS with `codesign`

### Review checklist


duncanista (approved):

juan-fernandez (approved):
thanks for this!

.github/workflows/ci.yml
```diff
@@ -510,7 +510,7 @@ jobs:
       - run: yarn install --immutable
       - run: yarn build
       - name: Create standalone binary
-        run: yarn dist-standalone -t node22-linux-x64 -o datadog-ci_linux-x64
+        run: yarn dist-standalone -t node26-linux-x64 -o datadog-ci_linux-x64
```
ava-silver, line 513: are there any concerns about using node 26 for this given it's not LTS until october? I'm not sure if theres another way but wanted to raise it regardless

also I guess we have the standalone binary tests which should help catch anything major before release

scripts/node-download.mjs
```diff
@@ -0,0 +1,107 @@
+import {execFileSync} from 'node:child_process'
+import {chmodSync, createWriteStream} from 'node:fs'
+import path from 'node:path'
+import {pipeline} from 'node:stream/promises'
+
+export const NODE_VERSION = 'v26.1.0'
+export const NODE_DOWNLOAD_BASE_URL = 'https://nodejs.org/dist'
+export const PKG_FETCH_RELEASE = 'https://github.com/Drarig29/pkg-fetch/releases/download/v1.0'
```
ava-silver, line 8: I assume this is a temporary workaround? If we plan on keeping this around, at some point do we want to migrate the patch to be under the official `DataDog` Github org?

Drarig29:
Closed in favor of https://github.com/DataDog/datadog-ci/pull/2317/s

Issue #2302, opened 2026-05-04
Security vulnerability in pinned dependency fast-xml-parser


trygve-aaberge-adsk:
### Bug description

The latest version of `@datadog/datadog-ci-base` depends on the specific version `fast-xml-parser@5.5.9`. This version of the package has reported security issues: https://github.com/advisories/GHSA-gh4j-gqv2-49f6

Could you upgrade the dependency?

It looks like this has been an issue several times before (#914, #1414, #2077). Is there a reason it needs to be pinned to a specific version and not use a `^`?

### Command

None

ava-silver:
Previously we had pinned specific versions due to supply chain attacks like the recent Axios incident: https://github.com/DataDog/datadog-ci/pull/2195

That being said now that we're fully vendoring dependencies (also as a follow-up item), I could see the possibility of loosening the version constraints. I'll discuss with the team and get back to you!

Drarig29:
Hi @trygve-aaberge-adsk! Which version of `@datadog/datadog-ci-base` are you using? And is it a direct or transitive dependency for you?

trygve-aaberge-adsk:
> Previously we had pinned specific versions due to supply chain attacks like the recent Axios incident: [#2195](https://github.com/DataDog/datadog-ci/pull/2195)

Hm, okay. I'm not sure if I agree that it makes sense for all packages to pin their dependencies to specific versions though. When using a range for transitive dependencies, the consumers of the package can apply whatever restrictions they want (e.g. pnpms minimumReleaseAge). If you pin dependencies, you should at least be quick to handle vulnerability alerts for those packages. This one has been open for two weeks already.

> That being said now that we're fully vendoring dependencies (also as a follow-up item), I could see the possibility of loosening the version constraints. I'll discuss with the team and get back to you!

This may be a discussion for another place, but are you saying you're going to bundle all the dependencies code instead of having regular npm dependencies? I'm not sure what you mean by loosening the constraints if you're vendoring the dependencies.

> Hi [@trygve-aaberge-adsk](https://github.com/trygve-aaberge-adsk)! Which version of `@datadog/datadog-ci-base` are you using? And is it a direct or transitive dependency for you?

I'm using the latest version, `5.16.0`. It's a transitive dependency through `serverless-plugin-datadog`, which is the direct dependency we have.

ava-silver:
[...]

PR #2305, opened 2026-05-04
chore: loosen base package dependency version constraints

ava-silver:
### What and why?

Loosens version constraints in `@datadog/datadog-ci-base` from exact pinned versions to `^` (semver-compatible) ranges. This comes up in context of #2302 -- when users depend on the base package as a library, exact pins mean they get stuck on vulnerable patch versions until we release an update. With `^` ranges, their lockfiles can resolve newer compatible versions independently.

The `@datadog/datadog-ci` main package and plugin packages continue to use exact pinning (they're vendored), so this only affects library consumers of the base package.

### How?

**Core change:**
- Added a syncpack semver group for `@datadog/datadog-ci-base` that uses `^` ranges, placed before the catch-all exact-pin rule in `.syncpackrc.json`
- Changed all `dependencies` and `devDependencies` in `packages/base/package.json` from `x.y.z` to `^x.y.z`

**TypeScript fixes (required by the `^` ranges):**

The loosened ranges caused yarn to resolve newer minor versions of `@types/async-retry` (1.4.8 → 1.4.9) and `@types/deep-extend` (0.4.31 → 0.4.32), which ship stricter type signatures:

- `@types/async-retry@1.4.9`: `onRetry` callback parameter changed from `(error: Error)` to `(e: unknown)`. In `upload.ts`, the public `UploadOptions.onRetry` type is kept as `Error` (no API break) and an internal cast handles the mismatch. In the 9 command/plugin files that call `retryRequest` directly with inline `onRetry` callbacks, `e.message` accesses are updated to use `instanceof Error` guards.
- `@types/deep-extend@0.4.32`: generics now require `extends object`. Added the constraint to `resolveConfigFromFile<T>` and handled the `undefined` case from `getConfig`.

### Review checklist


ava-silver:
* **#2305** <a href="https://app.graphite.com/github/pr/DataDog/datadog-ci/2305?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/datadog-ci/2305?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2306** <a href="https://app.graphite.com/github/pr/DataDog/datadog-ci/2306?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

packages/base/src/helpers/upload.ts
```diff
@@ -73,7 +73,7 @@ export const upload =
     opts.onUpload()
     try {
       await retryRequest(() => uploadMultipart(requestBuilder, payload, opts.useGzip ?? false), {
-        onRetry: opts.onRetry,
+        onRetry: opts.onRetry as ((e: unknown, attempt: number) => any) | undefined,
```
Drarig29: **nit** Can we change the definition of `UploadOptions.onRetry` instead?

Drarig29 (approved):

PR #2321, opened 2026-05-18
[dev] Add `AGENTS.md` and change default JSON formatter

Drarig29:
### What and why?

Claude uses a non-standard `CLAUDE.md` file, while the standard is `AGENTS.md`. And it's a [common practice](https://www.reddit.com/r/ClaudeCode/comments/1r9zx34/pointing_claudemd_to_agentsmd/) to refer to `AGENTS.md` with `@AGENTS.md` in the Claude non-standard file.

### How?

- Point `CLAUDE.md` to `AGENTS.md`
- Change default JSON formatter because agents keep formatting with the wrong formatter in my IDE

### Review checklist


CLAUDE.md
ava-silver, line 1: what if we instead just make `CLAUDE.md` a symlink?

ava-silver (approved):
can we make AGENTS.md/CLAUDE.md (or any other agent markdown) only owned by the admins rather than documentation as well? since this isn't really public documentation

CLAUDE.md
Drarig29: `@` instructs Claude Code to inline the file as a memory into the context, so it's not a tool call. And that's a pattern we use in multiple internal Datadog repos.

PR #2322, opened 2026-05-19
[chore] Add Alpine x64 and arm64 standalone binaries

### LINKED ISSUE #1474: musl based binary for alpine ?

We use datadog in our CI and having NodeJS is not ideal for us. 

Could you add a musl based binary in the release page that can run in alpine ?

### Solution

Most of the times, it's just a matter of compiling the sw without modifications, but it depends on the sw

### Additional context

_No response_

### Command

None

Drarig29:
### What and why?

Closes #1474

This PR adds support for Alpine standalone binaries. However, `pkg --sea` doesn't support it and fails with `Unsupported OS: alpine`, so we have to use the ["Standard mode"](https://yao-pkg.github.io/pkg/guide/sea-vs-standard) of `pkg` which downloads binaries from https://github.com/yao-pkg/pkg-fetch.

### How?

I tried patching `@yao-pkg/pkg` again to use `process.execPath` as the base Node binary but it resulted in a segfault. So I chose to use "Standard mode" for Alpine only.

### Review checklist


juan-fernandez (approved):

.github/workflows/ci.yml
ava-silver, line 1: we probably should add the alpine checksums to `upload-checksums` -- also do we want to update the RELEASING.md to change the number of expected binaries?

package.json
```diff
@@ -10,6 +10,7 @@
     "compare-binary-size": "FORCE_COLOR=1 node bin/compare-binary-size.js",
     "dev": "yarn tsc:build --watch",
     "dist-standalone": "FORCE_COLOR=1 yarn bundle:sea && pkg --sea packages/datadog-ci/dist/bundle.js",
+    "dist-standalone:legacy": "FORCE_COLOR=1 yarn bundle:sea && pkg packages/datadog-ci/dist/bundle.js",
```
ava-silver: 💭 `legacy` sounds like it's deprecated, but it is a permanent solution -- what are your thoughts on something like `dist-standalone:no-sea` or `dist-standalone:standard` ([based on the docs](https://yao-pkg.github.io/pkg/guide/sea-vs-standard))

joepeeples (approved):

ava-silver (approved):

PR #2328, opened 2026-05-20
deployment correlate-image: include ci_env in request payload

DaniFdz:
## What this changes

Two related fixes to the `deployment correlate` and `deployment correlate-image` commands:

1. `correlate-image` now sends a `ci_env` attribute on its request payload, mirroring the sibling `correlate` (GitOps) command. The backend uses it to derive `pipeline_correlation_id` and `job_correlation_id`.
2. Both commands now read the GitHub Actions numeric check run ID from the runner diagnostic log files (via `getGithubJobIDFromLogs`) before building the CI span tags, so `ci.job.id` carries the numeric ID instead of the workflow job name.

## Why

The backend (`/api/v2/ci/deployments/correlate-image` in dd-source) needs `ci_env` to derive CI correlation IDs. Same wire shape as `correlate` so we are not inventing a new format.

For GitHub Actions, the numeric check run ID is not in any standard env var. It only lives in the runner diagnostic log files. The CI visibility commands (`junit upload`, `trace`) already use this approach. Without it, `ci.job.id` ends up being the workflow job name (for example `"build"`), the backend cannot parse that as an int64, and `job_correlation_id` stays empty on every GitHub deployment.

## Example payload

```json
{
  "data": {
    "type": "ci_deployment_correlate_image",
    "attributes": {
      "commit_sha": "abcdef",
      "repository_url": "https://github.com/DataDog/example",
      "image": "my-image:latest",
      "ci_env": {
        "ci.job.id": "29891234567",
        "ci.pipeline.id": "12345",
        "ci.provider.name": "github",
        "GITHUB_SERVER_URL": "https://github.com",
        "GITHUB_REPOSITORY": "DataDog/example",
        "GITHUB_RUN_ID": "12345",
        "GITHUB_RUN_ATTEMPT": "1"
      }
    }
  }
}
```

## Companion backend PRs

- DataDog/dd-source#443772 (writer for `correlate-image` + `ci.job.id` fix)
- DataDog/dd-source#445158 (same `ci.job.id` fix on the existing `correlate` endpoint)
- DataDog/dd-go#238027 (reader on the ArgoCD path)

Safe to ship before the backends land. Older endpoint versions ignore unknown attributes.

## Test plan

- `yarn test packages/plugin-deployment/src/__tests__/correlate-image.test.ts` passes, including new coverage for `ci_env` population.
- `yarn lint` clean.
- Validated end-to-end in staging.

### DIFF
--- a/packages/plugin-deployment/src/commands/correlate.ts
+++ b/packages/plugin-deployment/src/commands/correlate.ts
@@ -1,7 +1,7 @@
 import {DeploymentCorrelateCommand} from '@datadog/datadog-ci-base/commands/deployment/correlate'
 import {FIPS_ENV_VAR, FIPS_IGNORE_ERROR_ENV_VAR} from '@datadog/datadog-ci-base/constants'
 import {getDatadogSite} from '@datadog/datadog-ci-base/helpers/api'
-import {getCISpanTags, getGithubJobIDFromLogs, getGithubJobNameFromLogs} from '@datadog/datadog-ci-base/helpers/ci'
+import {getCISpanTags} from '@datadog/datadog-ci-base/helpers/ci'
 import {toBoolean} from '@datadog/datadog-ci-base/helpers/env'
 import {enableFips} from '@datadog/datadog-ci-base/helpers/fips'
 import {gitRepositoryURL, gitLocalCommitShas, gitCurrentBranch} from '@datadog/datadog-ci-base/helpers/git/get-git-data'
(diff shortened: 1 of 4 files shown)


packages/plugin-deployment/src/commands/correlate.ts
Aaron-9900, line 1: This one is legacy. I'd avoid updating it in any way

Aaron-9900 (approved):

packages/plugin-deployment/src/commands/correlate.ts
DaniFdz: Reverted in `9db901b3`. Kept the change scoped to `correlate-image.ts` only.

Side note: dd-source#445158 still applies the same `ci.job.id` fix on the GitOps `correlate` endpoint backend. If `correlate` is fully deprecated and we expect no more customer traffic on it, we can close #445158 and roll it back. Otherwise the backend fix is harmless and helps any legacy customers still on the GitOps flow. Up to you.

brett0000FF (approved):

rodrigo-roca (approved):

Issue #829, opened 2023-04-11
Global sourcemaps for microfrontend support


mortargrind:
### Feature request description

Hi,

We have a frontend application where we have “federated” microfrontends: essentially each microfrontend is managed by a different squad and can release new versions independent to each other. These microfrontends are not seperated via page/route boundaries so at any point in time there might be multiple microfrontend apps present on the page (I will call microfrontends “MFE”s throughout this post)

The app itself is using DataDog’s RUM SDK and “the shell” frontend is initializing it. We want to see each error thrown by each part of the application to have human readable stack traces (since we are minifying the code, they are not human-readable by default).

- Each MFE is completely self-managed (including their teams, repos and build pipelines)
- DataDog RUM SDK can only be initialized once and there can be only a single app version value (which is the shell frontend’s version)
- Sourcemap upload command requires a single app/service version to associate the sourcemaps with, regular “upload sourcemaps after you build your app so DataDog can display the human readable stack traces” is not applicable to our situation.

I’ve already went through some of the open issues under the DataDog RUM SDK GitHub repo (such as [this](https://github.com/DataDog/browser-sdk/issues/1280), [this](https://github.com/DataDog/browser-sdk/issues/1225) and [this](https://github.com/DataDog/browser-sdk/issues/1777)) I got the sense that the true MFE support (both for this case and where multiple “init” calls to the RUM SDK is needed) won’t be happening anytime soon.

### Solution

Since we don’t want to wait that long to have human readable errors on DataDog and preferably don’t want to maintain a really complex ad-hoc CI/CD solution where we would try to sync the associated sourcemaps and MFE versions across the organization, I have another short term solution that might be maybe done by the DataDog team.

In modern web applications almost all the asset files we deal with is being produced by a build process and the name of the files are being determined by their content. If you ignore hash collision probability; essentially we can say that each file is unique by their name. 

If a consumer is in this situation; there can be new opt-in things they can do for uploading sourcemaps, where instead of giving a release version to the upload command, they can just pass an indicator that tells DataDog that these sourcemaps they upload are “global” sourcemaps for the given application. So when DataDog encounters an error that is arising from that file, on top of just checking the sourcemaps associated with that release, it can also check the global one if there’s any match.

It would look something like this:

```
datadog-ci sourcemaps upload ./build --service my-service --minified-path-prefix https://static.datadog.com --global-release
```

[...]

PR #2337, opened 2026-05-29
[RUM-16533] Allow JS sourcemaps upload without a version (--global-release, --exclude)

buranmert:
### What and why?

Jira: [RUM-16533](https://datadoghq.atlassian.net/browse/RUM-16533) · GitHub: [#829](https://github.com/DataDog/datadog-ci/issues/829)

Micro-frontend apps and modern bundlers (content-hashed filenames) can't tie their sourcemaps to a single RUM `version`: each micro frontend releases independently, and per-version uploads duplicate huge numbers of files. Today `sourcemaps upload` **requires** `--release-version`, which makes this impossible.

This PR lets customers upload **version-less** sourcemaps that the backend matches regardless of the RUM `version` tag, plus an `--exclude` flag so builds mixing hashed and fixed filenames can be uploaded in separate invocations.

### How?

- **`--global-release`** (or simply omitting `--release-version`): sends a `__GLOBAL__` sentinel version that the backend recognizes as a global release. Passing both `--release-version` and `--global-release` is rejected. The command output shows `Version: __GLOBAL__ (global release)`.
- **`--exclude=<glob>`**: a glob (relative to the upload directory) of files to ignore. Resolved once to the set of minified files and applied as a filter on each collected sourcemap's minified file path, so exclusion behaves identically whether a sourcemap is discovered via a `sourceMappingURL` comment or the legacy `*.js.map` glob.
- README and command usage examples updated.

### Review checklist

[RUM-16533]: https://datadoghq.atlassian.net/browse/RUM-16533?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/base/src/commands/sourcemaps/README.md
+++ b/packages/base/src/commands/sourcemaps/README.md
@@ -32,7 +32,7 @@ The folder structure should match the structure of the served static files.
 
 * `--service` (required) should be set as the name of the service you're uploading sourcemaps for, and Datadog will use this service name to find the corresponding sourcemaps based on the `service` tag set on the RUM SDK.
 
-* `--release-version` (required) is similar and will be used to match the `version` tag set on the RUM SDK.
+* `--release-version` (required, unless uploading a global release) is similar and will be used to match the `version` tag set on the RUM SDK. See [Global (version-less) sourcemaps](#global-version-less-sourcemaps) to upload without a version.
 
 * `--minified-path-prefix` (required) should be a prefix common to all your JS source files, depending on the URL they are served from. The prefix can be a full URL or an absolute path.
 Example: if you're uploading `dist/file.js` to `https://example.com/static/file.js`, you can use `datadog-ci sourcemaps upload ./dist --minified-path-prefix https://example.com/static/` or `datadog-ci sourcemaps upload ./dist --minified-path-prefix /static/`.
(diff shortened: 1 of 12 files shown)


OliviaShoup (approved):

Drarig29 (commented):
I think we should simplify the API shape here and avoid the `global release` terminology.

The first-class RUM concept is `version`, not `release`, and this mode is not really a “global release” anyway: to me, a release sounds like an "immutable set of sourcemaps for a service at a given point in time", but we would keep appending sourcemaps to this "global release" over time.

I think the docs should describe the new behavior as "version-independent sourcemaps": sourcemaps that are matched independently of the RUM `version` tag.

I also think we should choose one UX for this behavior. Supporting both a new flag and allowing to omit `--release-version` makes it unclear which one is the intended path, and it adds complexity to the API. Given that accidentally uploaded version-independent sourcemaps are inspectable and removable from the [Explore RUM Debug Symbols](https://app.datadoghq.com/source-code/setup/rum) page, I’d lean toward making `--release-version` optional and removing the extra switch entirely.

We also need to make sure `version: __GLOBAL__` sourcemaps **can be viewed in the [Explore RUM Debug Symbols](https://app.datadoghq.com/source-code/setup/rum) page**.

My suggested direction would be:

- Remove `--global-release`.
- Keep `__GLOBAL__` sentinel (I don't mind the naming for the internal implementation)
- Make omitting `--release-version` the single way to upload version-independent sourcemaps.
- Update docs to talk about “version-independent sourcemaps” instead of “global release”

In the README:

```md
* `--release-version` is used to match the `version` tag set on the RUM SDK. If omitted, sourcemaps are uploaded as version-independent sourcemaps. See [Version-independent sourcemaps](#version-independent-sourcemaps).

...

### Version-independent sourcemaps

Some applications, such as micro-frontends (see [#829](https://github.com/DataDog/datadog-ci/issues/829)), cannot tie their sourcemaps to a single RUM `version`. When minified file URLs are unique across builds, typically because their filenames include a content hash, you can upload sourcemaps without binding them to a specific RUM version.

Version-independent sourcemaps are matched independently of the `version` tag set on the RUM SDK.
```

Example:
```sh
# Upload the content-hashed files as version-independent sourcemaps, skipping main.js
datadog-ci sourcemaps upload ./build --service my-service --minified-path-prefix https://static.datadog.com --exclude "**/main.js"

# Upload the non-hashed files with a RUM version
datadog-ci sourcemaps upload ./build/main.js.map --service my-service --minified-path-prefix https://static.datadog.com --release-version 1.234
```

And for the renderer:
```sh
Version: <none> (version-independent)
```

Also, in version-independent mode, could we emit a warning at the end when some file names do not contain a hash-like segment? (example regex: `(?:^|[.-])[a-f0-9]{8,}(?=[.-]|$)`)

Something along these lines:

```
Warning: 2 sourcemaps will be matched independently of RUM version, but their minified JS URLs do not look hash-based:
  - https://static.example.com/main.js
  - https://static.example.com/bundle.min.js

This may cause Datadog to match errors with the wrong sourcemap. Consider excluding them with `--exclude` and uploading them separately with a `--release-version`.
```

buranmert:
this is just a PoC and it was supposed to stay as a draft.
thanks for the reviews anyway!

buranmert:
we decided to proceed with a different solution after internal discussions. closing this PR

PR #2339, opened 2026-06-04
Add config file support to deployment gate command

GabrielAnca:
The deployment gate command only supports pre-created gates, where rules must be configured in Datadog ahead of time. This PR adds a `--config-file` option that accepts a path to a JSON file containing inline rule definitions, allowing gates to be evaluated without requiring any prior setup in Datadog.

Passing configuration via a JSON file is the standard pattern in this CLI for structured input (used by synthetics, SBOM, and others) — it avoids shell escaping of complex nested objects and keeps deployment config readable alongside the code.

When provided, the parsed configuration is included in the evaluation request. The command validates that the file exists, is valid JSON, and contains a non-empty `rules` array before proceeding.

### DIFF
--- a/packages/base/src/commands/deployment/gate.ts
+++ b/packages/base/src/commands/deployment/gate.ts
@@ -75,8 +75,8 @@ export class DeploymentGateCommand extends BaseCommand {
     description:
       'When true, the script will consider the gate as failed when timeout is reached or unexpected errors occur calling the Datadog APIs',
   })
-  protected configFilePath = Option.String('--config-file', {
-    description: 'Path to a JSON file containing gate configuration',
+  protected configFilePath = Option.String('--config', {
+    description: 'Path to the configuration file',
     validator: t.isString(),
   })
   // monitorsQueryVariable is hidden because it's not available yet
--- a/packages/base/src/commands/deployment/gate.ts
+++ b/packages/base/src/commands/deployment/gate.ts
@@ -26,6 +26,10 @@ export class DeploymentGateCommand extends BaseCommand {
         'datadog-ci deployment gate --service payments-backend --env prod',
       ],
       [
+        'Evaluate a gate with inline rule definitions',
+        'datadog-ci deployment gate --service payments-backend --env prod --version 1.2.3 --config-file ./gate-config.json',
+      ],
+      [
         'Evaluate a deployment gate with custom timeout',
         'datadog-ci deployment gate --service payments-backend --env prod --timeout 7200',
       ],
@@ -70,6 +74,10 @@ export class DeploymentGateCommand extends BaseCommand {
   protected failOnError = Option.Boolean('--fail-on-error', false, {
     description:
       'When true, the script will consider the gate as failed when timeout is reached or unexpected errors occur calling the Datadog APIs',
+  })
+  protected configFilePath = Option.String('--config-file', {
+    description: 'Path to a JSON file containing gate configuration',
+    validator: t.isString(),
   })
   // monitorsQueryVariable is hidden because it's not available yet
   protected monitorsQueryVariable = Option.String('--monitors-query-variable', '', {
--- a/packages/base/src/commands/deployment/gate.ts
+++ b/packages/base/src/commands/deployment/gate.ts
@@ -27,7 +27,7 @@ export class DeploymentGateCommand extends BaseCommand {
       ],
       [
         'Evaluate a gate with inline rule definitions',
-        'datadog-ci deployment gate --service payments-backend --env prod --version 1.2.3 --config-file ./gate-config.json',
+        'datadog-ci deployment gate --service payments-backend --env prod --version 1.2.3 --config ./gate-config.json',
       ],
       [
         'Evaluate a deployment gate with custom timeout',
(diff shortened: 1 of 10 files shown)


Aaron-9900 (approved):

packages/plugin-deployment/src/__tests__/fixtures/config-valid.json
```diff
@@ -0,0 +1,23 @@
+{
+  "dry_run": false,
```
Drarig29: I'd like all root level properties to be camelCase, to align with [`apiKey`, `appKey`, `datadogSite`](https://github.com/DataDog/datadog-ci/blob/master/packages/plugin-synthetics/src/__tests__/config-fixtures/config-with-global-keys.json).
```suggestion
  "dryRun": false,
```

For the `rules` array, since it's directly forwarded to the backend. I don't mind the snake case

packages/plugin-deployment/README.md
```diff
@@ -88,6 +88,7 @@ datadog-ci deployment gate --service payments-backend --env prod
 - `--identifier` (default: `default`): Deployment Gate identifier. For example, `pre`.
 - `--version` (**required** if your gate has faulty deployment detection rules): Version that is being deployed. For example, `v1.0.3`.
 - `--apm-primary-tag`: APM primary tag (only for gates with faulty deployment detection rules). For example, `region:us-central-1`.
+- `--config-file`: Path to a JSON file containing gate configuration. When provided, rules are evaluated inline without requiring a pre-created gate in Datadog.
```
Drarig29:
```suggestion
- `--config`: Path to the configuration file. Use this file to define inline rules without requiring a pre-created gate in Datadog.
```

packages/base/src/commands/deployment/gate.ts
```diff
@@ -71,6 +75,10 @@ export class DeploymentGateCommand extends BaseCommand {
     description:
       'When true, the script will consider the gate as failed when timeout is reached or unexpected errors occur calling the Datadog APIs',
   })
+  protected configFilePath = Option.String('--config-file', {
+    description: 'Path to a JSON file containing gate configuration',
+    validator: t.isString(),
+  })
```
Drarig29: I'd like to reintroduce a base command with base options to align all commands' configuration at some point, so I suggest reusing `--config` like other commands.
```suggestion
  protected configFilePath = Option.String('--config', {
    description: 'Path to the configuration file',
    validator: t.isString(),
  })
```

packages/base/src/commands/deployment/gate.ts
```diff
@@ -25,6 +25,10 @@ export class DeploymentGateCommand extends BaseCommand {
         'Evaluate a deployment gate for payments-backend service in prod environment',
         'datadog-ci deployment gate --service payments-backend --env prod',
       ],
+      [
+        'Evaluate a gate with inline rule definitions',
+        'datadog-ci deployment gate --service payments-backend --env prod --version 1.2.3 --config-file ./gate-config.json',
```
Drarig29:
```suggestion
        'datadog-ci deployment gate --service payments-backend --env prod --version 1.2.3 --config ./gate-config.json',
```

packages/plugin-deployment/README.md
Drarig29: Can you also add an example of such inline rules? (or redirect to docs)

packages/plugin-deployment/src/__tests__/fixtures/config-valid.json
GabrielAnca: Updated. I updated it to have camelCase everywhere except for the 'options'. These options change more frequently, so leaving it open would help the customers being able to adapt to changes without having to upgrade the CLI.

packages/plugin-deployment/README.md
GabrielAnca: Updated with an example and link to docs. However, the docs are still missing documentation for this feature. It's the chicken and the egg problem: the docs will also include an example with datadog-ci (and the Datadog/deployment-gate-github-action -I'm working on a PR there too) 😅 I won't merge this PR until I have the docs ready so that I minimize the time where the three are out of date from each other

Drarig29 (approved):

packages/plugin-deployment/src/__tests__/fixtures/config-valid.json
Drarig29: Nice, perfect! Sounds good to me :)

OliviaShoup (approved):

packages/plugin-deployment/README.md
```diff
@@ -110,3 +129,4 @@ Additional helpful documentation, links, and articles:
 [2]: https://docs.datadoghq.com/continuous_delivery/
 [3]: https://docs.datadoghq.com/continuous_delivery/deployments/argocd#correlate-deployments-with-ci-pipelines
 [4]: https://docs.datadoghq.com/deployment_gates/
+[5]: https://docs.datadoghq.com/deployment_gates/setup
```
GabrielAnca: To-do: update this when https://github.com/DataDog/documentation/pull/37364 is approved. If current path is accepted, we'll have to update to

```suggestion
[5]: https://docs.datadoghq.com/deployment_gates/setup/jit
```

GabrielAnca: While that's the final link, let's point to the api spec for now, because that's already online. I'll follow up updating this when the docs are live (because they have a circular dependency on datadog-ci being released)


```suggestion
[5]: https://docs.datadoghq.com/api/latest/deployment-gates/trigger-a-deployment-gate-evaluation/
```

Drarig29: Sounds good!

PR #2340, opened 2026-06-05
[SVLS-9256] auto-mark dd env sticky on slot instrumentation

ava-silver:
### What and why?

In Azure App Service, a slot swap swaps app settings between slots by default. `DD_ENV` should be a *sticky* slot setting so a slot keeps its own environment tag after a swap, instead of the production slot's `DD_ENV` bleeding into staging.

This makes `aas instrument` register `DD_ENV` as a sticky slot setting whenever a slot is instrumented with `--env`. Non-slot instrumentation is unaffected.

### How?

- `common.ts`:
  - `StickySlotSettings` describes the sticky names a site needs (`resourceGroup`, `webAppName`, `names`).
  - `aggregateStickyBySite` collapses the per-resource sticky settings reported during instrumentation into one entry per site, unioning their names. `slotConfigNames` is a single site-level resource shared by all of a site's slots, so it keys by web app name (App Service names are globally unique) and writes each site once -- avoiding the concurrent read-modify-writes (and Azure 409s) that would happen if multiple slots of the same app each wrote it.
  - `registerStickySlotSettings` performs that single read-modify-write, appending any missing names via `updateSlotConfigurationNames`. It no-ops when the names are already sticky, logs the action, and honors dry-run.
- `instrument.ts`:
  - `processWebApp` now returns a `ProcessResult` (`{success, sticky?}`) instead of a bare boolean, reporting the sticky settings the resource needs on its parent site (just `DD_ENV`, and only for slots with `--env`).
  - `processSubscription` aggregates those results and registers the sticky settings once per site after all resources are processed.

### Review checklist

### DIFF
--- a/packages/plugin-aas/src/commands/instrument.ts
+++ b/packages/plugin-aas/src/commands/instrument.ts
@@ -11,7 +11,6 @@ import {
   WINDOWS_RUNTIME_EXTENSIONS,
 } from '@datadog/datadog-ci-base/commands/aas/common'
 import {AasInstrumentCommand} from '@datadog/datadog-ci-base/commands/aas/instrument'
-import {DATADOG_SITE_US1} from '@datadog/datadog-ci-base/constants'
 import {getDatadogSite} from '@datadog/datadog-ci-base/helpers/api'
 import {newApiKeyValidator} from '@datadog/datadog-ci-base/helpers/apikey'
 import {renderError, renderSoftWarning} from '@datadog/datadog-ci-base/helpers/renderer'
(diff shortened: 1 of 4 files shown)


ava-silver:
* **#2357** <a href="https://app.graphite.com/github/pr/DataDog/datadog-ci/2357?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2340** <a href="https://app.graphite.com/github/pr/DataDog/datadog-ci/2340?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/datadog-ci/2340?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

packages/plugin-aas/src/commands/instrument.ts
```diff
@@ -166,7 +169,11 @@ This flag is only applicable for containerized .NET apps (on musl-based distribu
       }
       config.isDotnet ||= isDotnet(site)
       config.isMusl &&= config.isDotnet && isContainer
-      await this.instrumentSidecar(aasClient, config, resourceGroup, webApp, isContainer, existingEnvVars)
+      await Promise.all([
+        this.instrumentSidecar(aasClient, config, resourceGroup, webApp, isContainer, existingEnvVars),
+        this.makeStickySlotEnvVars(aasClient, resourceGroup, webApp, config),
```
avangelillo: Will we ever want to instrument a web app and not make the slot env vars sticky?  Is it worth moving this into `instrumentSidecar` /  `instrumentExtension`

ava-silver: No, but also the current functions of `instrumentSidecar`/`instrumentExtension` are for conceptually splitting out the logic needed to update the config/extension, rather than for splitting out the entire instrumentation process. we currently also have this same pattern for updating tags, so I'm inclined to say its fine here, especially since it makes parallelization much easier than putting the calls into the `instrument*` bodies

Drarig29 (approved):

TalUsvyatsky (approved):

PR #2342, opened 2026-06-08
Use Datadog registry for Cloud Run sidecar default

AliDatadog:
## Summary
- Change the Cloud Run `serverless-init` sidecar default from `gcr.io/datadoghq` to `registry.datadoghq.com`.
- Update the Cloud Run plugin README plus affected tests and snapshots.

## Why
The container registry guidance recommends `registry.datadoghq.com` for simple Datadog image pulls while keeping GCR available as an explicit registry choice for Google deployments. This keeps the CLI default aligned with that guidance without removing the `--image` override.

## Validation
- `yarn install`
- `yarn build`
- `yarn lint`
- `yarn test packages/plugin-cloud-run/src/__tests__/instrument.test.ts packages/plugin-cloud-run/src/__tests__/uninstrument.test.ts`


AliDatadog:
Closing this draft after re-checking it against the container registry selection guidance. Cloud Run is a GCP-specific deployment path, so keeping the existing gcr.io/datadoghq/serverless-init default is the better fit than switching the command default to registry.datadoghq.com. The Linux Cloud Run E2E jobs also failed on the changed default.

PR #2344, opened 2026-06-08
[K9VULN-15920] Warn when env tag is passed to sbom upload

divyav-datadog:
### What and why?

The `--env` flag is ignored server-side for non-SCI SBOM uploads (env is always normalized to `"ci"`). This adds a client-side warning so users know the flag has no effect rather than silently ignoring it.

### How?

- `packages/plugin-sbom/src/renderer.ts` — added `renderEnvFlagDeprecationWarning()` that prints a yellow warning to stderr
- `packages/plugin-sbom/src/commands/upload.ts` — calls the warning when `--env` is set to anything other than the default `"ci"`

### DIFF
--- a/packages/plugin-sbom/src/commands/upload.ts
+++ b/packages/plugin-sbom/src/commands/upload.ts
@@ -96,10 +96,10 @@ export class PluginCommand extends SbomUploadCommand {
     const service = 'datadog-ci'
 
     if (this.env !== 'ci') {
-      this.context.stderr.write(renderEnvFlagDeprecationWarning())
+      this.context.stdout.write(renderEnvFlagDeprecationWarning())
     }
 
-    const environment = this.env
+    const environment = 'ci'
 
     if (!this.basePath || !this.basePath.length) {
       this.context.stderr.write('Missing basePath\n')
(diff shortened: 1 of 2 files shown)


packages/plugin-sbom/src/commands/upload.ts
```diff
@@ -94,6 +95,10 @@ export class PluginCommand extends SbomUploadCommand {
 
     const service = 'datadog-ci'
 
+    if (this.env !== 'ci') {
+      this.context.stderr.write(renderEnvFlagDeprecationWarning())
```
dastrong: We shouldn't write this to stderr, but instead stdout 

```suggestion
      this.context.stdout.write(renderEnvFlagDeprecationWarning())
```

packages/plugin-sbom/src/commands/upload.ts
```diff
@@ -94,6 +95,10 @@ export class PluginCommand extends SbomUploadCommand {
...
+    }
+
     const environment = this.env
```
dastrong: Should we override `environment` here with `ci` similarly to what's done on L96

```
const environment = 'ci'
```

dastrong (approved):

Issue #1991, opened 2025-11-19
Add JSON logging


rgoomar:
### Feature request description

Currently, the `Logger` helper only emits stdout text logs. There should be a way to specify JSON logging so that it can be easily consumed and attributed by DataDog itself. Currently, `ERROR: ...` type logs show up as `info` by default without a pipeline to parse it. 

### Command

None

PR #2347, opened 2026-06-10
feat(logger): add `jsonOutput` to internal logger

tsushanth:
## Why

Related to #1991.

The existing `Logger` only emits ANSI-coloured text. When the CLI runs inside a log pipeline (Datadog itself, or any other ingestion path), every line is unstructured — `error` records show up as `info` because the consumer can't tell them apart without parsing the colour codes or prefixes.

## What

- Add a third constructor shape that accepts `LoggerOptions` (`{ shouldIncludeTimestamp?, jsonOutput? }`). The legacy `boolean` form (`new Logger(write, INFO, true)`) still works unchanged, so existing call sites compile without edits.
- When `jsonOutput` is true, each log call writes a single-line JSON object:
  ```jsonl
  {"level":"error","message":"deploy failed","timestamp":"2026-06-10T15:09:00.123Z"}
  ```

  `level` is the canonical lowercase name. `timestamp` is only emitted when `shouldIncludeTimestamp` is also enabled (matching how the text-mode prefix works). ANSI colours are intentionally skipped so the line parses cleanly with `JSON.parse`.

- Expose `setJsonOutput(boolean)` alongside the existing `setShouldIncludeTime` so plugins can flip the format at runtime (e.g. based on a CLI flag or env var).

## Tests

`packages/base/src/helpers/__tests__/logger.test.ts` — 9 cases covering:

- Text mode: trailing newline, level gating, ISO timestamp prefix (both legacy boolean and options-bag form).
- JSON mode: level/message correctness across all four levels, no ANSI codes, timestamp inclusion vs omission, JSON→text runtime toggle.

`yarn build`, `yarn lint`, `yarn test packages/base/src/helpers/__tests__/logger.test.ts` — all clean.

## Follow-up

Wiring a `--json` (or `DATADOG_CI_JSON_LOG=1`) toggle into the individual plugin commands would close the loop end-to-end, but I've left that out of this PR since it touches every plugin and the right pattern (per-plugin flag vs. global) is a separate design call. Happy to do it as a follow-up if you want — flagging here so reviewers know it's intentional scope, not an oversight.

### DIFF
--- a/packages/base/src/helpers/__tests__/logger.test.ts
+++ b/packages/base/src/helpers/__tests__/logger.test.ts
@@ -1,0 +1,126 @@
+import {Logger, LogLevel} from '../logger'
+
+describe('Logger', () => {
+  const collect = () => {
+    const lines: string[] = []
+    const write = (s: string) => {
+      lines.push(s)
+    }
+
+    return {lines, write}
+  }
+
+  describe('text mode (default)', () => {
+    it('writes each level with a trailing newline', () => {
+      const {lines, write} = collect()
+      const logger = new Logger(write, LogLevel.DEBUG)
+
+      logger.debug('d')
+      logger.info('i')
+      logger.warn('w')
+      logger.error('e')
+
+      expect(lines).toHaveLength(4)
+      expect(lines.every((line) => line.endsWith('\n'))).toBe(true)
+    })
+
+    it('respects the configured log level', () => {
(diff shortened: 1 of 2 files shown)


Drarig29 (approved):
LGTM. Will merge and think about the best way to support this in all commands. We already have internal conversations ongoing for a shared base command class.

PR #2351, opened 2026-06-12
Split catalog-endpoints CI job into generate and upload

GabrielAnca:
The previous single-job approach used `runner:docker` (legacy EC2 runner) because `node:22` lacks Datadog's private CA certs needed for the CI Identities Fabric DNS path on Kubernetes runners. However, `runner:docker` is not available on protected branches, causing the job to get stuck.

The job is now split into two:

- **catalog-endpoints-generate** runs `yarn catalog-endpoints` on `node:22` with no CI Identities involved, and passes `endpoints.csv` as a GitLab artifact.
- **catalog-endpoints-upload** picks up the artifact and uploads it to S3 using the internal `registry.ddbuild.io/images/aws-cli` image, which already has Datadog CA certs and AWS CLI. Both jobs run on `arch:amd64`.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -50,7 +50,7 @@ catalog-endpoints-upload:
   only: ['master']
   needs: ['catalog-endpoints-generate']
   tags: ['arch:amd64']
-  image: registry.ddbuild.io/images/aws-cli:2.28.10
+  image: registry.ddbuild.io/images/aws-cli:2.33.12
   id_tokens:
     CI_IDENTITIES_GITLAB_ID_TOKEN:
       aud: ci-identities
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -32,22 +32,33 @@ publish-docker-image:
...
   only: ['master']
-  tags: ['runner:docker']
+  tags: ['arch:amd64']
   image: node:22
+  script:
+    - yarn install --immutable
+    - yarn catalog-endpoints
+  artifacts:
+    paths:
+      - endpoints.csv
```
Drarig29, line 45: Is this artifact declaration + `needs: ['catalog-endpoints-generate']` enough for GitLab to pass the file to `catalog-endpoints-upload`? No need to explicitly pull the artifact?

GabrielAnca: According to the [documentation](https://docs.gitlab.com/ci/yaml/#needsartifacts), it's enough. It downloads the artifacts from needs section by default

Drarig29 (approved):

PR #2355, opened 2026-06-12
Add YAML linting to CI

GabrielAnca:
Adds a YAML linting step to GitHub Actions CI to catch YAML syntax errors before they reach GitLab. This would have prevented the `.gitlab-ci.yml` syntax error we shipped recently (`"${CI_PROJECT_DIR}/..." assume-role` is invalid YAML).

Uses `eslint-plugin-yml` (1M weekly downloads, actively maintained) via a new `yarn lint:yaml` script. Linting covers all YAML files in the repo via `eslint-plugin-yml`'s `flat/standard` ruleset, with a few rules tuned:

- `yml/quotes` and `yml/plain-scalar` disabled — existing files use mixed quoting intentionally and the two rules conflict with each other
- `prettier/prettier` disabled for YAML — Prettier's single-quote preference clashes with YAML conventions across too many files
- `yml/no-empty-mapping-value` disabled

I added specific fixture files added to `globalIgnores` to prevent intentionally invalid files from failing CI.

Also bumps `actions/checkout` from v6.0.2 to v6.0.3 across all CI jobs

Happy to address the exceptions I made in separate PRs. I wanted to avoid making this one too complex.


eslint.config.mjs
```diff
@@ -106,6 +107,9 @@ export default defineConfig(
     'packages/datadog-ci/shims/injected-plugin-submodules.js',
     'packages/datadog-ci/shims/intl-collator.js',
     'bin/*.js',
+    '.github/dependabot.yml',
```
GabrielAnca, line 113: This file is commented out entirely. So it's invalid YAML

eslint.config.mjs
```diff
@@ -405,5 +413,16 @@ export default defineConfig(
     rules: {
       'no-restricted-imports': ['error', {paths: restrictedImports}],
     },
-  }
+  },
+  ...yml.configs['flat/standard'],
+  {
+    files: yamlFiles,
+    extends: [tseslint.configs.disableTypeChecked],
```
GabrielAnca, line 420: This avoids running TS checks on yaml files. If we don't do this, TypeScript lint fails on yaml files.

[Disabling TS tests in yaml files using the extends directive is the recommended approach](https://typescript-eslint.io/packages/typescript-eslint#flat-config-extends), even though I think it's not super easy to follow. Being able to do something at the level where we introduce the conflict `tseslint.configs.recommendedTypeChecked` would be better, but the current solution would be even harder to read (`tseslint.configs.recommendedTypeChecked.map((config) => config.files ? config : {...config, files: tsFiles})`)

Drarig29 (approved):

PR #2357, opened 2026-06-15
[SVLS-9186] add Windows Function App support to aas instrument

ava-silver:
### What and why?

`aas instrument` treated every site like a Web App, so Function Apps fell through the platform checks and got the wrong handling. This adds first-class Function App support:

- **Windows .NET Function Apps** install the same site extension as Windows Web Apps. On a deployment slot there's a wrinkle from [datadog-aas-extension#457](https://github.com/DataDog/datadog-aas-extension/pull/457): the Functions runtime holds file locks on the `SiteExtensions` directory, so Kudu's `MoveDirectory` step fails intermittently during the install. Setting `WEBSITE_PRIVATE_EXTENSIONS=0` makes the runtime release those locks. It is committed *before* the install and registered as a sticky slot setting so it doesn't follow a swap into production.
- **Windows Function Apps on Node or Java** aren't supported via the extension -- the command exits with an error pointing to the [Azure Functions docs](https://docs.datadoghq.com/serverless/azure_functions).
- **Windows Function Apps on a Consumption / Flex Consumption plan** aren't supported -- the [extension requires a Dedicated (App Service) or Premium plan](https://docs.datadoghq.com/serverless/azure_functions/dotnet_extension/), so the command exits with the same redirect rather than installing an extension that will never produce telemetry.
- **Linux Function Apps** aren't supported -- the command exits with the same redirect instead of silently falling through to the sidecar path.

Web App behavior is unchanged. Builds on the sticky-slot helpers from the parent PR.

### How?

- `common.ts`:
  - Adds `isFunctionApp`, `isWindowsFunctionApp`, `isConsumptionPlan`, the `WEBSITE_PRIVATE_EXTENSIONS` constant, and `AZURE_FUNCTIONS_DOCS_URL`.
  - `isConsumptionPlan` reads `site.sku` (the plan tier name, e.g. `Dynamic` for Consumption, `FlexConsumption` for Flex Consumption) off the site object the command already fetches -- no extra API call.
  - `getEnvVars(config, site, webApp)` derives containerization from the site and injects `WEBSITE_PRIVATE_EXTENSIONS=0` for Windows Function App slots, so the lock-release setting is part of the settings written before the install.
  - Generalizes the parent PR's `registerStickySlotSettings` into `mutateStickySlotSettings(..., mode: 'add' | 'remove', ...)`, so the same single-read-modify-write path both registers settings (instrument) and unregisters them (uninstrument).
  - Moves the shared `ProcessResult` type and a `stickySlotSettings` helper here so both commands use them.
- `instrument.ts`:
  - Rejects Linux Function Apps before the Windows branch. Inside the Windows Function App branch, rejects Consumption/Flex Consumption plans and non-.NET runtimes with the redirect.
  - `instrumentExtension`/`instrumentSidecar` take the resolved `site` and write app settings **before** the extension install (previously the write was dispatched and only awaited afterward). This gives the lock-release setting the ordering it needs.
  - On a .NET Function App slot, pins `WEBSITE_PRIVATE_EXTENSIONS` sticky alongside `DD_ENV`.
- `uninstrument.ts`:
  - Returns `ProcessResult` and, on a Windows Function App slot, removes the `WEBSITE_PRIVATE_EXTENSIONS` app setting and unregisters it sticky (`mutateStickySlotSettings` in `remove` mode, aggregated once per site). `DD_ENV` is intentionally left sticky.

### Testing

Validated end-to-end against a live Windows .NET (isolated) Function App on a Dedicated (App Service) plan: deployed an HTTP-triggered function, ran `aas instrument --windows-runtime dotnet`, drove traffic, and confirmed traces in Datadog.

While doing this I also confirmed the plan requirement empirically: on a **Consumption** plan the extension installs but never injects the profiler (`CORECLR_*` env vars are absent from the worker), so no traces flow -- which is why the command now rejects Consumption/Flex Consumption up front. On the Dedicated plan, traces flowed with `WEBSITE_USE_PLACEHOLDER_DOTNETISOLATED` left at its default, so no extra placeholder configuration is needed.

### Review checklist

### DIFF
--- a/packages/plugin-aas/src/common.ts
+++ b/packages/plugin-aas/src/common.ts
@@ -1,5 +1,5 @@
 import type {Site, SlotConfigNamesResource, WebSiteManagementClient} from '@azure/arm-appservice'
-import type {AasConfigOptions, WindowsRuntime} from '@datadog/datadog-ci-base/commands/aas/common'
+import type {AasConfigOptions, WebApp, WindowsRuntime} from '@datadog/datadog-ci-base/commands/aas/common'
 
 import {getBaseEnvVars} from '@datadog/datadog-ci-base/helpers/serverless/common'
 import chalk from 'chalk'
@@ -34,6 +34,10 @@ export const AAS_DD_SETTING_NAMES = [
   'WEBSITES_ENABLE_APP_SERVICE_STORAGE',
 ] as const
 
+// Disabling private site extensions releases the Functions runtime's locks on the SiteExtensions
+// directory before the extension install (datadog-aas-extension#457).
+export const WEBSITE_PRIVATE_EXTENSIONS = 'WEBSITE_PRIVATE_EXTENSIONS'
+
 /**
  * Detects the runtime of a Windows-based Web App
  * @param site The web app or slot
@@ -57,12 +61,17 @@ export const getWindowsRuntime = (site: Site, envVars: Record<string, string>): 
   return undefined
 }
 
-export const getEnvVars = (config: AasConfigOptions, isContainer: boolean): Record<string, string> => {
+export const getEnvVars = (config: AasConfigOptions, site: Site, webApp: WebApp): Record<string, string> => {
+  const isContainer = isLinuxContainer(site)
+  // Function App slots need private extensions disabled before the extension install.
+  const includePrivateExtensions = isWindowsFunctionApp(site) && !!webApp.slot
+
   // Get base environment variables
   let envVars = getBaseEnvVars(config)
   envVars = {
     WEBSITES_ENABLE_APP_SERVICE_STORAGE: 'true',
     DD_AAS_INSTANCE_LOGGING_ENABLED: (config.isInstanceLoggingEnabled ?? false).toString(),
+    ...(includePrivateExtensions ? {[WEBSITE_PRIVATE_EXTENSIONS]: '0'} : {}),
     ...envVars,
   }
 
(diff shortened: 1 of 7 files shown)


ava-silver:
* **#2357** <a href="https://app.graphite.com/github/pr/DataDog/datadog-ci/2357?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/datadog-ci/2357?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2340** <a href="https://app.graphite.com/github/pr/DataDog/datadog-ci/2340?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

packages/plugin-aas/src/common.ts
```diff
@@ -82,6 +91,12 @@ export const getEnvVars = (config: AasConfigOptions, isContainer: boolean): Reco
   return envVars
 }
 
+export const AZURE_FUNCTIONS_DOCS_URL = 'https://docs.datadoghq.com/serverless/azure_functions'
```
Lewis-E, line 94: Do you want `https://docs.datadoghq.com/serverless/azure_functions` for this url, or `https://docs.datadoghq.com/serverless/azure_functions/dotnet_extension`? The second page is specific to the extension installation, while the first (current value of AZURE_FUNCTIONS_DOCS_URL) is mostly about the compat layer.

ava-silver: good catch, for some of the warnings we do want it, but not for all. i'll differentiate the two

Lewis-E (approved):

Drarig29 (approved):

nina9753 (approved):

PR #2374, opened 2026-06-22
[RUM-16940] add `--debug-id` upload mode for web sourcemaps

amortemousque:
## Context

When an error spans multiple micro-frontends, unminification breaks because all frames are resolved using the top frame’s (service, version).

This PR adds support for `debug_id` sourcemaps by extracting the `debug_id` (a [TC39 proposal](https://github.com/tc39/ecma426/blob/main/proposals/debug-id.md)￼) from the source file and including it in the upload metadata. This enables the backend to resolve sourcemaps per frame using debug_id instead of a shared (service, version).

## What this PR does

Adds a `--debug-id` flag to `sourcemaps upload`. When set:

- `--service`, `--release-version`, and `--minified-path-prefix` are **forbidden** when --debug-id is present
- The CLI reads the `debug_id` from the bundle's `ddDebugId` runtime snippet (injected by the build plugin)
- The upload event metadata includes `debug_id` instead of `service`/`version`/`minified_url`

## Test plan

- [ ] `yarn build`
- [ ] `yarn test packages/base/src/commands/sourcemaps`
- [ ] Manual: `yarn launch sourcemaps upload <path> --debug-id  --dry-run`
- [ ] Manual: `yarn launch sourcemaps upload <path>` — validates and rejects missing flags

### DIFF
--- a/packages/base/src/commands/sourcemaps/debugId.ts
+++ b/packages/base/src/commands/sourcemaps/debugId.ts
@@ -1,21 +1,32 @@
 import fs from 'fs'
 
-import type {CommandContext} from '@datadog/datadog-ci-base'
+import type {Sourcemap} from './interfaces'
 
 const DD_DEBUG_ID_REGEX = /["']ddDebugId["']\s*:\s*["']([^"']+)["']/
 
-export const extractDebugId = (filePath: string, context: CommandContext): string | undefined => {
+export const extractDebugId = (filePath: string): string | undefined => {
   try {
     const source = fs.readFileSync(filePath, 'utf-8')
-    const match = source.match(DD_DEBUG_ID_REGEX)
-    if (match) {
-      return match[1]
+
+    return source.match(DD_DEBUG_ID_REGEX)?.[1]
+  } catch {
+    // Unreadable file: treated as having no debug ID.
+    return undefined
+  }
+}
+
+/**
+ * Adds the debug ID extracted from each payload's minified file onto the
+ * payload. Returns true if at least one payload has a debug ID.
+ */
+export const addDebugIdToPayloads = (payloads: Sourcemap[]): boolean => {
+  let hasAnyDebugId = false
+  for (const payload of payloads) {
+    payload.debugId = extractDebugId(payload.minifiedFilePath)
+    if (payload.debugId !== undefined) {
+      hasAnyDebugId = true
     }
-    context.stderr.write(`Debug ID not found in ${filePath}\n`)
-  } catch (err) {
-    const errorMsg = err instanceof Error ? err.message : String(err)
-    context.stderr.write(`Cannot extract Debug ID from ${filePath}: ${errorMsg}\n`)
   }
 
-  return undefined
+  return hasAnyDebugId
 }
--- a/packages/base/src/commands/sourcemaps/debugId.ts
+++ b/packages/base/src/commands/sourcemaps/debugId.ts
@@ -1,0 +1,21 @@
+import fs from 'fs'
+
+import type {CommandContext} from '@datadog/datadog-ci-base'
+
+const DD_DEBUG_ID_REGEX = /["']ddDebugId["']\s*:\s*["']([^"']+)["']/
+
+export const extractDebugId = (filePath: string, context: CommandContext): string | undefined => {
+  try {
+    const source = fs.readFileSync(filePath, 'utf-8')
+    const match = source.match(DD_DEBUG_ID_REGEX)
+    if (match) {
+      return match[1]
+    }
+    context.stderr.write(`Debug ID not found in ${filePath}\n`)
+  } catch (err) {
+    const errorMsg = err instanceof Error ? err.message : String(err)
+    context.stderr.write(`Cannot extract Debug ID from ${filePath}: ${errorMsg}\n`)
+  }
+
+  return undefined
+}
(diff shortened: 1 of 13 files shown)


Drarig29 (approved):

packages/base/src/commands/sourcemaps/upload.ts
```diff
@@ -421,14 +441,17 @@ export class SourcemapsUploadCommand extends BaseCommand {
         return UploadStatus.Skipped
       }
 
-      const payload = sourcemap.asMultipartPayload(
-        this.cliVersion,
-        this.service!,
-        this.releaseVersion!,
-        this.projectPath ?? ''
-      )
+      const debugId = this.debugId ? extractDebugId(sourcemap.minifiedFilePath, this.context) : undefined
```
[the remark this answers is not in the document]
amortemousque: Already handled — in `--debug-id` mode a payload is never uploaded without an identifier:

1. **Per file:** a sourcemap with no `ddDebugId` is skipped (`missing_debug_id`) in validation, before `asMultipartPayload` is built.
2. **All files:** if none yield a debug ID, the command aborts with exit 1 (`No debug ID found in any minified file. Aborting upload.`) before any upload.

packages/base/src/commands/sourcemaps/debugId.ts
```diff
@@ -0,0 +1,32 @@
+import fs from 'fs'
+
+import type {Sourcemap} from './interfaces'
+
+const DD_DEBUG_ID_REGEX = /["']ddDebugId["']\s*:\s*["']([^"']+)["']/
```
buranmert: can we use a more strict regex for UUID format like [this](https://www.postgresql.org/docs/current/datatype-uuid.html)? or do some bundlers use non-UUID debugIDs and we need to accept them too?

packages/base/src/commands/sourcemaps/debugId.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export const extractDebugId = (filePath: string): string | undefined => {
+  try {
+    const source = fs.readFileSync(filePath, 'utf-8')
+
+    return source.match(DD_DEBUG_ID_REGEX)?.[1]
```
buranmert, line 12: if we can assume that the debugID is in the last lines of a file, maybe we can use [this](https://github.com/DataDog/datadog-ci/blob/master/packages/base/src/commands/sourcemaps/utils.ts#L8) to avoid reading the whole file

buranmert (commented):
LGTM but i left 2 comments before approving

packages/base/src/commands/sourcemaps/debugId.ts
amortemousque: I’m reading the debugId from the snippet injected at the top of the file. There may already be other injected content before it, so I can’t assume it’s at a fixed position. I also considered streaming the file and stopping as soon as I found the match, but that would likely require running multiple regexes per file, which doesn’t seem worth the tradeoff. Wdyt?

packages/base/src/commands/sourcemaps/debugId.ts
amortemousque: I used the same pattern as the [format](https://github.com/DataDog/rum-events-format/pull/404#discussion_r3530632289)

BenoitZugmeyer (approved):

packages/base/src/commands/sourcemaps/upload.ts
```diff
@@ -380,6 +376,38 @@ export class SourcemapsUploadCommand extends BaseCommand {
     })
   }
 
+  private validateOptions(): boolean {
```
buranmert, line 379: (i chose a random line to leave this comment)

should we allow the customers to specify service,version,path and debug_id at the same time? or should we enforce one of these ways?

for example i'm a customer
yesterday i did `datadog-ci sourcemaps upload --service foo1 --version bar111 --debug_id 123-456` ✅ 
if today i do `datadog-ci sourcemaps upload --service foo2 --version bar222 --debug_id 123-456` this will be dropped in our ingestion pipeline as `debug_id` is a duplicate ❌ 

i'm wondering if this can create confusion
• IMO either we can disallow usage of service,version,path and debug_id at the same time
• or we can simply log a warning in terminal saying "service,version,path will be ignored as debug_id is already provided"

what do you think?

amortemousque: You’re right. I’ll update it to forbid using service and version alongside --debug-id.

amortemousque: Done https://github.com/DataDog/datadog-ci/pull/2374/changes/d5d3088628fd95f36d45b388c94544d35375bc87

buranmert (approved):

packages/base/src/commands/sourcemaps/interfaces.ts
```diff
@@ -53,20 +50,17 @@ export class Sourcemap {
...
     const metadata: {[k: string]: any} = {
       cli_version: cliVersion,
-      minified_url: this.minifiedUrl,
       project_path: projectPath,
-      service,
       type: 'js_sourcemap',
+      service,
       version,
+      minified_url: this.minifiedUrl,
+      debug_id: this.debugId,
```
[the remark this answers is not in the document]
amortemousque: This is the --minified-path-prefix option that is forbidden with --debug-id, not the minifiedUrl metadata. It’s true that minifiedUrl will be ignored with debug id during matching, but we don’t need extra logic to exclude it from the upload.

PR #2375, opened 2026-06-24
fix(trace): write diagnostics to stderr so they don't corrupt captured stdout

### LINKED ISSUE #2381: Route diagnostics and the version banner from stdout to stderr (and allow suppressing the banner)

`datadog-ci` writes several of its own diagnostics — and the `datadog-ci v<version>` banner — to **stdout**. This corrupts captured output: `datadog-ci trace -- <cmd>` runs the wrapped command with `stdio: ['inherit', 'inherit', 'pipe']`, so the child's stdout is fd 1 and its bytes are captured verbatim by `VAR=$(datadog-ci trace -- cmd)`. Any diagnostic the CLI also writes to fd 1 gets mixed into that capture. When the wrapped command emits a secret (e.g. `TOKEN=$(... vault write -field=token ...)`), the captured value gains an extra line of diagnostic text and is rejected downstream ("configured Vault token contains non-printable characters"). The corruption is intermittent and load-correlated (e.g. the `--no-fail` note only fires when the span report fails under intake throttling).

The diagnostics currently written to stdout that shouldn't be:

- the `--no-fail` note (`trace.ts`)
- the unsupported-CI-provider message (shared `CustomSpanCommand`, `helper.ts`)
- the missing-API-key message (shared `CustomSpanCommand`, `helper.ts`)
- the `datadog-ci v<version>` banner (printed on every invocation)

Separately, the banner is **spammy**: in CI where the CLI runs many times (e.g. many `trace --` wrap per step), it repeats on every call with no per-run value, and there's no way to turn it off.

### Solution

1. **Route all of these diagnostics, and the banner, to stderr.** This keeps them out of captured stdout, consistent with the adjacent error paths that already use `context.stderr`. The banner mirrors the resolved command's `--log-format`, so it stays a JSON line in JSON mode.
2. **Add `DD_CI_SKIP_VERSION_BANNER`** (`1`/`true`) to suppress the banner entirely. It follows the existing `DD_CI_`-prefixed toggle convention (e.g. `DD_CI_BYPASS_SITE_VALIDATION`) and the `1`/`true` truthiness used by `DD_BETA_COMMANDS_ENABLED`.

### Additional context

- The stdout→stderr move is a behavior change for anything that was scraping these lines from stdout, hence this issue.
- Default banner behavior is otherwise unchanged; it still prints (now on stderr) unless the variable is set.
- Implemented in #2375.

### Command

trace

eliebleton-manomano:
### What and why?

`datadog-ci trace -- <cmd>` runs the wrapped command with `stdio: ['inherit', 'inherit', 'pipe']`, so the child's stdout **is** fd 1 and its bytes are captured verbatim by `VAR=$(datadog-ci trace -- cmd)`. But `trace` also wrote some of its **own** diagnostics to fd 1 (`console.log` / `context.stdout.write`), so those lines got mixed into the capture.

When the wrapped command emits a secret — e.g. `TOKEN=$(… vault write -field=token …)` — the captured value gains a newline + diagnostic text and is rejected downstream (`configured Vault token contains non-printable characters`). It is intermittent and load-correlated: the `--no-fail` note only fires when the span report fails (e.g. under intake throttling).

These diagnostics belong on **stderr**, consistent with the adjacent error paths in the same files that already use `this.context.stderr.write`. The `datadog-ci v<version>` banner has the same problem — it's printed to stdout on every invocation, so it can corrupt the same captures — and on top of that it's noisy in CI where the CLI runs many times.

Resolves #2381.

### How?

Route three trace diagnostics from stdout → stderr:

- the `--no-fail` note (`trace.ts`) — was `console.log`, now `this.context.stderr.write` (with the trailing `\n` that `console.log` implied)
- the unsupported-CI-provider message (`helper.ts`, shared `CustomSpanCommand`)
- the missing-API-key message (`helper.ts`, shared `CustomSpanCommand`)

The `helper.ts` changes also harden the `span` command, which shares the `CustomSpanCommand` base. No change to stdio inheritance, the `childProcess.stderr` pipe, or the span/reporting logic.

Also route the **version banner** to stderr (`cli.ts`), mirroring the resolved command's `--log-format` so it stays a JSON line in JSON mode; builtin commands and parse errors fall back to text format. And add **`DD_CI_SKIP_VERSION_BANNER`** (`1`/`true`) to suppress the banner entirely, following the existing `DD_CI_`-prefixed toggle convention (e.g. `DD_CI_BYPASS_SITE_VALIDATION`).

Adds regression tests: the `--no-fail` note lands on stderr (not stdout), the shared CI-provider assertion is updated to match, and the banner is suppressed when `DD_CI_SKIP_VERSION_BANNER` is `1`/`true`.

**Judgment call left out (happy to add if preferred):** the `--dry-run` payload write in `helper.ts` is also capturable, but moving it would change `span --dry-run | jq` behaviour and is lower risk (only under `--dry-run`), so it's left on stdout.

### Review checklist

### DIFF
--- a/packages/base/src/commands/trace/trace.ts
+++ b/packages/base/src/commands/trace/trace.ts
@@ -55,7 +55,10 @@ export class TraceCommand extends CustomSpanCommand {
       stdio: ['inherit', 'inherit', 'pipe'],
     })
     const chunks: Buffer[] = []
-    childProcess.stderr.pipe(this.context.stderr)
+    // `{end: false}`: don't end `context.stderr` when the child closes, so the post-child
+    // `--no-fail` diagnostic below can still be written without a "write after end" error
+    // on regular Writable streams (`process.stderr` tolerates it, but others don't).
+    childProcess.stderr.pipe(this.context.stderr, {end: false})
     const stderrCatcher: Promise<string> = new Promise((resolve, reject) => {
       childProcess.stderr.on('data', (chunk) => chunks.push(Buffer.from(chunk)))
       childProcess.stderr.on('error', (err) => reject(err))
@@ -85,7 +88,7 @@ export class TraceCommand extends CustomSpanCommand {
 
     if (res !== 0) {
       if (this.noFail) {
-        console.log('note: Not failing since --no-fail provided')
+        this.context.stderr.write('note: Not failing since --no-fail provided\n')
 
         return exitCode
       }
--- a/packages/datadog-ci/src/cli.ts
+++ b/packages/datadog-ci/src/cli.ts
@@ -78,10 +78,11 @@ if (require.main === module) {
     // Parse error: defer to cli.run below for clipanion's error rendering.
   }
 
-  // Route the banner through the resolved command's logger so it respects `--log-format`.
-  // Builtin commands (e.g. --help) and command line parse errors fall back to a default text logger.
-  const versionLogger =
-    command instanceof BaseCommand ? command.logger : new Logger((s) => context.stdout.write(s), LogLevel.INFO)
+  // Write the banner to stderr so it never corrupts captured stdout (e.g. `VAR=$(datadog-ci ...)`),
+  // while mirroring the resolved command's `--log-format` so it stays a JSON line in JSON mode.
+  // Builtin commands (e.g. --help) and command line parse errors fall back to text format.
+  const jsonOutput = command instanceof BaseCommand ? command.logger.isJsonOutput() : false
+  const versionLogger = new Logger((s) => context.stderr.write(s), LogLevel.INFO, {jsonOutput})
 
   printVersion(versionLogger)
 
--- a/packages/base/src/__tests__/version.test.ts
+++ b/packages/base/src/__tests__/version.test.ts
@@ -38,4 +38,21 @@ describe('printVersion', () => {
       process.argv = original
     }
   })
+
+  it.each(['1', 'true', 'TRUE'])('does not print when DD_CI_SKIP_VERSION_BANNER is %p', (value) => {
+    const original = process.env.DD_CI_SKIP_VERSION_BANNER
+    process.env.DD_CI_SKIP_VERSION_BANNER = value
+
+    try {
+      const {lines, logger} = collect(false)
+      printVersion(logger)
+      expect(lines).toHaveLength(0)
+    } finally {
+      if (original === undefined) {
+        delete process.env.DD_CI_SKIP_VERSION_BANNER
+      } else {
+        process.env.DD_CI_SKIP_VERSION_BANNER = original
+      }
+    }
+  })
 })
--- a/packages/base/src/commands/span/__tests__/span.test.ts
+++ b/packages/base/src/commands/span/__tests__/span.test.ts
@@ -30,8 +30,12 @@ describe('span', () => {
 
     test('no-name', async () => {
       const env = {GITLAB_CI: '1'}
-      const {code} = await runCLI(['--duration', '10000'], env)
+      const {context, code} = await runCLI(['--duration', '10000'], env)
       expect(code).toBe(1)
+      // Validation errors are diagnostics: they belong on stderr, not stdout (which must stay
+      // clean for `$(datadog-ci trace span ...)` captures).
+      expect(context.stderr.toString()).toContain('The span name must be provided.')
+      expect(context.stdout.toString()).toBe('')
     })
 
     test('no-time-info', async () => {
--- a/packages/base/src/commands/span/span.ts
+++ b/packages/base/src/commands/span/span.ts
@@ -45,7 +45,7 @@ export class SpanCommand extends CustomSpanCommand {
     this.tryEnableFips()
 
     if (!this.name) {
-      this.context.stdout.write(`The span name must be provided.\n`)
+      this.context.stderr.write(`The span name must be provided.\n`)
 
       return 1
     }
@@ -55,7 +55,7 @@ export class SpanCommand extends CustomSpanCommand {
       (!this.startTimeInMs && this.endTimeInMs) ||
       (this.durationInMs && (this.startTimeInMs || this.endTimeInMs))
     ) {
-      this.context.stdout.write(`Either duration or start and end time must be provided.\n`)
+      this.context.stderr.write(`Either duration or start and end time must be provided.\n`)
 
       return 1
     }
@@ -65,13 +65,13 @@ export class SpanCommand extends CustomSpanCommand {
     }
 
     if (!this.durationInMs) {
-      this.context.stdout.write(`The span duration must be provided or start-time and end-time.\n`)
+      this.context.stderr.write(`The span duration must be provided or start-time and end-time.\n`)
 
       return 1
     }
 
     if (this.durationInMs < 0) {
-      this.context.stdout.write(`The span duration must be positive / end time must be after start time.\n`)
+      this.context.stderr.write(`The span duration must be positive / end time must be after start time.\n`)
 
       return 1
     }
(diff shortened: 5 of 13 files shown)


juan-fernandez (commented):
Leaving one inline review comment for the stream lifetime issue.

packages/base/src/commands/trace/trace.ts
```diff
@@ -85,7 +85,7 @@ export class TraceCommand extends CustomSpanCommand {
 
     if (res !== 0) {
       if (this.noFail) {
-        console.log('note: Not failing since --no-fail provided')
+        this.context.stderr.write('note: Not failing since --no-fail provided\n')
```
juan-fernandez, line 99: [P2] This can write to an ended `stderr` stream. `childProcess.stderr.pipe(this.context.stderr)` uses the default `{end: true}`, so once the wrapped command exits it ends the destination stream. `process.stderr` tolerates a later write, but regular `Writable` streams used by programmatic Clipanion callers or tests will emit `write after end` when this diagnostic is written afterward. The test currently masks that by overriding `stderr.end`; the command should instead pipe with `{end: false}` before moving post-child diagnostics onto stderr.

juan-fernandez (commented):
thanks for the contribution! This makes a lot of sense. 

A few non-blocking suggestions that could make this safer and easier to maintain:

- Add a CLI-level regression test for the banner stream. The new tests cover `printVersion(logger)`, but the customer-facing behavior lives in `packages/datadog-ci/src/cli.ts`: the automatic banner should go to `stderr`, explicit `datadog-ci version` / `--version` should still go to `stdout`, and command stdout should stay clean for captures/pipes. A small integration-style test would protect the exact bug class this PR is fixing.

- Consider using the existing `toBoolean()` helper for `DD_CI_SKIP_VERSION_BANNER`. Right now the new env var accepts exactly `1` or lowercase `true`; most env parsing in this repo uses `toBoolean()`, which handles casing consistently.

- Consider preserving `printVersion` API compatibility. `printVersion` is exported from `@datadog/datadog-ci-base/version` and re-exported by `@datadog/datadog-ci`; changing it from `printVersion()` to `printVersion(logger)` could break external callers that import it directly. If that API surface matters, making the logger optional would reduce the blast radius.

- If `DD_CI_SKIP_VERSION_BANNER` is intended for customers rather than just as an escape hatch, it may be worth documenting it somewhere discoverable.

Drarig29:
Hi @eliebleton-manomano! Thanks for the contribution 🙏 

I just ran the CI, and there is a tiny fix to do on the standalone tests. Also, can you rebase on the `master` branch, we recently fixed a flaky test.

```
  ● standalone binary › synthetics › plugin can be loaded

    expect(received).toStrictEqual(expected) // deep equality

    - Expected  - 1
    + Received  + 1

      Object {
        "exitCode": 0,
    -   "stderr": "",
    +   "stderr": "datadog-ci v5.19.0",
        "stdout": StringContaining "All plugins are already baked into the standalone binary.",
      }

      152 |     it('plugin can be loaded', async () => {
      153 |       const result = await execBinary(['plugin', 'check', 'synthetics', 'run-tests'])
    > 154 |       expect(result).toStrictEqual({
          |                      ^
      155 |         exitCode: 0,
      156 |         stdout: expect.stringContaining('All plugins are already baked into the standalone binary.') as string,
      157 |         stderr: '',

      at standalone-e2e/standalone-smoke.test.ts:154:22
      at fulfilled (standalone-e2e/standalone-smoke.test.ts:5:58)
```

packages/datadog-ci/src/cli.ts
```diff
@@ -78,10 +78,11 @@ if (require.main === module) {
...
 
-  // Route the banner through the resolved command's logger so it respects `--log-format`.
-  // Builtin commands (e.g. --help) and command line parse errors fall back to a default text logger.
-  const versionLogger =
-    command instanceof BaseCommand ? command.logger : new Logger((s) => context.stdout.write(s), LogLevel.INFO)
+  // Write the banner to stderr so it never corrupts captured stdout (e.g. `VAR=$(datadog-ci ...)`),
+  // while mirroring the resolved command's `--log-format` so it stays a JSON line in JSON mode.
+  // Builtin commands (e.g. --help) and command line parse errors fall back to text format.
+  const jsonOutput = command instanceof BaseCommand ? command.logger.isJsonOutput() : false
+  const versionLogger = new Logger((s) => context.stderr.write(s), LogLevel.INFO, {jsonOutput})
```
rodrigo-roca, line 85: is there unit tests for this piece of new logic? If not can we add some?

rodrigo-roca (commented):
there is also the `span` subcommand which reuses most of the logic, and still use  context.stdout.write if I recall correctly. I think we are missing changes in that command, can you take a look at `span.ts`?

eliebleton-manomano:
Most of the concerns in Juan's comments have been addressed in the latest version. I'll work through the rest in short order.

>     * Consider preserving `printVersion` API compatibility. `printVersion` is exported from `@datadog/datadog-ci-base/version` and re-exported by `@datadog/datadog-ci`; changing it from `printVersion()` to `printVersion(logger)` could break external callers that import it directly. If that API surface matters, making the logger optional would reduce the blast radius.

The printVersion(logger) signature landed on master with the --log-format json feature (426c5480) — this PR doesn't change it. The only change here is an internal early-return for DD_CI_SKIP_VERSION_BANNER, so there's no API-surface change to guard against.

packages/base/src/commands/trace/trace.ts
eliebleton-manomano: Good catch. I believe this is now fixed and the test was adjusted accordingly.

packages/datadog-ci/src/cli.ts
eliebleton-manomano: Added `packages/datadog-ci/src/__tests__/version-banner.test.ts` — it spawns the real CLI entry point and asserts this logic end-to-end: the banner goes to stderr (never stdout, which is the bug class this fixes), `--log-format json` is mirrored as a JSON line, `version`/`--version` still print to stdout, and `DD_CI_SKIP_VERSION_BANNER` suppresses it.

An in-process `cli.run()` test wouldn't cover it, since the banner is emitted under `require.main === module`.

-- Claudelie

eliebleton-manomano:
Thanks for running the CI! 🙏

Fixed the standalone smoke test: `plugin check synthetics run-tests` now expects the `datadog-ci v<version>` banner on stderr. The banner intentionally moved off stdout (so it can't corrupt captures), and `run-tests` isn't a skip-listed arg like `--help`/`version`, so it now lands on stderr — the other smoke cases use `--help`/`version` and stay `stderr: ''`. That was the assertion behind the failing standalone jobs.

Also rebased onto the latest `master` to pick up the flaky-test fix.

-- Claudelie

eliebleton-manomano:
One more consistency fix for the `trace span` subcommand (shares the `CustomSpanCommand` base).

The inherited diagnostics (unsupported-CI-provider, missing-API-key) were already moved to stderr in this PR. But `span` also had its own four argument-validation messages — missing `--name`, conflicting/missing/negative duration — still written to `stdout` while returning a non-zero exit code. That's the same diagnostics-on-stdout issue this PR is fixing elsewhere (and `trace`'s analogous `"Missing command to run"` was already on stderr), so they could leak into a `$(datadog-ci trace span …)` capture.

Moved those four to `stderr` and added a regression assertion (the existing validation tests only checked the exit code, not the stream). The `[DRY-RUN]` payload stays on stdout, since that's the command's intended output.

-- Claudelie

Drarig29 (approved):
LGTM

juan-fernandez (approved):
awesome! Thanks!

PR #2380, opened 2026-06-25
feat(trace): add `--no-capture` flag to omit command arguments from reported spans

### LINKED ISSUE #2379: trace: add --no-capture flag to keep command arguments out of reported spans

`datadog-ci trace -- <cmd>` always reports the **full command line** — including every argument — in the span's `command` field, and uses that same string as the default span `name`.

When the wrapped command takes secrets as arguments (auth tokens, passwords, signed URLs, API keys), those values are sent to Datadog and become visible in the CI Visibility UI and in anyone's access to the trace. Example:

```bash
datadog-ci trace -- ./deploy.sh --token "$SECRET"
# reported command/name: "./deploy.sh --token <SECRET>"
```

There is currently no way to trace a command while keeping its arguments out of the reported span.

### Solution

Add a `--no-capture` boolean flag (default `false`) to the `trace` command. When set, only the **executable name** is reported in the `command` field instead of the full command line. The child process is still launched with all of its arguments — only the reported span is trimmed. When `--name` is not provided, the span name also falls back to the executable name so arguments don't leak through the default name; an explicit `--name` is still honored.

```bash
datadog-ci trace --no-capture -- ./deploy.sh --token "$SECRET"
# reported command/name: "./deploy.sh"
```

### Additional context

Alternatives considered: redacting via a pattern/denylist of argument names (more flexible but fragile and hard to get right), or env-var opt-in. A simple all-or-nothing `--no-capture` flag covers the common "these args are sensitive, don't record them" case with no room for partial leaks.

Scope note: `--no-capture` only affects the reported `command`/`name`. User-supplied `--tags`, `--measures`, and `DD_TAGS` are still reported as-is, since the user controls those explicitly.

eliebleton-manomano:
Closes #2379

### What and why?

`datadog-ci trace -- <cmd>` always reports the full command line, including every argument, in the span's `command` field (and uses it as the default `name`). When arguments carry secrets (tokens, passwords, signed URLs), those values are sent to Datadog and surface in the CI Visibility UI.

```bash
datadog-ci trace -- ./deploy.sh --token "$SECRET"
# reported command/name: "./deploy.sh --token <SECRET>"
```

### How?

Add `--no-capture` (default: false): when set, only the executable name is reported instead of the full command line. The child process is still launched with all of its arguments — only what we report is trimmed. When `--name` is not provided, the span name also falls back to the executable name so arguments don't leak via the default name; an explicit `--name` is still honored.

```bash
datadog-ci trace --no-capture -- ./deploy.sh --token "$SECRET"
# reported command/name: "./deploy.sh"
```

Scope: `--no-capture` only affects the reported `command`/`name`. User-supplied `--tags`, `--measures`, and `DD_TAGS` are reported as-is, since the user controls those explicitly.

Adds unit tests for all three behaviours (default captures the full line, `--no-capture` trims to the executable, `--no-capture` still honors `--name`) and documents the flag in the command README and `--help` examples.

### Review checklist


janine-c (approved):

juan-fernandez (approved):
thanks for this!

Drarig29 (approved):

PR #2388, opened 2026-06-26
feat(tag): comma-separated `--level` (e.g. `--level pipeline,job`)

### LINKED ISSUE #2387: feat(tag/measure): additive comma-separated `--level` (e.g. `--level pipeline,job`)
## Summary

Make `--level` on `datadog-ci tag` (and `datadog-ci measure`) **additive**: accept a comma-separated list so the same tags/measures are applied to several spans in one invocation, e.g. `--level pipeline,job`.

Today `--level` accepts exactly one of `pipeline | job | stage | step`, so applying the same tag to multiple span levels requires multiple invocations.

## Motivation

Tags do not propagate between spans, and the level they belong on isn't always singular:

- Some tags are useful at **every** level (e.g. `team`) — today you must call `tag` once per level.
- Some tags should "flow down" to a few levels (e.g. `job_maintainer` belongs on both the `job` and the `step` span).

A comma-separated `--level` lets users express this directly instead of scripting N invocations.

## Proposed behavior

- `--level` accepts one or more of `[pipeline, job, stage, step]` as a comma-separated list (e.g. `--level pipeline,job`).
- The full tag/measure set is sent to **each** listed span — one request per level.
- Duplicate levels are ignored; order is irrelevant (levels are processed by span depth).
- Validation is fail-fast: if any member is unknown or unsupported for the current provider, the command errors before sending anything.
- A single value (`--level pipeline`) and all existing usage are **unchanged**.
- `--no-fail` semantics are **unchanged** (narrow): it tolerates a failed submission request, but setup/config errors (missing API key, unsupported provider, GitHub diagnostic logs that can't be located) still exit non-zero. Per-level failures are reported with the offending level.

```bash
# apply team to both the pipeline and job spans in one call
datadog-ci tag --level pipeline,job --tags team:backend

# job_maintainer on both job and step
datadog-ci tag --level job,step --tags job_maintainer:alice
```

## Scope

- `tag` and `measure` (they share the level machinery).
- README + CLI help updated for both.

## Implementation notes

- The per-level loop is shared between the two commands.
- A per-execution cache avoids re-globbing and re-reading the GitHub Actions diagnostic logs once per level when several GitHub levels are requested (e.g. `--level job,step`); it is scoped to a single command run, never module-global.
- No behavioral change for the existing single-level path or for `--no-fail`.

I have a working implementation with extensive tests and would be happy to open a PR if this is something the team would accept.

-- Claudelie

eliebleton-manomano:
### What and why

Makes `--level` on `datadog-ci tag` **additive**: it now accepts a comma-separated list, so the same tags can be applied to several spans in one invocation (e.g. `--level pipeline,job`).

Tags don't propagate between spans, and the level they belong on isn't always singular — e.g. `team` is useful at every level, and `job_maintainer` belongs on both `job` and `step`. Today that requires N separate invocations.

Closes #2387.

```bash
# team on both the pipeline and job spans, one call
datadog-ci tag --level pipeline,job --tags team:backend
# job_maintainer on both job and step
datadog-ci tag --level job,step --tags job_maintainer:alice
```

### Behavior

- `--level` accepts one or more of `[pipeline, job, stage, step]` as a comma-separated list.
- The full tag set is sent to **each** listed span — one request per level.
- Duplicate levels are ignored; order is irrelevant (levels are processed by span depth).
- Validation is fail-fast: an unknown or provider-unsupported member errors before anything is sent.
- **Single-level usage is unchanged**, and so is `--no-fail`: it still tolerates a failed submission request but not setup/config errors (missing API key, unsupported provider, GitHub diagnostic logs that can't be located), which exit non-zero with the offending level named.

### Implementation

- A shared `processLevels()` helper drives the per-level loop for both commands; each level is enriched on its own copy of `ciEnv` so e.g. the step index never leaks into the job request.
- A per-execution `GithubLogCache` locates and reads the GitHub Actions diagnostic logs at most once even when several levels need them (e.g. `--level job,step`). It is scoped to a single command run, never module-global, so it can't leak across commands or tests.
- `tag` READMEs and CLI help updated.

### Tests

- `parseLevels` unit tests (single/multi/sort/dedupe/whitespace/empty/unknown/unsupported).
- `processLevels` tests: per-execution caching (logs read once across `job,step`), orchestration-id selection cached, and fatal-error propagation annotated with the level.
- Command tests for both `tag`: one request per level, dedupe, fail-fast, GitHub `job,step` enrichment, and `--no-fail` semantics.

`yarn build`, `yarn lint`, and the affected test suites pass.

-- Claudelie


jeff-morgan-dd (approved):

packages/base/src/commands/tag/README.md
```diff
@@ -14,10 +14,20 @@ For example:
 datadog-ci tag --level job --tags "go.version:`go version`"
 ```
 
-- `--level` Has to be one of `[pipeline, job, stage, step]`. It will determine in what span the tags will be added.
+- `--level` is one or more of `[pipeline, job, stage, step]`. It determines in what span(s) the tags will be added.
```
Drarig29: Now that `--level` can take multiple values, users could be tempted to do something like `--title '[stage, step]'`, it may be confusing. I think `{...}` reads more like a set of values.
```suggestion
- `--level` is a comma-separated list of `{pipeline,job,stage,step}`. It determines in what span(s) the tags will be added.
```

packages/base/src/commands/tag/README.md
```diff
@@ -14,10 +14,20 @@ For example:
...
   If `pipeline` is selected then the tags will be added to the pipeline trace span. If `job` is selected it will be added
```
Drarig29, line 19: **nit** Let's move this to a new paragraph
```suggestion

  If `pipeline` is selected then the tags will be added to the pipeline trace span. If `job` is selected it will be added
```

packages/base/src/commands/tag/README.md
```diff
@@ -14,10 +14,20 @@ For example:
...
   to the span for the currently running job. If `stage` is selected it will be added to the span for the currently
   running stage. If `step` is selected it will be added to the span for the currently running step.
+
+  The value is **additive**: pass a comma-separated list to add the same tag set to several spans at once, e.g.
```
Drarig29: **nit** let's emphasize it's the multiple **levels** that are interesting
```suggestion
    Pass a comma-separated list to add the same tags to multiple levels at once, e.g.
```

Drarig29 (commented):
LGTM for `tag`, just a few nits.

But for `measure`, can you really think of a measure that would be interesting to set with the **same value** at multiple levels? It sounds like it's more of a feature for `tag` only 🤔 

Numbers are meant to be aggregated, so I would expect this:
- Pipeline: `binary.size:1000`
  - Job 1: `binary.size:600`
  - Job 2: `binary.size:400`

And since we market [measures as "numerical tags"](https://docs.datadoghq.com/continuous_integration/pipelines/custom_tags_and_measures/#add-measures-to-pipeline-traces), I'd assume we overwrite the values instead of appending. I'll check with the team.

But long story short, it looks like you are only interested for the `tag` command, so I'd say it's safer to split the effort... because it's indeed a low-hanging fruit on the `tag` side.

Drarig29:
Also, I ran the CI and it's failing. You can run the tests locally with `yarn test`. And there is now a conflicting file.

Thanks again for the effort 🙏

eliebleton-manomano:
Thanks for the thorough review, @Drarig29!

- Applied all 3 README nits on `tag`: `{pipeline,job,stage,step}` set notation, moved the level-selection explanation to its own paragraph, and reworded the additive blurb to emphasize it's multiple **levels** being targeted.
- Dropped multi-level `--level` support from `measure` entirely (reverted to the original single-level behavior/README/tests). Agreed the additive-duplication semantics don't make sense for numeric measures — `tag` keeps the additive comma-separated `--level`, `measure` stays single-level only.
- Also rebased on `master` and fixed the CI failures you flagged (a couple of tests were implicitly relying on the local filesystem not having stray `Worker_*.log` files, which didn't hold on the hosted runners).

-- Claudelie

Drarig29 (approved):

rodrigo-roca (approved):

PR #2396, opened 2026-07-03
docs(trace): surface the `trace span` subcommand

eliebleton-manomano:
## What & why

The `trace span` subcommand — report a *standalone* custom span with a name and a duration (or start/end time), without wrapping a command — ships today and has its own [README](/packages/base/src/commands/span/README.md), and it's covered on the [official docs page](https://docs.datadoghq.com/continuous_integration/pipelines/custom_commands/). But it isn't discoverable from the repo's front-door README: the `trace` section only describes the command-wrapping form (`trace ... -- <cmd>`) and never mentions or links `span`. It's easy to forget the subcommand exists.

## Changes

- Root README: list `span` under the `trace` section, linking to its README.
- `trace` README: add a one-line cross-reference to the `trace span` subcommand.
- `span` README: spell out that `--start-time` / `--end-time` epoch milliseconds are UTC (they're serialized via `toISOString()`), so there's no timezone ambiguity for readers.

Docs only — no code or behavior change.

## Note on parameters

I checked the full option surface (`--name`, `--duration`, `--start-time`, `--end-time`, `--tags`, `--measures`, `--dry-run`) — all already documented in the span README. The only inherited-but-undocumented flags are `--fips` / `--fips-ignore-error`, but no command README in the repo documents those, so I left them out to stay consistent rather than diverge here.

-- Claudelie

### DIFF
--- a/packages/base/src/commands/span/README.md
+++ b/packages/base/src/commands/span/README.md
@@ -15,8 +15,8 @@ datadog-ci trace span --name "Say Hello" --duration 800 --tags responded-hello-t
 ```
 
 - `--name` is a human-readable name for the reported span.
-- `--start-time` the span start time in milliseconds since the unix epoch.
-- `--end-time` the span end time in milliseconds since the unix epoch.
+- `--start-time` the span start time in milliseconds since the unix epoch (UTC).
+- `--end-time` the span end time in milliseconds since the unix epoch (UTC).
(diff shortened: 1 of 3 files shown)


Drarig29 (approved):
Good point, LGTM. I just made the front README a bit less verbose

PR #2411, opened 2026-07-27
Improve git metadata API error messages

juan-fernandez:
### What and why?

Git metadata upload failures can surface generic errors that omit the failing endpoint and useful backend details. For example, an unexpected `/search_commits` response was rendered as `Invalid API response: [object Object]`, while request timeouts and `/packfile` failures also lacked consistent context.

### How?

Format errors at the shared GitDB request boundary used by both `/search_commits` and `/packfile`:

- Include the endpoint name for request failures and client-side timeouts.
- Include the HTTP status, status text, and serialized response body when a response is available.
- Include the same details in retry warnings and final errors.
- Keep malformed successful `/search_commits` responses as a separate validation error.
- Omit request configuration and headers so API credentials cannot appear in logs.

Add regression coverage for a retried `/search_commits` timeout, a `/packfile` HTTP error response, and a malformed successful `/search_commits` response.

Validated with:

- `yarn build`
- `yarn lint`
- `yarn test packages/base/src/commands/git-metadata/__tests__/gitdb.test.ts --runInBand --watchman=false`

### Review checklist

### DIFF
--- a/packages/base/src/commands/git-metadata/__tests__/gitdb.test.ts
+++ b/packages/base/src/commands/git-metadata/__tests__/gitdb.test.ts
@@ -1412,6 +1412,85 @@ describe('gitdb', () => {
     mocks.expectCalls()
   })
 
+  test('includes response details when the search commits response is invalid', async () => {
+    const mocks = new MockAll({
+      getConfig: [
+        {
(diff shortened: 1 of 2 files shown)


ava-silver (approved):
lgtm for admins

PR #2412, opened 2026-07-27
[SDCT-783] Add ci.pipeline.display_name tag for Buildkite

gbrandrea:
## What

  Adds a new, additive `ci.pipeline.display_name` tag for Buildkite, sourced from `BUILDKITE_PIPELINE_NAME` (the real pipeline name), alongside the existing `ci.pipeline.name` tag.

  ## Why

  `ci.pipeline.name` currently carries the Buildkite pipeline **slug** (`BUILDKITE_PIPELINE_SLUG`), not the real pipeline name. We can't change `ci.pipeline.name`'s value without a breaking change, since the backend computes pipeline fingerprinting from it today. This PR adds the real name as a new, separate tag instead. A future major-version release will eventually swap `ci.pipeline.name` over to the real name and retire this tag — that's out of scope here.

  Scope is intentionally limited to Buildkite only; no other CI provider is touched, and `ci.pipeline.name`'s source is untouched.


Aaron-9900 (approved):

packages/base/src/helpers/interfaces.ts
```diff
@@ -45,6 +46,7 @@ export interface Metadata {
     pipeline: {
       id?: string
       name?: string
+      display_name?: string
```
juan-fernandez: is this consumed by our backend somehow? I remember issues with incompatible reported data types

juan-fernandez (approved):
LGTM but let's confirm the new `display_name` in `Metadata` is not an issue

packages/base/src/helpers/interfaces.ts
juan-fernandez: the comment should rather be in this line, which is the runtime change: https://github.com/DataDog/datadog-ci/pull/2412/changes#diff-dc1fcf04816c2d334efaa97fa563c1431218db229a8e620d4023b5925304bafeR1032

juan-fernandez (changes requested):
so this would indeed break synthetics API. Blocking for the moment

packages/base/src/helpers/ci.ts
```diff
@@ -1026,6 +1029,7 @@ export const getCIMetadata = (tagSizeLimits?: {[key in keyof SpanTags]?: number}
       pipeline: removeUndefinedValues({
         id: tags[CI_PIPELINE_ID],
         name: tags[CI_PIPELINE_NAME],
+        display_name: tags[CI_PIPELINE_DISPLAY_NAME],
```
Drarig29: Let's remove this line (maybe keep a comment), and remove the property from `Metadata`. Indeed, `getCIMetadata` is only used by Synthetics and we have our own validation schema, so this could break API calls for Synthetics users using Buildkite

gbrandrea: Done! Will this work for you now?

Drarig29 (approved):
LGTM

packages/base/src/helpers/ci.ts
Drarig29: yep!

juan-fernandez (approved):
thanks!

PR #2420, opened 2026-07-31
coverage: upload the coverage config and CODEOWNERS as attachments

### LINKED ISSUE #2414: coverage: read code-coverage.datadog.yml from the working directory, and warn when no config is found
## Feature request description

`datadog-ci coverage upload` only picks up `code-coverage.datadog.yml` if it's committed. A config generated during the CI run is silently ignored without any warning and the coverage numbers just come back without any of our rules applied.

The lookup goes through git, not the filesystem. In [packages/plugin-coverage/src/commands/upload.ts](https://github.com/DataDog/datadog-ci/blob/master/packages/plugin-coverage/src/commands/upload.ts#L250):

```ts
const coverageConfig = await this.getRepoFile(COVERAGE_CONFIG_PATHS, reportedCommit)

private async getRepoFile(possiblePaths: string[], ref: string | undefined) {
  if (!this.git || !ref) return undefined
  for (const path of possiblePaths) {
    try {
      const sha = await getGitFileHash(this.git, path, ref)
      if (sha) return {path, sha}
    } catch (e) {
      this.logger.debug(`Error while trying to get repo file ${path} details at ${ref}: ${e}`)
    }
  }
  return undefined
}
```

and `getGitFileHash` in [packages/base/src/commands/git-metadata/git.ts](https://github.com/DataDog/datadog-ci/blob/master/packages/base/src/commands/git-metadata/git.ts#L109):

```ts
export const getGitFileHash = async (git, path, ref = 'HEAD'): Promise<string> => {
  return git.revparse([`${ref}:${path}`])
}
```

Two things make this hard to debug:

1. It fails silently - `git rev-parse <sha>:code-coverage.datadog.yml` resolves the blob in the committed tree, so an untracked (or even staged-but-uncommitted) file throws, the catch swallows it into `logger.debug`, and `getRepoFile` returns `undefined`. At default verbosity there's nothing in the output to tell you the config wasn't applied.
2. **The docs don't mention the constraint.** [Code Coverage Configuration](https://docs.datadoghq.com/code_coverage/configuration/) simply says to create the file "in the root of your repository" but doesn't specify it needs to be explicitly committed.

### Why this matters to us

We maintain a shared CircleCI orb used by a fleet of Python repos. We want to leverage one ignore list to produce the Datadog config at build time, so the repos stay consistent without each committing (and drifting) their own copy. That's not possible today. We are not willing to leverage the UI here as we want our config to live in the orbs source code.

## Solution

Any one of these would unblock us, in rough order of preference:

1. Fall back to the working directory when the git lookup misses - read `code-coverage.datadog.yml` from `basePath`/cwd and send its contents.
2. Add a new `--coverage-config <path>` flag that takes an explicit path and uploads the contents, leaving the existing git behaviour as the default.
3. At minimum, log a warning when no config is resolved — something like `No code-coverage config found at <sha>; using global config` + Promoting the `logger.debug` in the catch to a warning. 

## Additional context

- Seen with `datadog-ci` v5.13.1 (`datadog-ci_linux-x64` release binary), CircleCI, producing Cobertura XML.
- Invoked as `datadog-ci coverage upload --skip-git-metadata-upload <format> <dir>`.
- Logs:
```
Uploading coverage to Datadog...
datadog-ci v5.13.1
@datadog/datadog-ci-plugin-coverage v5.13.1
Tue Jul 28 2026 18:02:12 GMT+0000 (UTC) - Starting upload. 
Will look for code coverage report files in ., .
Uploading code coverage report file(s) in ./test-coverage-results/coverage.xml
✅ Uploaded 1 files in 0.153 seconds.
```
- Our ignore file that we are generating on the fly:
```
schema-version: v1
  ignore:
    - "**/test*"
    - "*****ds_jobs/**"
    - "troubleshooters/**"
    - "obprojects/**"
    - "**/venv/**"
    - "**/env/**"
    - "**/.venv/**"
    - "**/site-packages/**"
```
- Workaround for now is to commit the file per repo, which is what we'll do, but it means the ignore list lives in N places instead of one and will invariably drift over time.

I am happy to draft up a PR for this if there is a preference between 1 and 2.

## Command

`coverage`

nikita-tkachenko-datadog:
Fixes #2414

The issue offered three options and asked which was preferred. **This implements all three:** option 1 (fall back to the working directory), option 2 (an explicit `--coverage-config <path>` flag), and option 3 (warn when no config is resolved, and promote the swallowed `logger.debug`). It also goes one step further than any of them: the file's **content** is uploaded as an attachment, so the backend can apply a config that was never committed — options 1 and 2 alone would still have left the backend re-reading the file from the committed tree by path and SHA, which is the actual reason a generated config is ignored today. Thanks to @swarajrao for a very well-researched report; it made this straightforward to act on.

`coverage upload` resolves `code-coverage.datadog.yml` and `CODEOWNERS` through git only (`git rev-parse <ref>:<path>`), so a config generated during the CI run is silently ignored — the lookup error is swallowed into `logger.debug`, and at default verbosity nothing in the output says the config was not applied. This is the CLI half of "coverage config + CODEOWNERS as EVP attachments"; the backend services read the attachments in separate PRs.

### What changed

**Both files are now looked up in the committed tree *and* in the working directory**, and their content is uploaded as EVP attachments alongside the coverage reports whenever it can be read:

| form field | filename |
|---|---|
| `coverage_config` | `coverage_config.yml.gz` |
| `codeowners` | `codeowners.gz` |

The on-disk copy wins over the committed blob — regenerating the config at build time is exactly the use case in the issue. When content is attached, `config.sha` / `codeowners.sha` become the **git blob SHA-1 of the attached bytes** (`sha1("blob " + len + "\0" + content)`), so the value still matches `git hash-object` and the backend's content-addressed cache keeps working whether the file was committed or generated. Search roots for the working-directory lookup, in order: the git repository root, `--base-path`, then the current directory.

Files over **1 MiB uncompressed** are not attached (CODEOWNERS in a large monorepo can be arbitrarily big); the command warns and falls back to today's git-derived path and SHA. No new event tags, no schema change.

**New `--coverage-config <path>` flag** for a config that is generated, or lives outside the default locations. An explicitly requested file that cannot be read is a user error: the command writes a clear message to stderr and **exits non-zero before uploading anything**, git metadata included. Without the flag, discovery of both files stays best-effort and never fails the upload.

**The warnings the issue asks for.** When no config is resolved at all:

```
⚠️ No code coverage configuration found (looked in the files committed at <sha> and on disk in <dirs>);
   the organization-level configuration will be used. Pass --coverage-config <path> to upload a
   configuration that is not committed.
```

and the swallowed `logger.debug` in the git lookup is promoted to `logger.warn` when the file was explicitly requested. What was resolved is now reported at `info`, so the silence the issue complains about is gone:

```
Uploading the code coverage configuration code-coverage.datadog.yml (49 bytes, sha a5005d07...)
Uploading the CODEOWNERS file .github/CODEOWNERS (22 bytes, sha dc2328ea...)
```

`--dry-run` prefixes those with `[DRYRUN]`, so it reports which files would be attached.

**The attachment budget is computed instead of hardcoded.** `MAX_REPORTS_PER_REQUEST = 7` ("backend supports 10 attachments, to keep the logic simple we subtract 3") is replaced by:

```ts
reserved = [prDiff, commitDiff, fileFixes, coverageConfig, codeowners].filter(present).length
maxReportsPerRequest = Math.max(1, MAX_ATTACHMENTS_PER_REQUEST - reserved)
```

All five are resolved before chunking, so the budget is exact rather than an unconditional -3. I verified the real limit against the intake source rather than trusting the comment: `cicovreprt` is configured with `max-attachment-count: 10` and `max-content-size: 50MiB`, and **`event.json` does not count towards the limit** — the public path uses `V2MultiPartMessageDecoder`, which excludes the part named `event` before counting. Exceeding the count returns HTTP 400, which is in `errorCodesStopUpload`, so an over-budget request aborts the whole upload; getting this right matters.

**Chunking now respects the request size limit too.** The same intake rejects a body over 50 MiB with a 413 — and 413 is *not* in `errorCodesStopUpload`, so it is retried and then skipped, dropping that chunk's reports from a run that still prints `Uploaded N files`. Since this change makes requests carry more attachments than before, it also makes chunking size-aware: the budget is what remains of 48 MiB (50 MiB less a margin for `event.json` and the multipart framing) after the diffs, file fixes, config and CODEOWNERS. Report sizes come from disk, so they over-estimate the gzipped bytes actually sent — over-estimating only ever produces smaller chunks, never an oversized request. A single report larger than the budget still gets its own request, and warns. Verified on four 22 MB jacoco reports: previously one request of 88 MB, now two of 44 MB.

### Tests

216 unit tests in `packages/plugin-coverage` (was 152), covering: blob-SHA computation (asserted against `git hash-object`, including empty and binary content), search-root and candidate precedence, on-disk-wins, git-only fallback, the size cap at and just over the boundary, unreadable/missing/directory/dangling-symlink explicit paths, the oversized-explicit fallback to the git lookup, non-git operation, the `debug`→`warn` promotion, every branch of the budget computation, and the two new multipart parts. The count invariant is asserted over the **full cross-product** of the five optional attachments (32 combinations): every chunk satisfies `reports + reserved <= 10` and no report is dropped — an off-by-one there would 400, which aborts the whole upload. `api-multipart.test.ts` encodes a real multipart body rather than mocking `form-data`, pinning `name="coverage_config"; filename="coverage_config.yml.gz"` and `name="codeowners"; filename="codeowners.gz"` as they appear on the wire, since the backend matches on the filename.

Also exercised by hand in a throwaway repo: committed-and-identical (SHA equals `git hash-object`, no behaviour change), modified-on-disk, untracked-generated, missing explicit path (exit 1, nothing uploaded), a 1.32 MB CODEOWNERS, and 12 reports + config + CODEOWNERS in a non-git directory chunking into 8 + 4 — exactly 10 attachments per request.

`yarn lint`, `yarn tsc:build` and `yarn test packages/plugin-coverage` are clean.

### Note

The docs at [Code Coverage Configuration](https://docs.datadoghq.com/code_coverage/configuration/) still say to create the file "in the root of your repository" without mentioning the committed-only constraint (issue item 2). The plugin README is updated here; the public docs live in another repo and need a separate change.

### DIFF
--- a/packages/base/src/commands/coverage/upload.ts
+++ b/packages/base/src/commands/coverage/upload.ts
@@ -37,6 +37,10 @@ export class CoverageUploadCommand extends BaseCommand {
         'Upload all code coverage report files in current directory with extra verbosity',
         'datadog-ci coverage upload --verbose .',
       ],
+      [
+        'Upload all code coverage report files in current directory with a coverage configuration generated at build time',
+        'datadog-ci coverage upload --coverage-config build/code-coverage.datadog.yml .',
+      ],
     ],
   })
 
@@ -64,6 +68,11 @@ export class CoverageUploadCommand extends BaseCommand {
     description: 'The repository URL to retrieve git metadata from',
   })
   protected basePath = Option.String('--base-path', {description: 'The base path to the coverage report files'})
+
+  protected coverageConfigPath = Option.String('--coverage-config', {
+    description:
+      'The path to the code coverage configuration file to upload. Useful when the configuration is generated during the CI run instead of being committed. The command fails if the file cannot be read. When this option is not set, the configuration is looked up in the committed files and in the working directory.',
+  })
 
   protected ignoredPaths = Option.String('--ignored-paths', {
     description:
--- a/packages/plugin-coverage/README.md
+++ b/packages/plugin-coverage/README.md
@@ -31,6 +31,25 @@ datadog-ci coverage upload --flags type:unit-tests --flags jvm-21 unit-tests/cov
 - `--git-repository-url` is a string specifying the repository URL to retrieve git metadata from. If this is missing, the URL is retrieved from the local git repository.
 - `--disable-file-fixes` (default: `false`): disable the generation and upload of file fixes for code coverage.
 - `--file-fixes-search-path` is a string specifying the root directory used to scan source files for file fixes. By default, the repository root is used. This is useful for monorepos or when coverage reports only cover a subset of the codebase.
+- `--coverage-config` is a string specifying the path to the [code coverage configuration][2] file to upload. Use it when the configuration is generated during the CI run instead of being committed, or when it is not at one of the default locations. If the file cannot be read, the command fails without uploading anything.
+
+#### Code coverage configuration and CODEOWNERS
+
+The command looks for the [code coverage configuration][2] and the `CODEOWNERS` file, and uploads their
+content along with the coverage reports. Both lookups are best-effort: a missing file never fails the
+upload, and the command warns when no coverage configuration is found at all.
+
+Each file is looked up in two places:
+
+1. The files committed at the reported commit — `code-coverage.datadog.yml`, then `code-coverage.datadog.yaml`
+   for the configuration, and `.github/CODEOWNERS`, `CODEOWNERS`, then `docs/CODEOWNERS` for code owners.
+2. The working directory, under the repository root, then `--base-path`, then the current directory.
+
+**The working-directory copy wins**, so a configuration generated at build time is used even when it is
+not committed — which is also what `--coverage-config` is for when the generated file is somewhere else.
+A file that is only committed (for example in a sparse checkout) still applies: it is read from the
+repository instead of being uploaded. Files larger than 1 MiB are not uploaded; only their path and
+SHA are sent.
 
 #### Environment variables
 
@@ -66,5 +85,7 @@ Will upload code coverage report file packages/plugin-coverage/src/__tests__/fix
 Additional helpful documentation, links, and articles:
 
 - [Learn about Datadog Code Coverage][1]
+- [Code Coverage configuration][2]
 
 [1]: https://docs.datadoghq.com/code_coverage/
+[2]: https://docs.datadoghq.com/code_coverage/configuration/
--- a/packages/plugin-coverage/src/__tests__/api-multipart.test.ts
+++ b/packages/plugin-coverage/src/__tests__/api-multipart.test.ts
@@ -1,0 +1,113 @@
+// Asserts the bytes that actually go on the wire, with the real `form-data` and `zlib`.
+// The backend matches the coverage config and CODEOWNERS attachments by *filename*, and the intake
+// rejects a filename that does not match its attachment-name rules, so both are pinned here rather
+// than only asserted through a mocked `form.append`.
+
+import {gunzipSync, gzipSync} from 'zlib'
+
+import type {Payload} from '../interfaces'
+import type {RequestConfig} from '@datadog/datadog-ci-base/helpers/request'
+
+import {CODEOWNERS_ATTACHMENT_FILENAME, COVERAGE_CONFIG_ATTACHMENT_FILENAME, uploadCodeCoverageReport} from '../api'
+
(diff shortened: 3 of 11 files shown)


swarajrao (commented):
thank you for making this change! 😄

PR #2454, opened 2026-08-13
[sourcemaps] Add standalone debug ID injection command

jhssilva:
## Summary

- Adds `datadog-ci sourcemaps inject <path>` to inject deterministic Debug IDs into JavaScript bundles and adjust their sourcemaps before deployment.
- Uses the same SHA-256 UUID strategy, runtime snippet, and progressive Debug ID extraction behavior as Datadog build plugins.
- Keeps `sourcemaps upload --debug-id` read-only: it extracts `ddDebugId` from JavaScript and sends it as multipart event metadata.
- Shares bundle/sourcemap discovery between injection and upload so both commands operate on the same artifacts.

## Why

The Debug ID must be available to the browser runtime and included in sourcemap upload metadata. Because inserting the runtime snippet shifts generated JavaScript positions, the corresponding sourcemap must be adjusted so stack frames continue resolving to the correct original source.

The inject command also guarantees that the sourcemap's top-level `debug_id` matches the bundle. Upload matching remains based on `ddDebugId` in JavaScript, so sourcemaps produced directly by Datadog build plugins remain compatible without that field.

### Implementation note

We moved away from `magic-string` because composing its generated adjustment map with the customer's sourcemap required additional remapping dependencies. `webpack-sources` applies changes directly to the mapped source and emits the adjusted JavaScript and sourcemap together.

## How

- `generateDebugId(...)` applies the build-plugins deterministic SHA-256 UUID-v4-shaped strategy to the original JavaScript.
- `injectDebugIdSnippet(...)` preserves hashbangs and directive prologues and registers the ID in `DD_SOURCE_CODE_CONTEXT`.
- Existing bundle IDs are not reinjected; `inject` adds or corrects the sourcemap's `debug_id` when needed.
- Debug ID extraction scans progressively in bounded-memory chunks and stops as soon as it finds the ID.
- Writes are rollback-safe, and `--dry-run` performs complete transformation validation without modifying files.
- Local sourcemap references that resolve outside the requested directory are rejected.
- Original sourcemap extension metadata is preserved. Indexed sourcemaps fail with a clear unsupported message when mapping recomposition is required.

## Usage

```bash
npm run build
datadog-ci sourcemaps inject ./dist
# Regenerate SRI, compression, signatures, or checksums here when applicable.
datadog-ci sourcemaps upload ./dist --debug-id
```

Upload and deploy the exact injected JavaScript and sourcemap files. Run injection before byte-dependent post-processing such as SRI, compression, signatures, or checksum generation.

## Test plan

- [x] TypeScript validation
- [x] Sourcemap command test suites: 76 tests
- [x] ESLint and README usage validation
- [x] Build-plugins Debug ID parity vector and progressive scanner coverage
- [x] Real esbuild position-resolution E2E
- [x] Staging upload, browser error capture, Debug ID deobfuscation, and Git source links
- [x] Runtime registration, build-plugin compatibility, idempotency, dry-run validation, transactional rollback, mixed-batch failure isolation, and upload discovery

### DIFF
--- a/packages/base/package.json
+++ b/packages/base/package.json
@@ -123,7 +123,6 @@
     }
   },
   "dependencies": {
-    "@ampproject/remapping": "^2.3.0",
     "@antfu/install-pkg": "^1.1.0",
     "@types/datadog-metrics": "^0.6.1",
     "async-retry": "^1.3.3",
@@ -140,7 +139,6 @@
     "jest-diff": "^30.2.0",
     "js-yaml": "^4.3.1",
     "jszip": "^3.10.1",
-    "magic-string": "^1.2.0",
     "proxy-agent": "^6.4.0",
     "semver": "^7.6.3",
     "simple-git": "^3.36.0",
@@ -148,16 +146,17 @@
     "tiny-async-pool": "^2.1.0",
     "typanion": "^3.14.0",
     "undici": "^7.29.0",
-    "upath": "^2.0.1"
+    "upath": "^2.0.1",
+    "webpack-sources": "^3.5.1"
   },
   "devDependencies": {
-    "@jridgewell/trace-mapping": "^0.3.31",
     "@types/async-retry": "^1.4.8",
     "@types/debug": "^4.1.12",
     "@types/deep-extend": "^0.4.31",
     "@types/inquirer": "^8.2.6",
     "@types/js-yaml": "^4.0.9",
     "@types/semver": "^7.7.1",
-    "@types/tiny-async-pool": "^2.0.3"
+    "@types/tiny-async-pool": "^2.0.3",
+    "source-map": "^0.7.4"
   }
 }
--- a/packages/base/package.json
+++ b/packages/base/package.json
@@ -123,6 +123,7 @@
     }
   },
   "dependencies": {
+    "@ampproject/remapping": "^2.3.0",
     "@antfu/install-pkg": "^1.1.0",
     "@types/datadog-metrics": "^0.6.1",
     "async-retry": "^1.3.3",
@@ -139,6 +140,7 @@
     "jest-diff": "^30.2.0",
     "js-yaml": "^4.3.1",
     "jszip": "^3.10.1",
+    "magic-string": "^1.2.0",
     "proxy-agent": "^6.4.0",
     "semver": "^7.6.3",
     "simple-git": "^3.36.0",
@@ -149,6 +151,7 @@
     "upath": "^2.0.1"
   },
   "devDependencies": {
+    "@jridgewell/trace-mapping": "^0.3.31",
     "@types/async-retry": "^1.4.8",
     "@types/debug": "^4.1.12",
     "@types/deep-extend": "^0.4.31",
--- a/packages/base/package.json
+++ b/packages/base/package.json
@@ -124,6 +124,7 @@
   },
   "dependencies": {
     "@antfu/install-pkg": "^1.1.0",
+    "@babel/parser": "^7.29.0",
     "@types/datadog-metrics": "^0.6.1",
     "async-retry": "^1.3.3",
     "chalk": "^3.0.0",
--- a/packages/base/src/commands/sourcemaps/debugId.ts
+++ b/packages/base/src/commands/sourcemaps/debugId.ts
@@ -3,6 +3,9 @@ import fs from 'fs'
 
 import type {Sourcemap} from './interfaces'
 import type {Writable} from 'stream'
+import type {RawSourceMap} from 'webpack-sources'
+
+import {ReplaceSource, SourceMapSource} from 'webpack-sources'
 
 const DD_DEBUG_ID_REGEX =
   /["']ddDebugId["']\s*:\s*["']([a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12})["']/
@@ -42,6 +45,7 @@ const buildSnippet = (debugId: string): string =>
 const DEBUG_ID_COMMENT_PREFIX = '//# debugId'
 
 const HASHBANG_REGEX = /^#!.*(?:\r\n|\r|\n)/
+const SOURCE_MAPPING_URL_REGEX = /\/\/[#@] sourceMappingURL=.*(?:\r\n|\r|\n)?$/
 const USE_DIRECTIVE_REGEX =
   /^(?:\s|\/\*[\s\S]*?\*\/|\/\/.*(?:\r\n|\r|\n))*(?<useDirective>"use [^"]*"|'use [^']*');?(?:\r\n|\r|\n)?/
 
@@ -102,29 +106,18 @@ export const generateDebugId = (jsContent: string, sourcemapContent: string): st
 
 const PRE_INJECTION_SOURCE_NAME = 'pre-injection.js'
 
-// `@ampproject/remapping`'s conditional-exports/types shape confuses TS's `node16`
-// module resolution for a dynamic `import()` of a CJS-consumed ESM-only default
-// export, so the callable is re-declared locally rather than relying on its
-// inferred type from the dynamic import.
-type RemappingFn = (
-  input: string | Record<string, unknown>,
-  loader: (file: string) => Record<string, unknown> | undefined
-) => Record<string, unknown>
-
 // The snippet is inserted as early as possible (after an optional hashbang and/or
 // leading "use ...;" directives, which are kept first since they must remain the first
-// statements in the file), and a `//# debugId=` comment is appended at the very end.
-// Built via MagicString (rather than plain string concatenation) so the resulting
-// position map can be composed with the original sourcemap through `remapping`,
-// instead of hand-deriving how many lines/characters the injection shifted.
+// statements in the file), and a `//# debugId=` comment is appended near the end,
+// ahead of a final sourceMappingURL directive when one is present.
+// SourceMapSource starts with the original generated code and its map. ReplaceSource
+// applies insertions directly to that mapped source, preserving existing mappings
+// while shifting their generated positions.
 export const injectDebugIdSnippet = async (
   jsContent: string,
   sourcemapContent: string,
   debugId: string
 ): Promise<{js: string; sourcemap: string}> => {
-  const {default: MagicString} = await import('magic-string')
-  const remapping = (await import('@ampproject/remapping')).default as unknown as RemappingFn
-
   const hashbangMatch = jsContent.match(HASHBANG_REGEX)
   const hashbangPortion = hashbangMatch ? hashbangMatch[0] : ''
   const sourceWithoutHashbang = jsContent.slice(hashbangPortion.length)
@@ -132,28 +125,32 @@ export const injectDebugIdSnippet = async (
   const useDirectives = extractLeadingUseDirectives(sourceWithoutHashbang)
 
   const debugIdComment = `${DEBUG_ID_COMMENT_PREFIX}=${debugId}`
+  const sourceMappingURLMatch = jsContent.match(SOURCE_MAPPING_URL_REGEX)
+  const debugIdCommentPosition = sourceMappingURLMatch?.index ?? jsContent.length
 
-  const ms = new MagicString(jsContent)
-  ms.appendLeft(hashbangPortion.length, `${useDirectives}${buildSnippet(debugId)}\n`)
-  ms.append(`\n${debugIdComment}\n`)
+  const originalSourcemap = JSON.parse(sourcemapContent) as RawSourceMap
+  const source = new SourceMapSource(jsContent, PRE_INJECTION_SOURCE_NAME, originalSourcemap)
+  const injected = new ReplaceSource(source)
+  injected.insert(hashbangPortion.length, `${useDirectives}${buildSnippet(debugId)}\n`)
+  injected.insert(debugIdCommentPosition, `\n${debugIdComment}\n`)
 
-  // hires is required: with per-line (lo-res) mappings, magic-string emits a single
-  // anchor segment at column 0 of each unedited line. Real minified output frequently
-  // has no mapping at column 0 of its packed line (e.g. unmapped bundler/IIFE
-  // boilerplate before the first real token), so that anchor falls in a gap in the
-  // original sourcemap and the whole line's mapping is silently dropped on composition.
-  const adjustmentMap = ms.generateMap({source: PRE_INJECTION_SOURCE_NAME, includeContent: false, hires: true})
-  const originalSourcemap: Record<string, unknown> = JSON.parse(sourcemapContent)
-  const combinedMap = remapping(adjustmentMap as unknown as Record<string, unknown>, (file: string) =>
-    file === PRE_INJECTION_SOURCE_NAME ? originalSourcemap : undefined
-  )
-  combinedMap.debugId = debugId
-  combinedMap.debug_id = debugId
+  const {source: injectedSource, map} = injected.sourceAndMap({columns: true})
+  if (!map) {
+    throw new Error('Failed to generate the adjusted sourcemap')
+  }
+
+  const combinedMap = {...map, debugId, debug_id: debugId}
 
   return {
-    js: ms.toString(),
+    js: injectedSource.toString(),
     sourcemap: JSON.stringify(combinedMap),
   }
+}
+
+export interface InjectionResult {
+  failed: number
+  injected: number
+  skipped: number
 }
 
 /**
@@ -163,13 +160,16 @@ export const injectDebugIdSnippet = async (
  * are left without a debug ID, which `validatePayload` treats as a skip further
  * downstream, instead of aborting the whole batch.
  */
-export const generateAndInjectMissingDebugIds = async (
+export const injectMissingDebugIds = async (
   payloads: Sourcemap[],
   dryRun: boolean,
   stdout: Writable
-): Promise<void> => {
+): Promise<InjectionResult> => {
+  const result: InjectionResult = {failed: 0, injected: 0, skipped: 0}
+
   for (const payload of payloads) {
     if (payload.debugId !== undefined) {
+      result.skipped++
       continue
     }
 
@@ -178,7 +178,9 @@ export const generateAndInjectMissingDebugIds = async (
     try {
       jsContent = fs.readFileSync(payload.minifiedFilePath, 'utf-8')
       sourcemapContent = fs.readFileSync(payload.sourcemapPath, 'utf-8')
-    } catch {
+    } catch (error) {
+      result.failed++
+      stdout.write(`WARN: Failed to read ${payload.minifiedFilePath} or its sourcemap: ${error}\n`)
       continue
     }
 
@@ -187,6 +189,7 @@ export const generateAndInjectMissingDebugIds = async (
 
     if (dryRun) {
       payload.debugId = debugId
+      result.injected++
       stdout.write('Dry run: no files modified.\n')
       continue
     }
@@ -196,8 +199,12 @@ export const generateAndInjectMissingDebugIds = async (
       fs.writeFileSync(payload.minifiedFilePath, js)
       fs.writeFileSync(payload.sourcemapPath, sourcemap)
       payload.debugId = debugId
+      result.injected++
     } catch (error) {
+      result.failed++
       stdout.write(`WARN: Failed to inject debug ID into ${payload.minifiedFilePath}: ${error}\n`)
     }
   }
+
+  return result
 }
--- a/packages/base/src/commands/sourcemaps/debugId.ts
+++ b/packages/base/src/commands/sourcemaps/debugId.ts
@@ -9,16 +9,68 @@ import {ReplaceSource, SourceMapSource} from 'webpack-sources'
 
 const DD_DEBUG_ID_REGEX =
   /["']ddDebugId["']\s*:\s*["']([a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12})["']/
+const DEBUG_ID_COMMENT_REGEX = /\/\/[#@] debugId=([a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12})/
+const DEBUG_ID_VALUE_REGEX = /^[a-f0-9]{8}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{4}-[a-f0-9]{12}$/
+const DEBUG_ID_PREFIX_SCAN_BYTES = 1024 * 1024
+const DEBUG_ID_SUFFIX_SCAN_BYTES = 64 * 1024
+
+const extractDebugIdFromSource = (source: string): string | undefined =>
+  source.match(DD_DEBUG_ID_REGEX)?.[1] ?? source.match(DEBUG_ID_COMMENT_REGEX)?.[1]
 
 export const extractDebugId = (filePath: string): string | undefined => {
   try {
-    const source = fs.readFileSync(filePath, 'utf-8')
+    const fileDescriptor = fs.openSync(filePath, 'r')
+    try {
+      const fileSize = fs.fstatSync(fileDescriptor).size
+      const prefixSize = Math.min(fileSize, DEBUG_ID_PREFIX_SCAN_BYTES)
+      const prefix = Buffer.alloc(prefixSize)
+      const prefixBytesRead = fs.readSync(fileDescriptor, prefix, 0, prefixSize, 0)
+      const prefixDebugId = extractDebugIdFromSource(prefix.toString('utf-8', 0, prefixBytesRead))
+      if (prefixDebugId !== undefined || fileSize <= prefixSize) {
+        return prefixDebugId
+      }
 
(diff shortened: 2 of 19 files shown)


packages/base/src/commands/sourcemaps/debugId.ts
```diff
@@ -31,3 +33,125 @@ export const addDebugIdToPayloads = (payloads: Sourcemap[]): boolean => {
...
+  const sourcemapBytes = Buffer.from(sourcemapContent, 'utf-8')
+
+  const lengthPrefix = (length: number): Buffer => {
+    const buf = Buffer.alloc(8)
+    buf.writeBigUInt64LE(BigInt(length))
+
+    return buf
+  }
+
+  const hash = crypto.createHash('sha1')
```
buranmert: this is an unnecessarily expensive crypto hash, let's use xxHash instead

UPDATE: xxHash will require an extra 3rd party dep, maybe the cost diff between xxHash and sha1 is not worth installing one more 3rd party. let's keep `crypto.createHash` for now as it's the simplest solution for the moment.

packages/base/src/commands/sourcemaps/debugId.ts
```diff
@@ -31,3 +33,147 @@ export const addDebugIdToPayloads = (payloads: Sourcemap[]): boolean => {
...
+  ms.appendLeft(hashbangPortion.length, `${useDirective}${buildSnippet(debugId)}\n`)
+  ms.append(`\n${debugIdComment}\n`)
+
+  // hires is required: with per-line (lo-res) mappings, magic-string emits a single
+  // anchor segment at column 0 of each unedited line. Real minified output frequently
+  // has no mapping at column 0 of its packed line (e.g. unmapped bundler/IIFE
+  // boilerplate before the first real token), so that anchor falls in a gap in the
+  // original sourcemap and the whole line's mapping is silently dropped on composition.
+  const adjustmentMap = ms.generateMap({source: PRE_INJECTION_SOURCE_NAME, includeContent: false, hires: true})
+  const originalSourcemap: Record<string, unknown> = JSON.parse(sourcemapContent)
```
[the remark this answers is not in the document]
jhssilva: Fixed in 68b6fc8c — `generateAndInjectMissingDebugIds` now catches per-payload injection errors, logs a `WARN:`, and leaves `debugId` unset (skipped) instead of throwing. Verified with a regression test and a live staging upload of a mixed valid/malformed batch.

packages/base/src/commands/sourcemaps/debugId.ts
```diff
@@ -31,3 +33,147 @@ export const addDebugIdToPayloads = (payloads: Sourcemap[]): boolean => {
...
+  const hashbangPortion = hashbangMatch ? hashbangMatch[0] : ''
+  const sourceWithoutHashbang = jsContent.slice(hashbangPortion.length)
+
+  const useDirectiveMatch = sourceWithoutHashbang.match(USE_DIRECTIVE_REGEX)
+  const useDirective = useDirectiveMatch?.groups?.useDirective ? `${useDirectiveMatch.groups.useDirective};` : ''
+
+  const debugIdComment = `${DEBUG_ID_COMMENT_PREFIX}=${debugId}`
+
+  const ms = new MagicString(jsContent)
+  ms.appendLeft(hashbangPortion.length, `${useDirective}${buildSnippet(debugId)}\n`)
```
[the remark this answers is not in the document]
jhssilva: Fixed in 68b6fc8c — now loops the directive match to capture and repeat the full leading run of directives (not just the first) ahead of the injected snippet. Verified with a regression test (`"use asm"; "use strict";`) and trace-mapping position resolution.

packages/base/src/commands/sourcemaps/debugId.ts
```diff
@@ -31,3 +33,171 @@ export const addDebugIdToPayloads = (payloads: Sourcemap[]): boolean => {
...
+  const bytes = hash.digest().subarray(0, 16)
+  // eslint-disable-next-line no-bitwise
+  bytes[6] = (bytes[6] & 0x0f) | 0x50
+  // eslint-disable-next-line no-bitwise
+  bytes[8] = (bytes[8] & 0x3f) | 0x80
+
+  const hex = bytes.toString('hex')
+
+  return `${hex.slice(0, 8)}-${hex.slice(8, 12)}-${hex.slice(12, 16)}-${hex.slice(16, 20)}-${hex.slice(20)}`
+}
```
amortemousque: **💬 suggestion:** ‏Could be nice to have a consistent strategy with the injection plugin cf: https://github.com/DataDog/build-plugins/blob/3eb123ebd22be37d5ef624f0ffa455bf13309874/packages/plugins/rum/src/debugId.ts#L11

Why don't we have the same implementation?

amortemousque:
Interesting! I originally wanted to hold this for later, since MFE support will need a build plugin for other features anyway regardless.
A few concerns I don't have answers to yet:
- **Duplicated logic**: Snippet and debug-ID generation now are duplicated in datadog-ci and build-plugins . No single source of truth. Any future change to has to land in both places.
- **Parallel deploy/upload breaks the setup**: If sourcemap upload runs in parallel with (or after) deploy, the deployed bundle never gets the snippet injected, Datadog stores a sourcemap under a debug ID the running code never sends, and un-minification silently fails.
- **Breaks Subresource Integrity**: Mutating the JS after SRI hashes are computed invalidates it. The customer's scripts won't be loaded anymore.
- **Unexpected injection on the fallback path**: If a customer configures their bundler to inject only on specific chunks via build-plugins(deliberately excluding others), this fallback injects unconditionally on anything missing a debug ID — overriding that intent silently.

rtrieu (approved):

jhssilva:
@codex review

packages/base/src/commands/sourcemaps/debugId.ts
```diff
@@ -1,33 +1,347 @@
...
+
+  return directives
+}
+
+// SHA-1 over "js" + 8-byte little-endian length + JS bytes + "sourcemap" + 8-byte
+// little-endian length + sourcemap bytes, truncated to 16 bytes, formatted as a UUID
+// with the version nibble forced to 5 and variant bits forced to RFC4122. NOT the
+// v4-style forcing used by RN's own generateDebugId, which is an unrelated convention
+// specific to RN's MD5 scheme.
+export const generateDebugId = (jsContent: string, sourcemapContent: string): string => {
```
amortemousque: 💬 suggestion: ‏Could be nice to have a consistent strategy with the injection plugin cf: https://github.com/DataDog/build-plugins/blob/3eb123ebd22be37d5ef624f0ffa455bf13309874/packages/plugins/rum/src/debugId.ts#L11

Why don't we have the same implementation?

packages/base/src/commands/sourcemaps/debugId.ts
```diff
@@ -1,33 +1,347 @@
...
+    throw new Error('Sourcemap must be a JSON object')
+  }
+
+  return parsed as ParsedSourcemap
+}
+
+export const extractSourcemapDebugId = (sourcemapContent: string): string | undefined => {
+  const sourcemap = parseSourcemap(sourcemapContent)
+  const debugId = sourcemap.debugId
+  const legacyDebugId = sourcemap.debug_id
```
amortemousque: **❓ question:** ‏Why this legacyDebugId?

packages/base/src/commands/sourcemaps/debugId.ts
```diff
@@ -1,33 +1,347 @@
...
+/** Adds the extracted debug ID to each payload whose minified file contains one. */
+export const addDebugIdToPayloads = (payloads: Sourcemap[]): void => {
+  for (const payload of payloads) {
+    payload.debugId = extractDebugId(payload.minifiedFilePath)
+  }
+}
+
+// The runtime snippet read by the browser RUM SDK (packages/browser-core/src/domain/sourceCodeContext.ts),
+// matching the shape build-plugins already injects via getSourceCodeContextSnippet.ts.
+const buildSnippet = (debugId: string): string =>
```
amortemousque, line 95: **💬 suggestion:** ‏
We could add a cross-reference noting that this snippet mirrors https://github.com/DataDog/build-plugins/blob/c9384d115d53578f220cd5e1f29994acb96a1782/packages/plugins/rum/src/getSourceCodeContextSnippet.ts#L55 
This will make the relationship explicit and remind future maintainers to keep both implementations in sync.

packages/base/src/commands/sourcemaps/debugId.ts
```diff
@@ -1,33 +1,347 @@
...
+  sourcemapContent: string,
+  debugId: string
+): {js: string; sourcemap: string} => {
+  const hashbangMatch = jsContent.match(HASHBANG_REGEX)
+  const hashbangPortion = hashbangMatch ? hashbangMatch[0] : ''
+  const sourceWithoutHashbang = jsContent.slice(hashbangPortion.length)
+
+  const useDirectives = extractLeadingUseDirectives(sourceWithoutHashbang)
+
+  const debugIdComment = `${DEBUG_ID_COMMENT_PREFIX}=${debugId}`
```
amortemousque: **💬 suggestion:** ‏We don't need the debugId comment in our implementation.

packages/base/src/commands/sourcemaps/findSourcemaps.ts
amortemousque, line 1: **🥜 nitpick:** ‏‏ For clarity, since this file exports only one function, could we name it after that function`findSourcemaps`?

packages/base/src/commands/sourcemaps/inject.ts
```diff
@@ -0,0 +1,79 @@
...
+    if (!fs.existsSync(this.basePath)) {
+      this.context.stderr.write(`Path does not exist: ${this.basePath}\n`)
+
+      return 1
+    }
+
+    let discoveryFailures = 0
+    const payloads = await findSourcemaps(
+      this.basePath,
+      20,
```
amortemousque: **❓ question:** ‏This is an option of the upload why not keep it an option here too?

packages/base/src/commands/sourcemaps/upload.ts
```diff
@@ -140,10 +138,8 @@ export class SourcemapsUploadCommand extends BaseCommand {
     const initialTime = Date.now()
     const payloads = await this.getPayloadsToUpload(useGit)
 
-    if (this.debugId && payloads.length > 0 && !addDebugIdToPayloads(payloads)) {
-      this.context.stderr.write(renderNoDebugIdFound())
-
-      return 1
+    if (this.debugId && payloads.length > 0) {
+      addDebugIdToPayloads(payloads)
```
amortemousque: **💬 suggestion:** ‏Since injecting is no longer a fallback, I would keep the old behaviour here if no debug_id our found:
```ts
  if (this.debugId && payloads.length > 0 && !addDebugIdToPayloads(payloads)) {
      this.context.stderr.write(renderNoDebugIdFound())
      return 1
   }
```

packages/base/src/commands/sourcemaps/inject.ts
```diff
@@ -0,0 +1,79 @@
...
+        return new Sourcemap(minifiedFilePath, relativePath, sourcemapPath, relativePath)
+      },
+      (message) => {
+        discoveryFailures++
+        this.context.stdout.write(`WARN: ${message}\n`)
+      }
+    )
+
+    if (payloads.length === 0) {
+      this.context.stdout.write(`No JavaScript sourcemaps found in ${this.basePath}.\n`)
```
amortemousque: **💬 suggestion:** Messages are centralized in `renderer.ts`.

packages/base/src/commands/sourcemaps/debugId.ts
```diff
@@ -1,33 +1,347 @@
...
+const parseSourcemap = (sourcemapContent: string): ParsedSourcemap => {
+  const parsed = JSON.parse(sourcemapContent) as unknown
+  if (!parsed || typeof parsed !== 'object' || Array.isArray(parsed)) {
+    throw new Error('Sourcemap must be a JSON object')
+  }
+
+  return parsed as ParsedSourcemap
+}
+
+export const extractSourcemapDebugId = (sourcemapContent: string): string | undefined => {
```
amortemousque: **💬 suggestion:** ‏
The build plugin only injects the snippet containing the debug_id into the source file; it does not add it to the source map, since the backend does not use it there. Following the same approach could simplify the implementation and avoid unnecessary processing.

packages/base/src/commands/sourcemaps/validation.ts
```diff
@@ -48,6 +51,22 @@ export const validatePayload = (sourcemap: Sourcemap, stdout: Writable, debugIdR
...
+          `bundle debug ID ${sourcemap.debugId} does not match sourcemap debug ID ${sourcemapDebugId ?? '<missing>'}`
+        )
+      }
+    } catch (error) {
+      throw new InvalidPayload(
+        'invalid_debug_id_pair',
+        `Skipping sourcemap ${sourcemap.sourcemapPath} because its debug ID is inconsistent: ${error}`
+      )
+    }
+  }
```
amortemousque: **💬 Suggestion:** If we decide not to inject the debug_id into the source map, we can remove this validation. The source file’s debug_id is the only one we need.

packages/base/src/commands/sourcemaps/debugId.ts
jhssilva: Valid observation. Discussed with the team, and we would prefer to have the debug_id in the sourcemap. This way we can easily troubleshoot an issue or identify a map from their content if needed.

packages/base/src/commands/sourcemaps/debugId.ts
jhssilva: Great suggestion.

**Note:** The longer-term goal would be to share this logic between `build-plugins` and `datadog-ci`.

jhssilva: Added in 8a5aa01d: the runtime snippet now has an explicit pinned cross-reference to build-plugins so future changes can be kept in sync.

packages/base/src/commands/sourcemaps/findSourcemaps.ts
jhssilva: Done in 8a5aa01d: discovery.ts was renamed to findSourcemaps.ts and both command imports were updated.

packages/base/src/commands/sourcemaps/debugId.ts
jhssilva: Addressed in 8a5aa01d: generation now uses the build-plugins SHA-256 UUID strategy with crypto.createHash and no additional dependency.

packages/base/src/commands/sourcemaps/debugId.ts
jhssilva: Removed in 8a5aa01d. The DD_SOURCE_CODE_CONTEXT runtime snippet is now the only Debug ID marker added to JavaScript.

packages/base/src/commands/sourcemaps/inject.ts
jhssilva: Added in 8a5aa01d: sourcemaps inject now accepts --max-concurrency, matching upload.

packages/base/src/commands/sourcemaps/upload.ts
jhssilva: Restored in 8a5aa01d: upload --debug-id aborts with renderNoDebugIdFound when no discovered bundle contains an ID.

packages/base/src/commands/sourcemaps/inject.ts
jhssilva: Addressed in 8a5aa01d: the new command-level messages are centralized in renderer.ts.

packages/base/src/commands/sourcemaps/debugId.ts
jhssilva: Implemented in 8a5aa01d: inject guarantees the map debug_id matches the bundle, while upload remains JavaScript-authoritative and accepts build-plugin maps that omit it.

packages/base/src/commands/sourcemaps/debugId.ts
jhssilva: Aligned in 8a5aa01d: the CLI now uses the same SHA-256 UUID transformation as build-plugins, covered by a fixed parity vector.

packages/base/src/commands/sourcemaps/debugId.ts
jhssilva: Removed in 8a5aa01d. The CLI no longer reads or writes the camel-case debugId field; it writes only debug_id.

packages/base/src/commands/sourcemaps/debugId.ts
```diff
@@ -31,3 +79,240 @@ export const addDebugIdToPayloads = (payloads: Sourcemap[]): boolean => {
...
+  const hashbangMatch = jsContent.match(HASHBANG_REGEX)
+  const hashbangPortion = hashbangMatch ? hashbangMatch[0] : ''
+  const sourceWithoutHashbang = jsContent.slice(hashbangPortion.length)
+
+  const useDirectives = extractLeadingUseDirectives(sourceWithoutHashbang)
+
+  const originalSourcemap = parseSourcemap(sourcemapContent)
+  if (originalSourcemap.sections !== undefined) {
+    throw new Error('Indexed sourcemaps with "sections" are not supported by sourcemaps inject')
+  }
```
amortemousque: **💬 suggestion:** ‏This check could happen earlier, as a guard clause in `injectMissingDebugIds`, before we start processing the payload.

packages/base/src/commands/sourcemaps/debugId.ts
```diff
@@ -31,3 +79,240 @@ export const addDebugIdToPayloads = (payloads: Sourcemap[]): boolean => {
...
+        stdout.write(`WARN: Failed to record debug ID in ${payload.sourcemapPath}: ${error}\n`)
+      }
+      continue
+    }
+
+    let jsContent: string
+    let sourcemapContent: string
+    try {
+      jsContent = fs.readFileSync(payload.minifiedFilePath, 'utf-8')
+      sourcemapContent = fs.readFileSync(payload.sourcemapPath, 'utf-8')
```
amortemousque: **❓ question:** ‏Could we avoid re-reading the source map file, since it was already read a few lines above?

packages/base/src/commands/sourcemaps/debugId.ts
amortemousque: I’m not really sure how useful this would be for debugging. Also, injecting debug_id into the source map introduces a new set of potential issues, for example, the JavaScript and source map could end up out of sync. The build plugin doesn’t add it either. To me, the complexity isn’t worth the potential benefit at this stage, so I’m still in favor of removing it.

buranmert: tl;dr: IMO we should have debugID in the sourcemap.

let's say i'm a user, i build my project and upload my sourcemaps from my build.
then i look at my JS errors in Datadog UI and i see an error message saying "frames cannot be unminified because JS sourcemap with debugID: 1234-5678 cannot be found".
at this point, you can't understand what is actually missing unless you have debugID in your sourcemaps.

also, [the proposal has debugID in the sourcemap](https://github.com/tc39/ecma426/blob/main/proposals/debug-id.md#debug-ids-in-source-maps).

---

if the current implementation is proven too complex or slow, we can look for ways to improve though.

packages/base/src/commands/sourcemaps/debugId.ts
jhssilva: Implemented in 7693dc22. injectMissingDebugIds now rejects indexed maps immediately after the initial parse, before reading the JavaScript or generating/logging a debug ID. The lower-level guard remains for direct callers; existing-ID metadata-only reconciliation stays supported because it does not transform mappings.

packages/base/src/commands/sourcemaps/debugId.ts
jhssilva: Implemented in 7693dc22. Each sourcemap is now read and parsed once per payload, and the parsed object is reused for empty/indexed validation, metadata reconciliation, and injection.

packages/base/src/commands/sourcemaps/validation.ts
jhssilva: Comment above: [discussion_r3843548760](https://github.com/DataDog/datadog-ci/pull/2454#discussion_r3843548760)

packages/base/src/commands/sourcemaps/debugId.ts
```diff
@@ -31,3 +88,258 @@ export const addDebugIdToPayloads = (payloads: Sourcemap[]): boolean => {
...
+  return [
+    withVariant.slice(0, 8),
+    withVariant.slice(8, 12),
+    withVariant.slice(12, 16),
+    withVariant.slice(16, 20),
+    withVariant.slice(20, 32),
+  ].join('-')
+}
+
+const PRE_INJECTION_SOURCE_NAME = 'pre-injection.js'
```
buranmert: nit: i think we can prefix this with `dd` or `datadog`. if this file stays as a leftover for some reason, i prefer it showing why it existed in the first place.

buranmert (commented):
LGTM (i can't approve though)

ImaneLargou (approved):

packages/base/src/commands/sourcemaps/debugId.ts
jhssilva: **Note**: 
- This opens the door for a comment like "validate" which allows customers to troubleshoot sourcemaps missing with debug id

packages/base/package.json
```diff
@@ -155,6 +157,7 @@
     "@types/inquirer": "^8.2.6",
     "@types/js-yaml": "^4.0.9",
     "@types/semver": "^7.7.1",
-    "@types/tiny-async-pool": "^2.0.3"
+    "@types/tiny-async-pool": "^2.0.3",
+    "source-map": "^0.7.4"
```
Drarig29, line 161: Why is `source-map` needed as a dev dep?

packages/base/src/commands/sourcemaps/debugId.ts
Drarig29: > The goal would be to have a shared package between bundle-plugin and datadog-ci

Did you mean build-plugins instead of bundle-plugin? In which repo do you think the shared package should live? If you prefer in the datadog-ci repo, it should be possible to move the `sourcemaps` command to a `@datadog/datadog-ci-plugin-sourcemaps` NPM package, and have it both as a built-in plugin of datadog-ci and a dependency of build-plugins.

packages/base/package.json
```diff
@@ -124,6 +124,7 @@
   },
   "dependencies": {
     "@antfu/install-pkg": "^1.1.0",
+    "@babel/parser": "^7.29.0",
```
Drarig29, line 127: Why Babel 7 and [not Babel 8](https://babeljs.io/docs/v8-migration)? If we can, I'd rather start with Babel 8 to avoid introducing debt on day 1 😁

Drarig29 (commented):
I'm OK with adding `@babel/parser` [(bundlephobia)](https://bundlephobia.com/package/@babel/parser@8.0.4) and `webpack-sources` [(bundlephobia)](https://bundlephobia.com/package/webpack-sources@3.5.1) as dependencies as they are pretty small compared to the vendored `@datadog/datadog-ci` [(bundlephobia)](https://bundlephobia.com/package/@datadog/datadog-ci@5.23.0), even though it will make `@datadog/datadog-ci-base` [(bundlephobia)](https://bundlephobia.com/package/@datadog/datadog-ci-base@5.23.0) bigger.

We could reuse Babel for another project in Synthetics.

packages/base/package.json
jhssilva: It is imported directly by `debugId.test.ts` to generate and consume source maps while verifying mapping preservation. It may currently resolve transitively through `dd-trace`, but declaring it as a dev dependency avoids relying on hoisting and keeps it out of runtime dependencies.

packages/base/package.json
jhssilva: I checked Babel Parser 8.0.4, but its [package metadata](https://github.com/babel/babel/blob/v8.0.4/packages/babel-parser/package.json#L32-L34) requires Node `^22.18.0 || >=24.11.0`, while [`datadog-ci` supports Node `>=20`](https://github.com/DataDog/datadog-ci/blob/master/packages/datadog-ci/package.json#L19-L21). Using Babel 8 here would therefore break the current compatibility contract. Babel 7 remains maintained and supports our runtime range, so I think we should stay on 7 until `datadog-ci` drops Node 20.

packages/base/src/commands/sourcemaps/debugId.ts
jhssilva: Yes, I meant `build-plugins`; I corrected the typo. I had not settled on the repository or package boundary yet. A `@datadog/datadog-ci-plugin-sourcemaps` package makes sense if we want the complete sourcemaps workflow to have a single owner, especially since both repositories currently contain sourcemap-upload logic. I would explore that extraction in a follow-up rather than expand this PR. As part of that work, we can compare sharing the complete plugin with extracting only the Debug ID/source-code-context primitives.

Drarig29: (yep, this would be a separate PR 👍)

packages/base/package.json
Drarig29: sounds good!

Drarig29 (approved):
LGTM from admin side

PR #2463, opened 2026-08-21
[SVLS-9300] add ecs-fargate instrument command

ojproductions:
### What and why?

Adds `datadog-ci ecs-fargate instrument`, which instruments an AWS ECS Fargate task
definition with Datadog: it adds the `datadog-agent` sidecar, gives the application
containers the environment their tracers read, and registers the result as a new
revision. Container images are left untouched.

Fargate users currently have to hand-edit task definition JSON to add the sidecar, which
is easy to get wrong and hard to keep consistent across services.

### How?

New `plugin-ecs-fargate` package plus the base command definition, following the usual
base/plugin split. A run describes each task definition, transforms it, and registers
the new revision.

- Targeting: `--task-definition` (family, `family:revision`, or ARN), `--region`,
  `--profile`, and a `datadog-ci.json` config file.
- Rollout: `--ecs-service` points services at the revision that was just registered, so
  the change reaches running tasks without a manual deployment. `--cluster` is only
  needed when the services are not named by full ARN. Services are updated only after
  every task definition in the run is instrumented, so a partial failure does not leave
  half the fleet on a new revision.
- API key: `--api-key-secret-arn` keeps the key out of the task definition; falling back
  to `DD_API_KEY` writes it in plain text, which is warned about and validated first.
- Idempotency: the Agent container is matched by name and the `dd_sls_ci` version tag is
  excluded from the comparison, so re-running (or upgrading the CLI) does not burn a
  revision.
- `--dry-run` prints the diff and registers nothing.

Unified service tagging and the product toggles follow in the next PR; Windows tasks in
the one after. The `@aws-sdk/client-ecs` version is pinned to the one the other AWS
packages in the repo already use, so the lockfile addition stays small.

### Review checklist


ava-silver:
@codex review

packages/datadog-ci/README.md
```diff
@@ -113,6 +113,12 @@ The following `<scope>` and `<command>` values are available.
 
 - `upload`: Upload iOS dSYM files for Error Tracking (macOS only).
 
+#### `ecs-fargate`
+
+<sub>**README:** [📚](/packages/plugin-ecs-fargate) | **Documentation:** [🔗](https://docs.datadoghq.com/integrations/ecs_fargate/) | **Plugin:** `@datadog/datadog-ci-plugin-ecs-fargate`</sub>
```
OliviaShoup: this link 301-redirects to `/integrations/aws-fargate/` (confirmed via curl). it still works, but it's not the canonical URL

```suggestion
<sub>**README:** [📚](/packages/plugin-ecs-fargate) | **Documentation:** [🔗](https://docs.datadoghq.com/integrations/aws-fargate/) | **Plugin:** `@datadog/datadog-ci-plugin-ecs-fargate`</sub>
```

packages/plugin-ecs-fargate/README.md
```diff
@@ -0,0 +1,91 @@
...
+- `AWS_REGION` (or `AWS_DEFAULT_REGION`): the region to use when `--region` is not passed.
+
+### Arguments
+
+You can pass the following arguments to `instrument` to specify its behavior. `--fips` and `--fips-ignore-error` are also accepted, as they are on every command.
+
+<!-- BEGIN_USAGE:instrument -->
+| Argument | Shorthand | Description | Default |
+| -------- | --------- | ----------- | ------- |
+| `--dry` or `--dry-run` | `-d` | Preview changes running command would apply | `false` |
```
OliviaShoup:
```suggestion
| `--dry` or `--dry-run` | `-d` | Preview the changes the command would apply | `false` |
```

packages/plugin-ecs-fargate/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+#### Deploying the new revision
+
+Pass `--ecs-service` for each service that should run the revision the command just registered, and `--cluster` if those services are not in the `default` cluster. A service named by its full ARN already says which cluster it runs in, so `--cluster` can be left off; passing one that the ARN contradicts is an error rather than a silent choice between them. A run updates services in a single cluster, so ARNs naming more than one are reported too. Each service is matched to the task definition family it currently runs, so a run over several task definitions points each service at its own new revision, and a service already running the instrumented revision is left alone rather than redeployed. Updating a service starts an ECS deployment: the command returns as soon as ECS accepts it, and the rollout follows your service's deployment configuration.
```
OliviaShoup: "a silent choice between them" is a little hard to parse on first read. the antecedent for "them" (ARN's cluster vs. flag's cluster) isn't obvious. so suggesting this but no pressure

```suggestion
Pass `--ecs-service` for each service that should run the revision the command just registered, and `--cluster` if those services are not in the `default` cluster. A service named by its full ARN already says which cluster it runs in, so `--cluster` can be left off; passing a --cluster that contradicts the ARN's cluster is an error, not a silent override. A run updates services in a single cluster, so ARNs naming more than one are reported too. Each service is matched to the task definition family it currently runs, so a run over several task definitions points each service at its own new revision, and a service already running the instrumented revision is left alone rather than redeployed. Updating a service starts an ECS deployment: the command returns as soon as ECS accepts it, and the rollout follows your service's deployment configuration.
```

OliviaShoup (approved):
this looks great! left a few comments but nothing blocking

packages/plugin-ecs-fargate/src/task-definition.ts
```diff
@@ -0,0 +1,346 @@
...
+
+  if (!agentContainer.logConfiguration) {
+    warnings.push(
+      `The ${AGENT_CONTAINER_NAME} container has no logConfiguration, so its own logs will not be collected and a failing Agent will be hard to diagnose. Add one to the task definition.`
+    )
+  }
+
+  // ECS resolves secrets through the task's execution role, so a reference without a role in place
+  // registers a revision whose tasks cannot start.
+  if (settings.apiKeySecretArn && !taskDefinition.executionRoleArn) {
```
ava-silver: **P1:** Reject this configuration instead of warning. ECS resolves container secrets through the task execution role, so this revision cannot start without `executionRoleArn`; with `--ecs-service`, the command proceeds to deploy that unusable revision.

packages/plugin-ecs-fargate/src/aws.ts
```diff
@@ -0,0 +1,155 @@
...
+import {AWS_SHARED_CREDENTIALS_FILE_ENV_VAR, EXPONENTIAL_BACKOFF_RETRY_STRATEGY} from './constants'
+
+// TODO: the two credential helpers below are duplicated from plugin-lambda's `functions/commons.ts`
+// (which also prompts for an MFA code). Move them to a shared `helpers/serverless/aws.ts` in the base
+// package, alongside `AWS_SHARED_CREDENTIALS_FILE_ENV_VAR` and the retry strategy from `constants.ts`.
+
+/**
+ * Returns the credentials loaded from the given AWS named profile.
+ */
+export const getAWSProfileCredentials = async (profile: string): Promise<AwsCredentialIdentity | undefined> => {
```
ava-silver, line 32: **P1:** Support MFA-backed named profiles here. The analogous Lambda credential helper supplies `mfaCodeProvider`, but this `fromIni` setup does not, so `--profile` fails for profiles configured with `mfa_serial`. Please share or reproduce the Lambda credential setup, including its MFA callback.

packages/plugin-ecs-fargate/src/commands/instrument.ts
```diff
@@ -0,0 +1,288 @@
...
+    cluster: string | undefined,
+    services: string[],
+    instrumented: InstrumentedRevision[]
+  ): Promise<boolean> {
+    let deployed = true
+    for (const name of services) {
+      try {
+        const service = await describeService(client, cluster, name)
+        const family = taskDefinitionFamily(service.taskDefinition)
+        const target = instrumented.find((candidate) => candidate.family === family)
```
ava-silver: **P2:** Do not select the first result when multiple targets resolve to the same family. For example, `--task-definition app:1 --task-definition app:2` registers both revisions but deploys the first one, which can roll a service back. Reject duplicate resolved families or define an explicit selection rule.

ava-silver (commented):

The command needs stronger safeguards around deployable task definitions, MFA-backed AWS profiles, and duplicate task-definition families.

packages/plugin-ecs-fargate/src/task-definition.ts
```diff
@@ -0,0 +1,346 @@
...
+ * Builds the environment the Agent sidecar runs with.
+ */
+const getAgentEnvVars = (settings: InstrumentSettings): ManagedValues => ({
+  managed: {
+    [ECS_FARGATE_ENV_VAR]: 'true',
+    [SITE_ENV_VAR]: settings.site,
+    // The Agent's own trace intake, which is a separate switch from the tracers' `DD_TRACE_ENABLED`.
+    [DD_APM_ENABLED_ENV_VAR]: 'true',
+    [DD_USE_DOGSTATSD_ENV_VAR]: 'true',
+    [DD_ECS_TASK_COLLECTION_ENABLED_ENV_VAR]: 'true',
```
ava-silver, line 132: **P2:** Enabling ECS task collection requires permissions that this command neither supplies nor checks. The [Terraform module adds `ecs:ListClusters`, `ecs:ListContainerInstances`, and `ecs:DescribeContainerInstances` to the task role](https://github.com/DataDog/terraform-aws-ecs-datadog/blob/b4cf7d00146dff2e167b43403c6a2184ef442e19/modules/ecs_fargate/iam.tf#L95-L108), and the [CDK construct adds the same policy](https://github.com/DataDog/datadog-cdk-constructs/blob/abde8e1eb7d54f50c9f0a56f1054a5c21959c5bc/src/ecs/utils.ts#L79-L86). Without those permissions, ECS task collection is incomplete. Avoid enabling it automatically, or make the required task-role policy explicit to the user.

packages/plugin-ecs-fargate/src/task-definition.ts
```diff
@@ -0,0 +1,346 @@
...
+/**
+ * Builds the environment the Agent sidecar runs with.
+ */
+const getAgentEnvVars = (settings: InstrumentSettings): ManagedValues => ({
+  managed: {
+    [ECS_FARGATE_ENV_VAR]: 'true',
+    [SITE_ENV_VAR]: settings.site,
+    // The Agent's own trace intake, which is a separate switch from the tracers' `DD_TRACE_ENABLED`.
+    [DD_APM_ENABLED_ENV_VAR]: 'true',
+    [DD_USE_DOGSTATSD_ENV_VAR]: 'true',
```
ava-silver, line 131: **P2:** Match the DogStatsD origin-tagging defaults used by the other ECS instrumentation tools. The Terraform module defaults origin detection to enabled and cardinality to `orchestrator` ([configuration](https://github.com/DataDog/terraform-aws-ecs-datadog/blob/b4cf7d00146dff2e167b43403c6a2184ef442e19/modules/ecs_fargate/variables.tf#L152-L164), [Agent environment](https://github.com/DataDog/terraform-aws-ecs-datadog/blob/b4cf7d00146dff2e167b43403c6a2184ef442e19/modules/ecs_fargate/datadog.tf#L313-L333)); the [CDK construct injects the same variables](https://github.com/DataDog/datadog-cdk-constructs/blob/abde8e1eb7d54f50c9f0a56f1054a5c21959c5bc/src/ecs/fargate/environment.ts#L40-L44). Add `DD_DOGSTATSD_ORIGIN_DETECTION=true`, `DD_DOGSTATSD_ORIGIN_DETECTION_CLIENT=true`, and `DD_DOGSTATSD_TAG_CARDINALITY=orchestrator` so custom metrics retain equivalent origin tags and cardinality.

ava-silver (commented):

Two additional parity gaps surfaced when comparing this command with the corresponding Terraform module and CDK construct.

stzou (commented):
Thanks for adding ECS fargate support - just a couple comments around logging and traces.

APM/DogStatsD: right now app containers reach the Agent over the implicit `localhost:8126` tracer default, which only works because `awsvpc` shares one network namespace. Can we follow the Terraform module's pattern here?
- mount a shared dd-sockets volume
- set `DD_TRACE_AGENT_URL/DD_DOGSTATSD_URL` to the unix socket by default and fall back to UDP (`DD_AGENT_HOST=127.0.0.1`) only when the user opts out of the socket.

Logs: there's currently no path to get application logs into Datadog. Could we add an opt-in Fluent Bit sidecar (datadog-log-router) using the awsfirelens log driver, matching the Terraform module's dd_log_collection config, so app containers can route logs through it to Datadog's log intake.

Could also be worth adding `DD_INSTALL_INFO_TOOL / _TOOL_VERSION / _INSTALLER_VERSION` so we can track usage of this cli. If there's another way this is being tracked, feel free to leave it out. 

Where possible, we should try to have parity between this onboarding tool and existing ones so the onboarding experience is consistent regardless of method used.

packages/plugin-ecs-fargate/src/aws.ts
```diff
@@ -0,0 +1,166 @@
...
+  TaskDefinitionField,
+  UpdateServiceCommand,
+} from '@aws-sdk/client-ecs'
+import {fromIni, fromNodeProviderChain} from '@aws-sdk/credential-providers'
+import {input as promptInput} from '@inquirer/prompts'
+import {CredentialsProviderError} from '@smithy/property-provider'
+
+import {AWS_SHARED_CREDENTIALS_FILE_ENV_VAR, EXPONENTIAL_BACKOFF_RETRY_STRATEGY} from './constants'
+
+// TODO: the two credential helpers below are duplicated from plugin-lambda's `functions/commons.ts`.
```
ava-silver, line 19: yeah the tricky thing is we don't want to pull the AWS SDK into the base package -- we can revisit this if need be in the future

packages/plugin-ecs-fargate/src/aws.ts
```diff
@@ -0,0 +1,166 @@
...
+  // than letting the profile fail to load.
+  init.mfaCodeProvider = (mfaSerial) => promptInput(mfaCodeQuestion(mfaSerial))
+
+  try {
+    const credentialsProvider: AwsCredentialIdentityProvider = fromIni(init)
+
+    return await credentialsProvider()
+  } catch (err) {
+    if (err instanceof Error) {
+      throw Error(`Couldn't set AWS profile credentials. ${err.message}`)
```
ava-silver:
```suggestion
      throw Error(`Couldn't get AWS profile credentials. ${err.message}`)
```

packages/plugin-ecs-fargate/src/commands/instrument.ts
```diff
@@ -0,0 +1,299 @@
...
+    const [config, configErrors] = await this.ensureConfig()
+    if (configErrors.length > 0) {
+      for (const error of configErrors) {
+        this.context.stdout.write(renderError(error))
+      }
+
+      return 1
+    }
+
+    const region = config.region ?? AWS_REGION_ENV_VARS.map((envVar) => process.env[envVar]).find((value) => !!value)
```
ava-silver, line 57: would it be worth supporting multiple regions here (based on the task definition ARN)? although either way that'd be better as a follow-up PR

packages/plugin-ecs-fargate/src/commands/instrument.ts
```diff
@@ -0,0 +1,299 @@
...
+      if (revision) {
+        instrumented.set(revision.family, revision)
+      } else {
+        failed = true
+      }
+    }
+
+    if (failed) {
+      // Deploying changes what is running, so a run that could not instrument every task definition
+      // stops here rather than pointing some of the services at a new revision.
```
ava-silver: I'm not sure I understand why we'd want to do this rather than roll out the successful ones -- lets say they have 99 task definitions succeed, but one fails, we'd fail the entire fleet and nothing would get rolled out

I feel like we should probably be in line with the other commands and roll out the successful ones, so we could do each instrumentation in parallel and just wait for all the operations at the same time.

packages/plugin-ecs-fargate/src/commands/instrument.ts
```diff
@@ -0,0 +1,299 @@
...
+      this.context.stdout.write(renderError(error instanceof Error ? error.message : error))
+
+      return 1
+    }
+
+    const services = config.ecsServices ?? []
+    const instrumented: InstrumentedRevisions = new Map()
+    let failed = false
+    for (const taskDefinition of config.taskDefinitions ?? []) {
+      const revision = await this.instrument(client, taskDefinition, settings, services.length > 0)
```
ava-silver: this should probably use a `Promise.all()` to do these in parallel, right? (Assuming the ECS client has throttling/retry handling builtin)

packages/plugin-ecs-fargate/src/commands/instrument.ts
```diff
@@ -0,0 +1,299 @@
...
+   * @returns whether every service runs an instrumented revision.
+   */
+  private async deploy(
+    client: ECSClient,
+    cluster: string | undefined,
+    services: string[],
+    instrumented: InstrumentedRevisions
+  ): Promise<boolean> {
+    let deployed = true
+    for (const name of services) {
```
ava-silver: same with this for loop, it'd be great to parallelize these, although if we refactor to make each app individual then we probably get this for free

packages/plugin-ecs-fargate/src/commands/instrument.ts
ojproductions: yup I think followup PR is better

ojproductions:
@stzou I've added the dd-sockets change, though I think the logging changes would be better added to [this PR](https://github.com/DataDog/datadog-ci/pull/2465) (the next PR in the stack) where I add other configuration flags. 
Regarding `DD_INSTALL_INFO_TOOL / _TOOL_VERSION / _INSTALLER_VERSION` variables, we're already using the `dd_sls_ci` tag... should we be adding these new vars? if so this may also be a separate PR as none of the other cli commands add this. 
thoughts?

ava-silver (approved):

stzou (approved):
LGTM

For the install info vars, `dd_sls_ci` covers our need to track CLI-driven installs so we're fine leaving those out.

ava-silver (commented):
a couple last things i missed in my initial review now after solidifying on command guidance: https://github.com/DataDog/datadog-ci/pull/2500

packages/plugin-ecs-fargate/src/task-definition.ts
```diff
@@ -0,0 +1,471 @@
...
+    [DD_ECS_TASK_COLLECTION_ENABLED_ENV_VAR]: 'true',
+    ...(settings.apiKey ? {[API_KEY_ENV_VAR]: settings.apiKey} : {}),
+  },
+  defaults: {
```
ava-silver, line 135: The [lifecycle guidance](https://github.com/DataDog/datadog-ci/pull/2500) says omitted configuration must resolve to an explicit default or absence rather than preserve remote Datadog state. These values -- along with `DD_TRACE_ENABLED` and `DD_LOGS_INJECTION` below -- are defaults, so the same invocation can retain an earlier `false` or `low` value and yield a different Datadog configuration. Please either manage these values or add explicit options whose omitted values resolve deterministically.

packages/plugin-ecs-fargate/src/task-definition.ts
```diff
@@ -0,0 +1,471 @@
...
+    [DD_DOGSTATSD_ORIGIN_DETECTION_ENV_VAR]: 'true',
```
ava-silver, line 136: For ECS Fargate, Datadog's DogStatsD UDS documentation requires `pidMode: "task"` when `DD_DOGSTATSD_ORIGIN_DETECTION=true`. This transform enables origin detection but leaves `pidMode` unchanged, so metrics may not receive the task/container origin tags described in the README. Please apply `pidMode: "task"` as part of this desired state.

packages/plugin-ecs-fargate/src/task-definition.ts
```diff
@@ -0,0 +1,471 @@
...
+ * produces an identical result and does not burn a revision.
+ */
+export const instrumentTaskDefinition = (
+  taskDefinition: TaskDefinition,
+  settings: InstrumentSettings,
+  tags: Tag[] = []
+): InstrumentResult => {
+  const family = taskDefinition.family
+
+  const compatibilities = (taskDefinition.requiresCompatibilities ?? []).map((compatibility) =>
```
ava-silver: Windows Fargate definitions pass these checks, then receive the Linux Agent image, `/probe.sh`, and Unix-socket configuration. With `--ecs-service`, this can deploy a revision whose Agent cannot start. Since Windows support is deferred, please reject `runtimePlatform.operatingSystemFamily` values for Windows before registering a revision.

packages/base/src/commands/ecs-fargate/instrument.ts
```diff
@@ -0,0 +1,211 @@
...
+ * Derive the cluster from an ECS service ARN
+ */
+const clusterFromServiceArn = (service: string): string | undefined => {
+  if (!service.startsWith('arn:')) {
+    return undefined
+  }
+
+  const segments = service.split('/')
+
+  return segments.length === 3 ? segments[1] : undefined
```
ava-silver, line 61: This returns no cluster for a legacy short service ARN (`...:service/name`). ECS then resolves it in the default cluster, so a service in another cluster cannot be found despite the README saying a full ARN does not need `--cluster`. Please require `--cluster` for short-format ARNs and reserve automatic cluster extraction for long-format ARNs.

PR #2465, opened 2026-08-21
[SVLS-9300] add instrumentation configuration flags for ecs-fargate instrument

ojproductions:
### What and why?

Adds the flags that decide how an instrumented task reports to Datadog. Without them the
command produces telemetry that is hard to slice: everything lands under the task
definition family with no environment or version to correlate against.

### How?

Unified service tagging (`--service`, `--env`, `--version`, `--extra-tags`) is written
three ways, so the same task is named identically wherever it shows up:

- as `DD_SERVICE`, `DD_ENV`, `DD_VERSION`, and `DD_TAGS` on every container, which is
  what the tracers send;
- as `com.datadoghq.tags.*` Docker labels on the application containers, which is what
  the Agent reads to tag the metrics it collects about them from the outside. The Agent
  container is deliberately left unlabelled, so it does not report its own resource usage
  under your service;
- as tags on the revision itself.

An explicit flag overrides what a container already declares; the task definition family
is only a fallback, and `DD_SERVICE`, `DD_TRACE_ENABLED`, and `DD_LOGS_INJECTION` are
filled in only when the container has not made a choice itself.

Also the product toggles the tracers read (`--tracing`, `--log-level`, `--appsec`,
`--llmobs`, `--env-vars`) and the source code integration. Its Git tags are resolved even
on a dry run so the diff shows the `DD_TAGS` a real run would write, while the metadata
upload — a write to Datadog — is skipped.

### Review checklist


ava-silver:
@codex review

OliviaShoup (approved):

packages/plugin-ecs-fargate/src/task-definition.ts
```diff
@@ -355,17 +615,32 @@ export const instrumentTaskDefinition = (
   }
 
   const containers = taskDefinition.containerDefinitions ?? []
+  const borrowed = borrowedLogConfiguration(containers)
+  const firelens = settings.logCollection ? firelensLogConfiguration(settings) : undefined
```
ava-silver, line 662: When log collection is turned off, `firelens` becomes undefined and the existing router plus every existing `awsfirelens` configuration remain unchanged. The command therefore cannot converge back to its default log state, and no ECS uninstrument command exists to clean these CLI-managed artifacts. Add a cleanup path that removes the router and its routing configuration.

ava-silver (commented):

The command needs a consistent desired-state model for UST, product settings, and log collection. I also noted one documentation mismatch.

packages/plugin-ecs-fargate/src/task-definition.ts
```diff
@@ -119,48 +165,134 @@ const mergeManaged = <T>(
...
+ *
+ * An explicit `--service` takes precedence. Otherwise derived from the family the task definition is named after
+ */
+const getServiceTagEnvVars = (settings: InstrumentSettings, family?: string): ManagedValues => {
+  const managed: Record<string, string> = {}
+  const defaults: Record<string, string> = {}
+
+  if (settings.service) {
+    managed[SERVICE_ENV_VAR] = settings.service
+  } else if (family) {
```
ava-silver: Using the family as a default here preserves an existing application `DD_SERVICE`, but the Agent and revision tag use the family. A task with `DD_SERVICE=checkout` in a `payments-worker` family then emits traces as `checkout` and Agent/resource telemetry as `payments-worker`. Resolve UST once, then apply the same desired state to every Datadog-owned UST destination.

packages/plugin-ecs-fargate/src/task-definition.ts
```diff
@@ -119,48 +165,134 @@ const mergeManaged = <T>(
...
+    managed[DD_TRACE_AGENT_URL_ENV_VAR] = APM_SOCKET_URL
+    managed[DD_DOGSTATSD_URL_ENV_VAR] = DOGSTATSD_SOCKET_URL
+  } else {
+    managed[DD_AGENT_HOST_ENV_VAR] = AGENT_LOOPBACK_HOST
+  }
+
+  if (settings.tracing !== undefined) {
+    managed[DD_TRACE_ENABLED_ENV_VAR] = String(settings.tracing)
+  }
+  if (settings.appsec) {
```
ava-silver, line 299: `--no-appsec` resolves to `false`, but this add-only branch leaves an existing `DD_APPSEC_ENABLED=true` untouched. The same desired-state issue applies to disabling source-code integration and omitting other new settings. Resolve each setting to an enabled value or absence, then apply or remove its Datadog-owned fields consistently.

packages/base/src/commands/ecs-fargate/instrument.ts
```diff
@@ -132,6 +147,48 @@ export class EcsFargateInstrumentCommand extends BaseCommand {
...
+  private version = Option.String('--version', {
+    description:
+      'The value for the version tag. Use this to correlate spikes in latency, load, or errors to new versions. For example, `1.0.0`.',
+  })
+  private extraTags = Option.String('--extra-tags,--extraTags', {
+    description: 'Additional tags to add to the task in the format "key1:value1,key2:value2".',
+  })
+  private envVars = Option.Array('-e,--env-vars', {
+    description:
+      'Additional environment variables to set on every container in the task. Can specify multiple variables in the format `--env-vars VAR1=VALUE1 --env-vars VAR2=VALUE2`.',
```
ava-silver: This says every container, but with `--log-collection` the transform deliberately excludes `datadog-log-router` from `envVars`. Please describe this as applying to application containers and the Datadog Agent.

ava-silver (approved):

PR #2466, opened 2026-08-21
[SVLS-9300] support Windows containers in ecs-fargate instrument command

ojproductions:
### What and why?

Instruments Fargate tasks that run Windows containers. Until now the command would give
them the Linux Agent image, which does not start on Windows.

### How?

The task's `runtimePlatform` decides, and three things change from the Linux path:

- the Agent runs the `-servercore` build of the image, published as a manifest list so
  ECS pulls the variant matching the Windows Server version;
- the Agent container is given `C:\` as its working directory, which it needs and its
  image does not set;
- it gets no health check, because the Agent's probe is a shell script only the Linux
  image ships. A probe would report the Agent as permanently unhealthy rather than tell
  anyone anything, so the command warns instead of adding one that cannot pass.

An explicit `--agent-image` is still used exactly as given, so it has to point at a
`-servercore` tag. The application containers are instrumented exactly as they are on
Linux.

### Review checklist


ava-silver:
@codex review

OliviaShoup (approved):

packages/plugin-ecs-fargate/src/task-definition.ts
```diff
@@ -614,13 +643,22 @@ export const instrumentTaskDefinition = (
...
+      `Task definition ${family} runs Windows, which the ${LOG_ROUTER_CONTAINER_NAME} sidecar does not support. FireLens is Linux-only, so log collection cannot be enabled on this task.`
+    )
+  }
+
   const containers = taskDefinition.containerDefinitions ?? []
   const borrowed = borrowedLogConfiguration(containers)
   const firelens = settings.logCollection ? firelensLogConfiguration(settings) : undefined
 
   const existingAgent = containers.find((container) => container.name === AGENT_CONTAINER_NAME)
   const {container: agentContainer, warnings} = buildAgentContainer({
```
ava-silver, line 807: `windows` only changes the Agent definition, but the application containers still receive the default Unix-socket transport and shared socket volume below. Windows Fargate supports neither shared bind mounts nor UDS, so the registered task cannot carry traces or metrics. Resolve a Windows-specific TCP transport (or reject Windows tasks until it is supported) rather than retaining the Linux default.

packages/plugin-ecs-fargate/src/task-definition.ts
```diff
@@ -614,13 +643,22 @@ export const instrumentTaskDefinition = (
     )
   }
 
+  const windows = isWindowsTask(taskDefinition)
+  // FireLens does not support Windows Fargate
+  if (settings.logCollection && windows) {
```
ava-silver: When `--log-collection` is omitted, an existing `datadog-log-router` reaches line 704 and is preserved. That leaves a Datadog-owned FireLens sidecar on a Windows task even though FireLens is unsupported there. Resolve Windows to a clean no-log-router state, or fail before registering this invalid configuration.

packages/plugin-ecs-fargate/src/task-definition.ts
```diff
@@ -450,14 +469,22 @@ const buildAgentContainer = ({
...
-    interval: AGENT_HEALTH_CHECK_INTERVAL,
-    timeout: AGENT_HEALTH_CHECK_TIMEOUT,
-    retries: AGENT_HEALTH_CHECK_RETRIES,
-    startPeriod: AGENT_HEALTH_CHECK_START_PERIOD,
-  }
-  if (existing?.healthCheck && !sortedEqual(existing.healthCheck, healthCheck)) {
+  // The Agent's probe is a shell script that only exists in the Linux image, so a Windows task gets
+  // no health check rather than one that can never pass.
+  const healthCheck = windows
+    ? undefined
```
ava-silver, line 566: Removing the Agent health check leaves application `dependsOn` entries untouched. A task that has `{containerName: 'datadog-agent', condition: 'HEALTHY'}` then fails ECS validation because a `HEALTHY` dependency requires a health check on its target. Preserve customer-owned dependency semantics by rejecting this incompatible task with guidance, or deliberately migrate the identifiable Datadog dependency to `START`.

ava-silver (commented):

Windows support needs a platform-specific desired state before this can safely instrument Windows Fargate tasks.

packages/plugin-ecs-fargate/src/task-definition.ts
```diff
@@ -614,13 +643,22 @@ export const instrumentTaskDefinition = (
     )
   }
 
+  const windows = isWindowsTask(taskDefinition)
```
ava-silver: Passing Windows tasks through the existing case-sensitive environment merge can duplicate variables that Windows treats as the same name. For example, existing `dd_trace_enabled=false` receives default `DD_TRACE_ENABLED=true`, so the customer's setting is no longer reliably preserved. Thread this platform decision into environment merging and compare managed environment names case-insensitively for Windows.

ava-silver (approved):

packages/plugin-ecs-fargate/src/task-definition.ts
```diff
@@ -365,17 +388,17 @@ const withSocketVolume = (existing: Volume[] | undefined, socketEnabled: boolean
...
- * already declares.
+ * already declares. Windows resolves variable names without case, so the merge does too.
  */
-const toEnvironment = (values: ManagedValues, existing?: KeyValuePair[]): KeyValuePair[] => {
+const toEnvironment = (values: ManagedValues, {windows}: Platform, existing?: KeyValuePair[]): KeyValuePair[] => {
   const declared: [string, string | undefined][] = []
   for (const {name, value} of existing ?? []) {
     if (name !== undefined) {
       declared.push([name, value])
     }
```
ava-silver, line 398: Windows treats environment-variable names case-insensitively ([Microsoft documentation](https://learn.microsoft.com/en-us/windows/win32/procthread/environment-variables)). This platform-aware merge does not extend to the Agent API-key transition: `inheritedSecrets` and `inheritedEnvironment` still filter only exact `DD_API_KEY`. An existing `dd_api_key` is therefore retained when the command adds secret-backed `DD_API_KEY`; ECS `secrets` set environment variables ([AWS documentation](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/secrets-envvar-secrets-manager.html)). That leaves conflicting definitions of the same Windows variable and can expose the retained plaintext key in the rendered diff because `withMaskedApiKey` also only matches `DD_API_KEY`. Please apply the platform-aware comparison to API-key filtering and masking.

ava-silver (approved):

PR #2470, opened 2026-08-26
fix(sourcemaps): reject legacy uploads with debug ID injection

jhssilva:
### What and why?

Prevent service/version sourcemap uploads when a minified bundle contains a Datadog debug ID injection.

Debug ID and service/version matching are mutually exclusive. Allowing an injected bundle to use service/version metadata can produce sourcemaps that do not match runtime errors.

### How?

- Detect a `ddDebugId` literal in discovered minified files by reusing the existing `extractDebugId` helper (via `addDebugIdToPayloads`).
- Abort before uploading when a debug ID is found with service/version matching.
- Preserve the existing `--debug-id` upload behavior.

build-plugins always emits `ddDebugId` inside the `DD_SOURCE_CODE_CONTEXT` runtime snippet, so detecting the literal is equivalent to detecting the injection for real bundles.

### Testing

- Focused sourcemap tests: 95 passed
- `tsc --noEmit`
- `eslint` on changed files
- Local CLI dry-runs for both upload modes

### Review checklist

### DIFF
--- a/packages/base/src/commands/sourcemaps/README.md
+++ b/packages/base/src/commands/sourcemaps/README.md
@@ -58,7 +58,7 @@ This command will upload all JavaScript sourcemaps and their corresponding JavaS
 
 With `--debug-id`, this command uploads only bundles that already contain a debug ID. It does not modify build artifacts; run `sourcemaps inject` first when debug IDs are missing.
 
-Service/version uploads reject minified bundles containing the Datadog `DD_SOURCE_CODE_CONTEXT` runtime injection. Debug-ID and service/version matching are mutually exclusive; upload those bundles with `--debug-id`, or rebuild them without the injection before using service/version matching. A `ddDebugId` value by itself, without `DD_SOURCE_CODE_CONTEXT`, does not trigger this check.
+Service/version uploads reject minified bundles containing the Datadog `DD_SOURCE_CODE_CONTEXT` runtime injection. Debug ID and service/version matching are mutually exclusive; upload those bundles with `--debug-id`, or rebuild them without the injection before using service/version matching. A `ddDebugId` value by itself, without `DD_SOURCE_CODE_CONTEXT`, does not trigger this check.
 
 To upload the sourcemaps in the build folder, this command should be run:
 
--- a/packages/base/src/commands/sourcemaps/README.md
+++ b/packages/base/src/commands/sourcemaps/README.md
@@ -58,6 +58,8 @@ This command will upload all JavaScript sourcemaps and their corresponding JavaS
 
 With `--debug-id`, this command uploads only bundles that already contain a debug ID. It does not modify build artifacts; run `sourcemaps inject` first when debug IDs are missing.
 
+Service/version uploads reject minified bundles containing the Datadog `DD_SOURCE_CODE_CONTEXT` runtime injection. Debug-ID and service/version matching are mutually exclusive; upload those bundles with `--debug-id`, or rebuild them without the injection before using service/version matching. A `ddDebugId` value by itself, without `DD_SOURCE_CODE_CONTEXT`, does not trigger this check.
+
 To upload the sourcemaps in the build folder, this command should be run:
 
 ```bash
--- a/packages/base/src/commands/sourcemaps/README.md
+++ b/packages/base/src/commands/sourcemaps/README.md
@@ -58,7 +58,7 @@ This command will upload all JavaScript sourcemaps and their corresponding JavaS
 
 With `--debug-id`, this command uploads only bundles that already contain a debug ID. It does not modify build artifacts; run `sourcemaps inject` first when debug IDs are missing.
 
-Service/version uploads reject minified bundles containing the Datadog `DD_SOURCE_CODE_CONTEXT` runtime injection. Debug ID and service/version matching are mutually exclusive; upload those bundles with `--debug-id`, or rebuild them without the injection before using service/version matching. A `ddDebugId` value by itself, without `DD_SOURCE_CODE_CONTEXT`, does not trigger this check.
+Service/version uploads reject minified bundles containing the Datadog `ddDebugId` runtime injection. Debug ID and service/version matching are mutually exclusive; upload those bundles with `--debug-id`, or rebuild them without the injection before using service/version matching.
 
 To upload the sourcemaps in the build folder, this command should be run:
 
--- a/packages/base/src/commands/sourcemaps/__tests__/upload.test.ts
+++ b/packages/base/src/commands/sourcemaps/__tests__/upload.test.ts
@@ -376,7 +376,7 @@ describe('execute', () => {
     expect(context.stdout.toString()).not.toContain('[DRYRUN] Uploading sourcemap')
   })
 
-  test('service/version mode allows a standalone debug ID without DD_SOURCE_CODE_CONTEXT', async () => {
+  test('service/version mode rejects a standalone debug ID', async () => {
     await withTempDirectory(async (directory) => {
       fs.writeFileSync(
         upath.join(directory, 'bundle.js'),
@@ -389,10 +389,9 @@ describe('execute', () => {
 
       const {context, code} = await runCLI([directory])
 
-      expect(code).toBe(0)
-      expect(context.stdout.toString()).toContain('[DRYRUN] Uploading sourcemap')
-      expect(context.stdout.toString()).not.toContain('(debug ID:')
-      expect(context.stderr.toString()).not.toContain('A Datadog debug ID injection was found')
+      expect(code).toBe(1)
+      expect(context.stderr.toString()).toContain('A Datadog debug ID injection was found')
+      expect(context.stdout.toString()).not.toContain('[DRYRUN] Uploading sourcemap')
     })
   })
 
--- a/packages/base/src/commands/sourcemaps/__tests__/upload.test.ts
+++ b/packages/base/src/commands/sourcemaps/__tests__/upload.test.ts
@@ -1,10 +1,9 @@
 import fs from 'fs'
-import os from 'os'
 
 import chalk from 'chalk'
 import upath from 'upath'
 
-import {createCommand, makeRunCLI} from '@datadog/datadog-ci-base/helpers/__tests__/testing-tools'
+import {createCommand, makeRunCLI, withTempDirectory} from '@datadog/datadog-ci-base/helpers/__tests__/testing-tools'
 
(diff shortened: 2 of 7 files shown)


packages/base/src/commands/sourcemaps/__tests__/upload.test.ts
```diff
@@ -370,6 +370,39 @@ describe('execute', () => {
...
+
+      const {context, code} = await runCLI([directory])
+
+      expect(code).toBe(0)
+      expect(context.stdout.toString()).toContain('[DRYRUN] Uploading sourcemap')
+      expect(context.stdout.toString()).not.toContain('(debug ID:')
+      expect(context.stderr.toString()).not.toContain('A Datadog debug ID injection was found')
+    } finally {
+      fs.rmSync(directory, {recursive: true, force: true})
+    }
```
Drarig29: Could you move `withTempDirectory()` to `packages/base/src/helpers/__tests__/testing-tools.ts` and use it in both `packages/base/src/commands/sourcemaps/__tests__/debugId.test.ts` and `packages/base/src/commands/sourcemaps/__tests__/upload.test.ts`?

packages/base/src/commands/sourcemaps/debugId.ts
Drarig29: **nit** to see where the const comes from
```diff
- ({"ddDebugId":"${debugId}"},"DD_SOURCE_CODE_CONTEXT")
+ ({"ddDebugId":"${debugId}"},"${SOURCE_CODE_CONTEXT_MARKER}")
```

packages/base/src/commands/sourcemaps/README.md
```diff
@@ -58,6 +58,8 @@ This command will upload all JavaScript sourcemaps and their corresponding JavaS
 
 With `--debug-id`, this command uploads only bundles that already contain a debug ID. It does not modify build artifacts; run `sourcemaps inject` first when debug IDs are missing.
 
+Service/version uploads reject minified bundles containing the Datadog `DD_SOURCE_CODE_CONTEXT` runtime injection. Debug-ID and service/version matching are mutually exclusive; upload those bundles with `--debug-id`, or rebuild them without the injection before using service/version matching. A `ddDebugId` value by itself, without `DD_SOURCE_CODE_CONTEXT`, does not trigger this check.
```
Drarig29: > A `ddDebugId` value by itself, without `DD_SOURCE_CODE_CONTEXT`, does not trigger this check.

**question** IIUC, this won't trigger the check for backwards compatibility with older versions of the build-plugins, to preserve the existing datadog-ci `--debug-id` upload behavior.

Just for me to understand the bigger picture: are newer versions of the build-plugins going to also inject `DD_SOURCE_CODE_CONTEXT`, to support 2 user flows:
- Either they build their frontend with build-plugins and directly upload with `datadog-ci sourcemaps upload`
- Or they build their frontend without build-plugins, and first inject with `datadog-ci sourcemaps inject` then upload with `datadog-ci sourcemaps upload`?

Drarig29: **nit** for consistency

```suggestion
Service/version uploads reject minified bundles containing the Datadog `DD_SOURCE_CODE_CONTEXT` runtime injection. Debug ID and service/version matching are mutually exclusive; upload those bundles with `--debug-id`, or rebuild them without the injection before using service/version matching. A `ddDebugId` value by itself, without `DD_SOURCE_CODE_CONTEXT`, does not trigger this check.
```

packages/base/src/commands/sourcemaps/renderer.ts
```diff
@@ -51,6 +51,9 @@ export const renderFailedUpload = (sourcemap: Sourcemap, errorMessage: string) =
 
 export const renderNoDebugIdFound = () => 'No debug ID found in any minified file. Aborting upload.\n'
 
+export const renderSourceCodeContextFoundWithoutDebugIdFlag = () =>
+  'A Datadog debug ID injection was found in at least one minified file, but --debug-id was not set. Debug-ID uploads and service/version uploads are mutually exclusive. Re-run with --debug-id or rebuild the bundles without the Datadog debug ID injection. Aborting upload.\n'
```
Drarig29, line 55: **nit** for consistency

```suggestion
  'A Datadog debug ID injection was found in at least one minified file, but --debug-id was not set. Debug ID uploads and service/version uploads are mutually exclusive. Re-run with --debug-id or rebuild the bundles without the Datadog debug ID injection. Aborting upload.\n'
```

packages/base/src/commands/sourcemaps/upload.ts
```diff
@@ -139,9 +140,14 @@ export class SourcemapsUploadCommand extends BaseCommand {
...
     const payloads = await this.getPayloadsToUpload(useGit)
+    if (this.debugId) {
+      if (payloads.length > 0 && !addDebugIdToPayloads(payloads)) {
+        this.context.stderr.write(renderNoDebugIdFound())
 
-    if (this.debugId && payloads.length > 0 && !addDebugIdToPayloads(payloads)) {
-      this.context.stderr.write(renderNoDebugIdFound())
+        return 1
+      }
+    } else if (payloads.some(({minifiedFilePath}) => hasSourceCodeContext(minifiedFilePath))) {
```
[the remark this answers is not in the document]
jhssilva: Addressed in [caba181](https://github.com/DataDog/datadog-ci/pull/2470/commits/caba181b3f0ae19f05c5a11657d6e4b7dadddae1)

packages/base/src/commands/sourcemaps/README.md
jhssilva: Thanks for the review. 

Yes, this won't trigger a re-check for backward compatibility.

Context:
The datadog-ci (inject command) and bundle plugin inject on the minified file a trigger to populate the window.DD_SOURCE_CODE_CONTEXT when the DOM renders on the browser. 

If a minified file has window.DD_SOURCE_CODE_CONTEXT, the browser SDK will be trigger errors with debug id. If not, the browser sdk triggers errors with service and version. 

---
Regarding, the `support 2 user flows`
Yes, for both. Adding:
Current, the bundle plugin currently injects DD_SOURCE_CODE_CONTEXT and it can upload directly via the bundle or after with datadog-ci sourcemaps upload --debug-id flag.

packages/base/src/commands/sourcemaps/__tests__/upload.test.ts
jhssilva: Addressed in commit 4636984e: moved withTempDirectory() to the shared testing tools and now use it in both sourcemap test files. https://github.com/DataDog/datadog-ci/commit/4636984eda2e365f215014f2eef2e091b4a4cfb4

packages/base/src/commands/sourcemaps/debugId.ts
jhssilva: Addressed in commit 4636984e: buildSnippet() now reuses SOURCE_CODE_CONTEXT_MARKER. https://github.com/DataDog/datadog-ci/commit/4636984eda2e365f215014f2eef2e091b4a4cfb4

packages/base/src/commands/sourcemaps/README.md
Drarig29: Just a documentation idea for your team: I see https://docs.datadoghq.com/real_user_monitoring/guide/upload-javascript-source-maps and https://docs.datadoghq.com/real_user_monitoring/application_monitoring/browser/build_plugins aren't cross-linking each other and I couldn't find a central page with all available options and user flows. If I was a user, I would like to see the bigger picture to choose the best setup for my needs

Drarig29 (approved):
LGTM on admin side

packages/base/src/commands/sourcemaps/README.md
jhssilva: Thanks for pointing this out.  We'll be updating them.

drichards-87 (approved):

buranmert (approved):
cc @thomasbertet this should eliminate the case which we discussed about debugID - service/version mismatch

packages/base/src/commands/sourcemaps/debugId.ts
```diff
@@ -40,23 +42,59 @@ export const extractDebugId = (filePath: string): string | undefined => {
...
+      while (true) {
+        const {bytesRead} = await fileHandle.read(buffer, 0, SOURCE_CODE_CONTEXT_SEARCH_CHUNK_BYTES, position)
+        if (bytesRead === 0) {
+          return false
+        }
+
+        const searchableContent = overlap + buffer.toString('utf8', 0, bytesRead)
+        if (searchableContent.includes(SOURCE_CODE_CONTEXT_MARKER)) {
+          return true
+        }
```
amortemousque: **🔨 warning:** ‏DD_SOURCE_CODE_CONTEXT can contain fields other than debugId. For example, it’s also used for MFE attribution, so we can’t assume that the presence of DD_SOURCE_CODE_CONTEXT means debugId is enabled. We could check for `ddDebugId`

jhssilva: Thanks for the review. 
We're adding a verification if ddDebugId exists.

jhssilva: [f743819](https://github.com/DataDog/datadog-ci/pull/2470/commits/f743819e4a91514bc1c16b6ef571d94fee9cf257) Validated if debug id it's set inside the `DD_SOURCE_CODE_CONTEXT`

packages/base/src/commands/sourcemaps/debugId.ts
```diff
@@ -40,23 +50,59 @@ export const extractDebugId = (filePath: string): string | undefined => {
...
     return undefined
   }
 }
 
+export const extractDebugId = (filePath: string): string | undefined => searchFile(filePath, matchDebugId)
+
+// Default service/version uploads may contain many large bundles. Scan them asynchronously in
+// larger chunks so checking marker-free files does not block the event loop with thousands of
+// small reads. Keep an overlap so markers split across chunk boundaries are still detected.
+export const hasSourceCodeContext = async (filePath: string): Promise<boolean> => {
```
amortemousque: **💬 suggestion:** ‏We could use extractDebugId instead no?

jhssilva: Implemented [56c9ba2](https://github.com/DataDog/datadog-ci/pull/2470/commits/56c9ba22829bf40ec8276de3dc1b3dfb44940545)

amortemousque (approved):

PR #2455, opened 2026-08-13
[SVLS-9302] cloud run ssi e2e tests

ava-silver:
### What and why?

Verify that single-language tracer injection works on real Cloud Run services without an application-image rebuild. This PR changes E2E coverage only; it adds no customer-facing CLI behavior.

### How?

Each case deploys a tracer-free Java, Node.js, .NET, Python, Ruby, or PHP fixture from [serverless-ci#208](https://github.com/DataDog/serverless-ci/pull/208), instruments it, checks its configuration, sends traffic, checks for spans, and deletes the service. Fixture and tracer images intentionally use `latest`.

Instrumentation uses the same Datadog organization as the span checker, so CI's separate CI Visibility credentials do not redirect workload telemetry.

### Manual validation

A clean local run passes configuration, traffic, and span checks for .NET, Node.js, PHP, Python, and Ruby. Retained Cloud Run logs showed that the Ruby `latest` fixture still referenced a build from before the merged production-mode and route changes. Rebuilding that controlled artifact from `serverless-ci/main` resolves the request timeouts; Ruby now returns three HTTP 200 responses and produces spans.

Java remains blocked because its published init image lacks `/datadog-init/probe-server`. This is the only remaining runtime blocker.

### How to validate

Authenticate `gcloud` and application-default credentials. Set `GCP_PROJECT_ID`, `GCP_REGION`, and Datadog credentials: `DATADOG_API_KEY` and `DATADOG_APP_KEY` must belong to the same organization, with span-read permission for the application key. The GCP identity needs Cloud Run deployment/deletion permissions and access to the fixture images in `datadog-serverless-gcp-dev/e2e-workloads`.

```sh
DATADOG_CI_COMMAND='yarn launch' \
SKIP_CLOUD_RUN_TESTS=false IS_STANDALONE_BINARY=false \
yarn e2e:test e2e/serverless/cloud-run/cloud-run-ssi.test.ts --runInBand
```

This creates billable, publicly invokable temporary services and attempts to delete them after each case. Expect all six cases to pass configuration, traffic, and span checks after the Java init-image release. Check cleanup output for any service that needs manual deletion.

### Review checklist

- [ ] Feature or bugfix MUST have appropriate tests (unit, integration)

PR #2418, opened 2026-07-30
[SVLS-9302] enable Cloud Run automatic APM instrumentation

ava-silver:
### What and why?

Enable APM tracing for Cloud Run applications without rebuilding their images to install a tracer.

- Use `--tracing inject --language` for Java, Node.js, .NET, Python, Ruby, or PHP. Existing manual/disabled tracing modes and boolean aliases remain supported.
- Configure the tracer version, libc, readiness port, and memory or disk storage. Disk storage uses a 10 GiB Preview volume and requires Gen2 and at least the BETA launch stage.
- Remove injected tracer state when switching to manual/disabled tracing or running `uninstrument`, while preserving unrelated application configuration.

### How?

A managed `datadog-tracer` container copies the tracer into a shared volume. Startup dependencies wait for the tracer and Agent before starting the selected application container with its language-specific injection configuration.

### Manual validation

Python injection produced [traces on a disposable Cloud Run service](https://ddserverless.datadoghq.com/apm/traces?query=env%3Assi-validation%20service%3Assi-validate-python-ava-08131404&start=1786641090579&end=1786648290579&paused=true) using a probe-server-capable tracer image.

A clean local E2E run with published `latest` images passes configuration, traffic, and span checks for .NET, Node.js, PHP, Python, and Ruby. The Ruby fixture artifact was rebuilt from the merged [serverless-ci#208](https://github.com/DataDog/serverless-ci/pull/208) source after retained Cloud Run logs showed that `latest` still referenced a pre-production fixture build. Java remains blocked only by the missing init-image probe server.

### How to validate

Use a disposable, tracer-free Cloud Run service, authenticated `gcloud` and application-default credentials, and `DD_API_KEY`/`DD_SITE`. Set `PROJECT`, `REGION`, `SERVICE`, and `LANGUAGE` for that workload.

```sh
yarn launch cloud-run instrument \
  --project "$PROJECT" --region "$REGION" --service "$SERVICE" \
  --tracing inject --language "$LANGUAGE" --no-source-code-integration
```

The command defaults to the `latest` tracer image. Expect a ready revision, an unchanged application image, and a `datadog-tracer` container and volume. Send requests to the service and confirm traces in Datadog. Repeat with `--tracer-volume-medium disk` to validate the Preview storage mode.

```sh
yarn launch cloud-run uninstrument \
  --project "$PROJECT" --region "$REGION" --service "$SERVICE"
```

[...]

PR #2472, opened 2026-08-28
[SVLS-9316] cloud run multi language instrumentation

ava-silver:
### What and why?

Cloud Run customers should be able to enable automatic tracer injection without knowing their application's language.

- `--tracing inject` without `--language` uses the composite tracer to detect Java, Node.js, .NET, Python, Ruby, or PHP.
- `--language` preserves explicit single-language injection.
- Instrumentation validates all selected services before applying updates, skips no-op updates, and reports partial completion.

### How?

The command mounts the composite tracer at `/opt/datadog-packages`, activates it in the application container, and removes owned single-language or composite-tracer state during cleanup. Automatic detection uses a 1.5 GiB memory-backed volume and a 2 GiB tracer-container memory limit. Both injection modes support Preview disk storage.

### Manual validation

A local run with published `latest` images passed configuration, traffic, and span checks for Java, Node.js, .NET, Python, Ruby, and PHP. The run deleted its temporary services.

### How to validate

Use a disposable, tracer-free Cloud Run service with authenticated `gcloud`, application-default credentials, `DD_API_KEY`, and `DD_SITE`.

```sh
yarn launch cloud-run instrument \
  --project "$PROJECT" --region "$REGION" --service "$SERVICE" \
  --tracing inject --no-source-code-integration
```

Confirm that the ready revision mounts the composite tracer at `/opt/datadog-packages`, sends traces, skips a repeated identical command, and removes Datadog-owned resources after `cloud-run uninstrument`.

The [Cloud Run E2E prerequisites and command](https://github.com/DataDog/datadog-ci/pull/2455) include the automatic-detection Node.js case. Use `--testNamePattern='detects and injects'` to run it.

### Review checklist

- [x] Feature or bugfix has appropriate tests.

### DIFF
--- a/packages/plugin-cloud-run/src/service-config.ts
+++ b/packages/plugin-cloud-run/src/service-config.ts
@@ -121,7 +121,7 @@ export const instrumentServiceConfig = (service: IService, options: InstrumentSe
       )
     }
   } else {
-    if (!hasSsi && (hasTracerContainer || hasTracerVolume)) {
+    if (existingSsiMode === undefined && (hasTracerContainer || hasTracerVolume)) {
       const resources = [hasTracerContainer ? 'container' : undefined, hasTracerVolume ? 'volume' : undefined].filter(
         (resource): resource is string => resource !== undefined
       )
@@ -138,9 +138,6 @@ export const instrumentServiceConfig = (service: IService, options: InstrumentSe
       reservedContainerNames(options.sidecarName)
     )
     assertTracerReadinessPortAvailable(mainContainer, options.sidecarName, tracerReadinessPort, healthCheckPort)
-    if (!hasSsi) {
-      assertTracerMountPathAvailable(mainContainer)
-    }
     sourceTemplate =
       existingSsiMode !== undefined || hasTracerContainer
         ? removeExistingSsiState(sourceTemplate, mainContainer, existingSsiMode)
@@ -149,6 +146,7 @@ export const instrumentServiceConfig = (service: IService, options: InstrumentSe
       sourceTemplate.containers ?? [],
       reservedContainerNames(options.sidecarName)
     )
+    assertTracerMountPathAvailable(updatedMainContainer, getTracerMountPath(ssiConfig))
     assertInjectionEnvCanBeMerged(updatedMainContainer.env, ssiConfig)
     targetContainers = new Set([updatedMainContainer])
     envVarsByName[DD_TRACE_ENABLED_ENV_VAR] = {name: DD_TRACE_ENABLED_ENV_VAR, value: 'true'}
@@ -185,7 +183,7 @@ export const instrumentServiceConfig = (service: IService, options: InstrumentSe
   if (ssiConfig.kind === 'single-language' || ssiConfig.kind === 'multi-language') {
     const mainContainer = selectMainContainer(template.containers ?? [], reservedContainerNames(options.sidecarName))
     const isMultiLanguage = ssiConfig.kind === 'multi-language'
-    const mountPath = isMultiLanguage ? COMPOSITE_TRACER_MOUNT_PATH : TRACER_MOUNT_PATH
+    const mountPath = getTracerMountPath(ssiConfig)
     const configuredMainContainer = {
       ...mainContainer,
       env: isMultiLanguage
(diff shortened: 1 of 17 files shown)


ava-silver:
* **#2472** <a href="https://app.graphite.com/github/pr/DataDog/datadog-ci/2472?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/datadog-ci/2472?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2455** <a href="https://app.graphite.com/github/pr/DataDog/datadog-ci/2455?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2418** <a href="https://app.graphite.com/github/pr/DataDog/datadog-ci/2418?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

brett0000FF (approved):

packages/plugin-cloud-run/src/service-config.ts
```diff
@@ -360,42 +382,60 @@ const assertTracerReadinessPortAvailable = (
...
-  return removeDependencies(cleaned, MANAGED_TRACER_CONTAINER_NAMES)
-}
-
 const isManagedTracerContainer = (container: IContainer): boolean =>
   typeof container.name === 'string' && MANAGED_TRACER_CONTAINER_NAMES.has(container.name)
 
+const isCompleteManagedTracerContainer = (container: IContainer): boolean =>
+  isManagedTracerContainer(container) &&
+  (container.image === COMPOSITE_TRACER_IMAGE ||
+    /^gcr\.io\/datadoghq\/dd-lib-(?:java|js|dotnet|python|rb|php)-init:[\w.-]+$/.test(container.image ?? '')) &&
```
avangelillo: Claude says there may be a bug here


>Confirmed bug — packages/plugin-cloud-run/src/service-config.ts:432:
> 
> const isCompleteManagedTracerContainer = (container: IContainer): boolean =>
  isManagedTracerContainer(container) &&
  (container.image === COMPOSITE_TRACER_IMAGE ||
    /^gcr\.io\/datadoghq\/dd-lib-(?:java|js|dotnet|python|rb|php)-init:[\w.-]+$/.test(container.image ?? '')) &&
> 
> The regex matches Ruby's tracer image as dd-lib-rb-init, but the actual image built by buildSingleLanguageTracerImage (per LANGUAGE_METADATA.ruby.tracerLanguage = 'ruby') is dd-lib-ruby-init:<version>. Looks like rb got copy-pasted from the wrong metadata field — that's the repository name (dd-trace-rb), not the language segment used in the image tag.

packages/plugin-cloud-run/src/service-config.ts
```diff
@@ -28,8 +38,12 @@ const DISK_VOLUME_SIZE_LIMIT = '10Gi'
 const GEN2_EXECUTION_ENVIRONMENT = 2 as const // google.cloud.run.v2.ExecutionEnvironment.EXECUTION_ENVIRONMENT_GEN2
 const SSI_INJECTION_MODE_LABEL = 'dd_sls_injection_mode'
 const SINGLE_LANGUAGE_SSI_MODE = 'single_language'
+const MULTI_LANGUAGE_SSI_MODE = 'multi_language'
 const LEGACY_TRACER_CONTAINER_NAME = 'datadog-tracer-copy'
 const MANAGED_TRACER_CONTAINER_NAMES = new Set([TRACER_CONTAINER_NAME, LEGACY_TRACER_CONTAINER_NAME])
+const MANAGED_TRACER_MOUNT_PATHS = new Set([TRACER_MOUNT_PATH, COMPOSITE_TRACER_MOUNT_PATH])
+const MULTI_LANGUAGE_VOLUME_SIZE_LIMIT = '1.5Gi'
+const MULTI_LANGUAGE_TRACER_MEMORY_LIMIT = '2Gi'
```
avangelillo, line 46: Do either of these need to be configurable?

ava-silver: not right now, we don't have them configurable anywhere else, and a customer who wants a low-touch instrumentation option wouldn't be configuring these.

packages/plugin-cloud-run/src/service-config.ts
ava-silver: oh good callout -- im gonna rip out the hard coded regex and make it dynamic based on the language metadata we already use elsewhere

avangelillo (approved):

PR #2473, opened 2026-08-28
[SVLS-9308] enable Container Apps tracer injection

ava-silver:
### What and why?

Azure Container Apps users currently need to include an APM tracer in their application image. This adds automatic instrumentation for Java, Node.js, .NET, Python, Ruby, and PHP without rebuilding the image, while retaining the Agent sidecar for trace transport.

This aligns the CLI lifecycle with the [Azure Container Apps Terraform implementation](https://github.com/DataDog/terraform-azurerm-container-app-datadog/pull/56).

### How?

Use `--tracing inject --language <language>` to copy one tracer through an init container into a selected application container. Tracing defaults declaratively to `manual`; `disabled` turns it off. Container Apps accepts only `manual`, `inject`, and `disabled` because it has no legacy tracing flag to preserve.

Re-instrumentation removes the complete known Datadog-managed tracer projection before applying the requested state. Full `uninstrument` performs the same exhaustive tracer cleanup, then removes Agent instrumentation.

### Review checklist

### DIFF
--- a/packages/base/src/commands/container-app/instrument.ts
+++ b/packages/base/src/commands/container-app/instrument.ts
@@ -1,7 +1,7 @@
 import type {ContainerAppConfigOptions} from './common'
 
 import {Command, Option} from 'clipanion'
-import {isNumber} from 'typanion'
+import * as t from 'typanion'
 
 import {executePluginCommand} from '../../helpers/plugin'
 import {
@@ -11,8 +11,18 @@ import {
   DEFAULT_SIDECAR_NAME,
   SIDECAR_IMAGE,
 } from '../../helpers/serverless/constants'
+import {DEFAULT_TRACER_LIBC, LIBCS} from '../../helpers/serverless/ssi/injection-spec'
+import {
+  DEFAULT_TRACER_VERSION,
+  LANGUAGE_METADATA,
+  TRACER_IMAGE_TAG_REG_EXP,
+  type Language,
+} from '../../helpers/serverless/ssi/tracer'
+import {TRACING_INPUTS} from '../../helpers/serverless/ssi/tracing'
 
 import {ContainerAppCommand} from './common'
+
+export const CONTAINER_APP_LANGUAGES = [...(Object.keys(LANGUAGE_METADATA) as Language[]), 'go'] as const
 
 export const DEFAULT_SIDECAR_CPU = 0.5
 export const DEFAULT_SIDECAR_MEMORY = 1
@@ -50,14 +60,39 @@ export class ContainerAppInstrumentCommand extends ContainerAppCommand {
   })
   private sidecarCpu = Option.String('--sidecar-cpu', {
     description: `The number of CPUs to allocate to the sidecar container. Defaults to '${DEFAULT_SIDECAR_CPU}'.`,
-    validator: isNumber(),
+    validator: t.isNumber(),
   })
   private sidecarMemory = Option.String('--sidecar-memory', {
     description: `The amount of memory (in GiB) to allocate to the sidecar container. Defaults to '${DEFAULT_SIDECAR_MEMORY}'.`,
-    validator: isNumber(),
+    validator: t.isNumber(),
   })
   private sidecarImage = Option.String('--sidecar-image', SIDECAR_IMAGE, {
     description: `Override to pin a specific version tag or to use a mirrored image from a custom registry (e.g., ACR) to avoid pull rate limits. Defaults to '${SIDECAR_IMAGE}'`,
+  })
+  private tracing: ContainerAppConfigOptions['tracing'] = Option.String('--tracing', {
+    description:
+      'Configure APM instrumentation. Use "manual" when the tracer is installed, "inject" with --language for automatic instrumentation, or "disabled" to turn tracing off. The legacy values "true"/"1" and "false"/"0" map to "manual" and "disabled".',
+    validator: t.isEnum(TRACING_INPUTS),
+  })
+  private language: ContainerAppConfigOptions['language'] = Option.String('--language', {
+    description: `Set the application language for log parsing. With --tracing inject, this also selects the tracer. Possible values: ${CONTAINER_APP_LANGUAGES.map(
+      (language) => `"${language}"`
+    ).join(', ')}.`,
+    validator: t.isEnum(CONTAINER_APP_LANGUAGES),
+  })
+  private tracerVersion = Option.String('--tracer-version', {
+    description: `Set the tracer image tag for automatic instrumentation. Defaults to '${DEFAULT_TRACER_VERSION}'.`,
+    validator: t.cascade(t.isString(), t.matchesRegExp(TRACER_IMAGE_TAG_REG_EXP)),
+  })
+  private tracerLibc: ContainerAppConfigOptions['tracerLibc'] = Option.String('--tracer-libc', {
+    description: `Set the C standard library used by the application image. Possible values: ${LIBCS.map(
+      (libc) => `"${libc}"`
+    ).join(', ')}. Defaults to '${DEFAULT_TRACER_LIBC}'.`,
+    validator: t.isEnum(LIBCS),
+  })
+  private containerName = Option.String('--container-name', {
+    description:
+      'Select the application container to instrument when the Container App has multiple application containers.',
   })
 
   private sourceCodeIntegration = Option.Boolean('--source-code-integration,--sourceCodeIntegration', true, {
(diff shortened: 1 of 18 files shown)


ava-silver:
* **#2474** <a href="https://app.graphite.com/github/pr/DataDog/datadog-ci/2474?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2473** <a href="https://app.graphite.com/github/pr/DataDog/datadog-ci/2473?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/datadog-ci/2473?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

ava-silver:
@codex review

Drarig29 (approved):

packages/plugin-container-app/README.md
```diff
@@ -124,6 +149,11 @@ You can pass the following arguments to `instrument` to specify its behavior. Th
 | `--sidecar-cpu` |  | The number of CPUs to allocate to the sidecar container. | `0.5` |
 | `--sidecar-memory` |  | The amount of memory (in GiB) to allocate to the sidecar container. | `1` |
 | `--sidecar-image` |  | Override to pin a specific version tag or to use a mirrored image from a custom registry (e.g., ACR) to avoid pull rate limits. | `index.docker.io/datadog/serverless-init:latest` |
+| `--tracing` |  | Configure APM instrumentation. Use "manual" when the tracer is installed, "inject" with --language for automatic instrumentation, or "disabled" to turn tracing off. The legacy values "true"/"1" and "false"/"0" map to "manual" and "disabled". |  |
```
jeff-morgan-dd: Non-blocking style suggestion: You could use backticks for the value examples here (`manual`, `inject`, etc). Also for the `--language`  mention in the description

packages/plugin-container-app/README.md
```diff
@@ -124,6 +149,11 @@ You can pass the following arguments to `instrument` to specify its behavior. Th
...
+| `--language` |  | Set the application language for log parsing. With --tracing inject, this selects a supported tracer. |  |
+| `--tracer-version` |  | Set the tracer image tag for automatic instrumentation. | `latest` |
+| `--tracer-libc` |  | Set the C standard library used by the application image. Possible values: "glibc", "musl". | `glibc` |
```
jeff-morgan-dd, line 152: Same note as above re: backticks for values

jeff-morgan-dd (approved):

packages/plugin-container-app/src/ssi.ts
```diff
@@ -0,0 +1,453 @@
...
+/** Removes exact native tracer fragments for every supported language. */
+export const removeLanguageInjectionEnv = (existingEnv: readonly EnvironmentVar[] | undefined): EnvironmentVar[] =>
+  (existingEnv ?? []).flatMap((variable) => {
+    if (!variable.name || variable.secretRef || !variable.value) {
+      return [variable]
+    }
+
+    const fragments = LANGUAGE_ENV_FRAGMENTS.filter(({name}) => name === variable.name)
+    const withoutTag = variable.name === DD_TAGS_ENV_VAR ? removeInjectionModeTag(variable.value) : variable.value
+    const value = fragments.reduce<string | undefined>(removeEnvFragment, withoutTag)
```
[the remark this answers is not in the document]
ava-silver: We are intentionally treating SSI state as a declarative Datadog-managed projection. Once existing SSI is recognized, every transition removes all exact known SSI fragments and deterministic resources that Datadog could have added, then rebuilds the requested state. Full uninstrumentation runs the same exhaustive removal without rebuilding. This handles stale state and customer drift without relying on fragile active-language inference. Exact fragment removal still preserves unrelated parts of compound environment values, while the managed names and /datadog-lib path are treated as reserved.

packages/plugin-container-app/src/commands/instrument.ts
```diff
@@ -210,9 +255,38 @@ export class PluginCommand extends ContainerAppInstrumentCommand {
...
+        config.containerName
+      )
+      assertLanguageInjectionEnvCanBeMerged(containerApp.template?.containers?.[targetIndex]?.env, ssiConfig.spec)
+      if (ssiExists) {
+        sourceApp = removeSsiState(containerApp)
+      } else {
+        assertSsiResourcesCanBeAdded(containerApp, targetIndex, config.sidecarName!)
+      }
+    } else if (ssiExists) {
+      sourceApp = removeSsiState(containerApp)
```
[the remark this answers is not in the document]
ava-silver: For Container Apps, omission now intentionally defaults to `manual`; it is not a preserve operation. This command did not previously expose a dedicated tracing option, so we can define the new option declaratively with only three states: `manual`, `inject`, and `disabled`. A customer without a tracer still emits no traces in manual mode, while a customer-installed tracer is enabled by default. Re-running without `--tracing inject` therefore intentionally removes CLI-managed injection and returns the app to manual tracing. The help and README now state this default, and the command warns when an omitted value removes existing automatic injection. Cloud Run keeps preserve-on-omission because its historical public `--tracing` behavior requires compatibility.

packages/base/src/helpers/serverless/ssi/tracing.ts
```diff
@@ -0,0 +1,21 @@
...
+export type TracingInput = (typeof TRACING_INPUT_METADATA)[number]['input']
+export type TracingMode = (typeof TRACING_INPUT_METADATA)[number]['mode']
+
+export const TRACING_INPUTS = TRACING_INPUT_METADATA.map(({input}) => input)
+
+const TRACING_MODE_BY_INPUT = Object.fromEntries(
+  TRACING_INPUT_METADATA.map(({input, mode}) => [input, mode])
+) as Record<TracingInput, TracingMode>
+
+export const normalizeTracingInput = (input: TracingInput | undefined): TracingMode | undefined =>
```
ojproductions: normalizeTracingInput and TRACING_INPUTS are unused. They exist to map true/1 to manual and false/0 to disabled, but Container Apps then filters those aliases out in CONTAINER_APP_TRACING_INPUTS. Either call this helper from the command so the aliases work, or drop it until a second command needs it.

packages/base/src/commands/container-app/instrument.ts
```diff
@@ -50,15 +52,39 @@ export class ContainerAppInstrumentCommand extends ContainerAppCommand {
...
-    validator: isNumber(),
+    validator: t.isNumber(),
   })
   private sidecarImage = Option.String('--sidecar-image', SIDECAR_IMAGE, {
     description: `Override to pin a specific version tag or to use a mirrored image from a custom registry (e.g., ACR) to avoid pull rate limits. Defaults to '${SIDECAR_IMAGE}'`,
   })
+  private tracing: ContainerAppConfigOptions['tracing'] = Option.String('--tracing', {
+    description:
+      'Configure APM instrumentation. Use `manual` when the tracer is installed, `inject` with `--language` for automatic instrumentation, or `disabled` to turn tracing off. Defaults to `manual`.',
+    validator: t.isEnum(CONTAINER_APP_TRACING_INPUTS),
```
ojproductions: This enum rejects --tracing true, false, 1, and 0. Other instrumentations (e.g Cloud Run) already maps those to manual and disabled

packages/plugin-container-app/src/ssi.ts
```diff
@@ -0,0 +1,475 @@
...
+          ', '
+        )}.`,
+      ],
+      warnings: [],
+    }
+  }
+  if (!isTracerInjectionLanguage(config.language)) {
+    return {
+      kind: 'errors',
+      errors: [`--tracing inject supports only these languages: ${TRACER_INJECTION_LANGUAGES.join(', ')}.`],
```
ojproductions: --tracing inject --language go hits this generic list even though Go can’t be injected at startup. Suggest to reject go with an error that tells the user to install dd-trace-go and use --tracing manual.

packages/plugin-container-app/src/commands/instrument.ts
ojproductions, line 1: instrumentSidecar still diffs the live config after listSecrets, so DD_API_KEY can show up in stdout on dry-run and apply. This PR redacts that in uninstrument (redactSecrets around the generateConfigDiff call) but not here. suggest to diff a redacted copy in this file too, and keep real secret values only on the Azure update.

ojproductions (commented):
Looks good overall but found a few things

packages/base/src/commands/container-app/instrument.ts
ava-silver: Yep, Container Apps intentionally does not accept the boolean aliases. Unlike Cloud Run, it had no historical dedicated `--tracing` option to preserve, so the new declarative interface exposes only `manual`, `inject`, and `disabled`, with omission defaulting to `manual`, which was analogous to what we were implementing before without a CLI option. Cloud Run keeps `true`/`1` and `false`/`0` solely for backward compatibility. I removed the unused alias-normalization machinery rather than enabling those aliases here.

ojproductions (approved):
lgtm

PR #2477, opened 2026-08-31
Strip characters XML cannot represent before building the JUnit report

### LINKED ISSUE #2475: [synthetics] A control character in a browser error description silently discards the entire JUnit report
### Summary

`synthetics run-tests --jUnitReport <path>` writes **no report at all** when any browser test result carries a `browser_errors[].description` containing raw response bytes. The whole document is discarded, not just the offending value.

The run prints:

```
❌ Couldn't write the JUnit report to synthetics-junit.xml:
Error: Invalid character in string: https://example.com/?ci=js/3.12.2&amp;q=<token><binary…> at index 81
```

In our case the description is a request to a third-party fingerprinting endpoint followed by its raw (compressed/encrypted) response body.

### Impact

- **The entire batch's report is lost**, not just the test that produced the bad string.
- **It is silent on green runs.** The write is wrapped in a `try/catch` that only logs, so the CLI still exits `0` and CI stays green while the JUnit consumer (Buildkite Test Engine, GitLab, …) receives nothing. One of our synthetics suites had been uploading zero test-analytics data for months before anyone noticed — every run looked fine.
- **It is misleading on red runs.** The message is printed last, immediately above the step's exit line, so it reads as the cause of a failure it cannot produce. We had engineers chasing "the build failed because the JUnit report failed to write" when the real cause was elsewhere.
- Passing tests are affected too — `browser_errors` are attached to a test case regardless of step outcome, so a fully green browser suite can still lose its report.

### Root cause

`packages/plugin-synthetics/src/reporters/junit.ts` → `getBrowserTestErrors()` copies the description verbatim into the XML tree:

```ts
if (step.browser_errors?.length) {
  browserErrors.push(
    ...step.browser_errors.map((e) => ({
      $: {type: e.type, name: e.name, step: step.description},
      _: e.description, // <-- not sanitized
    }))
  )
}
```

`runEnd()` then passes the whole document to `xml2js`'s `Builder`, which rejects C0 control characters (illegal in XML 1.0) and throws for the entire document:

```ts
try {
  this.writeReportToFile(this.builder.buildObject(this.json))
} catch (e) {
  this.write(`\n❌ Couldn't write the JUnit report to ${this.destination}:\n${e}\n`)
}
```

`step.failure.message` and `step.warnings[].message` go through the same unsanitized path and look equally exposed.

### Minimal reproduction

Three lines. One control character anywhere in a text node throws away the whole document:

```js
const {Builder} = require('xml2js')
new Builder().buildObject({testsuites: {testcase: {browser_error: 'response body\x1f\x8b'}}})
// Error: Invalid character in string: response body at index 13
```

In a real run you hit this whenever a browser test step reports a `browser_error` whose `description` embeds a raw response body.

### Suggested fix

Sanitize text nodes before building. Characters legal in XML 1.0 are `#x9 | #xA | #xD | [#x20-#xD7FF] | [#xE000-#xFFFD] | [#x10000-#x10FFFF]`:

```ts
const stripInvalidXmlChars = (s: string) =>
  s.replace(/[^\x09\x0A\x0D\x20-\uD7FF\uE000-\uFFFD\u{10000}-\u{10FFFF}]/gu, '')
```

Applied to `e.description`, `step.failure.message` and `w.message`.

Two things would help even without full sanitization:

1. **Degrade per-value, not per-report.** Dropping one `<browser_error>` is far better than dropping every test result in the batch — consider a per-test-case try/catch so one bad string cannot take the report with it.
2. **Make the failure visible.** Since the write error never changes the exit code, a green run gives no signal that its report is missing. A non-zero exit under `--failOnCriticalErrors`, or at minimum a clearly-marked warning, would stop this failing silently.

### Versions

- `datadog-ci` **v5.22.0** (`datadog/ci` Docker image); no sanitization present on `master` either as of 2026-08-28
- Command: `synthetics run-tests --batchTimeout <ms> --failOnCriticalErrors --jUnitReport synthetics-junit.xml --no-selectiveRerun --runName <name> --search 'tag:"<tag>" state:live'`

arpitjain099:
### What and why?

Fixes #2475.

A browser test result can carry raw response bytes in `browser_errors[].description`. Those bytes reach the XML tree verbatim, and `xml2js` refuses to build a document containing C0 control characters, so `buildObject` throws and the `try/catch` around it drops the report for the whole batch. Since the catch only logs, the run still exits 0, and a green CI job quietly uploads nothing.

Rather than sanitize the three or four known fields, this walks the finished document once in `runEnd` and strips the characters XML 1.0 cannot represent from every string in it. That covers `browser_errors[].description`, `failure.message` and `warnings[].message`, and also anything else the API might put in a name, a step description or an attribute. Tab, newline, carriage return and astral characters are all kept.

I left the two secondary suggestions in the issue alone. Degrading per value instead of per report, and making the write failure visible in the exit code, are behaviour changes worth deciding separately; this one is contained to what the reporter emits.

### Test coverage

`stripInvalidXmlChars` gets its own two cases, and there is a reporter-level test that pushes a result whose browser error, failure message and warning all carry control characters through `resultEnd` and `runEnd`, then asserts the file was written, the text survived minus the bad bytes, and no "Couldn't write the JUnit report" line was printed. All three fail on master.

`yarn jest packages/plugin-synthetics/src/__tests__/reporters/junit.test.ts` is 23 passing. `default.test.ts` and `mobile/app-upload.test.ts` fail in my checkout with the same 54 failures and 74 snapshot mismatches before and after this change, so they look unrelated. Lint is clean on both files.

### DIFF
--- a/packages/plugin-synthetics/src/__tests__/reporters/junit.test.ts
+++ b/packages/plugin-synthetics/src/__tests__/reporters/junit.test.ts
@@ -10,7 +10,7 @@ import type {PluginCommand as RunTestsCommand} from '../../commands/run-tests'
 import type {Device, Result, ServerTest} from '../../interfaces'
 import {ExecutionRule} from '../../interfaces'
 import type {Args, XMLTestCase} from '../../reporters/junit'
-import {getDefaultSuiteStats, getDefaultTestCaseStats, JUnitReporter} from '../../reporters/junit'
+import {getDefaultSuiteStats, getDefaultTestCaseStats, JUnitReporter, stripInvalidXmlChars} from '../../reporters/junit'
 
 import {
   BATCH_ID,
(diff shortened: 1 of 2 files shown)


Drarig29:
@codex review

arpitjain099:
The Categorize PR check needs a label I cannot set from outside the org. This one belongs under `synthetics`.

Drarig29 (approved):

PR #2478, opened 2026-09-01
[SYNTH-28953] Show each Test Suite member's own name/type in the default reporter

Drarig29:
### What and why?

`datadog-ci synthetics run-tests` was showing the suite's own name for every result of a triggered [Test Suite](https://docs.datadoghq.com/synthetics/test_suites/) member, instead of each member's own name. The batch result for a member already carries its own `test_name`/`test_type` directly from the backend, but `getResultFromBatch()` ignored them and only matched against the locally pre-fetched `Test` list, where a suite member is only known by public ID (via the suite's `memberPublicIds`) and inherits the suite's placeholder name.

Fixes https://datadoghq.atlassian.net/browse/SYNTH-28953

### How?

`BaseResultInBatch` now types the `test_name`/`test_type` fields the batch endpoint already returns per result. `getResultFromBatch()` builds each result's display `Test` from those fields directly, falling back to the pre-fetched test (or a minimal placeholder) only for the fields the batch result doesn't carry (`config`/`locations`/`options`). This also gets rid of the old `.find(...)!` non-null assertion, which could throw if a result's public ID didn't match anything in the known list — that path is now a normal fallback instead of a crash.

### Review checklist

### DIFF
--- a/packages/plugin-synthetics/src/interfaces.ts
+++ b/packages/plugin-synthetics/src/interfaces.ts
@@ -6,6 +6,7 @@ import type {ProxyConfiguration} from '@datadog/datadog-ci-base/helpers/utils'
 import type {Writable} from 'stream'
 
 export type SupportedReporter = 'junit' | 'json' | 'default'
+export type TestType = 'api' | 'browser' | 'mobile' | 'network'
 
 export interface MainReporter {
   log(log: string): void
@@ -116,7 +117,7 @@ export interface RawPollResult {
 
 export interface RawPollResultTest {
   id: string
-  type: 'browser' | 'api' | 'mobile'
+  type: TestType
   subtype?: string
   config: {
     request?: {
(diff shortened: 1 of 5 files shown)


packages/plugin-synthetics/src/interfaces.ts
```diff
@@ -262,7 +262,9 @@ export interface BaseResultInBatch {
   max_retries: number | null
   selective_rerun?: SelectiveRerunDecision
   status: Status
+  test_name: string
   test_public_id: string
+  test_type: 'api' | 'browser' | 'mobile'
```
gregoirechauvet: **suggestion:** I believe `network` runs are also supported?

gregoirechauvet (approved):
Looks good to me

packages/plugin-synthetics/src/interfaces.ts
Drarig29: true, I confirmed locally 👍

PR #2479, opened 2026-09-01
[SYNTH-28953] Align the JSON report's exposed fields with the JUnit report

Drarig29:
### What and why?

Builds on #2478 (already merged). Two changes, same underlying motivation: don't let raw backend data flow into datadog-ci's outputs unchecked.

First, `getResultFromBatch()` merged the poll-results endpoint's data onto a test with `deepExtend({}, test, pollResult?.test)` — a full deep merge of whatever the backend happened to send. We already know that's more than our types declare (poll results carry a full `options` object that our `PollResult['test']` type doesn't even mention), so anything the backend adds there in the future would silently show up wherever `result.test` gets read or serialized.

Second, and the main point of this PR: the JSON report (`--jsonReport`) currently writes the *entire* `Result`/`Test` object to disk, while the JUnit report only ever surfaces a specific, hand-picked set of fields. There's no reason for the two to diverge that much — as a starting point for GA, the JSON report should expose exactly what JUnit already exposes.

### How?

Added `utils/paths.ts`: a small dot-path picker (`pickPaths`) with `[]` array-wildcard support (e.g. `'result.steps[].failure.message'` means "every step's failure message"), plus `withPaths()` for merging in only specific paths from a source without touching anything else. This is custom — `get-value`/`set-value` (already a dependency, already used the same way in `deploy-tests-lib.ts`) only get/set a single fixed path with numeric array indices, with no way to project over every array element or to build a new object from multiple paths at once. What's borrowed from them is just the dot-path string convention already established in this package, not any code.

`batch.ts` now merges poll-result data via `withPaths(test, pollResult?.test, POLLED_TEST_FIELDS)`, an explicit 8-path list (`subtype`, `config.request.*`, `config.steps[].subtype`), each commented with which reporter/check actually reads it.

`reporters/json.ts` gets its own `JSON_REPORTER_EXPOSED_FIELDS` list, deliberately *not* derived from JUnit's hand-picked fields — it's a separate list that currently happens to match, so the two are free to drift independently later (e.g. if JSON ever needs something JUnit doesn't, or vice versa). Every result is projected through it before being pushed into the report.

This is a real, intentional change to the JSON report's output shape: it no longer contains the test's `config`/`locations`/full `options`, or anything else JUnit doesn't already show.

### How to read this PR

It's one commit but touches a new mechanism plus two call sites, so read it in this order rather than top-to-bottom by filename:

1. **`utils/paths.ts` + `__tests__/utils/paths.test.ts`** — the mechanism itself. `pickPaths()` builds a new object containing only the given paths; `withPaths()` uses it to override just those paths on an existing object. Worth reading the tests alongside the implementation — a couple of them (the null-intermediate and stray-empty-object cases) exist because the first version of this got those wrong.
2. **`batch.ts`** — the smaller of the two call sites. `POLLED_TEST_FIELDS` replaces the old `deepExtend`, with a comment on each path saying which reporter/check reads it.
3. **`reporters/json.ts`** — the actual point of the PR. `JSON_REPORTER_EXPOSED_FIELDS` is the new allowlist; everything else in this file is just `pickPaths()` being called before a result is pushed.
4. **`__tests__/reporters/json.test.ts`** — shows the shape of the change most concretely: one test asserts the exact projected output for a fixture result, the other asserts specific fields (`config`, `locations`, `options.device_ids`) are gone.
5. **`__tests__/batch.test.ts`** — mostly mechanical updates; a few assertions had been quietly relying on `deepExtend` leaking a mocked `pollResult.test.public_id` onto the real result, which `withPaths` (correctly) no longer does.

### Review checklist

### DIFF
--- a/packages/plugin-synthetics/src/__tests__/batch.test.ts
+++ b/packages/plugin-synthetics/src/__tests__/batch.test.ts
@@ -284,6 +284,51 @@ describe('waitForResults', () => {
     )
   })
 
+  test('should use the polled subtype/config for a triggered suite member, which is not known ahead of the batch results', async () => {
+    mockApi({
+      getBatchImplementation: async () => ({
+        ...deepExtend({}, batch),
+        results: [{...getPassedResultInBatch(), test_name: 'Member test', test_public_id: 'member-pid'}],
+      }),
+      pollResultsImplementation: async () => [
+        {
+          ...deepExtend({}, pollResult),
+          test: {
+            subtype: 'dns',
+            config: {request: {host: 'example.com', dnsServer: '1.1.1.1'}},
+          },
+        } as PollResult,
+      ],
+    })
+
+    const suiteTest: Test = {
+      ...apiTest,
+      public_id: 'suite-pid',
+      type: 'suite',
+      memberPublicIds: ['member-pid'],
+    } as Test
+
+    const results = await waitForResults(
+      api,
+      trigger,
+      [suiteTest],
+      {
+        batchTimeout: 120000,
+        datadogSite: DEFAULT_COMMAND_CONFIG.datadogSite,
+        failOnCriticalErrors: false,
+        subdomain: DEFAULT_COMMAND_CONFIG.subdomain,
+      },
+      mockReporter
+    )
+
+    const test = results[0].test as Test & {subtype: string}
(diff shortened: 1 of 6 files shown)


Drarig29:
@codex review

PR #2485, opened 2026-09-03
[SVLS-9308] full language coverage for container app single-language ssi e2e

ava-silver:
### What and why?

Azure Container Apps automatic tracer injection had end-to-end coverage only for Node.js. Runtime-specific injection failures for Java, .NET, Python, Ruby, and PHP could go undetected.

### How?

Parameterize the SSI case across all six supported runtimes, using the matching tracer repository and native environment fragment for each workload. Run the independent cases concurrently and verify injection, telemetry, idempotent reinstrumentation, cleanup, and resource deletion.

### Test plan

- [x] Manual Azure Container Apps SSI E2E: all six runtimes passed, including trace delivery and cleanup.

### Review checklist

- [x] Feature or bugfix MUST have appropriate tests (unit, integration)

PR #2486, opened 2026-09-03
[SVLS-9313] Add multi-language ssi support for container apps

ava-silver:
### What and why?

Azure Container Apps tracer injection required customers to specify an application language. This enables automatic runtime detection when they use `--tracing inject` without `--language`, while preserving explicit single-language injection.

### How?

The command copies the Azure composite tracer through the existing init-container lifecycle and activates `auto_inject`. It preserves declarative cleanup across manual, disabled, single-language, and multi-language transitions, and moves the shared composite contract into `packages/base` for other serverless plugins to reuse.

### Manual validation

A live Azure Container Apps test deployed a tracer-free Node.js 22 application, injected the production composite image, confirmed `_dd.injection.mode:serverless`, received traces in Datadog, and verified `uninstrument` cleanup.

### Review checklist

### DIFF
--- a/e2e/serverless/container-app/container-app-ssi.test.ts
+++ b/e2e/serverless/container-app/container-app-ssi.test.ts
@@ -23,7 +23,7 @@ const SSI_CASES = [
   })),
   {
     kind: 'multi-language',
-    testName: 'auto-detected Node.js',
+    testName: 'auto-detected Node.js at the 2-GiB ephemeral-storage boundary',
     applicationImage: 'dde2etfcapp.azurecr.io/node-ssi:latest',
   },
 ] as const
@@ -61,6 +61,7 @@ describeOrSkip('container-app automatic APM instrumentation', () => {
         ` --extra-tags "one_e2e_run_id:${runId}"` +
         ` --tracing inject` +
         (ssiCase.kind === 'single-language' ? ` --language "${ssiCase.language}"` : '') +
+        (ssiCase.kind === 'multi-language' ? ' --sidecar-cpu 0.25 --sidecar-memory 0.5' : '') +
         ` --no-source-code-integration`
 
       let lifecycleError: Error | undefined
--- a/e2e/serverless/container-app/container-app-ssi.test.ts
+++ b/e2e/serverless/container-app/container-app-ssi.test.ts
@@ -13,11 +13,20 @@ import {
   verifyUninstrumented,
 } from './container-app-verifier'
 
-const SSI_CASES = ssiCases.map(({fixtureImageName, nativeEnv, ...ssiCase}) => ({
-  ...ssiCase,
-  applicationImage: `dde2etfcapp.azurecr.io/${fixtureImageName}:latest`,
-  nativeEnv: {name: nativeEnv.name, fragment: nativeEnv.value},
-}))
+const SSI_CASES = [
+  ...ssiCases.map(({fixtureImageName, nativeEnv, ...ssiCase}) => ({
+    ...ssiCase,
+    kind: 'single-language' as const,
+    testName: ssiCase.language,
+    applicationImage: `dde2etfcapp.azurecr.io/${fixtureImageName}:latest`,
+    nativeEnv: {name: nativeEnv.name, fragment: nativeEnv.value},
+  })),
+  {
+    kind: 'multi-language',
+    testName: 'auto-detected Node.js',
+    applicationImage: 'dde2etfcapp.azurecr.io/node-ssi:latest',
+  },
+] as const
 
 const assertCommandSucceeded = (action: string, result: {exitCode: number; stdout: string; stderr: string}): void => {
   if (result.exitCode !== 0) {
@@ -36,11 +45,11 @@ describeOrSkip('container-app automatic APM instrumentation', () => {
   const resourceGroup = process.env.AZURE_RESOURCE_GROUP!
 
   it.concurrent.each(SSI_CASES)(
-    'injects, retries, traces, and removes the $language tracer',
+    'injects, retries, traces, and removes the $testName tracer',
     async (ssiCase) => {
-      const {language, applicationImage, tracerRepository, nativeEnv} = ssiCase
+      const {applicationImage} = ssiCase
       const runId = crypto.randomBytes(4).toString('hex')
-      const appName = `one-e2e-capp-ssi-${language}-${runId}`
+      const appName = `one-e2e-capp-ssi-${ssiCase.kind === 'single-language' ? ssiCase.language : 'auto'}-${runId}`
       const instrumentCommand =
         `${DATADOG_CI_COMMAND} container-app instrument` +
         ` -s "${subscriptionId}"` +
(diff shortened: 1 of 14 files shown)


ava-silver:
* **#2486** <a href="https://app.graphite.com/github/pr/DataDog/datadog-ci/2486?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/datadog-ci/2486?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2485** <a href="https://app.graphite.com/github/pr/DataDog/datadog-ci/2485?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

hestonhoffman (approved):

e2e/serverless/container-app/container-app-verifier.ts
```diff
@@ -34,6 +34,11 @@ const SHARED_VOLUME_NAME = 'shared-volume'
 const DD_API_KEY_SECRET_NAME = 'dd-api-key'
 const TRACER_NAME = 'datadog-tracer'
 const TRACER_MOUNT_PATH = '/datadog-lib'
+const COMPOSITE_TRACER_MOUNT_PATH = '/opt/datadog-packages'
+const COMPOSITE_TRACER_IMAGE = 'datadoghq.azurecr.io/dd-lib-composite-init:latest'
+const COMPOSITE_PRELOAD = `${COMPOSITE_TRACER_MOUNT_PATH}/datadog-apm-inject/stable/inject/launcher.preload.so`
+const NODE_TRACER_IMAGE = 'datadoghq.azurecr.io/dd-lib-js-init:latest'
```
ojproductions: NODE_TRACER_IMAGE variable is unused

e2e/serverless/container-app/container-app-verifier.ts
```diff
@@ -235,6 +240,52 @@ export const verifySsiInstrumented = (
   expect(tags.one_e2e_created).toBeDefined()
 }
 
+export const verifyMultiLanguageSsiInstrumented = (
```
ojproductions, line 243: This skips verifyDatadogEnv, sidecar LD_PRELOAD isolation, and tags.service / env / version / dd_sls_ci that verifySsiInstrumented already checks. A sidecar that received composite activation, or a missing DD_SERVICE, would still pass.

packages/plugin-container-app/src/__tests__/ssi.test.ts
```diff
@@ -204,6 +257,29 @@ describe('Container Apps automatic APM instrumentation', () => {
       ).not.toContainEqual(expect.objectContaining({volumeName: TRACER_VOLUME_NAME}))
     })
 
+    test('adds composite activation to the selected application container', () => {
```
ojproductions: The single-language test above asserts the sidecar has no tracer volume. Should also assert the sidecar also has no composite mount, LD_PRELOAD, or DD_INJECT_SENDER_TYPE.

packages/base/src/helpers/serverless/ssi/composite.ts
```diff
@@ -0,0 +1,28 @@
+import type {EnvFragment} from './env'
+import type {TracerRegistry} from './tracer'
+
+export const COMPOSITE_TRACER_MOUNT_PATH = '/opt/datadog-packages'
+export const COMPOSITE_TRACER_COMPLETION_MARKER = `${COMPOSITE_TRACER_MOUNT_PATH}/.datadog-composite-copy-finished`
+
+export interface CompositeInjectionSpec {
+  readonly image: string
+  readonly mountPath: string
+  readonly completionMarker: string
```
ojproductions: completionMarker is never read; Cloud Run still uses getTracerCopyCompletionMarker()

packages/base/src/helpers/serverless/ssi/tracer.ts
```diff
@@ -4,7 +4,8 @@ export const SINGLE_LANGUAGE_TRACER_REGISTRIES = [
   'datadoghq.azurecr.io',
 ] as const
 
-export type SingleLanguageTracerRegistry = (typeof SINGLE_LANGUAGE_TRACER_REGISTRIES)[number]
+export type TracerRegistry = (typeof SINGLE_LANGUAGE_TRACER_REGISTRIES)[number]
```
ojproductions: This list now backs the composite registry too. Would renaming SINGLE_LANGUAGE_TRACER_REGISTRIES to TRACER_REGISTRIES be more accurate?

e2e/serverless/container-app/container-app-ssi.test.ts
```diff
@@ -164,4 +169,88 @@ describeOrSkip('container-app automatic APM instrumentation', () => {
     },
     1_200_000
   )
+
+  it('detects Node.js, injects its tracer, and removes the composite', async () => {
```
ojproductions: This copies the create/instrument/traffic/uninstrument lifecycle from it.concurrent.each above, but drops the re-instrument retry. Parameterize that case (omit --language, swap the verifier) so both paths stay in sync.

ojproductions (commented):
AI Assisted

The implementation matches the ssi contract well. the main gaps are duplicated e2e lifecycle that already drifted (no retry, weaker verifier) and missing sidecar / shared-env assertions that the single-language tests already cover.

Highest severity is medium on both axes — fold the auto-detect e2e into the existing table and reuse verifyDatadogEnv plus sidecar isolation before merge.

packages/plugin-container-app/src/ssi.ts
```diff
@@ -451,25 +512,42 @@ const getInitContainerLanguage = (container: InitContainer): Language | undefine
...
 
-const buildTracerInitContainer = (spec: LanguageInjectionSpec): InitContainer => ({
+const buildTracerInitContainer = (image: string, mountPath: string): InitContainer => ({
   name: TRACER_CONTAINER_NAME,
-  image: spec.image,
+  image,
   command: ['/datadog-init/copy-lib.sh'],
-  args: [TRACER_MOUNT_PATH],
+  args: [mountPath],
   resources: {cpu: 0.25, memory: '0.5Gi', ephemeralStorage: '1Gi'},
```
ojproductions: These numbers are carried over from the single-language init container, but the composite image is much bigger. Two things: the Azure SDK marks ephemeralStorage read-only and drops it from the request, so this 1Gi never reaches Azure. And [Azure sizes a replica's temporary disk by its CPU](https://learn.microsoft.com/en-us/azure/container-apps/storage-mounts) — 1 GiB at 0.25 vCPU, 2 GiB at 0.5 — with the EmptyDir volume sharing that space. If the image doesnt fit, Azure evicts the replica rather than failing the deploy, and this gets hidden by the e2e app. should we measure the published image on Azure, then either raise the init container's CPU (that's what raises the disk limit) to account for the larger size? or maybe warn when the app has too little?

ojproductions (commented):
One last thing, but I think everything else is good

packages/plugin-container-app/src/ssi.ts
ava-silver: Good catch. It turns out that `ephemeralStorage` is response-only: the Azure REST schema and `@azure/arm-appcontainers` mark it read-only, so the SDK omits it from update requests. It cannot reserve 1 GiB.

Azure derives temporary storage from total replica CPU: 1 GiB at <=0.25 vCPU, 2 GiB at <=0.5, and 4 GiB at <=1. Init-container CPU does not independently raise that quota, so increasing it is not a viable fix. The default instrumented configuration has a 0.25-vCPU app and a 0.5-vCPU sidecar, for 0.75 vCPU and a 4-GiB quota.

I also measured the current Linux/amd64 composite image. `copy-lib.sh` copies only `package/`, not the full image, but that package occupies about 1 GiB on disk. A 1-GiB tier would therefore be unsafe, but the default 4-GiB configuration has sufficient headroom.

ojproductions (approved):
🚢

avangelillo (approved):

PR #2504, opened 2026-09-10
[SVLS-9317] automatic instrumentation linux code web apps

ava-silver:
### What and why?

Adds automatic (single-step) APM instrumentation for code-based Linux Azure Web Apps via a new `--apm-enabled` flag on `datadog-ci aas instrument` (SVLS-9317). Today, getting APM on a Linux web app means adding the Datadog tracer to the application itself. With `--apm-enabled`, customers get tracing on supported runtimes (.NET, Node.js, Python, PHP, Java) with no code changes, and `aas uninstrument` removes everything the new flow adds.

### How?

- On `--apm-enabled`, `instrument` resolves the app runtime from `linuxFxVersion`, downloads the matching tracer as an OCI fleet package from `install.datadoghq.com`, and stages it onto the app through the Kudu OneDeploy API under `/home/data/datadog-tracer/`. Injection happens through app settings (for example, `NODE_OPTIONS` for Node.js); the existing Datadog sidecar continues to ship telemetry.
- `uninstrument` removes the staged tracer and the injection app settings.
- The new Kudu client tolerates apps with no deployment history (Kudu answers 404 for `GET /api/deployments/latest`), polls deployments for up to 20 minutes since OneDeploy extraction onto the Azure Files-backed `/home` mount is slow on a busy plan, and includes the Kudu deployment log when a deployment fails.
- New e2e coverage in `e2e/serverless/aas/aas-linux-ssi.test.ts` runs in parallel with the existing Linux suite: it provisions a zero-dependency Node app, instruments it, asserts request spans, and uninstruments.

Manual validation beyond CI: instrumented a scratch Node web app in the e2e Azure subscription from this branch and confirmed request spans with the expected service/env/version/tags in Datadog.

### Review checklist


ava-silver:
* **#2504** <a href="https://app.graphite.com/github/pr/DataDog/datadog-ci/2504?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/datadog-ci/2504?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

drichards-87 (approved):

packages/plugin-aas/src/ssi-env.ts
```diff
@@ -0,0 +1,151 @@
...
+  const result = value
+    .split(' ')
+    .filter((part) => !remove(part))
+    .join(' ')
+    .trim()
+
+  return result || undefined
+}
+
+export const getAasSsiSpecEnv = (runtime: AasCodeRuntime, root: string): readonly EnvFragment[] =>
```
ojproductions: This export isnt used anywhere, you can remove this

packages/base/src/commands/aas/instrument.ts
```diff
@@ -63,6 +63,9 @@ export class AasInstrumentCommand extends AasCommand {
   private sidecarImage = Option.String('--sidecar-image', SIDECAR_IMAGE, {
     description: `Override to pin a specific version tag or to use a mirrored image from a custom registry (e.g., ACR) to avoid pull rate limits. Only applies to Linux web apps. Defaults to '${SIDECAR_IMAGE}'`,
   })
+  private apmEnabled = Option.Boolean('--apm-enabled', false, {
```
ojproductions, line 66: So far we've been using --tracing inject / manual / disabled as defined in the ssi contract rather than this --apm-enabled boolean flag. Is this a deliberate change? If not, we should align it

packages/plugin-aas/src/commands/instrument.ts
```diff
@@ -237,7 +248,41 @@ export class PluginCommand extends AasInstrumentCommand {
...
+        } else {
+          const staging = await resolveAasTracerStaging(runtime)
+          const baseEnvVars = getEnvVars({...config, isDotnet: false}, linuxSite, webApp)
+          // Merge before publishing so environment conflicts abort before any app mutation.
+          ssiEnvVars = mergeAasSsiEnv({...existingEnvVars, ...baseEnvVars}, runtime, staging.root)
+          await staging.publish(await getKuduClient(aasClient, resourceGroup, webApp))
+        }
+      } else if (removesSsi) {
+        ssiEnvVars = {...withoutSsiEnvVars, ...getEnvVars({...config, isDotnet: false}, linuxSite, webApp)}
+      }
```
ojproductions, line 307: **omitted flag removes injection with no warning, and does not delete staged files**
Container Apps warns: "Tracing defaults to manual… Use --tracing inject to retain…." Here removesSsi rewrites app settings and never calls Kudu deleteDirectory. Re-running aas instrument without the flag (today’s documented sidecar command) silently strips NODE_OPTIONS / CORECLR_* while leaving /home/data/datadog-tracer on disk.
Fix: Warn, then delete AAS_SSI_STAGING_ROOT on code-based Linux when leaving inject (same exhaustive cleanup as uninstrument, minus sidecar removal).

packages/plugin-aas/src/commands/uninstrument.ts
```diff
@@ -213,7 +209,42 @@ export class PluginCommand extends AasUninstrumentCommand {
...
+    const settings =
+      (
+        await (webApp.slot
+          ? client.webApps.listApplicationSettingsSlot(resourceGroup, webApp.name, webApp.slot)
+          : client.webApps.listApplicationSettings(resourceGroup, webApp.name))
+      ).properties ?? {}
+    await this.removeAasSsiEnvVars(config, webApp, client, resourceGroup, settings)
+    if (hasStagedAasTracer(settings) && !this.dryRun) {
+      await (await getKuduClient(client, resourceGroup, webApp)).deleteDirectory(AAS_SSI_STAGING_ROOT)
+    }
```
ojproductions, line 245: **uninstrument skips tracer delete when env markers are already gone**
Delete is gated on `hasStagedAasTracer(settings)`. env no longer contains /home/data/datadog-tracer/, so uninstrument also leaves the tree.
suggested fix: On Linux code apps (not SITECONTAINERS / DOCKER| / COMPOSE|), always delete AAS_SSI_STAGING_ROOT. Do not require env markers. Keep the containerized skip so Kudu is not required for sidecar-only containers.

packages/plugin-aas/src/ssi-stage.ts
```diff
@@ -0,0 +1,83 @@
...
+        method: 'GET',
+        url: thirdParty(`${FLEET_REGISTRY}/${repository}/blobs/${fleetPackage.layer.digest}`),
+        headers: {Accept: fleetPackage.layer.mediaType},
+        responseType: 'arraybuffer',
+        dispatcher: getProxyDispatcher(),
+        timeout: 120_000,
+      })
+      const zip = await buildFleetPackageZip(layer.data, fleetPackage.layer.digest, runtime)
+      await kudu.publish(root, zip)
+      if (!(await kudu.hasArtifacts(spec.artifacts))) {
```
ojproductions: This checks spec startup files only, not the rest of the staged tree, and does not retry after extract. The same CIFS /home mount already needs retries on delete; a false negative fails the run, and a partial tree that happens to contain init.js gets activated.

packages/plugin-aas/src/kudu.ts
```diff
@@ -0,0 +1,129 @@
...
+          continue
+        }
+        throw error
+      }
+    }
+
+    const parsedData =
+      typeof response.data === 'string'
+        ? response.data
+          ? (JSON.parse(response.data) as T)
```
ojproductions, line 79: Every Kudu body is JSON.parsed, including zip PUT responses. HTML/empty 403 pages become parse errors instead of an SCM-access message.

e2e/serverless/aas/aas-verifier.ts
```diff
@@ -120,6 +185,10 @@ export const verifyLinuxUninstrumented = (appName: string, rg: string, subscript
   const tags = app.tags || {}
   expect(Object.keys(tags)).not.toContain('service')
   expect(Object.keys(tags)).not.toContain('dd_sls_ci')
+  if (expectSsi) {
+    expect(tags.dd_sls_injection_mode).toBeUndefined()
+    verifyStagedTracer(appName, rg, false)
+  }
```
ojproductions, line 212: Uninstrument only asserts the SSI tag and that /home/data/datadog-tracer is gone. NODE_OPTIONS (and other injection settings) can remain and this still passes.

packages/plugin-aas/src/commands/instrument.ts
```diff
@@ -237,7 +248,41 @@ export class PluginCommand extends AasInstrumentCommand {
...
+          this.context.stdout.write(
+            `${this.dryRunPrefix}Staging the ${runtime.language} tracer for ${renderWebApp(webApp)}\n`
+          )
+        } else {
+          const staging = await resolveAasTracerStaging(runtime)
+          const baseEnvVars = getEnvVars({...config, isDotnet: false}, linuxSite, webApp)
+          // Merge before publishing so environment conflicts abort before any app mutation.
+          ssiEnvVars = mergeAasSsiEnv({...existingEnvVars, ...baseEnvVars}, runtime, staging.root)
+          await staging.publish(await getKuduClient(aasClient, resourceGroup, webApp))
+        }
```
ojproductions, line 299: **--dry-run --apm-enabled does not preview or validate injection**
Dry-run logs “Staging the … tracer” and skips mergeAasSsiEnv. ssiEnvVars stays undefined, so instrumentSidecar previews sidecar/agent settings only. Env conflicts (EnvFragmentConflictError, LD_PRELOAD 1024-byte limit) are not detected.
suggested fix: Always resolve runtime, merge env, and print the resulting settings. Gate only publish() and Azure writes on dryRun.

packages/plugin-aas/src/ssi-env.ts
```diff
@@ -0,0 +1,151 @@
...
+  const spec = getAasInjectionSpec(runtime, root)
+  const env = removeLegacyAasDotnetEnv(removeAasSsiEnv(current))
+
+  for (const fragment of spec.env) {
+    env[fragment.name] = mergeEnvFragment(env[fragment.name], fragment)
+  }
+  env.DD_TAGS = mergeInjectionModeTag(env.DD_TAGS)
+  env.DD_TRACE_ENABLED = 'true'
+
+  return env
```
ojproductions, line 68: **Key Vault-backed injection settings are merged as literals**
Container Apps rejects secretRef before merge (plugin-container-app/src/ssi.ts:451-454). AAS app settings use @Microsoft.KeyVault(...). If NODE_OPTIONS or JAVA_TOOL_OPTIONS is a Key Vault reference, mergeEnvFragment concatenates the reference with --require / -javaagent, which Azure will not resolve.

ojproductions (commented):
Happy path is sound, however left a few comments regarding the public command api and cleanup

packages/base/src/commands/aas/instrument.ts
ava-silver: Yes, deliberate. The flag shape follows whether tracing was already default-on for the platform:

- Cloud Run and Container Apps both had `DD_TRACE_ENABLED=true` as an established default before SSI (Container Apps got it from the shared instrumentation helper when we refactored it to use the same logic as Cloud Run). Both use the `--tracing` enum with declarative semantics -- `manual` as the default keeps tracing on, `inject` adds the tracer, `disabled` turns it off. Cloud Run additionally preserves its legacy `true`/`1`/`false`/`0` inputs; Container Apps only accepts `manual`/`inject`/`disabled` since it had no legacy tracing flag. A default-off boolean opt-in would have been a behavior regression on both.
- AAS never set anything tracing-related before, so tracer injection there is opt-in and default-off. That matches the planned net-new command contract of an explicit `--apm-enabled` toggle (mirroring `datadog_enable_apm` in Terraform and `APMFeatureConfig.isInstrumentationEnabled` in CDK). And since code-based Linux Web Apps detect the runtime from `linuxFxVersion`, there's no `--language` input for the `--tracing inject --language` shape to compose with.

Omitting `--apm-enabled` keeps the app settings identical to before; the only new omission behavior is declaratively removing SSI state that a previous `--apm-enabled` run added.

packages/plugin-aas/src/commands/uninstrument.ts
ava-silver: Agreed, we shouldn't gate the delete on env markers. I'll change uninstrument to attempt the staged-tree delete on code-based Linux apps regardless of markers (keeping the containerized skip so Kudu isn't required for sidecar-only containers). If SCM is inaccessible we'll log and continue rather than warn or fail -- at that point the files are stale and inactive, and anyone who had access to stage them likely has access to delete them.

packages/plugin-aas/src/ssi-env.ts
ava-silver: Good catch. I'll detect the `@Microsoft.KeyVault(` prefix on injection-target settings and reject with a clear error, mirroring the container-app secretRef rejection.

packages/plugin-aas/src/kudu.ts
ava-silver: I looked into this and I don't think it's a real issue:

- Non-2xx responses never reach the parse. `httpRequest` throws a `RequestError` on any non-2xx (helpers/request.ts checks `response.ok`), so an HTML 403 page surfaces as "Request failed with status code 403" through the retry loop's catch, not as a JSON parse error. 403 is also not in the retryable set, so it fails fast.
- The zip PUT returns an empty body, which is already handled. Kudu `ZipController.CreateDirectoryPutResponse` returns `Request.CreateResponse(HttpStatusCode.OK)` with no content, and the empty-string case takes the `undefined` branch without parsing. This path also ran successfully in the e2e test and in live manual validation, which would have failed on a parse error.

The only case that could hit a parse error is a 2xx with a non-JSON body (for example a proxy auth interstitial), and in that state SCM calls would be failing anyway. I'd rather keep the immediate SyntaxError there than silently treat it as absent data, which would turn it into a misleading "deployment incomplete" downstream. Happy to revisit if you have a repro where this actually produces a parse error.

packages/plugin-aas/src/commands/instrument.ts
ava-silver: Partially addressed: omission now warns "Removing automatic tracer injection from <app>. Use --apm-enabled to keep it." (mirroring the Container Apps warning) so the removal is no longer silent, but we intentionally keep the staged files. Deleting thousands of small files off the Azure Files /home mount takes up to a couple minutes, and a customer running plain `aas instrument` is still using Datadog -- inert tracer files on disk do no harm, and keeping them makes a later `--apm-enabled` run re-activate instantly instead of re-uploading. Full cleanup is the `uninstrument` contract: it removes the injection settings and deletes the whole staged tree (now unconditionally on code-based Linux apps, no longer gated on env markers).

packages/plugin-aas/src/ssi-stage.ts
ava-silver: Partially addressed: the startup-file check now retries (3 attempts, 5s apart) after extraction, so a CIFS enumeration false negative no longer fails the run.

Checking only the startup-critical files is deliberate, though. The layer digest is verified at download and the zip API provides transport integrity, so a partial tree that happens to contain the startup files but is missing other files isn't a credible failure mode -- and recursively checksumming the staged tree adds up to a minute or two of Azure Files latency to every run for no real protection. If activation succeeds (startup files present) the tracer loads from the same verified package contents.

packages/plugin-aas/src/ssi-env.ts
```diff
@@ -0,0 +1,164 @@
...
+    version: 'latest',
+    libc: runtime.libc,
+    root,
+  })
+
+// Azure resolves Key Vault references at runtime, so fragment-merging into one would corrupt the
+// reference and the app would receive the raw text.
+const KEY_VAULT_REFERENCE_PREFIX = '@Microsoft.KeyVault('
+
+export const mergeAasSsiEnv = (
```
avangelillo, line 46: Is there anything that doesn't get appropriately cleaned up tied to a specific version on re instrumenting?

Claude notes:
> Stale staged tracer versions are never cleaned up on re-instrument
>  (packages/plugin-aas/src/ssi-stage.ts)
>  Each instrument --apm-enabled run stages the tracer under a version+digest-suffixed path
>  (getStagedRoot), and mergeAasSsiEnv correctly repoints env vars to the new path — but
>  nothing deletes the previous staged directory. deleteDirectory is only ever called from
>  uninstrument. Repeatedly re-running instrument as tracer versions get bumped will leave
>  orphaned tracer trees accumulating forever on the app's Azure Files-backed /home quota
>  until a full uninstrument.

ava-silver: Yep, this is intentional, see this dicussion: https://github.com/DataDog/datadog-ci/pull/2504#discussion_r4005860749

we clean up config on re-instruments, but only do slow cleanup operations like deleting tracer files on `uninstrument` since stale tracer files won't cause any harm and they allow us to quickly add apm back in if the customer wants it.

avangelillo (approved):

ojproductions (approved):
lgtm

PR #2510, opened 2026-09-11
Use glob's unbundled entry point to avoid a stale bundled minimatch

GabrielAnca:
### What and why?

`glob`'s default entry point is a pre-bundled minified file that has its own copy of `minimatch` baked in at build time. That embedded copy is currently a vulnerable 10.2.2 version, and bumping or deduping the `minimatch` version in our own dependency tree has no effect on it, since the bundled file never actually resolves `minimatch` through node's module resolution.

### How?

Switch our shared glob helper (used by `plugin-junit`, `plugin-sarif`, `plugin-coverage`, and `plugin-synthetics`) to import from `glob/raw` instead of `glob`. That entry point is unbundled and resolves `minimatch` normally, so the patched version we already depend on is what actually runs.

[See glob's source where the default and `raw` exports are defined](https://github.com/isaacs/node-glob/blob/e80cb38ae60d6cbff9e75f39032a994858994d35/package.json#L28-L48).

### Review checklist

### DIFF
--- a/e2e/synthetics.test.ts
+++ b/e2e/synthetics.test.ts
@@ -12,6 +12,12 @@ describe('synthetics', () => {
       }
     )
 
+    // Assert the 2 tests defined in `e2e/fixtures/tests.synthetics.json` were resolved
+    // through the glob pattern defined in `e2e/fixtures/global.config.json`
+    expect(result.stdout).toContain('pwd-mwg-3p5')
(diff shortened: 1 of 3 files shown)


Drarig29 (approved):
Confirmed with a new E2E test that this will work as intended, even once our own packages are bundled.

PR #2512, opened 2026-09-11
[SVLS-8827] fix npmMinimalAgeGate units

ava-silver:
### What and why?

`npmMinimalAgeGate: "2d"` is silently parsed as `2` **minutes**, not 2 days. Yarn 4.10.3 declares this setting as a `NUMBER` (minutes) and coerces values with `parseInt`, so the age gate added in #2246 has effectively been disabled. (Duration strings like `"2d"` only work on Yarn >= 4.12, where the setting became a `DURATION` type.)

As a result, lockfile resolution can pin versions published less than 2 days ago. Those versions cannot be fetched through the internal npm mirror, which refuses tarballs still in their 2-day cooldown window (`403: package published 1 day ago, within cooldown window`). This currently breaks internal consumers that run `yarn install --immutable` against the mirror.

### How?

- Set `npmMinimalAgeGate: 2880` (2 days in minutes). A bare number is interpreted as minutes on both Yarn 4.10 and 4.12+, so the value stays correct across upgrades.
- Re-resolve the three packages currently pinned to in-cooldown versions (`baseline-browser-mapping`, `electron-to-chromium`, `update-browserslist-db`); with a working gate, Yarn resolves the newest versions at least 2 days old.

Validated with `yarn install --immutable` in a clean clone pointed at the internal mirror: all 1031 packages fetch successfully (previously 403 on the three re-pinned tarballs).

Related: https://github.com/ddoghq/serverless-ci/pull/238 (routes the lambda-layer-catalog campaign image through the internal mirror)

### Review checklist

  - N/A: config-only change; validated manually as described above

### DIFF
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -4,6 +4,6 @@ enableGlobalCache: false
 
 nodeLinker: node-modules
 
-npmMinimalAgeGate: "2d"
+npmMinimalAgeGate: 2880
 
 yarnPath: .yarn/releases/yarn-4.10.3.cjs
--- a/.yarnrc.yml
+++ b/.yarnrc.yml
@@ -4,6 +4,7 @@ enableGlobalCache: false
 
 nodeLinker: node-modules
 
-npmMinimalAgeGate: 2880
+# TODO: replace this with `2d` when we upgrade Yarn to >= 4.12
+npmMinimalAgeGate: 2880  # 2 days in minutes.
 
 yarnPath: .yarn/releases/yarn-4.10.3.cjs
--- a/yarn.lock
+++ b/yarn.lock
@@ -5413,11 +5413,11 @@ __metadata:
   linkType: hard
 
 "baseline-browser-mapping@npm:^2.11.20":
-  version: 2.11.22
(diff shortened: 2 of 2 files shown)


ava-silver:
* **#2512** <a href="https://app.graphite.com/github/pr/DataDog/datadog-ci/2512?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/datadog-ci/2512?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

.yarnrc.yml
```diff
@@ -4,6 +4,6 @@ enableGlobalCache: false
 
 nodeLinker: node-modules
 
-npmMinimalAgeGate: "2d"
+npmMinimalAgeGate: 2880
```
Drarig29:
```suggestion
# TODO: replace this with `2d` when we upgrade Yarn to >= 4.12
npmMinimalAgeGate: 2880  # 2 days in minutes.
```

Drarig29 (approved):
LGTM, just a nit for posterity

PR #2515, opened 2026-09-14
Elf-symbols: Add unwind information

jhssilva:
## Summary

- detect ELF files containing a non-empty `.eh_frame` section
- preserve `.eh_frame` contents when extracting debug information with `objcopy`
- send `generate_cfi_cache: true` in the existing `elf_symbol_file` upload metadata

## Motivation

The ELF-symbol uploader uses `objcopy --only-keep-debug`. This retains the `.eh_frame` section header but converts its contents to `NOBITS`, so the backend can generate a symcache for symbolication but cannot generate the separate CFI cache used for reliable minidump stack walking.

This PR preserves unwind information and explicitly requests CFI-cache generation through the existing ELF intake. It does not introduce a separate command, attachment type, or event type. Existing clients that omit the field retain the current behavior because the backend treats it as false.

A future change may expose an opt-out CLI flag that sends `generate_cfi_cache: false` and avoids preserving `.eh_frame` when minidump unwinding is not needed.

## Validation

- 38 focused ELF upload and parsing tests pass
- Prettier check passes
- ESLint passes with no errors
- affected TypeScript package compiles
- staging E2E previously confirmed ELF symcache symbolication; this change addresses the missing CFI artifact found during that test


jhssilva:
Closing this draft because preserving unwind information should be scoped to a dedicated minidump symbol upload command, rather than changing every profiler ELF upload. The replacement implementation is being developed locally with explicit minidump intake routing.

Drarig29 (approved):
LGTM from admin POV

maycmlee (approved):

Gandem (approved):
LGTM. 

Completely orthogonal, but raising so we don't forget: @nsavoire should we pull the same fix as https://github.com/DataDog/datadog-agent/pull/56694 in datadog-ci?

packages/base/src/commands/elf-symbols/upload.ts
```diff
@@ -205,6 +206,7 @@ export class ElfSymbolsUploadCommand extends BaseCommand {
       symbol_source: this.getElfSymbolSource(elfFileMetadata),
       filename: upath.basename(elfFileMetadata.filename),
       overwrite: this.replaceExisting,
+      generate_cfi_cache: this.includeUnwindInfo,
```
nsavoire, line 209: Perhaps this should conditioned on `hasEhFrame`:
```suggestion
      generate_cfi_cache: this.includeUnwindInfo && elfFileMetadata.hasEhFrame,
```

nsavoire:
> LGTM.
> 
> Completely orthogonal, but raising so we don't forget: @nsavoire should we pull the same fix as [DataDog/datadog-agent#56694](https://github.com/DataDog/datadog-agent/pull/56694) in datadog-ci?

@Gandem 
I opened [fix(elf-symbols): Preserve dynamic symbol headers on older binutils](https://github.com/DataDog/datadog-ci/pull/2527/s). Ideally the same flag preservation should be applied to `.eh_frame`.

packages/base/src/commands/elf-symbols/upload.ts
jhssilva: Thanks for your comment Nicolas. 
I'd keep this tied to includeUnwindInfo, the .eh_frame metadata isn't the only possible CFI source. 
Furthermore, the backend handles missing unwind information gracefully.

packages/base/src/commands/elf-symbols/elf.ts
```diff
@@ -606,6 +617,12 @@ export const copyElfDebugInfo = async (
   // Remove .gdb_index section as it is not needed and can be quite big
   let options = `${bfdTargetOption} --only-keep-debug ${compressDebugSectionsOption} --remove-section=.gdb_index`
 
+  if (includeUnwindInfo && elfFileMetadata.hasEhFrame) {
+    // --only-keep-debug turns allocated unwind sections into NOBITS sections. Preserve the contents so downstream
+    // processors can generate call frame information used to unwind optimized crash stacks.
+    options += ' --set-section-flags .eh_frame=alloc,readonly,contents'
```
nsavoire: I merged https://github.com/DataDog/datadog-ci/pull/2527, so you should be able to re-use `getObjcopySectionFlags` to get the correct section flags.

jhssilva: Added here [cc3643e](https://github.com/DataDog/datadog-ci/pull/2515/commits/cc3643ebb59744ef0f166343192190513bad6efe)
**Note**: Tested end to end with minidumps, CFI cache it's generated.

nsavoire: Thanks !

nsavoire (approved):

